PR #1, opened 2021-12-03
cxx-qt-gen: add support for QColor

ahayzen-kdab:
Change-Id: I205cb109a0017817dd535bf158c503158e54b956


ahayzen-kdab:
A version using UniquePtr as the return type is here #70 this is now the preferred route.

PR #3, opened 2021-12-06
Add Qt6 support

ahayzen-kdab:
This patch adds "Qt6" as a feature to cxx-qt-lib as to support the
change in QString binary size in Qt6.

Change-Id: I68c214eee48ec24b78ee3121892ed2689786dc21

### DIFF
--- a/examples/qt_types_standalone/Cargo.toml
+++ b/examples/qt_types_standalone/Cargo.toml
@@ -18,7 +18,7 @@ crate-type = ["staticlib"]
 [dependencies]
 cxx = "1.0"
 cxx-qt = { path = "../../cxx-qt" }
-cxx-qt-lib = { path = "../../cxx-qt-lib" }
+cxx-qt-lib = { path = "../../cxx-qt-lib", features=["Qt6"] }
 
 [build-dependencies]
 clang-format = { path = "../../clang-format" }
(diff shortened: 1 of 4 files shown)


examples/qt_types_standalone/Cargo.toml
```diff
@@ -18,7 +18,7 @@ crate-type = ["staticlib"]
 [dependencies]
 cxx = "1.0"
 cxx-qt = { path = "../../cxx-qt" }
-cxx-qt-lib = { path = "../../cxx-qt-lib" }
+cxx-qt-lib = { path = "../../cxx-qt-lib", features=["Qt6"] }
```
ahayzen-kdab, line 21: check if feature is the best way or if we can have something similar to defines work in C++.

eg can we replicate #[cfg(target_os = "macos")] with --cfg 'qt_major_version="5"' at the rustc level

https://doc.rust-lang.org/reference/conditional-compilation.html#the-cfg-attribute
https://doc.rust-lang.org/rustc/command-line-arguments.html#--cfg-configure-the-compilation-environment

ahayzen-kdab:
Since moving to use UniquePtr this has simplified the changes required, they are here now https://github.com/KDAB/cxx-qt/pull/74

PR #4, opened 2021-12-06
locking: add guards to rustObj inside C++ and postEvent property changes

ahayzen-kdab:
Change-Id: Iff2de681e5ff8e553c1bf0e7a466af2f9926687d


ahayzen-kdab:
We are going to use the combined one instead #5

PR #5, opened 2021-12-06
locking: introduce locking for m_rustObj and use a single lambda queue

ahayzen-kdab:


cxx-qt-gen/src/cxx_qt.h
```diff
@@ -73,53 +84,100 @@ class CxxQObject : public QObject
...
+  static const QEvent::Type ProcessQueueEvent;
 
 public:
   CxxQObject(QObject* parent = nullptr)
     : QObject(parent)
   {}
   virtual ~CxxQObject() = default;
 
-  void requestUpdate()
+  void requestEmitSignal(std::function<void()> signalFunctor)
```
LeonMatthesKDAB, line 95: This could actually be renamed to something like: `runOnGUIThread`. Because that's pretty much what this does now, right?

LeonMatthesKDAB: Maybe I'm missing something here, but why are all the functions implemented in the header file, and not a separate `.cpp` file?

cxx-qt-gen/src/cxx_qt.h
```diff
@@ -73,53 +84,100 @@ class CxxQObject : public QObject
...
-      // If we flip this flag before updateState then worst case we get an extra
-      // event with nothing to actually process whereas if we do it afterwards
-      // then we might miss an update request.
+      // If we flip this flag before takeQueue then worst case we get an
+      // extra event with nothing to actually process whereas if we do it
+      // afterwards then we might miss a queue item to process.
       m_waitingForUpdate.store(false, std::memory_order_relaxed);
 
-      updateState();
+      for (auto item : takeQueue()) {
```
LeonMatthesKDAB, line 170: Maybe use a structured binding here, i.e. `const auto& [type, functor] : takeQueue()`. Makes it more readable.

cxx-qt-gen/src/cxx_qt.h
```diff
@@ -73,53 +84,100 @@ class CxxQObject : public QObject
...
+    const std::lock_guard<std::mutex> guard(m_queueMutex);
+
+    if (!m_waitingForUpdate.exchange(true, std::memory_order_relaxed)) {
+      QCoreApplication::postEvent(this, new QEvent(ProcessQueueEvent));
+    }
+
+    m_queue.push_back(std::make_pair(QueueEvent::EmitSignal, signalFunctor));
+  }
+
+  void requestPropertyChange(std::function<void()> propertyFunctor)
```
LeonMatthesKDAB, line 109: What purpose does this function serve? It's pretty much a duplicate of requestEmitSignal.
Also, the UpdatePropertyChange type is never really read. So is there any point in having it around?

cxx-qt-gen/src/cxx_qt.h
```diff
@@ -73,53 +84,100 @@ class CxxQObject : public QObject
...
+    }
+
+    // Compress request updates to only one
+    //
+    // TODO: should we compress events? what happens if we change a
+    // property/emit signal, request update, change property/emit signal,
+    // request update? The second request update won't happen?
+    // Or should the request update always happen after the property/signal
+    // changes?
+    if (std::none_of(
```
LeonMatthesKDAB, line 142: I think this compression is really problematic with the current function interface.
The updateFunctor could theoretically do whatever, so simply dropping updateRequests, without checking whether the functor does the same thing (which you basically can't) is potentially problematic.

Either the updateFunctor is created within the requestUpdate method, which means it wouldn't need the argument at all, or we consolidate `requestUpdate`, `requestEmitSignal` and `requestPropertyChange` into a single function.
Like I mentioned earlier, this could be called e.g. `runOnGUIThread`.

In this case the m_queue could just be a queue of functions, instead of a pair of events and functions.
Currently either the function isn't really necessary, because the EventType already defines what ought to happen, or the eventtype is not really needed because what happens is entirely defined by the functor, so why have both?

LeonMatthesKDAB: If you want to keep the event-information, it would also be possible to use a `std::queue<std::variant<QueueEvent, std::function>>` in which case the action to be done is either defined by the QueueEvent, or by the function.

This would likely be more efficient for the updateRequest case, where the action to be taken is already known and we don't need to go through the std::function. However I'm also fine with just using std::functions for now, as that's more generic.

LeonMatthesKDAB (changes requested):
Generally moving in the right direction, I just think the interface with the eventqueue should be properly moved to a higher layer of abstraction as "A queue of code to be run on the GUI thread".

cxx-qt-gen/src/cxx_qt.h
ahayzen-kdab: Renamed to `runOnGUIThread` and moved to .cpp file. I think we were wondering if they would be inlined or templated.

cxx-qt-gen/src/cxx_qt.h
ahayzen-kdab: Now we don't have a pair, but do you want this to change? should it be a `const auto&` ?

```
    for (auto item : takeQueue()) {
      item();
    }
```

cxx-qt-gen/src/cxx_qt.h
ahayzen-kdab: The only purpose was to allow for requestUpdate to track it's enum, so now that's removed, these methods are merged.

cxx-qt-gen/src/cxx_qt.h
LeonMatthesKDAB: Not sure if the compiler can optimize this away, but simply using `auto` would likely cause the compiler to copy every std::function. So using `const auto& item` would indeed be preferred here.

cxx-qt-gen/src/cxx_qt.h
ahayzen-kdab: As discussed, removed the compression of events for now (if the user wants to compress events they can do that in their code for now). So now we have `runOnGUIThread` and just `std::queue<std::function<void()>>`

cxx-qt-gen/src/cxx_qt.cpp
```diff
@@ -55,8 +56,16 @@ extern "C"
     if (self->isNull())
       return false;
 
-    self->data()->requestUpdate();
-    return true;
+    const auto ret = QMetaObject::invokeMethod(
+      self->data(), "requestUpdate", Qt::DirectConnection);
```
LeonMatthesKDAB, line 60: `requestUpdate` is now removed from the CxxQObject. Is this a problem for this call?

cxx-qt-gen/src/cxx_qt.h
ahayzen-kdab: Changed :-)

cxx-qt-gen/src/cxx_qt.cpp
ahayzen-kdab: This is calling the `requestUpdate` that is generated in `gen_cpp` (it's a little confusing), so this is the one in the generation https://github.com/ahayzen-kdab/cxx-qt/blob/locking-5-lambda-queue/cxx-qt-gen/src/gen_cpp.rs#L773

Or this one in the tests https://github.com/ahayzen-kdab/cxx-qt/blob/locking-5-lambda-queue/cxx-qt-gen/test_outputs/basic_update_requester.cpp#L31

The idea is that

- The user has the RequestUpdater object in Rust (which can be moved across threads)
- They trigger request_update on the Rust side
- This then triggers requestUpdate on the C++ object
- This then calls runOnGUIThread and queues the updateState call
- The updateState calls then locks and goes back to Rust from the GUI thread

LeonMatthesKDAB: Ah, okay, didn't see that :)
I think we're ready to merge now :+1:

LeonMatthesKDAB (approved):
All issues fixed :+1:

PR #7, opened 2021-12-08
Add minimal example for showcases

LeonMatthesKDAB:


examples/minimal_cxx_qt_qml/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+        string: String,
+    }
+
+    #[derive(Default)]
+    struct RustObj;
+
+    impl RustObj {
+        #[invokable]
+        fn increment_number(&self, cpp: Pin<&mut CppObj>) {
+            let value = cpp.number() + 1;
```
ahayzen-kdab, line 22: Wonder if this should use the wrapper instead, as that's what will happen in the long term. As for that specific case you can get away with using the CppObj, but as soon as you try to do multiple write changes you'll get compile issues.

A wrapper route might look like this

```rust
let mut wrapper = CppObjWrapper::new(cpp);
let value = wrapper.number() + 1;
wrapper.set_number(value);
```

Eventually this would be transparent to the user though and instead of `Pin<&mut CppObj>` they would have a `CppObjectWrapper` etc.

examples/minimal_cxx_qt_qml/src/main.qml
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        Button {
+            text: "Increment Number"
+
+            onClicked: myObject.incrementNumber()
+        }
+    }
+
+    Component.onCompleted: myObject.sayHi(myObject.string, myObject.number)
```
ahayzen-kdab, line 44: Could maybe have a button to trigger this? To demo console printing from the rust side?

examples/minimal_cxx_qt_qml/src/tests/myobject/tst_myobject.qml
```diff
@@ -0,0 +1,80 @@
...
+
+        }
+    }
+
+    function test_increment() {
+        const myObject = createTemporaryObject(componentMyObject, null, {});
+        compare(myObject.incrementNumber(1), 2);
+    }
+
+    function test_increment_self() {
```
ahayzen-kdab, line 35: I think increment_number_self has been removed?

examples/minimal_cxx_qt_qml/src/tests/myobject/tst_myobject.qml
```diff
@@ -0,0 +1,80 @@
...
+        id: componentSpy
+
+        SignalSpy {
+
+        }
+    }
+
+    function test_increment() {
+        const myObject = createTemporaryObject(componentMyObject, null, {});
+        compare(myObject.incrementNumber(1), 2);
```
ahayzen-kdab, line 32: This fails as increment_number on the rust side doesn't return any more and doesn't take a parameter.

So instead use
```qml
myObject.incrementNumber();
compare(myObject.number(), 2);
```

examples/minimal_cxx_qt_qml/src/tests/myobject/tst_myobject.qml
```diff
@@ -0,0 +1,80 @@
+// SPDX-FileCopyrightText: 2021 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+// SPDX-FileContributor: Gerhard de Clercq <gerhard.declercq@kdab.com>
```
ahayzen-kdab, line 3: Please add yourself to SPDX-FileContributor here any elsewhere where it makes sense :-)

examples/minimal_cxx_qt_qml/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+# SPDX-FileCopyrightText: 2021 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+# SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+# SPDX-FileContributor: Gerhard de Clercq <gerhard.declercq@kdab.com>
+#
+# SPDX-License-Identifier: MIT OR Apache-2.0
+[package]
+name = "example-minimal-cxx-qt-qml"
+version = "0.1.0"
+authors = ["Andrew Hayzen <andrew.hayzen@kdab.com>", "Gerhard de Clercq <gerhard.declercq@kdab.com>"]
```
ahayzen-kdab, line 9: Consider adding yourself here as well as the SPDX lines.

ahayzen-kdab (changes requested):
Looks good, some comments about the tests to fix :-)

examples/minimal_cxx_qt_qml/src/lib.rs
ahayzen-kdab: So maybe it's OK to leave as you have currently. As eventually the user won't know there is a wrapper.

And maybe this indicates this task needs to be done soonish :-)

ahayzen-kdab: This will be done in #8

LeonMatthesKDAB: Ah, I didn't know that could cause such problems, I'll update it to the Wrapper type.
If we want to demonstrate that code, it better use the appropriate calls.

examples/minimal_cxx_qt_qml/src/tests/myobject/tst_myobject.qml
LeonMatthesKDAB: Yup, must have missed this

examples/minimal_cxx_qt_qml/src/lib.rs
LeonMatthesKDAB: Fixed

examples/minimal_cxx_qt_qml/src/tests/myobject/tst_myobject.qml
LeonMatthesKDAB: Done

examples/minimal_cxx_qt_qml/src/tests/myobject/tst_myobject.qml
LeonMatthesKDAB: Done

examples/minimal_cxx_qt_qml/src/tests/myobject/tst_myobject.qml
LeonMatthesKDAB: Done

examples/minimal_cxx_qt_qml/src/main.qml
LeonMatthesKDAB: Done

examples/minimal_cxx_qt_qml/Cargo.toml
LeonMatthesKDAB: Done

ahayzen-kdab (approved):
LGTM :-)

PR #41, opened 2021-12-09
Demo threading

ahayzen-kdab:


examples/demo_threading/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+            let update_requester = cpp.update_requester();
+
+            // Prepare our processing thread which builds average, count, total
+            let run_processing = async move {
+                let mut sensors: HashMap<Uuid, f64> = HashMap::new();
+
+                loop {
+                    let mut changed = false;
+
+                    Delay::new(Duration::from_millis(8)).await;
```
ahayzen-kdab, line 205: This causes CPU usage, instead using a blocking channel where recv() waits or CondVar might be a better solution.

ahayzen-kdab:
Should network and processing thread can be merged, so that we can respond failure to the network requests and remove a channel + thread.

ahayzen-kdab:
> Should network and processing thread can be merged, so that we can respond failure to the network requests and remove a channel + thread.

Done :-)

ahayzen-kdab:
So this works with Arc and Mutex quite nicely, but not sure how much it shows the benefits of Rust and the possible lock-free routes with channels like before ? Wonder if there is a solution between the original and this one ...

examples/demo_threading/src/lib.rs
```diff
@@ -0,0 +1,334 @@
...
+#[serde(rename_all = "snake_case")]
+enum RequestCommand {
+    Disconnect,
+    Power { value: f64 },
+}
+
+#[derive(Deserialize, Serialize)]
+struct Request {
+    command: RequestCommand,
+    uuid: Option<Uuid>,
```
ahayzen-kdab, line 35: Uuid isn't option anymore, so this could be changed

ahayzen-kdab: Done

examples/demo_threading/src/lib.rs
```diff
@@ -0,0 +1,335 @@
...
+                join_handle_timeout: None,
+                join_handle_update: None,
+            }
+        }
+    }
+
+    impl RustObj {
+        async fn handle_connection(mut stream: TcpStream, network_tx: SyncSender<NetworkChannel>) {
+            let mut buf = vec![0u8; 1024];
+            let _ = stream.read(&mut buf).await.unwrap();
```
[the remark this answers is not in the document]
LeonMatthesKDAB: We should just generally do proper error handling for all user input. We do want to demonstrate Rusts robustness compared to C++ after all.
So the last thing we need is our demo exploding when our network driver has a hickup :roll_eyes:

examples/demo_threading/src/lib.rs
```diff
@@ -0,0 +1,335 @@
...
+                            .into(),
+                    }
+                }
+                Err(_) => Status::ErrorInvalidRequest.into(),
+            };
+
+            stream
+                .write(serde_json::to_string(&response).unwrap().as_bytes())
+                .await
+                .unwrap();
```
LeonMatthesKDAB, line 185: Again, no unwrap here, please, just using `.ok()` here should be fine.

examples/demo_threading/src/lib.rs
```diff
@@ -0,0 +1,335 @@
...
+                sensors: 0,
+                total_use: 0.0,
+            }
+        }
+    }
+
+    struct RustObj {
+        qt_rx: Receiver<Data>,
+        qt_tx: SyncSender<Data>,
+        join_handle_network: Option<JoinHandle<()>>,
```
LeonMatthesKDAB, line 132: These three fields can probably be moved into a single:
`join_handles: Option<[JoinHandle<()>;3]>` member

examples/demo_threading/client.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env python3
```
LeonMatthesKDAB, line 1: The behavior of this client script is very different from what I expected.
I would expect this to spin up a single client that runs for as-long as the user wants.

That would make it easy to spin up as many clients as needed, pause them, kill them, observe their behavior, etc.
We could also have a "nastyclient.py" that tries to inject false events, scrambled json, etc.

LeonMatthesKDAB (changes requested):
Overall Rust code structure seems reasonable. I think we still need to think through the role the client should play though.

examples/demo_threading/src/lib.rs
```diff
@@ -0,0 +1,335 @@
...
+            }
+        }
+    }
+
+    impl RustObj {
+        async fn handle_connection(mut stream: TcpStream, network_tx: SyncSender<NetworkChannel>) {
+            let mut buf = vec![0u8; 1024];
+            let _ = stream.read(&mut buf).await.unwrap();
+            let trimmed = std::str::from_utf8(&buf)
+                .unwrap()
```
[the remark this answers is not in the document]
ahayzen-kdab: Done

examples/demo_threading/src/lib.rs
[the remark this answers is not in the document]
ahayzen-kdab: Done

examples/demo_threading/src/lib.rs
ahayzen-kdab: Done

examples/demo_threading/src/lib.rs
ahayzen-kdab: Done

examples/demo_threading/src/lib.rs
```diff
@@ -0,0 +1,355 @@
...
+                        if changed {
+                            let total_use = sensors.values().fold(0.0, |acc, x| acc + x.power);
+                            let sensors = sensors.len() as u32;
+                            let average_use = if sensors > 0 {
+                                total_use / (sensors as f64)
+                            } else {
+                                0.0
+                            };
+
+                            qt_tx
```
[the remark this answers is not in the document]
ahayzen-kdab: Changed to use Mutex for now.

LeonMatthesKDAB (approved):
Looks much better now.

PR #47, opened 2022-01-06
Add QRectF support

ahayzen-kdab:


examples/qt_types_standalone/src/main.cpp
```diff
@@ -241,6 +241,41 @@ TEST_CASE("Can copy a value QPointF on the Rust side")
   CHECK(c.y() == 4.56);
 }
 
+TEST_CASE("Can construct a QRectF on the Rust side")
+{
+  const auto r = construct_qrectf();
+  CHECK(r.x() == 1.23);
```
LeonMatthesKDAB, line 247: Maybe we should be using epsilon-based checks here as well? Like we do in read_qrectf().
Theoretically these floating point numbers can compile down to slightly different binary representations, so they won't be exactly equal :thinking:

LeonMatthesKDAB (commented):
Generally looks good, using epsilon-based comparison for the float checks might save us some trouble in the future.

examples/qt_types_standalone/src/main.cpp
ahayzen-kdab: Fixed to use `qFuzzyCompare` and then added a commit to change the QPointF and QSizeF instances.

examples/qt_types_standalone/src/main.cpp
```diff
@@ -287,14 +322,14 @@ TEST_CASE("Can copy a QSizeF on the Rust side")
...
+  CHECK(qFuzzyCompare(c.height(), 4.56));
 }
 
 TEST_CASE("Can copy a value QSizeF on the Rust side")
 {
   const auto s = QSizeF(1.23, 4.56);
   const auto c = copy_value_qsizef(s);
-  CHECK(c.width() == 1.23);
-  CHECK(c.height() == 4.56);
+  CHECK(qFuzzyCompare(c.width(), 1.23));
```
LeonMatthesKDAB, line 333: I think the fuzzyCompare here is actually not necessary, as we're constructing & checking in C++, but it also shouldn't hurt.

LeonMatthesKDAB (approved):

examples/qt_types_standalone/src/main.cpp
LeonMatthesKDAB: :+1:

PR #50, opened 2022-01-07
Add QSize support

ahayzen-kdab:


cxx-qt-gen/src/cxx_qt.cpp
```diff
@@ -178,6 +179,36 @@ extern "C"
 
 namespace {
 
+// We do these checks to ensure that we can safely store a QSize
+// inside a block of memory that Rust thinks contains two i32-s.
+// We also make sure that i32 and int are equivalent.
+
+static_assert(sizeof(int) == 4);
```
LeonMatthesKDAB, line 186: Hm, these are technically redundant, because we already have these sizeof(int) checks at QPoint.
But probably good to assert it again, so it's clear this is needed for both

LeonMatthesKDAB (approved):
LGTM

cxx-qt-gen/src/cxx_qt.cpp
ahayzen-kdab: Right good point, we are checking "everything" again for every type. Probably makes sense incase QPoint changes etc. And they are only static_asserts so not expensive ?

Issue #10, opened 2021-12-08
Add basic Qt types


ahayzen-kdab:
- [X] QColor - #1 #70
- [x] QDate - #52
- [x] QDateTime - #64
- [x] QTime (?) - #63
- [x] QPoint - #45
- [X] QPointF
- [x] QRect - #51
- [x] QRectF - #47
- [x] QSize - #50
- [x] QSizeF - #46
- [X] QString
- [X] QVariant - Note we need to add support for each new type #11
- [x] QUrl - #62

https://doc.qt.io/qt-5/qtqml-typesystem-basictypes.html

- [X] Ensure all types have the same `impl Default` as Qt and any relevant get/set methods. (split into #53 )

LeonMatthesKDAB:
I'd consider adding #55 as a task here as well.

ahayzen-kdab:
> I'd consider adding #55 as a task here as well.

I think this should be combined in a second pass with things like #53.

Otherwise we have meta issues open for ages, lets keep this task just for adding basic "support" for the types, then use the other issues to track the default value + useful methods for Rust etc.

ahayzen-kdab:
All the basic types have landed! :partying_face:

PR #51, opened 2022-01-07
cxx-qt-lib: add QRect support

ahayzen-kdab:
Related to #10

Change-Id: I9fcb9e5592b06d6282ef4bf162e12f871d381337


ahayzen-kdab:
Stacked on top of #47, will be marked as ready once that lands.

cxx-qt-lib/src/qrect.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    pub fn set_height(&mut self, h: i32) {
+        self.y2 = self.y1 + h - 1;
+    }
+
+    pub fn set_width(&mut self, w: i32) {
+        self.x2 = self.x1 + w - 1;
+    }
+
+    pub fn set_x(&mut self, x: i32) {
```
LeonMatthesKDAB, line 56: Add explaining documentation comment, why setting the x value will change the width.
Same for y + height

LeonMatthesKDAB (commented):
Just need to add some explaining docs comments for some of the QRect weirdness.
Otherwise looks good.

cxx-qt-lib/src/qrect.rs
ahayzen-kdab: Done :-)

LeonMatthesKDAB (approved):
LGTM

PR #52, opened 2022-01-07
Add QDate support

ahayzen-kdab:


cxx-qt-lib/src/qdate.rs
```diff
@@ -0,0 +1,194 @@
...
+                        29
+                    } else {
+                        28
+                    }
+                } else {
+                    QDate::days_in_usual_month(month)
+                })
+    }
+
+    pub fn year(&self) -> i32 {
```
ahayzen-kdab, line 81: wonder if we should somehow call the QDate::year on the C++ side and use it's response? Then we could do the same with year, month, day, set_date. This would allow for removing the duplication of Qt C++ code.

ahayzen-kdab: Done this is the latest commit now.

cxx-qt-gen/src/cxx_qt.cpp
```diff
@@ -118,6 +119,35 @@ extern "C"
 
 namespace {
 
+// We do these checks to ensure that we can safely store a QDate
+// inside a block of memory that Rust thinks contains one i64.
+// We also make sure that i64 and qin64 are equivalent.
```
ahayzen-kdab, line 124: typo should be qint64

cxx-qt-gen/src/cxx_qt.cpp
```diff
@@ -118,6 +119,35 @@ extern "C"
...
+static_assert(alignof(qint64) <= 8);
+
+static_assert(sizeof(QDate) == 8);
+static_assert(alignof(QDate) <= 8);
+
+// Our Rust code assumes that QDate is trivial. Because it is trivial to move,
+// we don't need to use Pin. Because it is trivial to destruct we do not
+// need a special C++ function to destruct the object.
+
+static_assert(std::is_trivially_move_assignable<QDate>::value == true);
```
ahayzen-kdab, line 136: don't need the == true

LeonMatthesKDAB (approved):
LGTM

PR #56, opened 2022-01-12
ci: add github actions CI for ctest

ahayzen-kdab:
Change-Id: I688fadf16830b9bdeef6100b5d578ccdf9941023

### DIFF
--- a/.github/workflows/github-cxx-qt-tests.yml
+++ b/.github/workflows/github-cxx-qt-tests.yml
@@ -1,0 +1,35 @@
+# SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+# SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+#
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+name: cxx-qt tests
+on: [push, pull_request]
+jobs:
+  ctest:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      # Install Rust toolchain
+      - uses: actions-rs/toolchain@v1
+        with:
+          components: clippy, rustfmt
+          toolchain: stable
+      # FIXME: FUTURE: Remove PPA for Qt 5.15 once 22.04 is available
+      # https://launchpad.net/~savoury1/+archive/ubuntu/qt-5-15
+      - run: sudo add-apt-repository ppa:savoury1/qt-5-15
+      # Install clang-format, Qt, reuse, valgrind
+      - run: sudo apt-get update
+      - run: sudo apt-get install -y clang-format-12 libssl-dev pkg-config python3-pip qtbase5-dev qtdeclarative5-dev qtquickcontrols2-5-dev qml-module-qtquick2 qml-module-qtquick-controls2 qml-module-qttest valgrind
+      - run: sudo ln -s /usr/bin/clang-format-12 /usr/local/bin/clang-format
+      - run: pip3 install reuse
+      # Ensure we have a clean build folder
+      - run: rm -rf build/
+      - run: mkdir -p build/
+      # Configure and build CMake then run ctest
+      #
+      # Note that each github action run command is executed from the same directory even after
+      # performing cd so for the ctest line we need to cd within the same command
+      - run: cmake -S. -Bbuild/ -G "Unix Makefiles"
+      - run: cmake --build build/ --config Debug --target all -j$(nproc)
+      - run: (cd build/ && QT_QPA_PLATFORM=offscreen ctest -j$(nproc) -C Debug -T test --output-on-failure)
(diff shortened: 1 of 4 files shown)


valgrind_suppressions.txt
```diff
@@ -14,4 +14,20 @@
    fun:dlopen@@GLIBC_*
    fun:*QLibraryPrivate*load_sysEv
    fun:*QLibraryPrivate*loadEv
+}
+{
+   qt_quick_test_setup_with_platform_offscreen
+   Memcheck:Leak
+   match-leak-kinds: definite
+   fun:_Znwm
+   obj:/usr/lib/x86_64-linux-gnu/qt5/plugins/platforms/libqoffscreen.so
```
ahayzen-kdab, line 23: wonder if we should remove the arch from the path here, so it can work on other platforms ? eg just have `*/plugins/platforms/libqoffscreen.so` or something

ahayzen-kdab: Done lets see if that still passes

ahayzen-kdab:
Note you can see here if it's passing or failing https://github.com/ahayzen-kdab/cxx-qt/actions  As with clang-format-rs, once this is merged for future pull requests the pass/fail status will appear here.

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -0,0 +1,35 @@
...
+on: [push, pull_request]
+jobs:
+  ctest:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      # Install Rust toolchain
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
```
LeonMatthesKDAB, line 16: You should be able to use `components: rustfmt, clippy` with the `actions-rs/toolchain@v1` and omit the next two `run:` commands.

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -0,0 +1,35 @@
...
+      - run: pip3 install reuse
+      # Ensure we have a clean build folder
+      - run: rm -rf build/
+      - run: mkdir -p build/
+      # Force configure of cmake
+      - run: cmake -H. -Bbuild/ -G "Unix Makefiles"
+      # Build the cmake project
+      - run: cmake --build build --config Debug --target all -j$(nproc)
+      # Run ctest
+      - run: (cd build && QT_QPA_PLATFORM=offscreen ctest -j$(nproc) -C Debug -T test --output-on-failure)
```
LeonMatthesKDAB, line 35: Any reason the `cd` and `ctest` is done in one line?
It would be cleaner if it were two separate run commands.

We could also simply `cd` into the build folder after `mkdir` and then replace the build & test commands with:

```
- run: cmake .. -G "Unix Makefiles"
- run: cmake --build . --config Debug --target all -j$(nproc)
- run: QT_QPA_PLATFORM=offscreen ctest -j$(nproc) -C Debug -T test --output-on-failure
```
Which would also remove the `-H.` option from the cmake command, which is a undocumented feature of CMake and might be removed in the future (see: https://stackoverflow.com/questions/31090821/what-does-the-h-option-means-for-cmake)

LeonMatthesKDAB (changes requested):
Overall looks good, just some simplifications that can still be done.

.github/workflows/github-cxx-qt-tests.yml
ahayzen-kdab: Done :-)

.github/workflows/github-cxx-qt-tests.yml
ahayzen-kdab: Good idea lets try that :-) I was only worried if the cd doesn't stay across run commands. I have some seem systems where changing directory isn't preserved across commands ...

ahayzen-kdab: So yes seems that all commands are run from the same directory... even after a `cd build/` this happens. When from previous runs it looks like it should be looking in the dir `/home/runner/work/cxx-qt/cxx-qt`
```
Run cmake .. -G "Unix Makefiles"
CMake Error: The source directory "/home/runner/work/cxx-qt" does not appear to contain CMakeLists.txt.
Specify --help for usage, or press the help button on the CMake GUI.
Error: Process completed with exit code 1.
```

ahayzen-kdab: So reverted back to using `cd build && ..`, I have dropped the `-H.` this time that vscode uses ;-)

The other option would be to put this all inside something like a `bash -c "..."` or something, but how it is now i think makes sense still

LeonMatthesKDAB: Ah, okay, then we can leave it as-is, but please add a comment explaining the situation, so no one tries to simplify this again ;)

ahayzen-kdab: Added a comment :-)

PR #62, opened 2022-01-17
Add support for QUrl

ahayzen-kdab:


cxx-qt-lib/src/qurl.rs
```diff
@@ -0,0 +1,242 @@
...
+    // scheme: Option<String>,
+    // userName: Option<String>,
+}
+
+#[repr(C)]
+pub struct Url {
+    pub d: Box<UrlImpl>,
+}
+
+impl Url {
```
ahayzen-kdab, line 189: Should we have a `as_string() -> &String` method here? So they UrlImpl doesn't need to be used in Rust?

examples/basic_cxx_qt_qml/src/tests/qttypes/tst_qttypes.qml
```diff
@@ -411,6 +410,47 @@ TestCase {
...
+        const spy = createTemporaryObject(componentSpy, null, {
+            signalName: "urlChanged",
+            target: mock,
+        });
+        compare(mock.url, "https://github.com/KDAB");
+
+        compare(spy.count, 0);
+        mock.url = "https://github.com/KDAB/cxx-qt";
+        tryCompare(spy, "count", 1);
+        compare(mock.url, "https://github.com/KDAB/cxx-qt");
```
LeonMatthesKDAB, line 428: Haha, good Url ;)

cxx-qt-lib/src/qurl.rs
```diff
@@ -0,0 +1,175 @@
...
+        }
+    }
+
+    // TODO: other QUrl methods
+    //
+    // fragment: Option<String>,
+    // host: Option<String>,
+    // password: Option<String>,
+    // path: Option<String>,
+    // port: i32,
```
LeonMatthesKDAB, line 149: The port should probably be an Option<u16>

cxx-qt-lib/src/qurl.rs
```diff
@@ -0,0 +1,175 @@
...
+            // Safety: TODO
+            inner: unsafe {
+                let mut ptr = MaybeUninit::<cxx::UniquePtr<QUrl>>::zeroed();
+                qurl_init_from_qurl(&mut ptr, qurl);
+                ptr.assume_init()
+            },
+        }
+    }
+
+    pub fn from_string(s: String) -> Self {
```
LeonMatthesKDAB, line 132: Why does this take a `String`, instead of a `&str`?

examples/basic_cxx_qt_qml/src/mock_qt_types.rs
```diff
@@ -38,6 +39,7 @@ mod mock_qt_types {
                 size: QSize::new(1, 3),
                 sizef: QSizeF::new(1.0, 3.0),
                 time: QTime::new(1, 2, 3, 4),
+                url: Url::from_string("https://github.com/KDAB".to_owned()),
```
LeonMatthesKDAB, line 42: to_owned() could be removed here if Url::from_string took a `&str`.

cxx-qt-lib/src/qurl.rs
```diff
@@ -0,0 +1,175 @@
...
+    /// Retrieve the UniquePtr to the Qt QUrl of this Rust Url
+    /// so that this object can be passed back to C++.
+    fn to_unique_ptr(self) -> cxx::UniquePtr<QUrl> {
+        self.inner
+    }
+}
+
+impl From<&QUrl> for Url {
+    fn from(qurl: &QUrl) -> Self {
+        qurl.to_rust()
```
LeonMatthesKDAB, line 173: When #9 is implemented, maybe we should reconsider whether to use a From<>, or a custom ToRust Trait.
Currently, the to_rust method is somewhat unnecessary, as we could just use .into() and implement the conversion directly in the trait impl.

cxx-qt-lib/src/qurl.rs
LeonMatthesKDAB: Probably rename to from_str then as well

LeonMatthesKDAB (commented):
Open Question: Why does from_string need to take a String? Can't it just take a `&str`?

Otherwise just a few minor nitpicks.

cxx-qt-lib/src/qurl.rs
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    #[doc(hidden)]
+    unsafe fn __drop(mut repr: MaybeUninit<*mut c_void>) {
+        unique_ptr_qurl_drop(&mut repr)
+    }
+}
+
+pub struct Url {
+    // Note that once map_qt_value is removed later, this can become private again
```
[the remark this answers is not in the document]
ahayzen-kdab: Done

cxx-qt-lib/src/qurl.rs
ahayzen-kdab: Done

examples/basic_cxx_qt_qml/src/mock_qt_types.rs
ahayzen-kdab: done

cxx-qt-lib/src/qurl.rs
ahayzen-kdab: Right, maybe note that in #9 or somewhere :-) Although I think i was thinking of just removing implementing a trait and instead just making the macro expansion put `to_rust()` where it needs to.

cxx-qt-lib/src/qurl.rs
ahayzen-kdab: Done

LeonMatthesKDAB (approved):
LGTM

PR #64, opened 2022-01-17
Add support for QDateTime

ahayzen-kdab:


cxx-qt-lib/src/qdatetime.rs
```diff
@@ -0,0 +1,267 @@
...
+
+    pub fn date(&self) -> &QDate {
+        &self.d.date
+    }
+
+    pub fn time(&self) -> &QTime {
+        &self.d.time
+    }
+
+    pub fn set_date(&mut self, date: QDate) {
```
ahayzen-kdab, line 213: Should we allow mutating of this DateTime or should they always need to construct a new one ?

cxx-qt-lib/src/qdatetime.rs
```diff
@@ -0,0 +1,198 @@
...
+        // Safety: TODO
+        unsafe { qdatetime_get_date(&self.inner) }
+    }
+
+    pub fn time(&self) -> QTime {
+        // Safety: TODO
+        unsafe { qdatetime_get_time(&self.inner) }
+    }
+
+    pub fn set_date(&mut self, date: &QDate) {
```
LeonMatthesKDAB, line 169: More Rust-y way would be to take a `QDate` instead of a `&QDate`, but because the C++ code needs a const reference anyway, I think taking a reference is fine.

cxx-qt-lib/src/qdatetime.rs
```diff
@@ -0,0 +1,198 @@
...
+            inner: unsafe {
+                let mut ptr = MaybeUninit::<cxx::UniquePtr<QDateTime>>::zeroed();
+                qdatetime_init_from_qdatetime(&mut ptr, qdatetime);
+                ptr.assume_init()
+            },
+        }
+    }
+
+    /// Construct a Rust DateTime from a given QDate and QTime
+    pub fn from_date_time(date: &QDate, time: &QTime) -> Self {
```
LeonMatthesKDAB, line 148: Maybe we should rename this to `from_date_and_time`.

I expected it to actually take a DateTime at first glance.

LeonMatthesKDAB (commented):
Looks good, only question: do we want to rename `from_date_time` to `from_date_and_time`

cxx-qt-lib/src/qdatetime.rs
ahayzen-kdab: Done

LeonMatthesKDAB (approved):
LGTM

Issue #12, opened 2021-12-08
Move CxxQObject to be a member rather than inherited


ahayzen-kdab:
We should change CxxQObject to be a member of the generated class rather than inherited.

This will then allow us to reuse the same object for other classes we generate (eg non-QObject ones).

This could look something like `ThreadHelper m_threadHelper` and then `m_threadHelper->runOnGUIThread(...)`.

PR #67, opened 2022-01-25
cxx-qt-gen: add support for declaring and emitting signals 

### LINKED ISSUE #31: Ability to declare and emit signals
* Have the ability to write enums for signals
 * ~~Have a handleSignal trait for listening to signals~~
 * CppObj has emit_queued(Signal) if there is a signal enum, and then uses runOnGUIThread
 * CppObj has unsafe emit_now(Signal) that is called now
 * ~~If there is a handleSignal then in constructor of C++ obj we need to connect signals to a method which proxies to RustObj~~
 * Extract phase needs
   * signals enum
   * if there is a trait

Rust API would use enums, then have an emit method in the wrapper. We use named struct like enums so that the names can be used for the C++ signal arg.

```
enum Signal {
  Ready,
  DataChanged { first: i32, second: i32 }
}

impl RustObj {
  #[invokable]
  fn invokable(&self, cpp: &mut CppObj) {
    cpp.emit_queued(Signal::Ready { 1, 2 });
    cpp.emit_queued(Signal::DataChanged { 1, 2 });
  }
}
```

Internally the emit method destructures the enum and call a relevant C++ method.

```
impl CppObj {
  fn emit_queued(&mut self, signal: Signal) {
    match signal {
      Signal::Ready => self.cpp.as_mut().emit_ready(),
      Signal::DataChanged(first, second) => self.cpp.as_mut().emit_data_changed(first, second),
    }
  }
}
```

The public C++ side would then appear as

```
Q_SIGNALS:
  void ready();
  void dataChanged(qint32 first, qint32 second);
```

Then the C++ side would have methods for each of these signals to emit them on the GUI thread (these are called from Rust).

```
void MyObject::emitReady() {
  runOnGUIThread([&]() { Q_EMIT ready(); };
}

void MyObject::emitDataChanged(qint32 first, qint32 second) {
  runOnGUIThread([&]() { Q_EMIT dataChanged(); };
}
```

The flow is the following for a generic "MySignal" with "number".

```
# A signal being emitted from Rust
CppObj::emit_queued(Signal::MySignal { number }) -> MyObject::emit_my_signal(number) -> runOnGUIThread -> Q_EMIT MyObject::mySignal(number)
```

ahayzen-kdab:
Closes #31

Change-Id: Id3781124f0090c0f69517d346ccefb33b6df000d

### DIFF
--- a/cxx-qt-gen/src/gen_rs.rs
+++ b/cxx-qt-gen/src/gen_rs.rs
@@ -422,6 +422,67 @@ pub fn generate_qobject_cxx(
         }
     }
 
+    // Add signals emitters
+    for signal in &obj.signals {
+        let signal_ident_cpp = &signal.signal_ident.cpp_ident;
+        let signal_ident_rust_str = &signal.signal_ident.rust_ident.to_string();
+
+        let queued_ident_cpp = &signal.emit_ident.cpp_ident;
+        let queued_ident_rust_str = &signal.emit_ident.rust_ident.to_string();
+
+        if signal.parameters.is_empty() {
+            cpp_functions.push(quote! {
+                #[rust_name = #signal_ident_rust_str]
+                fn #signal_ident_cpp(self: Pin<&mut #class_name>);
+                #[rust_name = #queued_ident_rust_str]
+                fn #queued_ident_cpp(self: Pin<&mut #class_name>);
+            });
+        } else {
+            // For immediate parameters we want by-ref or primitive by-value
+            let parameters = signal
+                .parameters
+                .iter()
+                .map(|parameter| {
+                    let ident = &parameter.ident;
+                    let param_type = parameter.type_ident.qt_type.cxx_bridge_type_ident();
+                    if parameter.type_ident.qt_type.is_ref() {
+                        quote! {
+                            #ident: &#param_type
+                        }
+                    } else {
+                        quote! {
+                            #ident: #param_type
+                        }
+                    }
+                })
+                .collect::<Vec<TokenStream>>();
+            // For queued parameters we want by-value (or UniquePtr<T>)
+            let parameters_queued = signal
+                .parameters
+                .iter()
+                .map(|parameter| {
+                    let ident = &parameter.ident;
+                    let param_type = parameter.type_ident.qt_type.cxx_bridge_type_ident();
+                    if parameter.type_ident.qt_type.is_opaque() {
+                        quote! {
+                            #ident: UniquePtr<#param_type>
+                        }
+                    } else {
+                        quote! {
+                            #ident: #param_type
+                        }
+                    }
+                })
+                .collect::<Vec<TokenStream>>();
+            cpp_functions.push(quote! {
+                #[rust_name = #signal_ident_rust_str]
+                fn #signal_ident_cpp(self: Pin<&mut #class_name>, #(#parameters),*);
+                #[rust_name = #queued_ident_rust_str]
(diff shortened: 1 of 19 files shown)


cxx-qt-gen/test_outputs/basic_only_signals.cpp
```diff
@@ -0,0 +1,41 @@
...
+void
+MyObject::emitReady()
+{
+  runOnGUIThread([=]() { Q_EMIT ready(); });
+}
+
+void
+MyObject::emitDataChanged(qint32 first, const QVariant& second)
+{
+  runOnGUIThread([=]() { Q_EMIT dataChanged(first, second); });
```
ahayzen-kdab, line 32: Can we avoid a copy of a the QVariant here and instead move into the lambda ? As from Rust this is already a copy of QVariant that we "own".

This is the case for both opaque and non-opaque as we pass a Signal which has by-value items and opaque types are converted using to_unique_ptr.

ahayzen-kdab: Should we just pass UniquePtr's for opaque types from the Rust side?

cxx-qt-gen/test_outputs/basic_only_signals.cpp
```diff
@@ -0,0 +1,41 @@
...
+MyObject::invokable()
+{
+  const std::lock_guard<std::mutex> guard(m_rustObjMutex);
+  m_rustObj->invokableWrapper(*this);
+}
+
+void
+MyObject::emitReady()
+{
+  runOnGUIThread([=]() { Q_EMIT ready(); });
```
ahayzen-kdab, line 26: We could instead use `QMetaObject::invokeMethod(this, "ready", Qt::QueueConnection)` which with possible changes in #12 could allow for removing `runOnGUIThread` etc if UpdateRequester also used this route.

ahayzen-kdab:
For signals we have the following choices.

- Use const ref for Qt types, this allows CXX / immediate to be able to directly call them from Rust, queued can take a unique_ptr then move it and create a ref
- Use by-value for Qt types, this means we use unique_ptr or value across CXX for both queued and immediate - but immediate must also have a wrapper.

Eg the Qt signals could be
```
void dataChanged(qint32 first, QVariant second);
void dataChanged(qint32 first, const QVariant& second);
```

The emit methods would be the following for both cases

```
void emitDataChanged(qint32 first, std::unique_ptr<QVariant> second);
```

Then either use `[second = std::move(second)]() { *second }` in the lambda or `[second = std::move(second)]() { &*second }`

cxx-qt-gen/src/gen_rs.rs
```diff
@@ -669,6 +730,79 @@ fn generate_property_methods_rs(obj: &QObject) -> Result<Vec<TokenStream>, Token
...
+                match signal {
+                    #(#queued_cases)*
+                }
+            }
+        });
+    }
+
+    if !immediate_cases.is_empty() {
+        signal_methods.push(quote! {
+            pub unsafe fn emit_immediate(&mut self, signal: Signal) {
```
ahayzen-kdab, line 795: If we use a `std::recursive_mutex` instead of a normal mutex to protect the RustObj, then can `unsafe` be removed here ? As then if a signal causes an invokable to be called from the same thread that is "fine" ? Or does this conflict with Rust's ownership/threading rules as it can't tell that halfway through invokable A we will jump to invokable B then back to A ?

https://en.cppreference.com/w/cpp/thread/recursive_mutex

LeonMatthesKDAB (approved):
LGTM

PR #68, opened 2022-01-26
Automatically convert between Rust and C++ types

### LINKED ISSUE #9: Automatically perform the to/from Rust type conversion inside the CppObjWrapper
Once we have #8 the user will always be using the CppObjWrapper. So then in the CppObjWrapper getter/setters etc change the types so that the parameters and return types are the "Rust" types and any conversion is performed inside the CppObjWrapper.

Eg instead of the user needing to change QVariant -> Variant. The CppObjWrapper can do the to_rust() conversion and return Variant.

ahayzen-kdab:
Closes #9

Change-Id: Ia840de3f5c19b9387f2914b4322a0df82e9627e3


ahayzen-kdab:
A non-UniquePtr variant is here #69

PR #69, opened 2022-01-28
Automatically convert between Rust and C++ types (second variant)

### LINKED ISSUE #9: Automatically perform the to/from Rust type conversion inside the CppObjWrapper

ahayzen-kdab:
Closes #9


ahayzen-kdab:
I was having issues making this build until i cleared out the `/target` folder. I think a recent cxx version has fixed a bug in the generation as it was causing a rustc compiler error. So we might need to figure out the version that was causing this or just pin to >= 1.0.63 or whichever version it is.

cxx-qt-gen/src/extract.rs
```diff
@@ -330,6 +338,11 @@ fn extract_type_ident(
     // Extract the Qt type this is used in C++ and Rust generation
     let qt_type = extract_qt_type(&idents, ty, cpp_namespace_prefix, qt_ident)?;
 
+    // Check if this Qt type is allowed to be a ref mut
+    if is_mut && is_ref && !qt_type.ref_mut_is_valid() {
+        return Err(ExtractTypeIdentError::InvalidType(ty.span()));
```
LeonMatthesKDAB, line 343: InvalidType here is a bit of a stretch. But I think we'll need to spend some time on better compiler errors anyway. So this is something for later.

LeonMatthesKDAB (approved):
LGTM

cxx-qt-gen/src/extract.rs
ahayzen-kdab: Right, i think we'll have a round of improving error messages later.

PR #70, opened 2022-01-31
Add qcolor using unique ptr

ahayzen-kdab:


ahayzen-kdab:
Note that parts of the extract/gen_rs/gen_cpp phases will be improved in #69 and some code has been moved into this pull request from there.

cxx-qt-lib/src/qcolor.rs
```diff
@@ -8,41 +8,41 @@
...
+        green: i32,
+        blue: i32,
+        alpha: i32,
     );
     #[link_name = "cxxqt1$qcolor$get$alpha"]
-    fn qcolor_get_alpha(this: &QColor) -> u16;
+    fn qcolor_get_alpha(this: &QColor) -> i32;
     #[link_name = "cxxqt1$qcolor$get$red"]
-    fn qcolor_get_red(this: &QColor) -> u16;
+    fn qcolor_get_red(this: &QColor) -> i32;
```
LeonMatthesKDAB, line 33: funny how in C++ we call this parameter `self`, because `this` is a reserved keyword, and in Rust we use `this`, as `self` is reserved :smile: .
Anyway, good name.

cxx-qt-lib/src/qcolor.rs
```diff
@@ -60,153 +60,175 @@ pub struct QColor {
...
+    #[link_name = "cxxqt1$unique_ptr$qcolor$get"]
+    fn unique_ptr_qcolor_get(this: *const MaybeUninit<*mut c_void>) -> *const QColor;
+    #[link_name = "cxxqt1$unique_ptr$qcolor$release"]
+    fn unique_ptr_qcolor_release(this: *mut MaybeUninit<*mut c_void>) -> *mut QColor;
+    #[link_name = "cxxqt1$unique_ptr$qcolor$drop"]
+    fn unique_ptr_qcolor_drop(this: *mut MaybeUninit<*mut c_void>);
+}
+
+// Safety: TODO
+unsafe impl UniquePtrTarget for QColor {
```
LeonMatthesKDAB, line 198: This impl, as well as the extern "C" block should probably be placed right after the QColor definition, not the Color definition. It might be very easy to miss that one "Q" and think this is actually implemented on "Color".

So either move this up, or move the QColor impl down.

cxx-qt-lib/src/qcolor.rs
```diff
@@ -8,41 +8,41 @@
 #![allow(improper_ctypes)]
```
LeonMatthesKDAB, line 8: This file feels much cleaner now after all these changes :+1:

cxx-qt-lib/src/qcolor.rs
```diff
@@ -60,153 +60,175 @@ pub struct QColor {
...
-        };
-    };
-}
-
 impl QColor {
     /// `QColor` is not constructible via `new`.
-    /// Instead, use the [`let_qcolor!`] macro.
+    /// Instead, use [`Color::from_rgba`].
+    //
+    // FIXME: Do we still need this?
```
LeonMatthesKDAB, line 67: I think the "new" function here can be removed.
You should never be able to call it, so it shouldn't exist.

cxx-qt-lib/src/qcolor.rs
```diff
@@ -60,153 +60,175 @@ pub struct QColor {
...
+    // Note that once map_qt_value is removed later, this can become private again
+    #[doc(hidden)]
+    pub inner: cxx::UniquePtr<QColor>,
+}
+
+impl Color {
+    /// Construct a Rust Color from an existing QColor, this is a copy operation.
+    pub fn from_qcolor(qcolor: &QColor) -> Self {
+        Self {
+            // Safety: TODO
```
LeonMatthesKDAB, line 102: For these `Safety: TODO` comments here, we can probably link to the design documents in the book that we're planning to add, once that is available.

LeonMatthesKDAB (changes requested):
I generally like the code a lot better with the opaque-unique-ptr approach. :+1: 
Good job!

There are 2 minor things I'd like to still see adressed:
1. remove QColor::new() - It shouldn't be called, so it shouldn't exist.
2. Clearly separate where the implementation of QColor (and its traits) starts vs. the Color impls

cxx-qt-gen/src/cxx_qt.cpp
```diff
@@ -351,17 +350,53 @@ extern "C"
...
-    switch (self.spec()) {
-      case QColor::Rgb:
-        return QColorSpec::Rgb;
-      default:
-        return QColorSpec::Unsupported;
-    }
+    self.setAlpha(alpha);
+  }
+
+  void cxxqt1$qcolor$set$red(QColor& self, int red) noexcept
```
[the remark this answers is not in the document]
ahayzen-kdab: Could you open an issue for this ? Especially the unique_ptr ones are just repeats I almost wondered about templates but the rust side needs different function names i think for the binding to work?

cxx-qt-lib/src/qcolor.rs
ahayzen-kdab: Done

cxx-qt-lib/src/qcolor.rs
ahayzen-kdab: Right :-) I felt like i was going to need to write the same statement here many times, linking to the book is a good idea later so lets do that.

cxx-qt-lib/src/qcolor.rs
ahayzen-kdab: Hah !

cxx-qt-lib/src/qcolor.rs
ahayzen-kdab: Right, and it will allow us to remove the Private struct in the lib.rs in the next pull request, done :-)

cxx-qt-gen/src/cxx_qt.cpp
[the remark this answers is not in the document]
LeonMatthesKDAB: https://github.com/KDAB/cxx-qt/issues/72
Done

LeonMatthesKDAB (approved):
LGTM

PR #71, opened 2022-01-31
 Use unique_ptr for all opaque types

ahayzen-kdab:


cxx-qt-gen/src/extract.rs
```diff
@@ -517,7 +517,14 @@ fn extract_invokable(
     let return_is_opaque = if let Some(return_type) = &return_type {
         matches!(
```
LeonMatthesKDAB, line 518: couldn't this use `return_type.qt_type.is_opaque()`?
This feels like code duplication of QtType::is_opaque() :thinking:

cxx-qt-gen/test_outputs/basic_only_invokable_return.cpp
```diff
@@ -24,14 +24,14 @@ QString
 MyObject::helloMessage(const QString& msg)
 {
   const std::lock_guard<std::mutex> guard(m_rustObjMutex);
-  return rustStringToQString(m_rustObj->helloMessage(msg));
+  return std::move(*m_rustObj->helloMessageWrapper(msg));
```
LeonMatthesKDAB, line 27: this feels a lot nicer :+1:

cxx-qt-lib/src/qstring.rs
```diff
@@ -42,60 +43,7 @@ pub struct QString {
     _pinned: PhantomData<PhantomPinned>,
 }
 
-// TODO: figure out how to make Syntax and Example compile as code
```
LeonMatthesKDAB, line 45: This whole file cleaned up nicely :)

cxx-qt-lib/src/qvariant.rs
```diff
@@ -99,167 +95,76 @@ pub struct QVariant {
...
-    type Id = type_id!("QVariant");
-    type Kind = cxx::kind::Opaque;
+    #[doc(hidden)]
+    unsafe fn __drop(mut repr: MaybeUninit<*mut c_void>) {
+        unique_ptr_qvariant_drop(&mut repr)
+    }
 }
 
-pub enum VariantImpl {
+pub enum VariantEnum {
```
LeonMatthesKDAB, line 167: I'd prefer this be called "VariantValue" or something similar.

cxx-qt-lib/src/qvariant.rs
```diff
@@ -7,19 +7,16 @@
...
-    mem::{self, ManuallyDrop, MaybeUninit},
-    ops::{Deref, DerefMut},
-    pin::Pin,
+    mem::MaybeUninit,
 };
 
 #[repr(u8)]
 #[allow(dead_code)]
-enum QVariantType {
+enum QVariantEnum {
```
LeonMatthesKDAB, line 19: I think we should stick with the "QVariantType" name here.
Having an enum named "Enum" doesn't really add value to the name

ahayzen-kdab: Done

cxx-qt-gen/src/extract.rs
ahayzen-kdab: Done

PR #81, opened 2022-02-17
Split tests and examples

ahayzen-kdab:
Make the real examples more obvious and split the unit test only ones into another folder


tests/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+#
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+add_subdirectory(basic_cxx_qt)
+add_subdirectory(basic_cxx_only)
+add_subdirectory(qt_types_standalone)
+add_subdirectory(update_requester_standalone)
+
+# Create helper method which adds relevent tests for the given acceptance test
+function(add_acceptance_tests EXAMPLE_NAME)
```
LeonMatthesKDAB, line 12: Change EXAMPLE_NAME to TEST_NAME here?

examples/basic_cxx_qt_qml_plugin/src/sub.rs
```diff
@@ -1,39 +0,0 @@
-// SPDX-FileCopyrightText: 2021 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
```
LeonMatthesKDAB, line 1: Why is this file not needed anymore?

LeonMatthesKDAB (commented):
Generally, the split is good.
Don't know why so much code was deleted though :thinking: 
Was that duplicated before somehow?

examples/basic_cxx_qt_qml_plugin/src/sub.rs
ahayzen-kdab: The `basic_cxx_qt_qml` and `basic_cxx_qt_qml_plugin` used to have exactly the same tests, it seems pointless to test the same thing twice? So i've dropped that duplication here as well.

LeonMatthesKDAB: Ah, alright, that explains it.

LeonMatthesKDAB (approved):
LGTM
any day we can remove over 1000 lines of code duplication is a good day :)

LeonMatthesKDAB (approved):

PR #85, opened 2022-02-24
book: allow book to build when linkcheck is missing

ahayzen-kdab:
Change-Id: I3a0be40c9f77d71ce11e33b2521ce607e470f041


ahayzen-kdab:
Installing linkcheck is preferred #86

PR #89, opened 2022-02-25
book: Getting Started Guide

LeonMatthesKDAB:
To make it easier for newcomers to CXX-Qt to follow along, add a Getting-Started Guide in the book.


book/src/getting-started/1-qobjects-in-rust.md
```diff
@@ -0,0 +1,77 @@
...
+> Don't bring a knife to a gun fight.
+
+There are many bits of advice like that.
+With CXX-Qt, we aim to make it possible to use the right tool for each of the many jobs necessary to build a modern GUI application.
+
+So what is in our toolbox for a typical Qt application?
+- QML - A declarative, flexible, dynamically-typed, interpreted language that is purpose built to define reactive and beautiful GUI layouts and widgets with quick iteration speed.
+- C++ - The traditional back-end of Qt - A fast, low-level language with a strong type system. C++ offers a rich ecosystem, many Qt-specific libraries and bare-metal performance. The cost when using C++ is that it is slow to develop, very error-prone and can easily lead to memory-issues, which can instantly crash your application and cause security issues.
+
+Notably absent then is a back-end language that allows us to get rid of the issues C++ has and provides us with a safe way to write fast Back-end code.
```
ahayzen-kdab, line 23: back-end vs Back-end in the same sentence.

book/src/getting-started/1-qobjects-in-rust.md
```diff
@@ -0,0 +1,77 @@
...
+This of course is where Rust comes in.
+Whilst Rust doesn't have quite as rich of an ecosystem, it is typically faster to develop than C++, with easy dependency management, and most importantly, safe memory access.
+Therefore it is an ideal candidate to replace C++ for writing the back-end business-logic code that feeds the GUI with data.
+
+However, both C++ as well as QML still have their place in Qt applications.
```
ahayzen-kdab, line 28: not sure having "both" and "as well as" for the same pair works ? It needs to be "both C++ and QML ... " or "C++ as well as QML ... "

book/src/getting-started/1-qobjects-in-rust.md
```diff
@@ -0,0 +1,77 @@
...
+For that reason Rust, C++ and QML should all be able to be used to complement each other.
```
ahayzen-kdab, line 29: Oxford comma ... "Rust, C++, and QML" ?

book/src/getting-started/1-qobjects-in-rust.md
```diff
@@ -0,0 +1,77 @@
...
+CXX-Qt aims to make it easy to integrate all three languages easily with each other, through the use of Qts [meta object system](https://doc.qt.io/qt-5/metaobjects.html).
+
+Qts design is inherently object-oriented, which is true both for C++ and QML as well.
```
ahayzen-kdab, line 32: Qt -> Qt's ?

book/src/getting-started/1-qobjects-in-rust.md
```diff
@@ -0,0 +1,77 @@
...
+Therefore, in order to integrate well with Qt, Rust needs to be able to extend the Qt object system with its own QObject subclasses and instances.
+This is exactly what CXX-Qt allows you to do.
+
+As Rust doesn't offer classes with inheritance and polymorphism, CXX-Qt uses a Rust module when defining new QObject subclasses.
+
+These CXX-Qt modules consist of multiple parts:
```
ahayzen-kdab, line 38: There are more parts later like the Signals enum :-) but this doesn't exist yet. And we automatically generate things in the macros like Property enum

ahayzen-kdab: (So this is probably fine for now)

book/src/getting-started/1-qobjects-in-rust.md
```diff
@@ -0,0 +1,77 @@
...
+- A `Data` struct
+    - Defines which Properties will be in the QObject subclass.
+    - Needs to implement the `Default` trait.
+    - This data will live in a C++ subclass that is generated from the CXX-Qt module
+- A `RustObj` struct
```
ahayzen-kdab, line 43: Also needs to implement Default trait.

book/src/getting-started/1-qobjects-in-rust.md
```diff
@@ -0,0 +1,77 @@
...
+    - A normal Rust struct
+    - One struct instance is created per class instance
+    - Contains any Rust-only data
+- The `impl` of the `RustObj` struct (optional):
+    - Contains any Rust code
+    - Functions marked with `#[invokable]` will be callable from QML and C++
```
ahayzen-kdab, line 49: This might change in the future, do we want to be specific of the macro name or do we want to say that you use attributes on functions to mark them as invokable and then link to another book page ? Not sure which is best, this also seems fine :-)

book/src/getting-started/1-qobjects-in-rust.md
```diff
@@ -0,0 +1,77 @@
...
+
+CXX-Qt will then expand this Rust module into two separate parts:
+1. A C++ subclass of QObject with the same name as the module
```
ahayzen-kdab, line 52: Even though you are listing two items, i sometimes visually prefer non numbered lists :thinking:

book/src/getting-started/1-qobjects-in-rust.md
```diff
@@ -0,0 +1,77 @@
...
+2. The Rust struct `RustObj`
+
+<div style="padding: 1rem; text-align: center;">
+
+![Overview of CXX-Qt module generation](../images/cxx_qt_module_generation.png)
```
ahayzen-kdab, line 57: I'd prefer not to have binary images without their source in the code (in case it needs to be adjusted later). I can recreate this image in something like Dia with a SVG output if that's useful ?

book/src/getting-started/1-qobjects-in-rust.md
```diff
@@ -0,0 +1,77 @@
...
+
+</div>
+
+CXX-Qt also generates the code needed for interaction of the C++ QObject subclass and the `RustObj` struct using the [CXX library](https://cxx.rs/).
+For more details, see the [Concepts: Bridge](../concepts/bridge.md) page.
+CXX-Qt also wraps some Qt types for us, so they can be easily used by the Rust side.
```
ahayzen-kdab, line 63: we have two sentences starting with "CXX-Qt also" next to each other. Maybe the first one can just be "CXX-Qt generates ..."

book/src/getting-started/1-qobjects-in-rust.md
```diff
@@ -0,0 +1,77 @@
...
+See the [Concepts: Qt Types](../concepts/types.md) page for the list of available types.
+
+The important take away here is the duality of any subclass generated by CXX-Qt.
```
ahayzen-kdab, line 66: Do we need to note (or have you already?) that the properties defined on the Data struct actually end up as members on the C++ QObject and are owned there? Then RustObj needs to access the properties via the CppObj (this part can be mentioned later).

book/src/getting-started/1-qobjects-in-rust.md
ahayzen-kdab: Also the file is missing it's equivalent `.license` file with licensing information (hence why reuse is failing in CI).

book/src/getting-started/1-qobjects-in-rust.md
LeonMatthesKDAB: Ah, stupid german brain capitalizing all the Nouns ;)

book/src/getting-started/1-qobjects-in-rust.md
LeonMatthesKDAB: I'll remove the both :+1:

book/src/getting-started/1-qobjects-in-rust.md
LeonMatthesKDAB: Done

book/src/getting-started/1-qobjects-in-rust.md
```diff
@@ -0,0 +1,77 @@
...
+- C++ - The traditional back-end of Qt - A fast, low-level language with a strong type system. C++ offers a rich ecosystem, many Qt-specific libraries and bare-metal performance. The cost when using C++ is that it is slow to develop, very error-prone and can easily lead to memory-issues, which can instantly crash your application and cause security issues.
+
+Notably absent then is a back-end language that allows us to get rid of the issues C++ has and provides us with a safe way to write fast Back-end code.
+This of course is where Rust comes in.
+Whilst Rust doesn't have quite as rich of an ecosystem, it is typically faster to develop than C++, with easy dependency management, and most importantly, safe memory access.
+Therefore it is an ideal candidate to replace C++ for writing the back-end business-logic code that feeds the GUI with data.
+
+However, both C++ as well as QML still have their place in Qt applications.
+For that reason Rust, C++ and QML should all be able to be used to complement each other.
+CXX-Qt aims to make it easy to integrate all three languages easily with each other, through the use of Qts [meta object system](https://doc.qt.io/qt-5/metaobjects.html).
```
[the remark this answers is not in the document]
LeonMatthesKDAB: Done

book/src/getting-started/1-qobjects-in-rust.md
LeonMatthesKDAB: Done

book/src/getting-started/1-qobjects-in-rust.md
LeonMatthesKDAB: :+1: I'll add a commented note to add the Signals enum once we've added them.

I don't think the Properties enum is something we should add here.
It isn't something the user needs to manually add, which is what I want to describe in this section.

book/src/getting-started/1-qobjects-in-rust.md
LeonMatthesKDAB: Ah, right, Done

book/src/getting-started/1-qobjects-in-rust.md
LeonMatthesKDAB: That does indeed look better :+1:

book/src/getting-started/1-qobjects-in-rust.md
LeonMatthesKDAB: Changed the second one to:
"Additionally, CXX-Qt wraps ..."

book/src/getting-started/1-qobjects-in-rust.md
LeonMatthesKDAB: Hm, then I didn't make that clear enough, will add a sentence about it in this paragraph.

book/src/getting-started/1-qobjects-in-rust.md
LeonMatthesKDAB: I think we can keep it this way.
If we want to rename something like this we'll need to make sure our docs match up.

book/src/getting-started/1-qobjects-in-rust.md
LeonMatthesKDAB: A Dia file would be good :+1:

ahayzen-kdab: As discussed this might reuse the existing one.

LeonMatthesKDAB: Changed it to reuse the existing diagram

book/src/getting-started/2-our-first-cxx-qt-module.md
```diff
@@ -0,0 +1,85 @@
...
+As with all things Rust, we'll first want to create a cargo project.
+```bash
+cargo new --lib cxx-qt-getting-started
+```
+
+As outlined in the previous section, to define a new QObject subclass, we'll need to create a Rust module.
+So let's go into the `src/lib.rs` file.
+We'll modify this file until it looks like this:
+
+```rust,noplayground
```
ahayzen-kdab, line 19: should these have `ignore` as well ?

book/src/getting-started/2-our-first-cxx-qt-module.md
```diff
@@ -0,0 +1,85 @@
...
+{{#include ../../../examples/qml_minimal/src/lib.rs:book_data_struct}}
+```
+That means the newly created QObject subclass will have two properties as members: `number` and `string`. For names that contain multiple words, like `my_number`, CXX-Qt will again perform the snake_case to camelCase conversion to fit with C++/QML naming conventions.
+
+Note that the data types we use here are normal Rust data types.
+CXX-Qt will automatically convert these types to their C++/Qt equivalent.
+In our case that means:
+- `number: i32` -> `int number`
+- `string: String` -> `QString string`\
+For more details on the available types, see the [Qt types page](../concepts/types.md).
```
ahayzen-kdab, line 45: the formatting looks odd here. Think you need a blank line at the end of the list ?

Also wonder if there is any difference between these two (note i normally indent the bullet point as well.
```
- thing
- thing
```
vs
```
  * thing
  * thing
```

LeonMatthesKDAB: Yup, the trailing backslash is intentional here.
It makes the "For more details" line indented on the same level as the enumeration.

I think it looks okay, because it saves us from adding another entire paragraph.

Was considering using footnotes as well, but thought this was an okay compromise.

ahayzen-kdab: OK lets keep it like this.

book/src/getting-started/2-our-first-cxx-qt-module.md
LeonMatthesKDAB: Yup, all changed.

book/src/getting-started/3-exposing-to-qml.md
```diff
@@ -0,0 +1,47 @@
...
+For every QObject subclass that is defined in Rust, CXX-Qt will generate a corresponding C++ class.
+This class is included by the first code snippet.
+They will always be in the `cxx-qt-gen/include/` include path and use the snake_case naming convention.
+
+The second code snippet then exports the class to QML.
+This works the same as it would for any other QObject subclass, as that is exactly what `MyObject` is, as far as Qt is concerned.
+The only thing to note here is that the class is generated in the `cxx_qt::my_object` namespace.
+Where `my_object` is the name of the Rust module we defined earlier.
+
+As we want to include the `main.qml` file inside the [Qt resource system](https://doc.qt.io/qt-5/resources.html), we'll have to add a `qml.qrc` file as well:
```
ahayzen-kdab, line 38: is it obvious that the qml.qrc is in the src (?) directory?

book/src/getting-started/3-exposing-to-qml.md
```diff
@@ -0,0 +1,47 @@
...
+```
+<RCC version="1.0">
+    <qresource prefix="/">
+        <file>main.qml</file>
+    </qresource>
+</RCC>
+```
+
+And that's it already. We can now [use our cool new class from QML](./4-qml-gui.md).
```
ahayzen-kdab, line 47: Not sure "already" is the perfect word :thinking:  maybe just saying "And that's it, we can now ..."

book/src/getting-started/4-qml-gui.md
```diff
@@ -0,0 +1,36 @@
...
+As noted in the [QObjects in Rust](./1-qobjects-in-rust.md) chapter, we always want to use "the right tool for the right job".
+For a small modern GUI in Qt, that definitely means using QML.
+It's powerful, flexible, declarative, and allows us to iterate very quickly.
+
+So let's add a `main.qml` next to our two other files in the `src` folder:
+```qml,ignore
+{{#include ../../../examples/qml_minimal/src/main.qml:book_main_qml}}
+```
+
+If you're not familiar with QML, I recommend you take a look at the [Qt QML intro](https://doc.qt.io/qt-5/qmlapplications.html).
```
ahayzen-kdab, line 19: Wonder if the perspective should be "I" or "we"  ?

book/src/getting-started/5-cmake-integration.md
```diff
@@ -0,0 +1,153 @@
...
+- Add `clang-format` and `cxx-qt-build` as build-dependencies.
+
+In the end, your `Cargo.toml` should look like this:
+```toml,ignore
+[package]
+name = "cxx-qt-getting-started"
+version = "0.1.0"
+edition = "2021"
+
+# Don't touch this, the CMake file expects a static lib with a specific name
```
ahayzen-kdab, line 41: obviously they don't have to put this comment, this was more a note for ourselves, but maybe still useful to have :-)

book/src/getting-started/5-cmake-integration.md
```diff
@@ -0,0 +1,153 @@
...
+[build-dependencies]
+clang-format = "0.1"
+cxx-qt-build = "0.2"
+```
+
+We'll then also need to add a script named `build.rs` next to our `Cargo.toml`:
+```rust,ignore
+{{#include ../../../examples/qml_minimal/build.rs:book_build_rs}}
+```
+This is what generates the C++ code for our `MyObject` class at compile-time.
```
ahayzen-kdab, line 60: It's probably important to note that you need to specify all the files in which you specify the macro, otherwise it won't work. I've had it a few times where i've forgotten to add a new file to the list and wondered why things error oddly :-)

book/src/getting-started/5-cmake-integration.md
```diff
@@ -0,0 +1,153 @@
...
+cxx_qt_include(${APP_NAME})
+
+# Link to generated rust library
+cxx_qt_link_rustlib(${APP_NAME})
+```
+
+Which will do the code generation and include it into the C++ build.
+
+An important thing to note here is that CMake must be able to resolve the call to `include(CxxQt)`.
+For this to work, you'll want to clone the [CXX-Qt repository](https://github.com/KDAB/cxx-qt/) and add the `CxxQt.cmake` file to the `CMAKE_MODULE_PATH` CMake variable.
```
ahayzen-kdab, line 132: Do we want to note that something like `list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/../../cmake")` is a way of doing that? (this is what we do in the existing CMakeLists.txt to add it to the module path)/

book/src/getting-started/3-exposing-to-qml.md
LeonMatthesKDAB: Yup, that section used to be shorter before I remembered to add the qml.qrc file ^^

book/src/getting-started/5-cmake-integration.md
LeonMatthesKDAB: Done

book/src/getting-started/5-cmake-integration.md
LeonMatthesKDAB: Done

book/src/getting-started/5-cmake-integration.md
LeonMatthesKDAB: I think this is something for the `concepts` chapter to take care of.
Here I just want to show the "easiest" option.
For something unfamiliar with CMake, I think this should be easy enough.
If we add this info to the dedicated CMake section , we can link to it from here.

ahayzen-kdab: We do have the `/concepts/cmake.html` section now, so lets add it there...

ahayzen-kdab:
LGTM :-)

ahayzen-kdab (approved):

PR #93, opened 2022-02-25
book: refresh threading_abstract image so it include white line

ahayzen-kdab:
Change-Id: I41d72b7c78f6627930d09aba98fdea37de806249


LeonMatthesKDAB (approved):
LGTM.
Funny how some of the floats changed a tiny bit. Probably due to non-deterministic floating point math somewhere :sweat_smile:

PR #95, opened 2022-02-28
cargo: add more metadata to the cxx-qt crate

ahayzen-kdab:
Change-Id: If0c2bc3de31c71b739e0f708217bc3c63397e974


LeonMatthesKDAB (approved):
I agree with the version bump. Enough minor changes have accumulated by now.

PR #97, opened 2022-02-28
cxx-qt-gen: only generate Property enum when there are properties

### LINKED ISSUE #96: Trying to build a object with no Data properties fails
We fail to generate a C++ from Rust File with the error `explicit #[repr(...)] is required for enum without any variants`, is this due to the Property enum being empty in the CXX bridge?

Solutions are one of the following
  *  Add something like #[repr(i32)] to the enum (lose CXX automatic type choosing)
  * Remove enum Property when there aren't any
  * Inject fake value such as  "LastProperty" or "NoProperties"

ahayzen-kdab:
Closes #96

Change-Id: I30e46ed2b4fb4cefe30f733e78ca0c3bf2c07be4

### DIFF
--- a/cxx-qt-gen/src/gen_rs.rs
+++ b/cxx-qt-gen/src/gen_rs.rs
@@ -444,6 +444,15 @@ pub fn generate_qobject_cxx(
     // Build the namespace string, rust::module
     let namespace = obj.namespace.join("::");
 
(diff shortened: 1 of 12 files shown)


LeonMatthesKDAB (approved):
LGTM

PR #105, opened 2022-03-07
cxx-qt-lib: use cxx for bridging Qt types

### LINKED ISSUE #104: Use CXX itself for bridging cxx-qt-lib types instead of implement UniquePtrTarget
Now that we use UniquePtr<T> to move opaque C++ items around the bridge, we can implement their support (and trivial types) via CXX itself with custom build scripts.

This then allows us to

- remove lots unsafe code
- use CXX for defining the Qt API that we expose

To do this we need to change the build process to use the following

- cxx-qt-lib is built as a normal Rust library, but any cxx bridge files have .h / .cpp files generated for them and defined in an env var passed downstrream
- cxx-qt-build reads the env var and adds the files to the C++ sources to be compiled
- cxx-qt-gen changes the bridge definition to include the generic C++ sources

What does this look like?

For the cxx-qt-gen Rust sources they need the bridge to look like the following
```rust
#[cxx::bridge]
mod ffi {
    unsafe extern "C++" {
        ...
        include!("cxx-qt-lib/include/qt_wrapper.h");
        type QColor = cxx_qt_lib::QColorCpp;
        type QPoint = cxx_qt_lib::QPoint;

        fn cpp_test(color: &QColor, point: &QPoint);
    }

    extern "Rust" {
        fn test_color_invokable_wrapper(color: &QColor) -> UniquePtr<QColor>;
        fn test_point_invokable(point: &QPoint) -> QPoint;
    }
}
```

For trivial types in cxx-qt-lib this can look like the following
```rust
#[derive(Debug, Clone)]
#[repr(C)]
pub struct QPoint {
    x: i32,
    y: i32,
}

impl QPoint {
    pub fn new(x: i32, y: i32) -> Self {
        Self { x, y }
    }
}

unsafe impl ExternType for QPoint {
    type Id = type_id!("QPoint");
    type Kind = cxx::kind::Trivial;
}

#[cxx::bridge]
mod ffi {
    unsafe extern "C++" {
        include!("cxx-qt-lib/include/qt_wrapper.h");

        type QPoint = super::QPoint;

        // Expose any Qt API we want here

        fn x(self: &QPoint) -> i32;
        fn y(self: &QPoint) -> i32;

        #[rust_name = "set_x"]
        fn setX(self: &mut QPoint, x: i32);
        #[rust_name = "set_y"]
        fn setY(self: &mut QPoint, y: i32);
    }
}
```

For opaque types in cxx-qt-lib this can look like the following
```rust
#[cxx::bridge]
mod ffi {
    unsafe extern "C++" {
        include!("cxx-qt-lib/include/qt_wrapper.h");

        type QColor;

        fn red(self: &QColor) -> i32;

        #[namespace = "cxx_qt_lib::qcolor"]
        fn new_qcolor() -> UniquePtr<QColor>;
        #[namespace = "cxx_qt_lib::qcolor"]
        fn copy_qcolor(color: &QColor) -> UniquePtr<QColor>;
    }

    impl UniquePtr<QColor> {}
}

pub type QColorCpp = ffi::QColor;

pub struct QColor {
    inner: cxx::UniquePtr<QColorCpp>,
}

impl QColor {
    pub fn new() -> Self {
        Self::from(ffi::new_qcolor())
    }

    // copy from a reference
    pub fn from_qcolor(color: &QColorCpp) -> Self {
        Self {
            inner: ffi::copy_qcolor(color),
        }
    }

    // to/from a UniquePtr
    pub fn from(color: cxx::UniquePtr<QColorCpp>) -> Self {
        Self { inner: color }
    }

    pub fn to_unique_ptr(self) -> cxx::UniquePtr<QColorCpp> {
        self.inner
    }

    // Wrap any Qt API that we want
    pub fn red(&self) -> i32 {
        if let Some(inner) = self.inner.as_ref() {
            inner.red()
        } else {
            0
        }
    }
}
```

Then qt_wrapper.h then looks like the following with .cpp defining the methods
```cpp
#pragma once
#include <memory>

#include <QColor>
#include <QPoint>

namespace cxx_qt_lib {
namespace qcolor {

std::unique_ptr<QColor>
new_qcolor();
std::unique_ptr<QColor>
copy_qcolor(const QColor& color);

}
}
```

ahayzen-kdab:
Closes #104

Change-Id: Idd5f3ba71774bd836434cdbf6b8fd49f3b17ce1b


cxx-qt-lib/build.rs
```diff
@@ -0,0 +1,62 @@
...
+    .expect("Could not copy qt_types.h");
+    std::fs::copy(
+        format!("{}/{}", dir_manifest, "src/qt_types.cpp"),
+        format!("{}/qt_types.cpp", src_path),
+    )
+    .expect("Could not copy qt_types.cpp");
+
+    // Expose target directory to downstream crates
+    // https://github.com/rust-lang/cargo/issues/3544
+    println!("cargo:target_dir={}", dir_out);
```
ahayzen-kdab, line 61: is there any other way of doing this ? this is the only weird part of this change :-)

cxx-qt-lib/build.rs
```diff
@@ -0,0 +1,64 @@
...
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use quote::ToTokens;
+use std::io::Write;
+
+fn main() {
+    let opt = cxx_gen::Opt::default();
+
+    // Read the cargo folder and out folder
+    let dir_manifest = std::env::var("CARGO_MANIFEST_DIR").expect("Could not get manifest dir");
```
Be-ing, line 13:
```suggestion
    let mut dir_manifest = env::var("CARGO_MANIFEST_DIR").expect("Could not get manifest dir");
    if cfg!(windows) {
        dir_manifest = dir_manifest.replace('\\', "/");
    }
```
The `format!` macros use `/` separators but `CARGO_MANIFEST_DIR` uses `\` on Windows.

ahayzen-kdab: Ugh :-) @Be-ing Maybe split out your commit in #107 that has the manifest_dir() method into a separate pull request. Then we can land that first and rebase other things on top as they'll take longer to review :-)

cxx-qt-lib/build.rs
```diff
@@ -0,0 +1,62 @@
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use quote::ToTokens;
+use std::io::Write;
+
+fn main() {
```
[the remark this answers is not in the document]
ahayzen-kdab: Instead we now use `include_str` and then expose this as a const to cxx-qt-build :-)

cxx-qt-lib/build.rs
ahayzen-kdab: Instead we now use include_str and then expose this as a const to cxx-qt-build :-)

cxx-qt-lib/build.rs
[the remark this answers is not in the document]
ahayzen-kdab: Oops commented on the wrong one, this one is still valid.

cxx-qt-lib/build.rs
Be-ing: #108

ahayzen-kdab: @Be-ing if the CARGO_MANIFEST_DIR is never used by CMake and only used by Rust's `std::fs::File::Create(path)` etc, do those need to have the \ to / flip as well ?

cxx-qt-lib/src/lib.rs
```diff
@@ -49,6 +49,66 @@ pub use qvariant::{QVariant, Variant, VariantValue};
 mod map_qt_value;
 pub use map_qt_value::*;
 
+pub const QT_TYPES_HEADER: &str = include_str!("../include/qt_types.h");
+pub const QT_TYPES_SOURCE: &str = include_str!("qt_types.cpp");
+
+pub const QCOLOR_CXX_HEADER: &str =
+    include_str!(concat!(env!("OUT_DIR"), "/cxx-qt-lib/include/qcolor_cxx.h"));
```
ahayzen-kdab, line 56: Hopefully Windows is able to pick up these correctly as well, even if there is a mix of \ and /. Because we are only on the Rust side and not CMake will this be OK ? @Be-ing

cxx-qt-lib/build.rs
Be-ing: Good question. I don't know. I wouldn't assume that the Rust standard library's Path works with mixed \ and / separators though.

cxx-qt-lib/src/lib.rs
ahayzen-kdab: Seems this isn't, so we should use cfg to set the right path depending on the platform. https://github.com/rust-lang/rust/issues/75075 https://doc.rust-lang.org/stable/core/macro.include_str.html

Be-ing: It sure would be nice if there was a way to tell Cargo to replace \ with / in all its environment variables that represent filesystem paths.

cxx-qt-lib/build.rs
ahayzen-kdab: Done for now.

cxx-qt-lib/src/lib.rs
```diff
@@ -49,71 +49,262 @@ pub use qvariant::{QVariant, Variant, VariantValue};
...
+        "/"
+    };
+}
+
+#[cfg(windows)]
+macro_rules! sep {
+    () => {
+        "\\"
+    };
+}
```
Be-ing, line 68: Macros seem overcomplicated here. Can't you use [std](https://doc.rust-lang.org/stable/std/index.html)::[path](https://doc.rust-lang.org/stable/std/path/index.html)::[MAIN_SEPARATOR](https://doc.rust-lang.org/stable/std/path/constant.MAIN_SEPARATOR.html#)? Or failing that, conditionally define a static &str like https://github.com/RustAudio/rust-jack/blob/main/jack-sys/src/lib.rs#L6

Be-ing: Me, but for Windows \ path separators
https://xkcd.com/567/

ahayzen-kdab: Only string literals can be used with `concat!` and `include_str!` so `MAIN_SEPARATOR` cannot be used. The macro could maybe be improved so we do't need to put `sep!` constantly, i'll look at this later.

cxx-qt-lib/build.rs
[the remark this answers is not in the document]
ahayzen-kdab: Done

book/src/internal/build.md
```diff
@@ -0,0 +1,87 @@
...
+  1. CMake build is started, and triggers cargo build
+  2. CXX-Qt build generates C++ sources and static library
+  3. CMake builds the project C++ sources combined with generated sources
+  4. CMake links the target to the Rust static library and Qt
+
+## 1. Cargo build
+
+When cargo build is triggered, this causes dependencies to be downloaded and built. One of these is
+`cxx-qt-lib` which defines helper Qt types. It contains a `build.rs` which uses CXX to generate C++
+sources for the Qt types. These are stored into the `OUT_DIR` of the crate.
```
LeonMatthesKDAB, line 21:
``` md
These are stored into the `OUT_DIR` of the crate. 
```
to:
``` md
These are stored into a single JSON file in the OUT_DIR of the crate.
```

and

``` md
Therefore as `cxx-qt-lib` is built it uses `include_str!`
to expose the generates files as `const` strings in the crate.
```
to:
``` md
Therefore as `cxx-qt-lib` is built it uses `include_str!`
to expose the generated JSON into the QT_TYPES_CXX_JSON variable.
```
I think its helpful to name the variable here, as that makes it a lot easier to find the code where this happens.

Also, we still use "strings" plural in the CXX-Qt build section.

LeonMatthesKDAB (commented):
Looks good, great to see the large number of include_str! macros gone :)
Just update the book a bit, then we're good to merge.

book/src/internal/build.md
ahayzen-kdab: Fixed!

LeonMatthesKDAB (approved):
LGTM

PR #117, opened 2022-03-24
extract: passthrough all unknown types

### LINKED ISSUE #80: Support more types in the module
If you try to put a `const VAR: T = V` inside our macro module then it fails with unsupported type. We should pass these through like we do with `use` and `enum` etc.

ahayzen-kdab:
Closes #80


LeonMatthesKDAB (approved):
LGTM

PR #120, opened 2022-03-25
cxxqobject: remove CXXQObject and UpdateRequester

### LINKED ISSUE #12: Move CxxQObject to be a member rather than inherited

ahayzen-kdab:
Instead use invokeMethod with QueueConnections and remove remaining extern blocks.

Closes #12


cxx-qt-lib/src/lib.rs
```diff
@@ -67,244 +67,50 @@ macro_rules! sep {
     };
 }
 
+macro_rules! include_cxx_file {
```
Be-ing, line 70: It would help to have documentation for this macro regarding using `sep!` for path separators.

ahayzen-kdab: This macro has been removed now.

LeonMatthesKDAB (approved):
LGTM

PR #122, opened 2022-04-06
try enabling tests on Windows

Be-ing:
Maybe #105 fixed the link errors?

### DIFF
--- a/.github/workflows/github-cxx-qt-tests.yml
+++ b/.github/workflows/github-cxx-qt-tests.yml
@@ -72,7 +72,7 @@ jobs:
             os: windows-2022
             qt_version: 5
             # FIXME: many tests fail to link
-            ctest_args: --exclude-regex '^(cxx_qt_gen_cargo_tests|example_qml.*|qml.*tests|test.*|reuse_lint|cpp_clang_format|.*valgrind)$'
+            ctest_args: --exclude-regex '^(reuse_lint|cpp_clang_format|.*valgrind)$'
             exe_suffix: .exe
             qt_qpa_platform: windows
             compiler_cache_path: C:\Users\runneradmin\AppData\Local\Mozilla\sccache\cache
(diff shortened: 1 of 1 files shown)


Be-ing:
Well that didn't work. I still don't know what's going wrong with those link errors.

LeonMatthesKDAB:
Just a random thought: We're in some places defining `extern "C"` functions that take references as arguments.

GCC has been warning about this for a while, as C doesn't have references, and we'd technically have to use pointers.
Maybe this is related somehow?

Be-ing:
I don't see any compile warnings from GCC using GCC 11.2.1 locally on Fedora 35 nor on the Ubuntu CI.

ahayzen-kdab:
IIRC the warnings were actually from clang and hence why we used to have this warning disable https://github.com/KDAB/cxx-qt/blob/3828f36202ee2edba3e7cf418449e444ac6c61ba/cxx-qt-gen/src/cxx_qt.cpp#L42

The Windows issues appear to be maybe linker related but it'll probably be more obvious on an actual machine than trying to fight CI, we just need to spend some time investigating what is missing/wrong.

ahayzen-kdab:
> Can this be closed? We have Windows CI now, don't we?

We don't there are problems ... 
- QML tests aren't being run
- Cargo tests aren't being run
- For me the examples crash on exit but not for Be (maybe a MSVC 2019 vs 2022 thing or something has fixed it)

Be-ing:
We still need to figure out what to do about adding Qt to PATH to fix the DLL not found errors when running the tests... or figure out some hack around #224.

Be-ing:
However, the link errors with the tests have been fixed since overhauling the build system to compile the generated C++ with Cargo. The remaining issues are setting the required environment variables at runtime.

ahayzen-kdab:
@Be-ing this can be closed now as there are working Windows tests? :tada:

Be-ing:
Yes superseded  by #500

Issue #53, opened 2022-01-07
Ensure that all Qt types have a correct default, traits, and similar methods


ahayzen-kdab:
Ensure that our wrapped Qt types have defaults that are the same as Qt and relevant methods (eg getters and setters).

- [x] Default (potentially using the constructor on the C++ side)
- [x] Display + Debug (see #131 )
- [x] Ord, Eq, Partials etc #413  
- [x] Add support for operators to types (+ - * / eg for QPoint) - #448 
- [x] AsRef

LeonMatthesKDAB:
We should also check for which types we can make the members public. QPoint comes to mind. I think in the data-driven design of Rust this might be more idiomatic.

ahayzen-kdab:
For QUrl we have `from_str` which needs to be `Result<T, E>` to match the std trait. At the moment we use `Infallible` but this could use `QUrl::isValid()` to return an error if the url is not valid!

ahayzen-kdab:
Consider if we want by-ref or by-value in various parts of our API.

- Invokable parameters are by ref
- Invokable returns are by value
- Property setters are by ref for non-opaque and by value for opaque
- Property getters are by ref for non-opaque and by value for opaque
- Most of our methods on types take by ref

Related to #58 as well

ahayzen-kdab:
Also ensure that we have `Display` and `Debug` for types that make sense.

ahayzen-kdab:
Related to #55

ahayzen:
For Ord and Eq we should be able to make a single C++ template like `bool operatorEq(const T& a, const T& b)` and reuse the single C++ template for all types like we have with construct and drop etc. (Assuming we can't bind with CXX to the operator of the type)

ahayzen-kdab:
@vimpostor did you want to continue this issue?

Things that are likely left
- Check relevant types have Default (think most do?)
- Add PartialOrd/Ord to relevant types (again template likely will work well)
- We added `impl AsRef<[u8]>` to QByteArray, to assist with using it in general Rust APIs would this, or similar, be useful on any other types?
- Consider any other `std` traits that could be good ?

Member functions is split out into #55 and operators (/ * - +) are in that issue as well, could be a good one after this one.

vimpostor:
Yep, I will likely add `PartialOrd` next ;)

ahayzen-kdab:
[...]

Issue #16, opened 2021-12-08
Consider how (de)serialise can work with custom types


ahayzen-kdab:
For types such as QPoint consider how the user can use these with serde on the Data struct.

Options appear to be

- Implement derive of serde always
- Behind a feature implement serde derive helpers
- Don't implement and force the user to impl the (de)seralise trait for the struct

ahayzen-kdab:
`#[cfg_attr(feature = "serde_support", derive(Serialize, Deserialize))]` allows for conditional derive in cxx-qt-lib once #105 is done this should be simple. https://doc.rust-lang.org/reference/conditional-compilation.html

ahayzen-kdab:
Moving into #292

PR #131, opened 2022-05-08
cxx-qt-lib: Display and Debug for QString

Ayush1325:
Implement Display and Debug for QString. 

I think at least Debug is a fairly common thing that should be implemented for all the types in cxx-qt-lib.

### DIFF
--- a/cxx-qt-lib/src/types/qstring.rs
+++ b/cxx-qt-lib/src/types/qstring.rs
@@ -4,6 +4,8 @@
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
+use std::fmt::Display;
+
 #[cxx::bridge]
 mod ffi {
     unsafe extern "C++" {
@@ -60,5 +62,18 @@ impl crate::ToUniquePtr for &str {
 impl From<&QStringCpp> for String {
     fn from(qstring: &QStringCpp) -> Self {
         qstring.to_rust()
+    }
+}
+
+impl Display for QStringCpp {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let s: String = self.into();
+        write!(f, "{}", s)
+    }
+}
+
+impl std::fmt::Debug for QStringCpp {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self)
     }
 }


ahayzen-kdab:
Thanks for pushing this change!

Yes I think we should implement Display / Debug for our Qt types, we have an issue to add Default #53 and Serde (Serialize/Deserialize) #16. I have added Display / Debug to the #53 issue so we can track it.

For QString itself this is an interesting one, as since #69 we now transparently wrap opaque types (note this is not in a tagged release yet). What this means is that for the opaque types I don't think the developer will ever need to know that `QStringCpp` exists? They will only be using `String` ? As since that change `String` is used as the return type and `&str` can be used as the input type.

So I'm not sure that this is useful for `QStringCpp` or the other `*Cpp` types? I think it'd be more useful for the type used by the developer - eg QPoint or QColor (not QColorCpp) etc (unless i'm forgetting some use case :-) )

Ayush1325:
> Thanks for pushing this change!
> 
> Yes I think we should implement Display / Debug for our Qt types, we have an issue to add Default #53 and Serde (Serialize/Deserialize) #16. I have added Display / Debug to the #53 issue so we can track it.
> 
> For QString itself this is an interesting one, as since #69 we now transparently wrap opaque types (note this is not in a tagged release yet). What this means is that for the opaque types I don't think the developer will ever need to know that `QStringCpp` exists? They will only be using `String` ? As since that change `String` is used as the return type and `&str` can be used as the input type.
> 
> So I'm not sure that this is useful for `QStringCpp` or the other `*Cpp` types? I think it'd be more useful for the type used by the developer - eg QPoint or QColor (not QColorCpp) etc (unless i'm forgetting some use case :-) )

Well, in the example in the book, the invocable function takes `&QString` as input which is then converted to string. That was why I decided to implement `Debug` and `Display` for it. However, the examples in the /examples folder have `&str` inputs for such parameters so I guess this is no longer needed.

ahayzen-kdab:
> Well, in the example in the book, the invocable function takes `&QString` as input which is then converted to string.

Right, the book shows the last the tagged version of the code (eg what a developer would be using from crates.io), unfortunately this area has already changed in `main` so when the next version is tagged (eg 0.3) then the book will show how `&QString` isn't used any more.

I still think these changes are valid for other types though, eg QPoint, QColor etc as they don't use a std Rust type.

Ayush1325:
I am thinking of closing the pull request now since this doesn't seem required.

PR #133, opened 2022-05-23
ci: use release builds for tests

ahayzen-kdab:
This ensures that we pick up any failures from code being
optimised out inside Q_ASSERT etc.

Change-Id: Ie83484c98790c8db944b692650cdb9c6fb2d1d26


Be-ing:
Hmm, is this worth the extra CI build time?

ahayzen-kdab:
Ideally we somehow need to pick up errors like we have seen in #132 :-) I wonder with caching how much extra time it'll add ...

Be-ing (approved):
With caching this branch builds in about 25 minutes on CI, which is (approximately) the same as the master branch.

PR #134, opened 2022-05-24
GH Actions: update sccache to 0.3.0

Be-ing:
which was released a few weeks ago 04-05-2022


ahayzen-kdab:
Looks good, just needs a rebase.

Assume there isn't any way of easily automating this or being notified when we need to bump these?

Be-ing:
I recommend to subscribe to GitHub release notifications for every dependency you use:
image

ahayzen-kdab (approved):

PR #135, opened 2022-05-25
Demo threading visuals

ahayzen-kdab:


examples/demo_threading/src/lib.rs
```diff
@@ -164,6 +176,9 @@ mod energy_usage {
                 qt_rx,
                 qt_tx,
                 join_handles: None,
+                sensors: Arc::new(Mutex::new(Arc::new(
```
Be-ing, line 179: Why an Arc *inside* the Mutex instead of only outside?

ahayzen-kdab: And if anything it's better than COW as if nothing is holding another Arc then it doesn't need to clone the data :tada:

ahayzen-kdab:
Note to self once we have the UI it might be nice to split the folders into the following similar to the extension plugin example. However note that the CMake will require some changes as we don't have a way to specify the Rust cargo manifest path yet.
```
src:/
  core/  <- Rust
  helpers/  <- C++
  ui/ <- QML
  main.cpp
```

PR #144, opened 2022-06-26
Use OUT_DIR to compute target dir, instead of  using MANIFEST_DIR

### LINKED ISSUE #139: Failed to build when using workspace or changing rust outdir
The rust library will fail to build using workspace with the following error:
```
--- stderr
thread 'main' panicked at 'Could not create cpp_namespace_prefix file: Os { code: 2, kind: NotFound, message: "No such file or directory" }', .../cxx-qt-build-0.2.1/src/lib.rs:300:40
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
I found that you used the following function to calculate the output of CPP generation directory
```Rust
fn manifest_dir() -> String {
    let mut manifest_dir = env::var("CARGO_MANIFEST_DIR").expect("Could not get manifest dir");
    // CARGO_MANIFEST_DIR uses \ path separators on Windows, but the format! macros that
    // use the return value of this function use / separators. CMake's `file(STRINGS)` command
    // which is used to load list of generated file paths fails if \ and / path separators
    // are mixed. CMake also fails when using all \ separators because it treats them as escaping
    // the character after the \
    if cfg!(windows) {
        manifest_dir = manifest_dir.replace('\\', "/");
    }
    manifest_dir
}
```

It will fail if  developers change the rust library build directory. I think the best way to calculate the output directory is to use `env::var("OUT_DIR")`, such as: 
```Rust
fn get_output_dir() -> String {
    let mut out_dir = std::env::var("OUT_DIR").expect("Could not get out dir");
    out_dir += "/../../../..";
    if cfg!(windows) {
        out_dir = out_dir.replace("\\", "/");
    }
    out_dir
}
```

jason-yau:
fixed #139 
Then developers can specify target_dir in cmake, such as:
```cmake
    set(CARGO_TARGET_DIR "${CMAKE_BINARY_DIR}/target")
    if (CMAKE_BUILD_TYPE STREQUAL "Debug")
        set(CARGO_CMD cargo build --target-dir ${CARGO_TARGET_DIR})
        set(TARGET_DIR "debug")
    else ()
        set(CARGO_CMD cargo build --release --target-dir ${CARGO_TARGET_DIR})
        set(TARGET_DIR "release")
    endif ()
```

Then the library that rust generated will be in `${CARGO_TARGET_DIR}/target/librust.a` or `${CARGO_TARGET_DIR}/target/rust.lib`. The `cxx-qt-gen` and `cxx-qt-lib` folders where generated cpp sources and headers are placed is in `${CARGO_TARGET_DIR}/target` folder. 
This will not generate Rust target directory outside of the Qt project build directory. This allows us to always have one build directory, rather than Rust having one and Qt having one.

### DIFF
--- a/cxx-qt-build/src/lib.rs
+++ b/cxx-qt-build/src/lib.rs
@@ -108,6 +108,16 @@ fn manifest_dir() -> String {
     manifest_dir
 }
 
+fn out_dir() -> String {
+    let mut out_dir = env::var("OUT_DIR").expect("Could not get out dir");
+    if cfg!(windows) {
+        out_dir = out_dir.replace('\\', "/");
+    }
+    out_dir += "/../../../..";
+    println!("cargo:rerun-if-env-changed=OUT_DIR");
+    out_dir
+}
+
 /// Extract the cxx or cxx_qt module from a Rust file
 fn extract_modules(file_content: &str, rs_path: &str) -> ExtractedModule {
     let file = syn::parse_file(file_content).unwrap();
@@ -158,13 +168,10 @@ fn extract_modules(file_content: &str, rs_path: &str) -> ExtractedModule {
 
 /// Write the generated cpp and h files for a qobject out to files
 fn write_qobject_cpp_files(obj: CppObject, snake_name: &str) -> Vec<String> {
-    let manifest_dir = manifest_dir();
+    let out_dir = out_dir();
 
-    let h_path = format!(
-        "{}/target/cxx-qt-gen/include/{}.h",
-        manifest_dir, snake_name
-    );
-    let cpp_path = format!("{}/target/cxx-qt-gen/src/{}.cpp", manifest_dir, snake_name);
+    let h_path = format!("{}/cxx-qt-gen/include/{}.h", out_dir, snake_name);
+    let cpp_path = format!("{}/cxx-qt-gen/src/{}.cpp", out_dir, snake_name);
 
     let mut file = File::create(&h_path).expect("Could not create .h file");
     write!(file, "{}", obj.header).expect("Failed to write .h file");
@@ -182,6 +189,7 @@ fn gen_cxx_for_file(
     cpp_namespace_prefix: &[&'static str],
 ) -> Vec<String> {
     let manifest_dir = manifest_dir();
+    let out_dir = out_dir();
     let mut generated_cpp_paths = Vec::new();
 
     // TODO: in the future use the module path as the file path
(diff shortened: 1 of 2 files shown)


cxx-qt/src/lib.rs
```diff
@@ -10,8 +10,11 @@ use cxx_qt_gen::{extract_qobject, generate_qobject_rs};
 
 /// Read the C++ namespace prefix that cxx-qt-build has set for us
 fn read_cpp_namespace_prefix() -> Vec<String> {
-    let dir_target = std::env::var("CARGO_MANIFEST_DIR").expect("Could not get manifest dir");
-    let path = format!("{}/target/cxx-qt-gen/cpp_namespace_prefix.txt", dir_target);
+    let dir_target = std::env::var("OUT_DIR")
+        .map(|s| s + "/../../../..")
```
ahayzen-kdab, line 14: This code seems a bit magical here, what is it trying to do ?

I would have thought it should be looking inside `OUT_DIR/cxx-qt-gen/cpp_namespace_prefix.txt` ? Or are there problems going across the crates.

As one of our next steps for the CMake is to instead use a CLI tool to do the generation so that we don't have this mess with the paths and can avoid cargo build happening in the CMake configure stage. (see some things in #/106 but again we'll write up a github discussion shortly).

cxx-qt-build/src/lib.rs
```diff
@@ -108,6 +108,16 @@ fn manifest_dir() -> String {
     manifest_dir
 }
 
+fn out_dir() -> String {
+    let mut out_dir = env::var("OUT_DIR").expect("Could not get out dir");
+    if cfg!(windows) {
+        out_dir = out_dir.replace('\\', "/");
+    }
+    out_dir += "/../../../..";
```
ahayzen-kdab, line 116: Ah now I see why things later look strange, so we are abusing the OUT_DIR here, wonder if this has other implications and if we should go straight to the CLI tool for CMake to use :thinking: ( #/106 )

jason-yau: Although this path is strange, the only way I know to place Rust's target-dir in the Qt build-dir without other CLI is calculate the target-dir through OUT_DIR. If there is a better way, please tell me.

ahayzen-kdab: I worry that this will have problems when we reach cross compiling, eg `target/debug/build/<crate-hash>/out/` vs `target/aarch64-unknown-linux-gnu/debug/build/<crate-hash>/out/`.

What if you compared the OUT_DIR to the CARGO_MANIFEST_DIR could you then work out how many levels "up" it is from the OUT_DIR to the CARGO_MANIFEST_DIR..

Ah I think i see what's happening and why you wouldn't want to do that ^^, this is trying to put `cxx-qt-gen` into either `target/cxx-qt-gen` or `target/aarch64-unknown-linux-gnu/cxx-qt-gen` or where ever is specified ?

So then this probably a good first step until we build the CLI tool (which i'll write up the discussion for shortly).

ahayzen-kdab: So this is probably fine for now, eventually we'll move to a CLI tool as being discussed here https://github.com/KDAB/cxx-qt/discussions/146

PR #159, opened 2022-07-14
WIP: syntax: move to using an extern Qt block

ahayzen-kdab:
Change-Id: I8e523e41b127eb065f0fc82f432d24415f93506c

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,48 +36,50 @@ Below is an example of Rust code that exposes a QObject with two properties and 
 ```rust
 #[cxx_qt::bridge]
 mod my_object {
-    use serde::{Deserialize, Serialize};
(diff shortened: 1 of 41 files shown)


ahayzen-kdab:
We can't use `extern "Qt" { ... }` with "invalid" syntax (eg items not in https://docs.rs/syn/1.0.98/syn/struct.ItemForeignMod.html ) even though we are using a procedural attribute macro, we would potentially have to change to a function macro or a module ?

As the Rust compiler needs valid tokens to do an attribute macro.

LeonMatthesKDAB:
Also, see discussion here: https://github.com/KDAB/cxx-qt/discussions/156#discussioncomment-3149051

PR #160, opened 2022-07-15
cxx-qt-build: remove the option to specify a custom namespace prefix

ahayzen-kdab:
This will be removed from the build tool and instead later
this will be moved onto the macro.

For now just use cxx_qt as the default.


Be-ing (approved):
waiting for CI

PR #161, opened 2022-07-15
add CLI tool for C++ code generation

Be-ing:
implementing #146

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -152,14 +152,14 @@ add_subdirectory(examples)
 # Create helper method which adds relevent cargo tests for a given manifest
 function(add_test_cargo TEST_NAME_PREFIX MANIFEST_PATH ADD_DOCTESTS)
     # Add cargo as a test
-    add_test(NAME ${TEST_NAME_PREFIX}_cargo_tests COMMAND cargo test --all-targets --all-features --manifest-path ${MANIFEST_PATH})
+    add_test(NAME ${TEST_NAME_PREFIX}_cargo_tests COMMAND cargo test --all-targets --manifest-path ${MANIFEST_PATH})
     # Check if we should enable doc tests
     if (${ADD_DOCTESTS} STREQUAL "DOCTESTS_ON")
         # Add cargo docs as a test
         add_test(NAME ${TEST_NAME_PREFIX}_cargo_doc_tests COMMAND cargo test --doc --manifest-path ${MANIFEST_PATH})
     endif()
     # Add clippy as a test
-    add_test(NAME ${TEST_NAME_PREFIX}_cargo_clippy COMMAND cargo clippy --all-targets --all-features --manifest-path ${MANIFEST_PATH} -- -D warnings)
+    add_test(NAME ${TEST_NAME_PREFIX}_cargo_clippy COMMAND cargo clippy --all-targets --manifest-path ${MANIFEST_PATH} -- -D warnings)
     # Add rustfmt as a test
     add_test(NAME ${TEST_NAME_PREFIX}_cargo_fmt COMMAND cargo fmt --manifest-path ${MANIFEST_PATH} -- --check)
 endfunction()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -171,6 +171,7 @@ add_test_cargo(cxx_qt "${CMAKE_CURRENT_SOURCE_DIR}/cxx-qt/Cargo.toml" DOCTESTS_O
 add_test_cargo(cxx_qt_build "${CMAKE_CURRENT_SOURCE_DIR}/cxx-qt-build/Cargo.toml" DOCTESTS_ON)
 add_test_cargo(cxx_qt_gen "${CMAKE_CURRENT_SOURCE_DIR}/cxx-qt-gen/Cargo.toml" DOCTESTS_ON)
 add_test_cargo(cxx_qt_lib "${CMAKE_CURRENT_SOURCE_DIR}/cxx-qt-lib/Cargo.toml" DOCTESTS_ON)
+add_test_cargo(cxx_qt_cli "${CMAKE_CURRENT_SOURCE_DIR}/cxx-qt-cli/Cargo.toml" DOCTESTS_OFF)
 
 # Ensure test inputs and outputs are formatted
 file(GLOB CXX_QT_GEN_TEST_INPUTS ${CMAKE_CURRENT_SOURCE_DIR}/cxx-qt-gen/test_inputs/*.rs)
--- a/cxx-qt-build/src/lib.rs
+++ b/cxx-qt-build/src/lib.rs
@@ -216,7 +216,7 @@ impl GeneratedCpp {
 
         if let Some(cxx_qt_generated) = &self.cxx_qt {
             let header_path = PathBuf::from(format!(
-                "{}/{}.h",
+                "{}/{}.cxxqt.h",
                 include_directory_path.display(),
                 self.module_ident
             ));
@@ -228,7 +228,7 @@ impl GeneratedCpp {
             written_files.push(header_path);
 
             let cpp_path = PathBuf::from(format!(
-                "{}/{}.cpp",
+                "{}/{}.cxxqt.cpp",
                 source_directory_path.display(),
                 &self.module_ident
             ));
--- a/cxx-qt-build/src/lib.rs
+++ b/cxx-qt-build/src/lib.rs
@@ -9,6 +9,7 @@ use serde::{Deserialize, Serialize};
 use std::env;
 use std::fs::File;
 use std::io::Write;
+use std::path::PathBuf;
 use syn::*;
 
 use clang_format::ClangFormatStyle;
@@ -93,7 +94,7 @@ fn manifest_dir() -> String {
 }
 
 /// Extract the cxx or cxx_qt module from a Rust file
-fn extract_modules(file_content: &str, rs_path: &str) -> ExtractedModule {
+fn extract_modules(file_content: &str, rs_path: &impl AsRef<std::path::Path>) -> ExtractedModule {
     let file = syn::parse_file(file_content).unwrap();
 
     // Define a helper function that will ensure that we can extract at most one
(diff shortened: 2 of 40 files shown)


cxx-qt-cli/src/main.rs
```diff
@@ -0,0 +1,35 @@
...
+    #[clap(short, long, value_parser)]
+    input: PathBuf,
+
+    /// directory to output generated C++ files
+    #[clap(short, long, value_parser)]
+    output: PathBuf,
+}
+
+
+fn main() { let cli = Cli::parse();
```
ahayzen-kdab, line 23: formatting looks wrong here, which hints that you need to add the following to the root `CMakeLists.txt`

```cmake
add_test_cargo(cxx_qt_cli "${CMAKE_CURRENT_SOURCE_DIR}/cxx-qt-cli/Cargo.toml" DOCTESTS_ON)
```

cxx-qt-build/src/lib.rs
```diff
@@ -9,6 +9,7 @@ use serde::{Deserialize, Serialize};
 use std::env;
 use std::fs::File;
 use std::io::Write;
+use std::path::PathBuf;
```
ahayzen-kdab, line 12: Can we split this first commit of using `PathBuf` into a separate pull request, as we can likely land this first and faster, therefore reducing noise in this second CLI pull request?

cxx-qt-build/src/lib.rs
```diff
@@ -143,139 +144,146 @@ fn extract_modules(file_content: &str, rs_path: &str) -> ExtractedModule {
...
-    write!(file, "{}", obj.header).expect("Failed to write .h file");
-
-    let mut file = File::create(&cpp_path).expect("Could not create .cpp file");
-    write!(file, "{}", obj.source).expect("Failed to write .cpp file");
-
-    vec![h_path, cpp_path]
+pub struct GeneratedCpp {
+    cxx_qt: Option<CppObject>,
+    cxx: cxx_gen::GeneratedCode,
+    module_ident: String,
```
ahayzen-kdab, line 150: So the module ident is correct right now, however in the future we might support multiple objects being defined in one module and multiple modules in one file.

In this case, it's probably best just to match the input file name as the output filename and we'll use the namespace defined in the macro to avoid any collisions of similar names etc (i assume CXX does the same?).

So eg `lib.rs` -> `lib.h` or `my_object.rs` -> `my_object.h` etc.

cxx-qt-build/src/lib.rs
```diff
@@ -143,139 +144,146 @@ fn extract_modules(file_content: &str, rs_path: &str) -> ExtractedModule {
...
+            cpp.write_all(cxx_qt_generated.source.as_bytes())
+                .expect("Could not write cxx-qt source file");
+            written_files.push(cpp_path);
+        }
 
-    let mut cpp = File::create(&cpp_path).expect("Could not create cpp file");
-    cpp.write_all(&gen_result.implementation)
-        .expect("Could not write cpp file");
+        let header_path = PathBuf::from(format!(
+            "{}/{}_cxx.h",
```
ahayzen-kdab, line 242: i wonder if `_cxx` is the best naming, as if you call your module (or the input file, see other comment above) something with cxx at the end it'll collide. Whereas if this was `{name}.cxx.h` or `{name}.rs.h` it can never collide as the name can't have a `.` ?

cxx-qt-gen/src/gen_cpp.rs
```diff
@@ -1039,7 +1039,7 @@ pub fn generate_qobject_cpp(obj: &QObject) -> Result<CppObject, TokenStream> {
     // Generate C++ source part
     let source = formatdoc! {r#"
         #include "cxx-qt-gen/include/{ident_snake}.h"
-        #include "cxx-qt-gen/src/{ident_snake}.rs.h"
+        #include "cxx-qt-gen/include/{ident_snake}_cxx.h"
```
ahayzen-kdab, line 1042: won't the folder on both of these includes now go ? As in the CMake we'll add the `include/` folder to the include directories ? and the output won't necessarily be `cxx-qt-gen/include/` but it could be like `build/rust/include` or anything etc.

ahayzen-kdab: This then makes me wonder what should happen with `#include "cxx-qt-lib/include/qt_types.h"` maybe when we have the tool output the cxx-qt-lib stuff into a folder we can have it produce `cxx-qt-lib/qt_types.h` so then the include becomes like `#include <cxx-qt-lib/qt_types.h>` or something.

cxx-qt-cli/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+[[bin]]
+name = "cxxqtbridge"
+path = "src/main.rs"
+
+[dependencies]
+cxx-qt-build = { path = "../cxx-qt-build" }
+clap = { version = "3.2.12", features = [ "derive" ] }
+
+[features]
+# Requires nightly compiler until std::path::absolute is stabilized
```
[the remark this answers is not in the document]
Be-ing: That's why this is behind a feature that is disabled by default.

cxx-qt-build/src/lib.rs
Be-ing: Good point, `.cxx.{h/cpp}` would avoid that issue.

cxx-qt-build/src/lib.rs
ahayzen-kdab: I literally meant take out the first of three commits in this branch https://github.com/KDAB/cxx-qt/pull/161/commits/9fbd45f67f88f082098334c0900bec0ffb10f0e3 and submit by itself.

cxx-qt-build/src/lib.rs
ahayzen-kdab: CXX appears to do them per file, so i think it's best to copy them. https://github.com/dtolnay/cxx/blob/9e0d3680be8cdc42a545467228fda096f34ce241/gen/src/mod.rs#L130  https://github.com/dtolnay/cxx/blob/9e0d3680be8cdc42a545467228fda096f34ce241/gen/src/write.rs#L17

Otherwise if we lean more on their generation later we could have mismatches of files (ours generating many files theirs generating one etc). But something to think about going forward as we continue evolving if this will cause issues and separate files are beneficial.

cxx-qt-cli/Cargo.toml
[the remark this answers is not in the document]
ahayzen-kdab: Ah i missed that it is shadowing the `output_file_path` i thought it might fail without the feature :') ... So then this is fine maybe just put a `TODO: FUTURE:` or something so we can find it later when / if the feature becomes stable.

cxx-qt-build/src/lib.rs
Be-ing: You're right, cxx uses the Rust filename including `.rs` and adds `.h`/`.cpp`. So how about:

our generated files: filename.cxxqt.{h, cpp}
cxx generated files: filename.cxx.{h, cpp}

where `filename` includes the `.rs.`

cxx-qt-cli/src/main.rs
```diff
@@ -0,0 +1,36 @@
...
+    /// path to Rust file containing module with #[cxx_qt::bridge] attribute macro
+    #[clap(short, long, value_parser)]
+    input: PathBuf,
+
+    /// directory to output generated C++ files
+    #[clap(short, long, value_parser)]
+    output: PathBuf,
+}
+
+
```
ahayzen-kdab, line 22: rust format complains about too much spacing here (see CI failure).

CMakeLists.txt
```diff
@@ -152,14 +152,14 @@ add_subdirectory(examples)
 # Create helper method which adds relevent cargo tests for a given manifest
 function(add_test_cargo TEST_NAME_PREFIX MANIFEST_PATH ADD_DOCTESTS)
     # Add cargo as a test
-    add_test(NAME ${TEST_NAME_PREFIX}_cargo_tests COMMAND cargo test --all-targets --all-features --manifest-path ${MANIFEST_PATH})
+    add_test(NAME ${TEST_NAME_PREFIX}_cargo_tests COMMAND cargo test --all-targets --manifest-path ${MANIFEST_PATH})
```
ahayzen-kdab, line 155: unfortunate that we need to disable `--all-features`, as later we might have multiple features in CXX-Qt we want to test :thinking: maybe OK for now though.

cxx-qt-gen/test_outputs/properties.cpp
```diff
@@ -1,5 +1,5 @@
-#include "cxx-qt-gen/include/my_object.h"
-#include "cxx-qt-gen/src/my_object.rs.h"
+#include "cxx-qt-gen/include/my_object.cxxqt.h"
```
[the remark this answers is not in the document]
ahayzen-kdab: As this is generated though just flipping the order should be right.

cxx-qt-cli/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+description = "Command line tool for cxx-qt's C++ code generation step for integration into C++ build systems"
+repository = "https://github.com/KDAB/cxx-qt/"
+
+[[bin]]
+name = "cxxqtbridge"
+path = "src/main.rs"
+
+[dependencies]
+cxx-qt-build = { path = "../cxx-qt-build" }
+clap = { version = "3.2.12", features = [ "derive" ] }
```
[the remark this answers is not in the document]
Be-ing: https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277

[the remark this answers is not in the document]
ahayzen-kdab: Ah I see from https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility that `3.2.12` is the same as `^3.2.12` can we use the latter one ? As it's clearer that `^3.2.12` means `>=3.2.12, <4.0.0` whereas just using `3.2.12` looks like it's pinning to that version.

cxx-qt-gen/src/gen_cpp.rs
ahayzen-kdab: As discussed, lets figure this out in later changes.

PR #162, opened 2022-07-18
gen_rs: use CXX as module and move CXX-Qt to external module 

ahayzen-kdab:
Note that nested objects now need to be referred to by cxx_qt_
in the last module name to reach the correct CppObj.

This will be improved in the future when we move to the new API
which will store nested objects as a pointer directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,28 +53,28 @@ mod my_object {
     }
 
     #[derive(Default)]
-    struct RustObj;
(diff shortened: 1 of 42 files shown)


Be-ing (approved):
Seeing the diffs in the cxx-qt-gen test outputs was helpful to understand this change. Waiting for CI

PR #163, opened 2022-07-19
cxx-qt-lib: add option to output C++ to CXX_QT_LIB_OUT_DIR env var

Be-ing:
These C++ files won't build yet because the #include paths need to
be adjusted.

### DIFF
--- a/cxx-qt-lib/build.rs
+++ b/cxx-qt-lib/build.rs
@@ -5,6 +5,7 @@
 
 use quote::ToTokens;
 use serde::{Deserialize, Serialize};
+use std::{fs::File, io::Write, path::PathBuf};
 
 /// Representation of a generated CXX header, source, and name
 #[derive(Serialize, Deserialize)]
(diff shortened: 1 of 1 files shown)


Be-ing:
I'm thinking the include paths should be:
```cpp
#include <cxx-qt/qttypes.h>
```

ahayzen-kdab:
> I'm thinking the include paths should be:
> 
> ```c++
> #include <cxx-qt/qttypes.h>
> ```

Right, wonder if it should be `cxx-qt` or `cxx-qt-lib` ? As we might have `gen` files too ? eg `#include <cxx-qt-lib/qttypes.h>` and `#include <cxx-qt-gen/my_object.rs.cxxqt.h>` ?

cxx-qt-lib/build.rs
```diff
@@ -80,5 +81,62 @@ fn main() {
...
+            .write_all(cxx_gen::HEADER.as_bytes())
+            .expect("Could not write cxx.h file");
+
+        for class in generated {
+            let include_file_path = PathBuf::from(format!(
+                "{}/{}.h",
+                include_directory_path.display(),
+                class.name
+            ));
+            File::create(&include_file_path)
```
ahayzen-kdab, line 126: We have this file create, write_all, some form of failure block 4 times, DRY points to this being put in a method ?

Or also maybe the whole block needs to be split up, as the header and source paths are repeated for creating the directory, copying the qt_types, creating the individual files.  Maybe there should be one method that does each of those steps once but is passed as parameter to tell it the destination and source ?

ahayzen-kdab (approved):
LGTM :-)

PR #164, opened 2022-07-20
gen_rs: pass through ForeignMod to the CXX bridge for now

ahayzen-kdab:
Later we will pass through all non CXX-Qt items to the CXX bridge,
but this is a stepping stone that allows us to progress.

### DIFF
--- a/cxx-qt-gen/test_inputs/passthrough.rs
+++ b/cxx-qt-gen/test_inputs/passthrough.rs
@@ -21,6 +21,8 @@ pub mod my_object {
     // ItemForeignMod
     extern "C" {}
 
+    unsafe extern "C++" {}
+
     // ItemMacro
     macro_rules! macro1 {
         () => {
--- a/cxx-qt-gen/test_inputs/passthrough.rs
+++ b/cxx-qt-gen/test_inputs/passthrough.rs
@@ -21,6 +21,20 @@ pub mod my_object {
     // ItemForeignMod
     extern "C" {}
 
+    #[namespace = "namespace"]
+    extern "C" {}
+
+    #[namespace = "namespace"]
+    #[custom_attr = "test"]
+    extern "C" {}
+
+    unsafe extern "C++" {}
+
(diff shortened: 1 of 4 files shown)


ahayzen-kdab:
This doesn't work with `unsafe extern "C++"` ...

ahayzen-kdab:
> This doesn't work with `unsafe extern "C++"` ...

So `extern "T"` comes through as `ItemForeegnMod` but `unsafe extern "T"` comes through as `Verbatim`.

ahayzen-kdab:
> > This doesn't work with `unsafe extern "C++"` ...
> 
> So `extern "T"` comes through as `ItemForeegnMod` but `unsafe extern "T"` comes through as `Verbatim`.

This works now.

Just need to also support attributes on the extern block.

ahayzen-kdab:
> Just need to also support attributes on the extern block.

This is working now

cxx-qt-gen/test_inputs/passthrough.rs
```diff
@@ -21,6 +21,22 @@ pub mod my_object {
     // ItemForeignMod
     extern "C" {}
 
+    #[namespace = "namespace"]
+    extern "C" {}
+
+    #[namespace = "namespace"]
+    #[namespace = "namespace"]
```
Be-ing, line 28: Repeating the same name twice? Are you checking here that two attributes on consecutive lines are properly handled? If so, I think using two different attributes would make the test easier to understand.

ahayzen-kdab: Yup, i'm checking that multiple attributes are passed through, i'll rename one of them.

ahayzen-kdab: Fixed.

Be-ing (approved):
waiting for CI

ahayzen-kdab:
(Rebased as the other one landed)

Be-ing (approved):

PR #165, opened 2022-07-21
cxx-qt-lib: use generic templated method for converting returns

ahayzen-kdab:
This simplifies generation later as we don't need to generate
code to convert between the types, we only need to write the
input and output types.

### DIFF
--- a/cxx-qt-lib/include/qt_types.h
+++ b/cxx-qt-lib/include/qt_types.h
@@ -27,6 +27,30 @@
 namespace rust {
 namespace cxxqtlib1 {
 
+template<typename R, typename T>
+struct cxx_qt_convert
+{
+  R operator()(T val) { return val; }
+};
+
+template<typename R, typename T>
+struct cxx_qt_convert<R&, T>
+{
+  R operator()(T val) { return val; }
+};
+
+template<typename R, typename T>
+struct cxx_qt_convert<R, std::unique_ptr<T>>
+{
+  R operator()(std::unique_ptr<T> ptr) { return std::move(*ptr); }
+};
+
+template<typename R, typename T>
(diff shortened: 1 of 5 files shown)


Be-ing (approved):

cxx-qt-lib/include/qt_types.h
```diff
@@ -27,6 +27,30 @@
...
+template<typename R, typename T>
+struct cxx_qt_convert
+{
+  R operator()(T val) { return val; }
+};
+
+template<typename R, typename T>
+struct cxx_qt_convert<R&, T>
+{
+  R operator()(T val) { return val; }
```
LeonMatthesKDAB, line 39: This should return `R&`, shouldn't it?
Also, this would need to take a `T& val`

cxx-qt-lib/include/qt_types.h
```diff
@@ -27,6 +27,30 @@
...
+template<typename R, typename T>
+struct cxx_qt_convert<R, std::unique_ptr<T>>
+{
+  R operator()(std::unique_ptr<T> ptr) { return std::move(*ptr); }
+};
+
+template<typename R, typename T>
+struct cxx_qt_convert<R&, std::unique_ptr<T>>
+{
+  R operator()(const std::unique_ptr<T>& ptr) { return *ptr; }
```
LeonMatthesKDAB, line 51: Should return a R& as well.

LeonMatthesKDAB:
Also, we should see whether we can create a nice error message if conversion can't be done.

Possibly with some combination of [is_convertible](https://en.cppreference.com/w/cpp/types/is_convertible) and [static_assert](https://en.cppreference.com/w/cpp/language/static_assert)

PR #166, opened 2022-07-21
cxx-qt-lib: use unique ptr types and don't wrap

ahayzen-kdab:
This will greatly simplify the generation later as we'll be
able to remove the pattern matching. Also it requires the
developer to perform what can be costly conversions, so
this then is more Rust-like as it points to where close might be.


Be-ing:
The Rust is failing to link on Windows.

README.md
```diff
@@ -34,21 +34,47 @@ For more complex examples navigate to the [examples folder](./examples) where th
 Below is an example of Rust code that exposes a QObject with two properties and four invokable methods to Qt.
 
 ```rust
+use serde::{Deserialize, Serialize};
+
+#[derive(Deserialize, Serialize)]
+pub struct DataSerde {
```
Be-ing, line 40: Is the purpose of this struct to convert the QString to a Rust std::String? Besides that, it seems redundant.

ahayzen-kdab:
> The Rust is failing to link on Windows.

Yup, we are unsure why or what's changed specifically to break that so far...

README.md
ahayzen-kdab: Yes this will be required to convert from any `UniquePtr<T>` to `T` / Rust equivalent for serde. As `UniquePtr` doesn't implement (de)seralisation.

Also later on you would likely need this separate struct anyway as private members will be in the Data struct too, so eg Mutex and Channel etc might be in there.

But this is something to loop back around to after the API transition is done as we have a few ideas that might help there.

But for now, this is the route we'll have to go.

book/src/concepts/types.md
```diff
@@ -55,17 +55,15 @@ Note that when they are used as a parameter type in invokables they should be pa
...
+Custom opaque types represent an opaque C++ type, they can be used for properties, parameters or return types in invokables, and parameters in signals. However when the object is being passed from Rust to C++ (eg properties, return types, or signals) they must be `UniquePtr<T>`, and for other use cases they must be by reference (either `&T` or `Pin<&mut T>`).
 
-On the rust side they appear as the cxx_qt_lib helper type.
-
-Note that when they are used as a parameter type in invokables they should be passed as a reference, eg `color: &QColor`, and when they are a property or return type they should be a value, eg `QColor`. Also for strings `&str` should be used when passed as a reference and `String` when passed as a value.
+Note that when they are used as a parameter type in invokables they should be passed as a reference, eg `color: &QColor`, and when they are a property, or return type, or signal they should be a value, eg `UniquePtr<QColor>`.
```
Be-ing, line 60: This is a little confusing. The text says using a value, but the example is a `UniquePtr`. Must it be a `UniquePtr<QColor>` rather than `QColor`?

ahayzen-kdab: Yup, opaque types can only be passed around as `&T`, `Pin<&mut T>`, or `UniquePtr<T>`. And we cannot return a reference so we have to return by value which for this opaque type means `UniquePtr<T>`.

I'll modify the text so that it reads should be a unique ptr instead of value.

README.md
Be-ing: Okay. It would help to have a brief inline comment, particularly considering this is the first example shown. Maybe something like `This struct has the fields of the Data struct below but converted to Rust types`

ahayzen-kdab: Added some text.

Be-ing:
Please leave a comment when you get the Windows build working again.

CMakeLists.txt
```diff
@@ -16,6 +16,9 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
 option(QT_DEFAULT_MAJOR_VERSION "Major version of Qt to use (6 or 5)" 6)
 
+# Ensure we always export all symbols for Windows
+set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
```
Be-ing, line 20: Eh, this is hacky. If we want to use it, we should set the property on our CMake targets instead of globally.

ahayzen-kdab: It doesn't seem to help or change anything anyway

book/src/concepts/types.md
```diff
@@ -55,17 +55,15 @@ Note that when they are used as a parameter type in invokables they should be pa
...
+Custom opaque types represent an opaque C++ type, they can be used for properties, parameters or return types in invokables, and parameters in signals. However when the object is being passed from Rust to C++ (eg properties, return types, or signals) they must be `UniquePtr<T>`, and for other use cases they must be by reference (either `&T` or `Pin<&mut T>`).
 
-On the rust side they appear as the cxx_qt_lib helper type.
-
-Note that when they are used as a parameter type in invokables they should be passed as a reference, eg `color: &QColor`, and when they are a property or return type they should be a value, eg `QColor`. Also for strings `&str` should be used when passed as a reference and `String` when passed as a value.
+Note that when they are used as a parameter type in invokables they should be passed as a reference, eg `color: &QColor`, and when they are a property, or return type, or signal they should be a `UniquePtr`, eg `UniquePtr<QColor>`.
```
Be-ing, line 60: `property, return type, or signal`

`or` is repeated

cxx-qt-gen/src/gen_rs.rs
```diff
@@ -111,25 +114,33 @@ impl RustType for QtTypes {
...
-                TargetType::Cpp => quote! {cxx_qt_lib::QVariantCpp},
+                TargetType::Cpp => quote! {cxx_qt_lib::QVariant},
                 TargetType::Rust => quote! {cxx_qt_lib::QVariant},
             },
             Self::U8 => quote! {u8},
             Self::U16 => quote! {u16},
             Self::U32 => quote! {u32},
+            Self::UniquePtr { inner } => {
+                let inner = inner.cxx_qt_lib_type(target_type);
+                quote! {UniquePtr<#inner> }
```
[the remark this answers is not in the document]
ahayzen-kdab: I think the trailing space should actually be removed to match the others :-) ... i think because it's a macro rustfmt doesn't pick it up.

If we put spaces on either side then all of them should have it, but hopefully this code will go away in the future anyway :-)

Be-ing (approved):

PR #172, opened 2022-07-23
build Qt types autogenerated C++ code as a library

Be-ing:
Previously, the C++ code was autogenerated by cxx-qt-build from
applications' build.rs, then repeatedly recompiled for each
application. Now, the C++ code is generated once and built as
a library which is linked to each application. This is a
prerequisite for removing the need for applications to have
build.rs to instead use the CLI C++ code generator. Also, this
avoids redundant compilation of the autogenerated C++ which is
constant acrosss applications.


Be-ing:
Closing in favor of #174

PR #174, opened 2022-07-25
build cxx-qt-lib autogenerated C++ with Cargo instead of CMake

Be-ing:
as [proposed](https://github.com/KDAB/cxx-qt/pull/173#issuecomment-1193353750)

### DIFF
--- a/.github/workflows/github-cxx-qt-tests.yml
+++ b/.github/workflows/github-cxx-qt-tests.yml
@@ -76,7 +76,8 @@ jobs:
             qt_version: 5
             # FIXME: many tests fail to link
             # https://github.com/KDAB/cxx-qt/issues/111
-            ctest_args: --exclude-regex '^(cxx_qt_gen_cargo_tests|demo_threading_cargo_tests|example_qml.*|qml.*tests|test.*|reuse_lint|cpp_clang_format|.*valgrind)$'
+            # cxx_qt_lib_cargo_tests fails to run with a missing DLL error.
+            ctest_args: --exclude-regex '^(cxx_qt_gen_cargo_tests|demo_threading_cargo_tests|example_qml.*|qml.*tests|test.*|reuse_lint|cpp_clang_format|.*valgrind|cxx_qt_lib_cargo_tests)$'
             exe_suffix: .exe
             qt_qpa_platform: windows
             compiler_cache_path: C:\Users\runneradmin\AppData\Local\Mozilla\sccache\cache
@@ -90,7 +91,8 @@ jobs:
             qt_version: 6
             # FIXME: many tests fail to link
             # https://github.com/KDAB/cxx-qt/issues/111
-            ctest_args: --exclude-regex '^(cxx_qt_gen_cargo_tests|demo_threading_cargo_tests|example_qml.*|qml.*tests|test.*|reuse_lint|cpp_clang_format|.*valgrind)$'
+            # cxx_qt_lib_cargo_tests fails to run with a missing DLL error.
+            ctest_args: --exclude-regex '^(cxx_qt_gen_cargo_tests|demo_threading_cargo_tests|example_qml.*|qml.*tests|test.*|reuse_lint|cpp_clang_format|.*valgrind|cxx_qt_lib_cargo_tests)$'
             exe_suffix: .exe
             qt_qpa_platform: windows
             compiler_cache_path: C:\Users\runneradmin\AppData\Local\Mozilla\sccache\cache
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -180,6 +180,7 @@ add_test(NAME cxx_qt_gen_test_inputs_gen COMMAND rustfmt --check ${CXX_QT_GEN_TE
 add_test(NAME cxx_qt_gen_test_outputs_gen COMMAND rustfmt --check ${CXX_QT_GEN_TEST_OUTPUTS})
 
 # QML example has add_test in it's CMakeLists, so just add the cargo tests here
+add_test_cargo(qt-build "${CMAKE_CURRENT_SOURCE_DIR}/qt-build/Cargo.toml" DOCTESTS_ON)
 add_test_cargo(demo_threading "${CMAKE_CURRENT_SOURCE_DIR}/examples/demo_threading/Cargo.toml" DOCTESTS_OFF)
 add_test_cargo(qml_features "${CMAKE_CURRENT_SOURCE_DIR}/examples/qml_features/Cargo.toml" DOCTESTS_OFF)
 add_test_cargo(qml_extension_plugin "${CMAKE_CURRENT_SOURCE_DIR}/examples/qml_extension_plugin/core/Cargo.toml" DOCTESTS_OFF)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,19 +149,32 @@ endfunction()
 add_subdirectory(book)
 add_subdirectory(examples)
 
+# CARGO_TARGET_DIR is needed so the generated headers can be added to the
+# include paths.
+# QMAKE environment variable is needed by qt-build to ensure that Cargo
+# uses the same installation of Qt as CMake does.
+find_package(QT NAMES Qt6 Qt5 COMPONENTS Core Gui Test REQUIRED)
+find_package(Qt${QT_VERSION_MAJOR} COMPONENTS Core Gui Test REQUIRED)
+get_target_property(QMAKE Qt${QT_VERSION_MAJOR}::qmake IMPORTED_LOCATION)
+set(CARGO_ENV "QMAKE=${QMAKE};CARGO_TARGET_DIR=${CMAKE_CURRENT_SOURCE_DIR}/target")
+
 # Create helper method which adds relevent cargo tests for a given manifest
 function(add_test_cargo TEST_NAME_PREFIX MANIFEST_PATH ADD_DOCTESTS)
     # Add cargo as a test
     add_test(NAME ${TEST_NAME_PREFIX}_cargo_tests COMMAND cargo test --all-targets --manifest-path ${MANIFEST_PATH})
+    set_property(TEST ${TEST_NAME_PREFIX}_cargo_tests PROPERTY ENVIRONMENT ${CARGO_ENV})
     # Check if we should enable doc tests
     if (${ADD_DOCTESTS} STREQUAL "DOCTESTS_ON")
         # Add cargo docs as a test
         add_test(NAME ${TEST_NAME_PREFIX}_cargo_doc_tests COMMAND cargo test --doc --manifest-path ${MANIFEST_PATH})
+        set_property(TEST ${TEST_NAME_PREFIX}_cargo_doc_tests PROPERTY ENVIRONMENT ${CARGO_ENV})
     endif()
     # Add clippy as a test
     add_test(NAME ${TEST_NAME_PREFIX}_cargo_clippy COMMAND cargo clippy --all-targets --manifest-path ${MANIFEST_PATH} -- -D warnings)
+    set_property(TEST ${TEST_NAME_PREFIX}_cargo_clippy PROPERTY ENVIRONMENT ${CARGO_ENV})
     # Add rustfmt as a test
     add_test(NAME ${TEST_NAME_PREFIX}_cargo_fmt COMMAND cargo fmt --manifest-path ${MANIFEST_PATH} -- --check)
+    set_property(TEST ${TEST_NAME_PREFIX}_cargo_fmt PROPERTY ENVIRONMENT ${CARGO_ENV})
 endfunction()
 
 # Add cargo tests for all our manifests
--- a/cmake/CxxQt.cmake
+++ b/cmake/CxxQt.cmake
@@ -41,9 +41,18 @@ function(cxx_qt_generate_cpp GEN_SOURCES)
 
     file(MAKE_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/target/cxx-qt-gen")
 
+    get_target_property(QMAKE Qt${QT_VERSION_MAJOR}::qmake IMPORTED_LOCATION)
+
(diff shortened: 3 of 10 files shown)


.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -87,7 +87,8 @@ jobs:
             qt_version: 5
             # FIXME: many tests fail to link
             # https://github.com/KDAB/cxx-qt/issues/111
-            ctest_args: --exclude-regex '^(cxx_qt_gen_cargo_tests|demo_threading_cargo_tests|example_qml.*|qml.*tests|test.*|reuse_lint|cpp_clang_format|.*valgrind)$'
+            # cxx_qt_lib_cargo_tests fails to run with a missing DLL error.
```
ahayzen-kdab, line 90: can this commit come earlier in the chain ? As we should try to ensure that each commit passes tests so that when tools like `git bisect` are used things make sense. Or should i squash commit at the end ? (but i prefer to commit with the full chain normally...)

CMakeLists.txt
```diff
@@ -149,19 +149,28 @@ endfunction()
 add_subdirectory(book)
 add_subdirectory(examples)
 
+find_package(QT NAMES Qt6 Qt5 COMPONENTS Core Gui Test REQUIRED)
```
ahayzen-kdab, line 152: Can we have a comment above this block describing what it is doing ? It looks like it is
- Retrieving the install location of Qt
- Passing the qmake location to cargo later (but only for the tests?)
- Settings the target directory to be the `root folder/target`, this is then for the cxx-qt-lib cargo target folder for the tests?

And i assume this is just to help the tests and not relevant for real applications.

cmake/CxxQt.cmake
```diff
@@ -41,9 +41,14 @@ function(cxx_qt_generate_cpp GEN_SOURCES)
 
     file(MAKE_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/target/cxx-qt-gen")
 
+    get_target_property(QMAKE Qt${QT_VERSION_MAJOR}::qmake IMPORTED_LOCATION)
+
     # Run cargo during config to ensure the cpp source file list is created
     execute_process(
-        COMMAND ${CARGO_CMD}
+        COMMAND cmake -E env
```
ahayzen-kdab, line 48: add some comments here as to why we need and what are CARGO_TARGET_DIR and QMAKE

cmake/CxxQt.cmake
```diff
@@ -100,9 +106,15 @@ function(cxx_qt_link_rustlib APP_NAME)
     else()
         set(RUST_PART_LIB "${CMAKE_CURRENT_SOURCE_DIR}/target/${TARGET_DIR}/librust.a")
     endif()
+
+    get_target_property(QMAKE Qt${QT_VERSION_MAJOR}::qmake IMPORTED_LOCATION)
+
     add_custom_target(
         "${APP_NAME}_rustlib"
-        COMMAND ${CARGO_CMD}
+        COMMAND cmake -E env
```
ahayzen-kdab, line 114: same again

tests/basic_cxx_qt/CMakeLists.txt
```diff
@@ -16,9 +16,6 @@ set(CMAKE_AUTOMOC ON)
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
-find_package(QT NAMES Qt6 Qt5 COMPONENTS Core Gui Test REQUIRED)
```
ahayzen-kdab, line 19: Think these lines should still be there, as i know they come from the global root CMake scope, but it is just showing what the usecase might look like for the downstream applications. (and the other examples do have the find_package's)

qt-build/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+/// Helper for build.rs scripts using Qt
+/// ```
+/// let qt_modules = vec!["Core", "Gui"]
+///     .iter()
+///     .map(|m| String::from(*m))
+///     .collect();
+/// let qtbuild = qt_build::QtBuild::new(qt_modules).expect("Could not find Qt installation");
+/// ```
+pub struct QtBuild {
+    pub version: SemVer,
```
ahayzen-kdab, line 31: this field doesn't need to be `pub` ?

CMakeLists.txt
```diff
@@ -149,19 +149,28 @@ endfunction()
...
+find_package(Qt${QT_VERSION_MAJOR} COMPONENTS Core Gui Test REQUIRED)
+get_target_property(QMAKE Qt${QT_VERSION_MAJOR}::qmake IMPORTED_LOCATION)
+set(CARGO_ENV "QMAKE=${QMAKE};CARGO_TARGET_DIR=${CMAKE_CURRENT_SOURCE_DIR}/target")
+
 # Create helper method which adds relevent cargo tests for a given manifest
 function(add_test_cargo TEST_NAME_PREFIX MANIFEST_PATH ADD_DOCTESTS)
```
ahayzen-kdab, line 158: can we add a `add_test_cargo` line for `qt-build` like we have for the other Rust crates, so that we ensure that fmt/clippy/tests are run.

qt-build/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+# SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+# SPDX-FileContributor: Be Wilson <be.wilson@kdab.com>
+#
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+[package]
+name = "qt-build"
```
ahayzen-kdab, line 7: Once this is stablish API wise, wonder if this should be added to it's own repository ? Eg `qt-build-rs` like we did with `clang-format-rs` etc ?

qt-build/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+    /// as the build system for the whole build, prefer using `QT_VERSION_MAJOR` over the `QMAKE`
+    /// environment variable because it will account for different names for the qmake executable
+    /// that some Linux distributions use.
+    ///
+    /// However, when building a Rust staticlib that gets linked to C++ code by a C++ build
+    /// system, it is best to use the `QMAKE` environment variable to ensure that the Rust
+    /// staticlib is linked to the same installation of Qt that the C++ build system has
+    /// detected. With CMake, you can get this from the `Qt${QT_VERSION_MAJOR}::qmake`
+    /// target's `IMPORTED_LOCATION` property, for example:
+    /// ```
```
ahayzen-kdab, line 57: could put ```cmake as formatters might have colouring for cmake syntax ?

qt-build/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+    /// )
+    /// ```
+    #[allow(clippy::result_unit_err)]
+    pub fn new(mut qt_modules: Vec<String>) -> Result<Self, ()> {
+        if qt_modules.is_empty() {
+            qt_modules.push("Core".to_string());
+        }
+        println!("cargo:rerun-if-env-changed=QMAKE");
+        println!("cargo:rerun-if-env-changed=QT_VERSION_MAJOR");
+        fn verify_candidate(candidate: &str) -> Result<(&str, versions::SemVer), String> {
```
ahayzen-kdab, line 77: having a nested function seems slightly odd to me, can it be a private function implement on `QtBuild` ? Or a closure, but even a closure means this `new` function is large, whereas moving the function out keeps each of them small.

qt-build/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+    ///
+    /// execute_process(
+    ///     COMMAND cmake -E env
+    ///         "CARGO_TARGET_DIR=${CMAKE_CURRENT_BINARY_DIR}/cargo"
+    ///         "QMAKE=${QMAKE}"
+    ///         cargo build
+    ///     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+    /// )
+    /// ```
+    #[allow(clippy::result_unit_err)]
```
ahayzen-kdab, line 70: can we remove this clippy lint change (see further comments below...)

qt-build/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+    pub fn new(mut qt_modules: Vec<String>) -> Result<Self, ()> {
```
ahayzen-kdab, line 71: Add an enum of errors, so then change this to
```
pub enum QtBuildErrors {
  QMakeSetQtMissing,
  QtMissing
};

pub fn new(mut qt_modules: Vec<String>) -> Result<Self, QtBuildErrors>{
...
}
```
This then means uses of it can still do `qt_build::QtBuild::new(qt_modules).expect("Could not find Qt installation")` but also means we remove the panic and we have meaning to the error results.

qt-build/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+        if let Ok(qmake_env_var) = env::var("QMAKE") {
+            match verify_candidate(qmake_env_var.trim()) {
+                Ok((executable_name, version)) => {
+                    return Ok(Self {
+                        qmake_executable: executable_name.to_string(),
+                        version,
+                        qt_modules,
+                    });
+                }
+                Err(e) => panic!(
```
ahayzen-kdab, line 122: change this to return an error enum

qt-build/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+                        qmake_executable: executable_name.to_string(),
+                        version,
+                        qt_modules,
+                    });
+                }
+                Err(_) => continue,
+            }
+        }
+
+        Err(())
```
ahayzen-kdab, line 143: return an error enum

cxx-qt-lib/src/lib.rs
```diff
@@ -11,33 +11,6 @@ pub use types::*;
 // this is because include_str requires th correct and non-mixed path separators
```
ahayzen-kdab, line 11: remove this comment block as well ?

qt-build/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+            qt_modules.push("Core".to_string());
+        }
+        println!("cargo:rerun-if-env-changed=QMAKE");
+        println!("cargo:rerun-if-env-changed=QT_VERSION_MAJOR");
+        fn verify_candidate(candidate: &str) -> Result<(&str, versions::SemVer), String> {
+            match Command::new(&candidate)
+                .args(&["-query", "QT_VERSION"])
+                .output()
+            {
+                Err(_) => Err("qmake -query failed".to_string()),
```
ahayzen-kdab, line 82: So could even have these inner QMake errors be their own Enum and then nest that into the outer enum or share the same one.

qt-build/src/lib.rs
Be-ing: Downstream users might want to have some code conditional on the Qt version.

Be-ing: returning a const reference is a better idea

qt-build/src/lib.rs
Be-ing: This is intentionally within the `new` function because it isn't needed by anything else: https://www.reddit.com/r/rust/comments/1ob8l7/declaring_functions_within_functions/

qt-build/Cargo.toml
Be-ing: I don't mind either way.

qt-build/src/lib.rs
Be-ing: rustdoc doesn't have syntax highlighting for CMake, but I suppose it doesn't hurt to add that anyway

qt-build/Cargo.toml
Be-ing: FWIW I set the initial version of qt-build to 0.3 to match the rest of the cxx-qt crates, but if we want to move it out of this repo, maybe it should initially be published with version 0.1.

ahayzen-kdab: I think for now while things are being developed lets keep it together. If it is separated out the version doesn't really matter, as it'll likely head rapidly towards 1.0 anyway :-)

qt-build/src/lib.rs
ahayzen-kdab: OK, it was more a styling thing anyway of trying to keep functions short, lets go with it for now.

ahayzen-kdab (approved):
LGTM :-)

ahayzen-kdab (changes requested):
Hmm i think we either need to fix the locking issue or find a workaround, it's failure rate is too high to accept even for a flaky temporary thing.

I guess possible solutions are
- Add a commit to use cargo workspaces in this branch
- Let rust wait / keep trying until it's successful in copying the file
- Tell CMake to not run the execute_process in parallel to other CMake tasks

cxx-qt-lib/build.rs
```diff
@@ -3,138 +3,77 @@
...
+    // build system, in which case CARGO_TARGET_DIR will be set by
+    // the C++ build system.
+    println!("cargo:rerun-if-changed=include/qt_types.h");
+    println!("cargo:rerun-if-env-changed=CARGO_TARGET_DIR");
+    if let Ok(target_dir) = env::var("CARGO_TARGET_DIR") {
+        let manifest_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
+        std::fs::create_dir_all(&format!("{}/cxxbridge/cxx-qt-lib/include", target_dir)).unwrap();
+        // FIXME: Horrible hack around this sometimes failing because
+        // Windows doesn't allow multiple processes to access a file by default.
+        loop {
```
ahayzen-kdab, line 28: assuming this does work, can we only compile the `loop` for Windows as well ? or do we rely on the `break`, `continue`, and panic. And should there be a sleep to avoid burning CPU ?

Be-ing: I added a 100ms sleep, but I don't think compiling the loop only for Windows would be worth the trouble.

ahayzen-kdab (approved):
LGTM :-)

PR #177, opened 2022-08-01
cxx-qt-gen: start a new syntax folder to help processing syntax

ahayzen-kdab:
This starts to split up the roles, later there will be more things
in syntax and other separate folders.

### DIFF
--- a/cxx-qt-build/src/lib.rs
+++ b/cxx-qt-build/src/lib.rs
@@ -10,11 +10,12 @@ use std::env;
 use std::fs::File;
 use std::io::Write;
 use std::path::PathBuf;
-use syn::*;
 
 use clang_format::ClangFormatStyle;
 use cxx_qt_gen::{
-    extract_qobject, generate_format, generate_qobject_cpp, generate_qobject_cxx, CppObject,
+    extract_qobject, generate_format, generate_qobject_cpp, generate_qobject_rs,
+    syntax::{parse_qt_file, CxxQtItem},
+    CppObject,
 };
 
 /// Representation of a generated CXX header, source, and name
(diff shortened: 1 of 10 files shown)


Be-ing (approved):
More modular and better documented code :+1:

cxx-qt-build/src/lib.rs
```diff
@@ -93,57 +48,6 @@ fn manifest_dir() -> String {
     manifest_dir
 }
 
-/// Extract the cxx or cxx_qt module from a Rust file
-fn extract_modules(file_content: &str, rs_path: &impl AsRef<std::path::Path>) -> ExtractedModule {
```
LeonMatthesKDAB, line 97: It seems like you inlined this method into `GeneratedCpp::new`.
Any reason for this? It makes the `new` method pretty large.

ahayzen-kdab: The `extract_modules` is now effectively in `syntax::parse_qt_file`. This then responds with the enum that GeneratedCpp::new iterates through to find any CXX/CXX-Qt blocks. Before the logic of building the enum was in cxx-qt-build, now it's over in syntax of cxx-qt-gen and using syn in a more correct way. I have further changes later to make cxx-qt-gen much more modular, which then cxx-qt-build will use etc.

LeonMatthesKDAB: Hm, okay, that is indeed cleaner.

I just think that `new` function is a bit too large now and has things in it that are at a separate level of abstraction.
Just putting that check-loop in a function `fn ensure_single_bridge` or something like that I think would clean that up a bit :)
But not a high-priority.

ahayzen-kdab: Right, a separate refactor can be done in here too. And eventually we want to support multiple bridges per file ... :-)

PR #178, opened 2022-08-01
cxx-qt-gen: use impl cxx_qt::QObject<T> block for invokables

ahayzen-kdab:
This helps later as the invokables are implemented onto a different
struct, so this clearly defines the split between C++ and Rust
methods etc.

### DIFF
--- a/book/src/qobject/rustobj_struct.md
+++ b/book/src/qobject/rustobj_struct.md
@@ -20,7 +20,10 @@ The RustObj struct allows you to define the following items
 
 ## Invokables
 
-To define a method which is exposed to QML and C++, add a method on the `RustObj` struct and add the attribute `#[invokable]`. The parameters and return type are then matched to the Qt side. Also CXX-Qt automatically adds wrapper code around your invokable to automatically perform any conversion between the [C++ and Rust types](../concepts/types.md).
+A `impl cxx_qt::QObject<RustObj>` is used to define invokables, the `impl cxx_qt::QObject<RustObj>` defines that the methods are implemented onto the C++ QObject.
+Therefore they have access to both C++ and Rust methods. Also CXX-Qt adds wrapper code around your invokables to automatically perform any conversion between the [C++ and Rust types](../concepts/types.md).
+
+To mark a method as invokable simply add the `#[invokable]` attribute to the Rust method. This then causes `Q_INVOKABLE` to be set on the C++ definition of the method, allowing QML to call the invokable.
 
 Note to access properties on the C++ object use [Cpp Object](./cpp_object.md).
 
@@ -28,4 +31,6 @@ Note to access properties on the C++ object use [Cpp Object](./cpp_object.md).
 
 Unlike the [Data Struct](./data_struct.md) fields which are defined on the `RustObj` struct are not exposed as properties to Qt. These can be considered as "private to Rust" fields, and are useful for storing channels for threading or internal information for the QObject.
 
-Methods implemented on the `RustObj` that do not have an `#[invokable]` attribute are not exposed to C++ and are considered "private to Rust" methods. Similar to fields these are useful for threading and internal information.
+Methods implemented using `impl RustObj` (and not `impl cxx_qt::QObject<RustObj>`) are just normal Rust member methods.
+Therefore they do not have access to any C++ or QObject functionality (e.g. emitting Signals, changing properties, etc.)
+You will usually only need to use `impl RustObj` if you want to also use your RustObj struct as a normal Rust struct, that is not wrapped in a QObject.
--- a/cxx-qt-gen/src/extract.rs
+++ b/cxx-qt-gen/src/extract.rs
@@ -171,8 +171,10 @@ pub struct QObject {
     pub ident: Ident,
     /// All the methods that can also be invoked from QML
     pub(crate) invokables: Vec<Invokable>,
-    /// All the methods that cannot be invoked from QML
+    /// All the methods that cannot be invoked from QML (but are on the C++ object)
     pub(crate) normal_methods: Vec<ImplItemMethod>,
+    /// All the rust only methods
+    pub(crate) rust_methods: Vec<ImplItem>,
     /// All the properties that can be used from QML
     pub(crate) properties: Vec<Property>,
     /// All the signals that can be emitted/connected from QML
@@ -466,7 +468,7 @@ fn extract_invokables(
             .cloned()
             .collect();
 
-        // Skip non-invokable members
+        // Non invokable methods we just pass through as normal methods
         if filtered_attrs.len() == method.attrs.len() {
             normal_methods.push(method);
             continue;
@@ -847,6 +849,8 @@ pub fn extract_qobject(
     let mut object_invokables = vec![];
     // A list of the normal methods (i.e. not invokables) for the struct
     let mut object_normal_methods = vec![];
+    // A list of the rust only methods for the struct
+    let mut object_rust_methods = vec![];
     // A list of original trait impls for the struct (eg `impl Default for Struct`)
     let mut original_trait_impls = vec![];
     // A list of insignificant declarations for the mod that will be directly passed through (eg `use crate::thing`)
@@ -943,7 +947,7 @@ pub fn extract_qobject(
                 // Extract the path from the type (this leads to the struct name)
                 if let Type::Path(TypePath { path, .. }) = &mut *original_impl.self_ty {
                     // Check that the path contains segments
-                    if path.segments.len() != 1 {
+                    if path.segments.is_empty() {
                         return Err(Error::new(
                             original_impl.span(),
                             "Invalid path on impl block.",
@@ -1008,15 +1012,23 @@ pub fn extract_qobject(
                                     _others => original_trait_impls.push(original_impl.to_owned()),
                                 }
                             } else {
-                                let mut extracted = extract_invokables(
-                                    &original_impl.items,
-                                    cpp_namespace_prefix,
-                                    &qt_ident,
-                                )?;
-
-                                object_invokables.append(&mut extracted.invokables);
-                                object_normal_methods.append(&mut extracted.normal_methods);
+                                // Rust only methods are on the impl T block
+                                object_rust_methods.append(&mut original_impl.items);
                             }
+                        }
+                        // Invokables are defined in the impl cxx_qt::QObject<T> block
+                        "cxx_qt"
+                            if path.segments.len() > 1
+                                && path.segments[1].ident.to_string().as_str() == "QObject" =>
+                        {
+                            let mut extracted = extract_invokables(
+                                &original_impl.items,
+                                cpp_namespace_prefix,
+                                &qt_ident,
+                            )?;
+
+                            object_invokables.append(&mut extracted.invokables);
+                            object_normal_methods.append(&mut extracted.normal_methods);
                         }
                         _others => {
                             return Err(Error::new(
@@ -1130,6 +1142,7 @@ pub fn extract_qobject(
         ident: qt_ident,
         invokables: object_invokables,
         normal_methods: object_normal_methods,
+        rust_methods: object_rust_methods,
         properties: object_properties,
         signals: object_signals,
         signal_ident,
@@ -1394,7 +1407,8 @@ mod tests {
         assert!(invokable.mutable);
 
         // Check that the normal method was also detected
-        assert_eq!(qobject.normal_methods.len(), 1);
+        assert_eq!(qobject.normal_methods.len(), 4);
+        assert_eq!(qobject.rust_methods.len(), 1);
     }
 
     #[test]
@@ -1458,7 +1472,8 @@ mod tests {
         // Check that it got the inovkables and properties
         assert_eq!(qobject.invokables.len(), 0);
         assert_eq!(qobject.properties.len(), 1);
-        assert_eq!(qobject.normal_methods.len(), 2);
+        assert_eq!(qobject.normal_methods.len(), 0);
+        assert_eq!(qobject.rust_methods.len(), 2);
 
         // Check that there is a use, enum and fn declaration
         assert_eq!(qobject.original_passthrough_decls.len(), 13);
--- a/cxx-qt-gen/test_outputs/invokables.rs
+++ b/cxx-qt-gen/test_outputs/invokables.rs
@@ -156,8 +156,24 @@ mod cxx_qt_my_object {
             QString::from_str("static")
         }
 
+        pub fn cpp_context_method(&self) {
+            println!("C++ context method");
+        }
+
+        pub fn cpp_context_method_mutable(&mut self) {
+            println!("mutable method");
+        }
+
+        pub fn cpp_context_method_cpp_obj(&mut self, cpp: &mut CppObj) {
+            println!("cppobj");
+        }
+
+        pub fn cpp_context_method_return_opaque(&self) -> UniquePtr<QColor> {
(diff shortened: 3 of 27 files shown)


ahayzen-kdab:
@LeonMatthesKDAB  did we decide if we were going to call this `impl Qt<T>` or `impl cxx_qt<T>` ? As i think the `Qt` felt more natural for saying these are "Qt" invokables, but the `cxx_qt` was better to avoid clashes ? Maybe even `cxx_qt::invokables` could be considered as that then matches `cxx_qt::signals` etc ?

book/src/qobject/rustobj_struct.md
```diff
@@ -20,12 +20,12 @@ The RustObj struct allows you to define the following items
 
 ## Invokables
 
-To define a method which is exposed to QML and C++, add a method on the `RustObj` struct and add the attribute `#[invokable]`. The parameters and return type are then matched to the Qt side. Also CXX-Qt automatically adds wrapper code around your invokable to automatically perform any conversion between the [C++ and Rust types](../concepts/types.md).
+To define a method which is exposed to QML and C++, add a method on the `RustObj` struct, with `impl Qt<RustObj>`, and add the attribute `#[invokable]`. The parameters and return type are then matched to the Qt side. Also CXX-Qt automatically adds wrapper code around your invokable to automatically perform any conversion between the [C++ and Rust types](../concepts/types.md).
```
Be-ing, line 23:
```suggestion
To define a method which is exposed to QML and C++, add a method on the `RustObj` struct, with `impl Qt<RustObj>`, and add the attribute `#[invokable]`. The parameters and return type are then matched to the Qt side. Also CXX-Qt adds wrapper code around your invokable to automatically perform any conversion between the [C++ and Rust types](../concepts/types.md).
```

`automatically` is repeated

cxx-qt-gen/src/extract.rs
```diff
@@ -960,16 +951,20 @@ pub fn extract_qobject(
...
-                                let mut extracted = extract_invokables(
-                                    &original_impl.items,
-                                    cpp_namespace_prefix,
-                                    &qt_ident,
-                                )?;
-
-                                object_invokables.append(&mut extracted.invokables);
-                                object_normal_methods.append(&mut extracted.normal_methods);
+                                // Invokables aren't on the impl T block
+                                object_normal_methods.append(&mut original_impl.items);
```
Be-ing, line 955: Should using `#[invokable]` outside of `impl Qt<RustObj>` cause an error?

ahayzen-kdab: I think it fails to build as there is no macro called `invokable` :-)

ahayzen-kdab:
> @LeonMatthesKDAB did we decide if we were going to call this `impl Qt<T>` or `impl cxx_qt<T>` ? As i think the `Qt` felt more natural for saying these are "Qt" invokables, but the `cxx_qt` was better to avoid clashes ? Maybe even `cxx_qt::invokables` could be considered as that then matches `cxx_qt::signals` etc ?

So i'm wondering if this is the route (assuming valid syntax)

```rust
impl cxx_qt::invokables<MyObject> {
  #[invokable]
  pub fn my_invokable(self: Pin<&mut Self>) {

  }
}

impl MyObject {
  pub fn rust_only_method(&self) {

  }
}
```

@LeonMatthesKDAB where did our discussion end up on having like `#[invokable(return-as-value)]` and `#[property(as-value)]` stuff for `UniquePtr` end up? I can't see it now as there were options like `#[invokable(return QColor)]` as well...

LeonMatthesKDAB:
> So i'm wondering if this is the route (assuming valid syntax)
> ```
> impl cxx_qt::invokables<MyObject> {
>   #[invokable]
>   pub fn my_invokable(self: Pin<&mut Self>) {
> 
>   }
> }
> 
> impl MyObject {
>   pub fn rust_only_method(&self) {
> 
>   }
> }
> ```
We shouldn't use `impl invokables<MyObject>`, as it still makes sense to have non-invokable (aka not to C++ exposed) methods defined within the `impl Qt<MyObject>` (or whatever we end up calling it).
Whilst you can define Rust-only methods directly on MyObject, you might want to have a private method that is still capable of emitting a Signal for example. That you would implement on `Qt<MyObject>` but without `#[invokable]`.

But regarding the `cxx_qt::` prefix. That does make sense for consistencies sake.
So something like: `impl cxx_qt::QObject<MyObject>` or `impl cxx_qt::Qt<MyObject>`. My local playground project even uses `impl cxx_qt<MyObject>`, but maybe `impl cxx_qt::Qt<MyObject>` is a bit more consistent with the other macros we have.
Later we could add the ability to add a `use cxx_qt::Qt` or `use cxx_qt::prelude::*`, to reduce the boilerplate.

> @LeonMatthesKDAB where did our discussion end up on having like #[invokable(return-as-value)] and #[property(as-value)] stuff for UniquePtr end up? I can't see it now as there were options like #[invokable(return QColor)] as well...
> 
I'm unsure whether we reached a consensus on that or whether we were still just discussing options :thinking: 
Do you need a decision on this for this PR already?

examples/qml_with_threaded_logic/src/lib.rs
```diff
@@ -124,6 +115,17 @@ mod website {
         }
     }
 
+    impl RustObj {
+        fn process_event(&mut self, event: &Event, cpp: &mut CppObj) {
```
LeonMatthesKDAB, line 119: This needs to be in the `impl Qt<MyObject>` block, not in the `impl RustObj` block, as we need access to the C++ functionality.

cxx-qt-gen/src/extract.rs
```diff
@@ -417,23 +417,13 @@ fn extract_type_ident(
     })
 }
 
-/// The result of extracting invokables from an impl block.
-/// Only intended for internal use for now.
-struct ExtractedInvokables {
-    /// Impl methods that will also be exposed as invokables to Qt
-    invokables: Vec<Invokable>,
-    /// Impl methods that will only be visible on the Rust side
-    normal_methods: Vec<ImplItemMethod>,
```
LeonMatthesKDAB, line 426: As mentioned, we still need these `normal_methods`.

book/src/qobject/rustobj_struct.md
```diff
@@ -20,12 +20,12 @@ The RustObj struct allows you to define the following items
 
 ## Invokables
 
-To define a method which is exposed to QML and C++, add a method on the `RustObj` struct and add the attribute `#[invokable]`. The parameters and return type are then matched to the Qt side. Also CXX-Qt automatically adds wrapper code around your invokable to automatically perform any conversion between the [C++ and Rust types](../concepts/types.md).
+To define a method which is exposed to QML and C++, add a method on the `RustObj` struct, with `impl Qt<RustObj>`, and add the attribute `#[invokable]`. The parameters and return type are then matched to the Qt side. Also CXX-Qt adds wrapper code around your invokable to automatically perform any conversion between the [C++ and Rust types](../concepts/types.md).
```
LeonMatthesKDAB, line 23: Maybe split this up into two paragraphs:
1. What is `impl Qt<RustObj>` (aka, implementation on the C++ QObject that wraps the RustObj - It has access to both C++ and Rust methods).
2. How can I define an invokable on it (#[invokable])

book/src/qobject/rustobj_struct.md
```diff
@@ -20,12 +20,12 @@ The RustObj struct allows you to define the following items
...
 
 Note to access properties on the C++ object use [Cpp Object](./cpp_object.md).
 
 ## Private Methods and Fields
 
 Unlike the [Data Struct](./data_struct.md) fields which are defined on the `RustObj` struct are not exposed as properties to Qt. These can be considered as "private to Rust" fields, and are useful for storing channels for threading or internal information for the QObject.
 
-Methods implemented on the `RustObj` that do not have an `#[invokable]` attribute are not exposed to C++ and are considered "private to Rust" methods. Similar to fields these are useful for threading and internal information.
+Methods implemented on the `RustObj` (and not the `impl Qt<RustObj>`) that do not have an `#[invokable]` attribute are not exposed to C++ and are considered "private to Rust" methods. Similar to fields these are useful for threading and internal information.
```
LeonMatthesKDAB, line 31: Maybe rephrase this a bit to focus on the access to C++ features.
> Methods implemented using `impl RustObj` (and not `impl Qt<RustObj>`) are just normal Rust member methods.
> Therefore they do not have access to any C++/QObject functionality (e.g. emitting Signals, changing properties, etc.)
> You will usually only need to use `impl RustObj` if you want to also use your `RustObj` struct as a normal Rust struct, that is not wrapped in a QObject.

LeonMatthesKDAB (changes requested):
Going in the right direction. A bit of clarification regarding the purposes of `impl Qt<...>` vs `impl ...` are still necessary :)

examples/qml_with_threaded_logic/src/lib.rs
ahayzen-kdab: Actually this one is a rust private method (as Event isn't a supported type), what happens is that the handle_update_request calls this method and passes in CppObj. But i've added support for both anyway.

LeonMatthesKDAB: Yes, but it will in future still need to be implemented in the `impl MyObjectQt` block (e.g. `impl cxx_qt::QObject<T>`).

That doesn't mean it needs to be put in the CXX bridge, so it will still be rust-private.

book/src/qobject/rustobj_struct.md
ahayzen-kdab: Done

book/src/qobject/rustobj_struct.md
ahayzen-kdab: Done

cxx-qt-gen/src/extract.rs
ahayzen-kdab: Done

cxx-qt-gen/src/extract.rs
```diff
@@ -1008,16 +1014,24 @@ pub fn extract_qobject(
...
-                                object_invokables.append(&mut extracted.invokables);
-                                object_normal_methods.append(&mut extracted.normal_methods);
+                                // Rust only methods are on the impl T block
+                                object_rust_methods.append(&mut original_impl.items);
                             }
                         }
+                        // Invokables are defined in the impl Qt<T> block
+                        "cxx_qt"
+                            if path.segments.len() > 1
+                                && path.segments[1].ident.to_string().as_str() == "QObject" =>
```
LeonMatthesKDAB, line 1024: The premature optimizer monkey in my brain tells me this can probably be "simplified":
`path.segments.get(1).filter(|seg| seg.ident.to_string().as_str() == "QObject").is_some()`
Then we don't need the segments.len() check ;)

Anyway, interesting mental exercise, not important

cxx-qt-gen/test_outputs/invokables.h
```diff
@@ -36,6 +36,11 @@ class MyObject : public QObject
   Q_INVOKABLE qint32 invokableReturnPrimitive();
   Q_INVOKABLE QString invokableReturnStatic();
 
+  void cppOnlyMethod();
```
LeonMatthesKDAB, line 39: These 4 methods should actually not be generated into C++.
They just need to be in the `impl cxx_qt::QObject<T>` block because that's the only place, where, in future, you'll still have access to C++ functionality (e.g. signals, etc.).

cxx-qt-gen/test_outputs/invokables.rs
```diff
@@ -156,8 +165,33 @@ mod cxx_qt_my_object {
             QString::from_str("static")
         }
 
+        pub fn cpp_only_method_cpp_obj_wrapper(&mut self, cpp: std::pin::Pin<&mut FFICppObj>) {
```
LeonMatthesKDAB, line 168: We shouldn't generate a wrapper for this.
It should just be 1-to-1 translated over into the `impl RustObj` block, and later the `impl MyObjectQt` block.

cxx-qt/src/lib.rs
```diff
@@ -60,6 +64,38 @@ pub fn signals(_args: TokenStream, _input: TokenStream) -> TokenStream {
...
+///         #[invokable]
+///         fn invokable(&self, a: i32, b: i32) -> i32 {
+///             a + b
+///         }
+///     }
+/// }
+/// ```
+#[proc_macro]
+#[allow(non_snake_case)]
+pub fn QObject(_input: TokenStream) -> TokenStream {
```
LeonMatthesKDAB, line 95: This shouldn't be a macro, it needs to be a type:
``` rust
pub struct QObject<T> {
    phantom: std::marker::PhantomData<T>
}
```

examples/qml_with_threaded_logic/src/lib.rs
LeonMatthesKDAB: In my opinion, this still needs to be in `impl cxx_qt::QObject<RustObj>`, as it modifies a C++ property (e.g. `cpp.set_title`).

LeonMatthesKDAB (changes requested):

cxx-qt/src/lib.rs
ahayzen-kdab: We can't export non macro things in a proc-macro crate, so it has to be this.

cxx-qt-gen/test_outputs/invokables.rs
ahayzen-kdab: Eventually yup, this is just the first step, still using the existing wrapper generation. As this impl cxx_qt::QObject block is not implemented on the C++ object yet.

cxx-qt-gen/test_outputs/invokables.h
ahayzen-kdab: Eventually yes, when we move the implementation of the invokables onto the C++ object. For now we still have the proxies.

cxx-qt-gen/test_outputs/invokables.rs
ahayzen-kdab: OK, as discussed i see what you mean now, we need these only on the Rust side, but on the C++ object later.

cxx-qt-gen/test_outputs/invokables.h
ahayzen-kdab: Removed :-)

book/src/qobject/rustobj_struct.md
```diff
@@ -20,12 +20,17 @@ The RustObj struct allows you to define the following items
 
 ## Invokables
 
-To define a method which is exposed to QML and C++, add a method on the `RustObj` struct and add the attribute `#[invokable]`. The parameters and return type are then matched to the Qt side. Also CXX-Qt automatically adds wrapper code around your invokable to automatically perform any conversion between the [C++ and Rust types](../concepts/types.md).
+A `impl cxx_qt::QObject<RustObj>` is used to define invokables, the `impl cxx_qt::QObject<RustObj>` defines that the methods are implemented onto the C++ QObject.
+Therefore they have access to both C++ and Rust methods. Also CXX-Qt adds wrapper code to around your invokables to automatically perform any conversion between the [C++ and Rust types](../concepts/types.md).
```
LeonMatthesKDAB, line 24: `wrapper code to around` => `wrapper code around`

LeonMatthesKDAB (commented):
Just one typo, then we're good to go :)

book/src/qobject/rustobj_struct.md
ahayzen-kdab: Fixed

LeonMatthesKDAB (approved):
LGTM

PR #179, opened 2022-08-01
cxx-qt-gen: have MyObjectRust on C++ side and MyObjectQt on Rust

ahayzen-kdab:
This helps later to avoid collisions in C++ with RustObj, when the
CXX bridge can be used it is more explicit which type is being used,
and generally makes it more obvious which type is being referred to.

### DIFF
--- a/cxx-qt-gen/test_outputs/custom_default.rs
+++ b/cxx-qt-gen/test_outputs/custom_default.rs
@@ -3,7 +3,8 @@ mod my_object {
     unsafe extern "C++" {
         include!("cxx-qt-gen/include/my_object.cxxqt.h");
 
-        type MyObject;
+        #[cxx_name = "MyObject"]
+        type MyObjectQt;
 
         include!("cxx-qt-lib/include/qt_types.h");
         #[namespace = ""]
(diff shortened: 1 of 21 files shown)


Be-ing (approved):
yay disambiguation

cxx-qt-gen/test_outputs/custom_default.rs
```diff
@@ -34,22 +35,23 @@ mod my_object {
...
+        fn setPublic(self: Pin<&mut MyObjectQt>, value: i32);
 
         #[rust_name = "new_cpp_object"]
-        fn newCppObject() -> UniquePtr<MyObject>;
+        fn newCppObject() -> UniquePtr<MyObjectQt>;
     }
 
     extern "Rust" {
+        #[cxx_name = "MyObjectRust"]
         type RustObj;
```
LeonMatthesKDAB, line 48: this will in future be the real `MyObject`, right?

But I guess that's out of scope for now :thinking:

LeonMatthesKDAB (commented):
Like @Be-ing said, more disambiguation, more better :)

cxx-qt-gen/test_outputs/custom_default.rs
ahayzen-kdab: Right, one of the next changes is to drop the `Data` struct and just have the `RustObj` struct. Then add the qobject macro and use the RustObj struct ident as the QObject ident.... baby steps :-)

ahayzen-kdab: (and potentially either within those changes or separately moving the properties from C++ -> Rust)

PR #180, opened 2022-08-01
cxx-qt-gen: use cxx_qt::signals attribute to define signals

ahayzen-kdab:
This allows us later to pick which QObject signals refer to and
easily allows us to determine which blocks are cxx_qt and which
are for CXX.

### DIFF
--- a/cxx-qt/src/lib.rs
+++ b/cxx-qt/src/lib.rs
@@ -40,6 +40,25 @@ pub fn bridge(_attr: TokenStream, input: TokenStream) -> TokenStream {
     extract_and_generate(module, &cpp_namespace_prefix)
 }
 
+/// A macro which describes that an enum the definition of signals for a QObject
+///
+/// It should not be used by itself and instead should be used inside a cxx_qt::bridge definition.
+///
+/// # Example
+///
+/// ```ignore
+/// #[cxx_qt::bridge]
+/// mod my_object {
+///     #[cxx_qt::signals(MyObject)]
+///     enum MySignals {
+///         Ready,
+///     }
+/// }
+#[proc_macro_attribute]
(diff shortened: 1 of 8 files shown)


Be-ing (dismissed):
yay less magic

Be-ing (approved):

cxx-qt/src/lib.rs
```diff
@@ -40,6 +40,25 @@ pub fn bridge(_attr: TokenStream, input: TokenStream) -> TokenStream {
     extract_and_generate(module, &cpp_namespace_prefix)
 }
 
+/// A macro which describes that an enum the definition of signals for a QObject
+///
+/// It should not be used by itself and instead should be used inside a cxx_qt::bridge definition.
+///
+/// # Example
+///
+/// ```ignore
```
LeonMatthesKDAB, line 49: Missing a closing ``` here.

cxx-qt/src/lib.rs
```diff
@@ -40,6 +40,25 @@ pub fn bridge(_attr: TokenStream, input: TokenStream) -> TokenStream {
     extract_and_generate(module, &cpp_namespace_prefix)
 }
 
+/// A macro which describes that an enum the definition of signals for a QObject
```
LeonMatthesKDAB, line 43: We're either missing an `is` here, or we need a small rephrasing:
`A macro which describes that an enum defines the signals for a QObject`.

cxx-qt/src/lib.rs
ahayzen-kdab: #186

cxx-qt/src/lib.rs
ahayzen-kdab: #186

PR #181, opened 2022-08-02
cxx-qt-gen: add rust getters for immutable and mutable access

ahayzen-kdab:
This helps later when invokables use a pointer to the C++ object
rather than the Rust object, as then they can use self.rust().prop
or unsafe { self.as_mut().rust_mut().prop = value; } etc.

There will also be helpers implemented onto the C++ object so that
for normal usecases they use set_prop and get_prop etc.

### DIFF
--- a/cxx-qt-gen/src/gen_cpp.rs
+++ b/cxx-qt-gen/src/gen_cpp.rs
@@ -973,6 +973,8 @@ pub fn generate_qobject_cpp(obj: &QObject) -> Result<CppObject, TokenStream> {
         public:
             explicit {ident}(QObject *parent = nullptr);
             ~{ident}();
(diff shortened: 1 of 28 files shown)


ahayzen-kdab:
Do we need to worry about locking ? Can we hide these via friend from C++ ? etc.

Be-ing:
There are merge conflicts.

ahayzen-kdab:
And discussions to figure out, hence still a draft and not ready for review/merging yet :-)

ahayzen-kdab:
OK, so we are going with the rust/rust_mut methods. On C++ they are called unsafe_rust/unsafe_rust_mut to signify that there is no locking in these methods and therefore unsafe.

LeonMatthesKDAB (approved):
LGTM

PR #182, opened 2022-08-02
cxx-qt-lib: remove links from cargo as it's not used anymore

ahayzen-kdab:


ahayzen-kdab:
This is still required.

PR #187, opened 2022-08-04
cxx-qt-gen: add a parser module for creating an abstract QObject

ahayzen-kdab:
This module will replace the extract.rs, taking syn ItemMod as an
input and then rearranging the Item's into an abstract
representation of the QObject's in the module.

Later a generate module will then take this abstract representation
and convert it into a RustObject and a CppObject, which will then
be passed to a writer module to create source files.


cxx-qt-gen/src/extract.rs
```diff
@@ -828,278 +825,63 @@ pub fn extract_qobject(
...
     // A list of the invokables for the struct
-    let mut object_invokables = vec![];
+    let object_invokables = extracted.invokables;
     // A list of the normal methods (i.e. not invokables) for the struct
-    let mut object_normal_methods = vec![];
-    // A list of the rust only methods for the struct
-    let mut object_rust_methods = vec![];
-    // A list of original trait impls for the struct (eg `impl Default for Struct`)
-    let mut original_trait_impls = vec![];
+    let object_normal_methods = extracted.normal_methods;
```
Be-ing, line 863: I presume by "normal" you mean not Q_INVOKABLE? If so, I think "non_invokable" would be more self-explanatory than "normal".

ahayzen-kdab: These will eventually become properties implemented on the C++ object in Rust (but not exposed to C++), so it will mean that these methods will be in the context of C++ (eg self will be the C++ object) which is the same as the invokables, but allows you to have Rust only methods while being able to access C++ state (eg for calling signals etc).

Not sure what the best thing is to call them yet, they have been shoehorned into the old generation as "normal" methods :thinking:  Maybe `cpp_context_methods` or something?

Be-ing: Alternatively we could call them simply "methods" and "invokable methods".

ahayzen-kdab: Yup, probably best for now :-) As at least the rust_methods have gone now and are just passed straight across, so that helps a bit with the naming :')

cxx-qt-gen/src/extract.rs
```diff
@@ -1177,24 +957,30 @@ mod tests {
...
+        if let Item::Impl(trait_impl) = &qobject.original_passthrough_decls[0] {
+            if let Type::Path(TypePath { path, .. }) = &*trait_impl.self_ty {
+                assert_eq!(path.segments.len(), 1);
+                assert_eq!(path.segments[0].ident.to_string(), "Data");
+            } else {
+                panic!("Trait impl was not a TypePath");
+            }
         } else {
-            panic!("Trait impl was not a TypePath");
+            panic!("Expected a impl block");
```
Be-ing, line 971:
```suggestion
            panic!("Expected an impl block");
```

cxx-qt-gen/src/extract.rs
```diff
@@ -1177,24 +957,30 @@ mod tests {
...
+        if let Item::Impl(trait_impl) = &qobject.original_passthrough_decls[1] {
+            if let Type::Path(TypePath { path, .. }) = &*trait_impl.self_ty {
+                assert_eq!(path.segments.len(), 1);
+                assert_eq!(path.segments[0].ident.to_string(), "RustObj");
+            } else {
+                panic!("Trait impl was not a TypePath");
+            }
         } else {
-            panic!("Trait impl was not a TypePath");
+            panic!("Expected a impl block");
```
Be-ing, line 983:
```suggestion
            panic!("Expected an impl block");
```

cxx-qt-gen/src/parser/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+
+pub mod qobject;
+
+use qobject::{find_and_merge_cxx_qt_item, find_qobjects, ParserCxxQtData};
+use syn::{token::Brace, ItemMod, Result};
+
+/// A struct representing a module block with CXX-Qt relevant Items into ParserCxxQtData
+pub struct Parser {
+    /// The module which unknown (eg CXX) blocks are stored into
+    pub module: ItemMod,
```
Be-ing, line 14: When I first saw this variable name I presumed it was for the cxx-qt module, but the comment says it's the opposite... can you think of a more descriptive name than simply `module`? Something like `non_cxx_qt_module`, but less wordy? `passthrough_module`? `foreign_module`?

cxx-qt-gen/src/parser/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+pub mod qobject;
+
+use qobject::{find_and_merge_cxx_qt_item, find_qobjects, ParserCxxQtData};
+use syn::{token::Brace, ItemMod, Result};
+
+/// A struct representing a module block with CXX-Qt relevant Items into ParserCxxQtData
```
Be-ing, line 11:
```suggestion
/// A struct representing a module block with CXX-Qt relevant syn::Items parsed into ParserCxxQtData, to be used later to generate Rust & C++ code
```

cxx-qt-gen/src/parser/mod.rs
ahayzen-kdab: Right, so if you had like this

```
#[cxx_qt::bridge]
mod ffi {
  extern "Rust" {
    fn test();
  }

  #[cxx_qt::qobject]
  struct MyObject;
}
```

The split that occurs here is the following...

### module

Gets the original module data, and any non CXX-Qt items, note it potentially still has the cxx_qt::bridge attribute (depending how earlier macro phases work). This is because the parser phase is simply annotating the structure rather than rewriting syn items. Later the generate Rust side will change that to `#[cxx::bridge]`.
```
#[cxx_qt::bridge]
mod ffi {
  extern "Rust" {
    fn test();
  }
}
```

### cxx_qt_data

Gets any CXX-Qt items and will try to annotate them later with extra info rather than just storing the Item itself.

```
  #[cxx_qt::qobject]
  struct MyObject;
```

So probably `passthrough_module` or `non_annotated` or `insignificant` or something? :thinking:

cxx-qt-gen/src/parser/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+    /// The module which unknown (eg CXX) blocks are stored into
+    pub module: ItemMod,
+    /// Any CXX-Qt data that needs generation later
+    pub cxx_qt_data: ParserCxxQtData,
+}
+
+impl Parser {
+    /// Constructs a Parser object from a given ItemMod block
+    pub fn from(module: ItemMod) -> Result<Self> {
+        let mut new_module = module.clone();
```
Be-ing, line 22: This is taking ownership of `module` so I'm unclear what the need for cloning is?

ahayzen-kdab: Aha, an older version got all confused with mutable and immutable borrows, seems this new layout doesn't need this clone. Good spot! :-)

cxx-qt-gen/src/extract.rs
Be-ing: Okay let's go with just "methods" and "invokable methods"

ahayzen-kdab: Yup changing to that

cxx-qt-gen/src/parser/mod.rs
Be-ing: Let's go with `passthrough_module`.

cxx-qt-gen/src/extract.rs
ahayzen-kdab: Although "methods" and "invokables" might be simplest, as that's going to cause a lot of noise in the code we are about to remove to change to invokable_methods :-)

cxx-qt-gen/src/parser/mod.rs
Be-ing: A little more detail:
```suggestion
/// A struct representing a module block with CXX-Qt relevant syn::Items parsed into ParserCxxQtData, to be used later to generate Rust & C++ code. Items that are not handled specially are passed through for cxx to process.
```

cxx-qt-gen/src/parser/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use quote::{quote, ToTokens};
+    use syn::ItemMod;
+
+    #[test]
+    fn test_parser_from() {
```
Be-ing, line 66: This one function tests several example inputs. How about splitting each into a separate `#[test]` function?

Also, there's some duplicated code between the test cases that might be able to be extracted to a helper function.

cxx-qt-gen/src/parser/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+            mod ffi {}
+        }
+        .into_token_stream();
+        let module: ItemMod = syn::parse2(tokens).unwrap();
+        let parser = Parser::from(module);
+        assert!(parser.is_ok());
+        let parser = parser.unwrap();
+        assert_eq!(parser.module.attrs.len(), 1);
+        assert_eq!(parser.module.ident, "ffi");
+        assert_eq!(parser.module.content.unwrap().1.len(), 0);
```
Be-ing, line 79: Could these tests assert that the passed through module is identical to a specified `syn::ItemMod` rather than testing a few pieces of metadata about the module?

cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -0,0 +1,479 @@
...
+use syn::{
+    spanned::Spanned, Error, Ident, ImplItem, Item, ItemEnum, ItemImpl, ItemStruct, Result, Type,
+    TypePath,
+};
+
+#[derive(Default)]
+pub struct ParserCxxQtData {
+    /// Map of the QObjects defined in the module that need generation
+    pub qobjects: HashMap<Ident, ParserQObject>,
+    /// Any use blocks end up in the CXX-Qt generated module
```
Be-ing, line 21:
```suggestion
    /// Any `use` statements end up in the CXX-Qt generated module
```

cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -0,0 +1,479 @@
...
+    pub uses: Vec<Item>,
+}
+
+/// A representation of a QObject within a CXX-Qt ItemMod
+///
+/// This has initial splitting of Item's into relevant blocks, other phases will
+/// then mutate these Item's for generation purposes.
+#[derive(Default)]
+pub struct ParserQObject {
```
Be-ing, line 30:
```suggestion
pub struct ParsedQObject {
```

This feels a bit more intuitive to me, but I don't have a strong opinion on this either way.

cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -0,0 +1,479 @@
...
+    ///
+    /// In the future this will be removed
+    pub data_struct: Option<ItemStruct>,
+    /// This is the RustObj struct that stores the invokables for the QObject
+    pub rust_struct: Option<ItemStruct>,
+    /// This is the Signals enum that defines the Q_SIGNALS for the QObject
+    pub signal_enum: Option<ItemEnum>,
+    /// This is the list of methods that need to be implemented on the C++ object in Rust
+    ///
+    /// Ones marked as invokable will need to be exposed to C++ as Q_INVOKABLE
```
Be-ing, line 41: How are the invokable methods marked as such in this struct?

cxx-qt-gen/src/parser/qobject.rs
ahayzen-kdab: Makes sense i think, keep the main `Parse::from(...)` but then the inner things are `ParsedT`

cxx-qt-gen/src/extract.rs
Be-ing: I like "methods" and "invokables"

cxx-qt-gen/src/parser/qobject.rs
ahayzen-kdab: This hasn't happened yet but eventually we'll have like `ParsedInvokable` that has a bool or something to say `is_invokable` or be part of an enum etc. I'll change the comment to be a TODO.

cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -0,0 +1,479 @@
...
+    /// In the future this may be removed
+    pub update_requester_handler: Option<ItemImpl>,
+    /// Items that we don't need to generate anything for CXX or C++
+    /// eg impls on the Rust object or Default implementations
+    pub others: Vec<Item>,
+}
+
+/// Find the QObject's within the module
+///
+/// Returns a hashmap with the ident of the QObject as the key and an empty ParserQObject as the value
```
Be-ing, line 54: Returning empty ParserQObjects seems odd... I can tell from the code that uses this that the empty ParserQObjects are populated by `find_and_merge_cxx_qt_item`, but somehow it feels strange that `find_qobjects` and `find_and_merge_cxx_qt_item` are free functions with completely different names that operate on the same data structure. Just looking at the documentation of these functions and the function signatures, I can't tell there's any relationship between the functions without seeing how they're used in context.

cxx-qt-gen/src/syntax/attribute.rs
```diff
@@ -0,0 +1,144 @@
...
+    punctuated::Punctuated,
+    spanned::Spanned,
+    token::Comma,
+    token::Paren,
+    Attribute, Error, Ident, Result,
+};
+
+/// Representation of a list of idents in an attribute, eg attribute(A, B, C)
+pub struct AttributeList {
+    pub bracket: Paren,
```
Be-ing, line 21: brackets `[` and parens `(` are not the same things

cxx-qt-gen/src/parser/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+
+pub mod qobject;
+
+use qobject::{find_and_merge_cxx_qt_item, find_qobjects, ParsedCxxQtData};
+use syn::{token::Brace, ItemMod, Result};
+
+/// A struct representing a module block with CXX-Qt relevant syn::Item's
+/// parsed into ParsedCxxQtData, to be used later to generate Rust & C++ code.
+///
+/// syn::Item's that are not handled specially by CXX-Qt are passed through for CXX to process.
```
Be-ing, line 14:
```suggestion
/// A struct representing a module block with CXX-Qt relevant [syn::Item]'s
/// parsed into ParsedCxxQtData, to be used later to generate Rust & C++ code.
///
/// [syn::Item]'s that are not handled specially by CXX-Qt are passed through for CXX to process.
```

Rustdoc is smart enough to link keywords from other crates :)

cxx-qt-gen/src/parser/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+            }
+        });
+        let parser = Parser::from(module);
+        assert!(parser.is_ok());
+        let parser = parser.unwrap();
+        assert_ne!(parser.passthrough_module, module);
+        assert_eq!(parser.passthrough_module.attrs.len(), 1);
+        assert_eq!(parser.passthrough_module.ident, "ffi");
+        assert_eq!(parser.passthrough_module.content.unwrap().1.len(), 0);
+        assert_eq!(parser.cxx_qt_data.qobjects.len(), 1);
```
Be-ing, line 123:
```suggestion
        let parser = parser.unwrap();
        
        assert_ne!(parser.passthrough_module, module);
        
        assert_eq!(parser.passthrough_module.attrs.len(), 1);
        assert_eq!(parser.passthrough_module.ident, "ffi");
        assert_eq!(parser.passthrough_module.content.unwrap().1.len(), 0);
        assert_eq!(parser.cxx_qt_data.qobjects.len(), 1);
```

It's easy to miss that one of these lines is `assert_ne` whereas the rest are `assert_eq`. IMO `assert_neq` would have been a better name for the macro to avoid this, but alas, that's not what it is called.

Same applies to the following tests.

cxx-qt-gen/src/syntax/attribute.rs
```diff
@@ -0,0 +1,144 @@
...
+    fn parse(input: ParseStream) -> Result<Self> {
+        let content;
+        let paren = parenthesized!(content in input);
+        let items = content.parse_terminated(Ident::parse_any)?;
+        Ok(AttributeList { paren, items })
+    }
+}
+
+/// Returns the first attribute that matches a given path
+pub fn attribute_find_path<'a>(attrs: &'a [Attribute], path: &[&str]) -> Option<&'a Attribute> {
```
Be-ing, line 35: No need to deal with this at the moment, but these seem like generally useful functions that might be nice to upstream into syn.

cxx-qt-gen/src/syntax/path.rs
```diff
@@ -0,0 +1,143 @@
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use syn::{
+    spanned::Spanned, AngleBracketedGenericArguments, Error, GenericArgument, Ident, Path,
+    PathArguments, Result, Type, TypePath,
+};
```
Be-ing, line 9: same here, I'm wondering if some of this might be useful to upstream into syn

cxx-qt-gen/src/syntax/attribute.rs
ahayzen-kdab: So although these methods look useful to upstream syn, there are so many combinations of different nested items in syn types that it probably wouldn't be accepted. These methods i have here are helpers to reduce repeating code that CXX-Qt uses. And we basically make a load of assumptions to simplify the checks about what are inside the types.

cxx-qt-gen/src/parser/qobject.rs
ahayzen-kdab: I've changed these to be methods on the `ParsedCxxQtData` struct. So there is `find_qobject_keys` that searches for the qobject's and adds them to the hashmap. Then `parse_cxx_qt_item` that adds any CXX-Qt items to the qobjects or returns them if they are others.

cxx-qt-gen/src/syntax/attribute.rs
ahayzen-kdab: Changed to paren

cxx-qt-gen/src/parser/mod.rs
ahayzen-kdab: Done

cxx-qt-gen/src/parser/mod.rs
ahayzen-kdab: Done

cxx-qt-gen/src/parser/qobject.rs
Be-ing: I think that's a good solution.

cxx-qt-gen/src/parser/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// The module which unknown (eg CXX) blocks are stored into
+    pub passthrough_module: ItemMod,
+    /// Any CXX-Qt data that needs generation later
+    pub cxx_qt_data: ParsedCxxQtData,
+}
+
+impl Parser {
+    /// Constructs a Parser object from a given [syn::ItemMod] block
+    pub fn from(mut module: ItemMod) -> Result<Self> {
+        // let mut new_module = module.clone();
```
Be-ing, line 25: remove commented code

cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -0,0 +1,479 @@
...
+#[derive(Default)]
+pub struct ParsedCxxQtData {
+    /// Map of the QObjects defined in the module that need generation
+    pub qobjects: HashMap<Ident, ParsedQObject>,
+    /// Any `use` statements end up in the CXX-Qt generated module
+    pub uses: Vec<Item>,
+}
+
+impl ParsedCxxQtData {
+    /// Find the QObject's within the module and adds the keys into the QObject's HashMap
```
Be-ing, line 55:
```suggestion
    /// Find the QObjects within the module and adds the keys into the QObject's HashMap
```

cxx-qt-gen/src/parser/mod.rs
ahayzen-kdab: Done

cxx-qt-gen/src/parser/qobject.rs
ahayzen-kdab: done

cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -0,0 +1,479 @@
...
+                _others => {}
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Determine if the given [syn::Item] is a CXX-Qt related item
+    /// If it is then add the [syn::Item] into QObject's list
+    /// Otherwise return the [syn::Item] if it's not relevant
```
Be-ing, line 86:
```suggestion
    /// Determine if the given [syn::Item] is a CXX-Qt related item
    /// If it is then add the [syn::Item] into the qobjects HashMap
    /// Otherwise return the [syn::Item] to pass through to cxx
```

ahayzen-kdab: done

cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -0,0 +1,479 @@
...
+    /// In the future this may be removed
+    pub update_requester_handler: Option<ItemImpl>,
+    /// Items that we don't need to generate anything for CXX or C++
+    /// eg impls on the Rust object or Default implementations
+    pub others: Vec<Item>,
+}
+
+#[derive(Default)]
+pub struct ParsedCxxQtData {
+    /// Map of the QObject's defined in the module that need generation
```
Be-ing, line 48:
```suggestion
    /// Map of the QObjects defined in the module that will be used for code generation
```

ahayzen-kdab: done

Be-ing (approved):
waiting for CI

PR #188, opened 2022-08-05
Use assert_str_eq! for file comparisons

LeonMatthesKDAB:
Enables proper multi-line diffing.

Before:
Screenshot from 2022-08-05 12-00-00

After:
Screenshot from 2022-08-05 12-08-12

Need I say more? ;)


ahayzen-kdab (approved):
LGTM :-)

PR #189, opened 2022-08-05
cxx-qt-gen: add ParsedSignal to parser

ahayzen-kdab:
For now the extract phase still needs to convert this into it's own Signal. But later this will happen in the generate phase.

Requires #187

### DIFF
--- a/cxx-qt-gen/src/parser/parameter.rs
+++ b/cxx-qt-gen/src/parser/parameter.rs
@@ -1,0 +1,16 @@
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use syn::{Ident, Type};
+
+/// Describes a single parameter for a function
+pub struct ParsedFunctionParameter {
+    /// The [syn::Ident] of the parameter
+    pub ident: Ident,
+    /// The [syn::Type] of the parameter
+    pub ty: Type,
+    // TODO: later this will describe if the parameter has an attribute
+    // stating that the a conversion in C++ needs to occur (eg UniquePtr<T> to T)..
+}
--- a/cxx-qt-gen/src/parser/qobject.rs
+++ b/cxx-qt-gen/src/parser/qobject.rs
@@ -3,6 +3,7 @@
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
+use crate::parser::signals::ParsedSignalsEnum;
 use crate::syntax::{
     attribute::{attribute_find_path, attribute_tokens_to_ident},
     path::{path_angled_args_to_type_path, path_compare_str, path_to_single_ident},
@@ -26,8 +27,8 @@ pub struct ParsedQObject {
     pub data_struct: Option<ItemStruct>,
     /// This is the RustObj struct that stores the invokables for the QObject
     pub rust_struct: Option<ItemStruct>,
-    /// This is the Signals enum that defines the Q_SIGNALS for the QObject
-    pub signal_enum: Option<ItemEnum>,
+    /// This is a representation of the Signals enum that defines the Q_SIGNALS for the QObject
+    pub signals: Option<ParsedSignalsEnum>,
     /// This is the list of methods that need to be implemented on the C++ object in Rust
     ///
     /// TODO: in the future we will have a ParsedMethod this will have a field to say whether the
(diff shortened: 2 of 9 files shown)


ahayzen-kdab:
Wonder what phase should remove the attributes that aren't needed in the final Rust code. Eg have have things like `#[cxx_qt::signals(T)]` or `#[invokable]` etc etc, the Parser phase reads this info and describes this info in the structs. Then later generation phases should not need to read the attributes again ? Generation phase should be able to use the info in the structs.

So i've talked myself into parser phase removing any attributes which it parses into struct data...

ahayzen-kdab:
> Wonder what phase should remove the attributes that aren't needed in the final Rust code. Eg have have things like `#[cxx_qt::signals(T)]` or `#[invokable]` etc etc, the Parser phase reads this info and describes this info in the structs. Then later generation phases should not need to read the attributes again ? Generation phase should be able to use the info in the structs.
> 
> So i've talked myself into parser phase removing any attributes which it parses into struct data...

Changed it to do this now :-) So the parser phase takes syn as input, reads any of the CXX-Qt macros, creates a struct abstract representation of what the macros are describing and removes any macros it has processed.

cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -88,11 +88,11 @@ impl ParsedCxxQtData {
         Ok(match &item {
             Item::Enum(item_enum) => {
```
LeonMatthesKDAB, line 89: Can you split up the code for the different match parts into separate functions, i.e.:
``` rust
pub fn parse_cxx_qt_item(&mut self, item: Item) -> Result<Option<Item>> {
    Ok(match item {
        Item::Enum(item_enum) => self.parse_enum(item_enum),
        Item::Struct(item_struct) => self.parse_struct(item_struct),
    })
}
```
Currently the function contains a bit too many levels of abstraction.

cxx-qt-gen/src/parser/signals.rs
```diff
@@ -0,0 +1,145 @@
...
+use crate::parser::parameter::ParserParameter;
+use crate::syntax::fields::fields_named_to_ident_type;
+use syn::{Ident, ItemEnum, Result};
+
+/// Describes an individual Signal
+pub struct ParsedSignal {
+    /// The name of the signal
+    pub ident: Ident,
+    /// The parameters of the signal
+    pub parameters: Vec<ParserParameter>,
```
LeonMatthesKDAB, line 15: Shouldn't this be a `ParsedParameter`, not a `ParserParameter`?

cxx-qt-gen/src/parser/parameter.rs
```diff
@@ -0,0 +1,14 @@
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use syn::{Ident, Type};
+
+/// Describes a Parameter for a QObject
+pub struct ParserParameter {
```
LeonMatthesKDAB, line 9: See other comment, should probably be named `ParsedParameter`.

cxx-qt-gen/src/parser/parameter.rs
```diff
@@ -0,0 +1,14 @@
...
+    /// The [syn::Ident] of the parameter
```
LeonMatthesKDAB, line 10: this level of documentation is probably a bit overkill.
No need to state the obvious :)

cxx-qt-gen/src/parser/signals.rs
```diff
@@ -0,0 +1,145 @@
...
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use crate::parser::parameter::ParserParameter;
+use crate::syntax::fields::fields_named_to_ident_type;
+use syn::{Ident, ItemEnum, Result};
+
+/// Describes an individual Signal
+pub struct ParsedSignal {
+    /// The name of the signal
```
LeonMatthesKDAB, line 12: Again, overkill level of documentation, imho :)

cxx-qt-gen/src/parser/signals.rs
```diff
@@ -0,0 +1,145 @@
...
+    pub ident: Ident,
+    /// The parameters of the signal
+    pub parameters: Vec<ParserParameter>,
+    // TODO: later this will describe of the signal has an attribute
+    // stating that the signal exists in the base class
+}
+
+/// Describes a Signals block for a QObject
```
LeonMatthesKDAB, line 20:
```suggestion
/// Describes a Signals enum for a QObject
```

cxx-qt-gen/src/parser/signals.rs
```diff
@@ -0,0 +1,145 @@
...
+    /// Constructs a ParsedSignals object from a given [syn::ItemEnum] block
+    pub fn from(item: &ItemEnum, attr_index: usize) -> Result<Self> {
+        // Remove the attribute index as we have processed it
+        let mut item = item.clone();
+        item.attrs.remove(attr_index);
+
+        let signals = item
+            .variants
+            .iter()
+            .map(|variant| {
```
LeonMatthesKDAB, line 40: The contenst of this lambda should probably be a constructor in `ParsedSignal`.
E.g. `ParsedSignal::from_variant`.

cxx-qt-gen/src/parser/signals.rs
LeonMatthesKDAB: Also, maybe rename `ParsedSignals` to `ParsedSignalsEnum` :thinking: 

It's easy to mix up with `ParsedSignal`

cxx-qt-gen/src/parser/signals.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    #[test]
+    fn test_parsed_signals_from_empty() {
+        let e: ItemEnum = tokens_to_syn(quote! {
+            #[cxx_qt::signals(MyObject)]
+            enum MySignals {}
+        });
+        let signals = ParsedSignals::from(&e, 0);
+        assert!(signals.is_ok());
+        let signals = signals.unwrap();
```
LeonMatthesKDAB, line 86:
```suggestion
        let signals = ParsedSignals::from(&e, 0).unwrap();
```
It's okay to just unwrap, the test will just fail if it doesn't work.

cxx-qt-gen/src/parser/signals.rs
```diff
@@ -0,0 +1,145 @@
...
+    use syn::Type;
+
+    /// Helper which returns a f64 as a [syn::Type]
+    fn f64_type() -> Type {
+        tokens_to_syn(quote! { f64 })
+    }
+
+    #[test]
+    fn test_parsed_signals_from_empty() {
+        let e: ItemEnum = tokens_to_syn(quote! {
```
LeonMatthesKDAB, line 80: nice approach to just use `quote!` :ok_hand:

cxx-qt-gen/src/parser/signals.rs
LeonMatthesKDAB: Same in the other test cases.

cxx-qt-gen/src/parser/signals.rs
```diff
@@ -0,0 +1,145 @@
...
+            #[cxx_qt::signals(MyObject)]
+            #[after]
+            enum MySignals {}
+        });
+        let signals = ParsedSignals::from(&e, 1);
+        assert!(signals.is_ok());
+        let signals = signals.unwrap();
+        assert_eq!(signals.ident, "MySignals");
+        assert_eq!(signals.item.attrs.len(), 2);
+        assert_eq!(signals.signals.len(), 0);
```
LeonMatthesKDAB, line 105: Can you add a check, that the correct attribute is removed and the others are untouched?

cxx-qt-gen/src/parser/signals.rs
```diff
@@ -0,0 +1,145 @@
...
+    #[test]
+    fn test_parsed_signals_from_empty_attrs() {
+        let e: ItemEnum = tokens_to_syn(quote! {
+            #[before]
+            #[cxx_qt::signals(MyObject)]
+            #[after]
+            enum MySignals {}
+        });
+        let signals = ParsedSignals::from(&e, 1);
+        assert!(signals.is_ok());
```
LeonMatthesKDAB, line 101: Should this even parse?
Do we even want to allow additional attributes on a Signals enum? :thinking:

cxx-qt-gen/src/syntax/attribute.rs
```diff
@@ -31,11 +31,11 @@ impl Parse for AttributeList {
     }
 }
 
-/// Returns the first [syn::Attribute] that matches a given path
-pub fn attribute_find_path<'a>(attrs: &'a [Attribute], path: &[&str]) -> Option<&'a Attribute> {
-    for attr in attrs {
+/// Returns the index of the first [syn::Attribute] that matches a given path
+pub fn attribute_find_path<'a>(attrs: &'a [Attribute], path: &[&str]) -> Option<usize> {
```
ahayzen-kdab, line 35: lifetime can be removed, not sure why clippy didn't pick that up...

cxx-qt-gen/src/syntax/fields.rs
```diff
@@ -0,0 +1,121 @@
...
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use syn::{spanned::Spanned, Error, Field, Fields, FieldsNamed, Ident, Result, Type};
+
+/// In a group of [syn::Fields] extract the [syn::Ident] and [syn::Type] from
+/// any [syn::FieldsNamed] fields
+///
+/// If there are [syn::FieldsUnnamed] then an error occurs
+pub fn fields_named_to_ident_type(fields: &Fields) -> Result<Vec<(Ident, Type)>> {
```
LeonMatthesKDAB, line 12: This should probably just be a function on Parameter and return a Vec<Parameter>.
 e.g. `Parameter::from_fields`.

LeonMatthesKDAB (commented):
Going in a great direction.

But there's some stuff that could be cleaner.

cxx-qt-gen/src/parser/signals.rs
ahayzen-kdab: Oops, yup :-)

cxx-qt-gen/src/parser/signals.rs
ahayzen-kdab: Potentially, what if you want to add serialisation to your signals enum, it doesn't matter CXX-Qt itself as it's still an enum. But you might want to derive other info for that enum.

cxx-qt-gen/src/syntax/fields.rs
ahayzen-kdab: This might be used later for other things like parsing the struct with it's named fields (although we need the attributes info as well). But i can see us later having our own Vec<NamedField> thing which has the ident, type, attributes etc potentially we'll see. So any general syntax helpers that the parser (or other phases may use) i've been splitting into this syntax crate.

cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -27,7 +27,7 @@ pub struct ParsedQObject {
     /// This is the RustObj struct that stores the invokables for the QObject
     pub rust_struct: Option<ItemStruct>,
     /// This is the Signals enum that defines the Q_SIGNALS for the QObject
```
Be-ing, line 29: comment needs to be updated, this is no longer an enum

ahayzen-kdab: It's the representation of the enum :thinking:

cxx-qt-gen/src/parser/qobject.rs
ahayzen-kdab: Done, only left the use one as it's simple.

cxx-qt-gen/src/parser/parameter.rs
ahayzen-kdab: Left for now, this might become more complicated later :-)

cxx-qt-gen/src/parser/signals.rs
ahayzen-kdab: Left for now, this might become more complicated later :-)

cxx-qt-gen/src/parser/signals.rs
ahayzen-kdab: Done

cxx-qt-gen/src/parser/signals.rs
ahayzen-kdab: Done

cxx-qt-gen/src/parser/signals.rs
ahayzen-kdab: Done

cxx-qt-gen/src/parser/qobject.rs
ahayzen-kdab: Changed the wording a bit.

cxx-qt-gen/src/parser/parameter.rs
```diff
@@ -0,0 +1,14 @@
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use syn::{Ident, Type};
+
+/// Describes a Parameter for a QObject
```
Be-ing, line 8: Do you intend to reuse this struct for more than parsing the signals? If so, I think this documentation could be improved. I think you mean parameters for QObject methods? If this will only be used for signals, then let's rename it to a more specific name, perhaps ParsedSignalParameters.

ahayzen-kdab: It'll likely be used for parameters in signals, methods, and invokables in the future (like the existing equivalent in extract.rs)

Be-ing: How about renaming it to ParsedFunctionParameter? And the documentation: `Describes a single parameter for a function`

ahayzen-kdab: Potentially :thinking: guess it's obvious enough that signals are actually functions and not enums (like on the Rust side). And then extra info will go into these later like describing any conversion that needs to occur on the C++ side etc.

Be-ing: I got confused reading this because a "parameter for a QObject" isn't really a thing, but a parameter for a QObject's method is.

ahayzen-kdab: Fixed :-)

Be-ing (approved):

PR #192, opened 2022-08-05
Auto-cancel previous workflows for the same PR.

LeonMatthesKDAB:
Should free up some MacOS runners.


Be-ing (approved):
I don't understand why GitHub doesn't make this the default :shrug:

PR #193, opened 2022-08-05
cxx-qt-gen: remove the cxx_qt::bridge in the parser

ahayzen-kdab:
This allows us to process the namespace later and means we don't need
to remove the macro sometimes in the generation.

### DIFF
--- a/cxx-qt-gen/src/parser/mod.rs
+++ b/cxx-qt-gen/src/parser/mod.rs
@@ -5,8 +5,9 @@
 
 pub mod qobject;
 
+use crate::syntax::attribute::attribute_find_path;
 use qobject::ParsedCxxQtData;
-use syn::{token::Brace, ItemMod, Result};
+use syn::{spanned::Spanned, token::Brace, Error, ItemMod, Result};
 
 /// A struct representing a module block with CXX-Qt relevant [syn::Item]'s
 /// parsed into ParsedCxxQtData, to be used later to generate Rust & C++ code.
@@ -24,6 +25,17 @@ impl Parser {
     pub fn from(mut module: ItemMod) -> Result<Self> {
         let mut cxx_qt_data = ParsedCxxQtData::default();
         let mut others = vec![];
+
+        // Remove the cxx_qt::bridge attribute
+        if let Some(index) = attribute_find_path(&module.attrs, &["cxx_qt", "bridge"]) {
+            // TODO: parse any namespace and store into the ParsedCxxQtData
+            module.attrs.remove(index);
+        } else {
+            return Err(Error::new(
+                module.span(),
+                "Tried to parse a module which doesn't have a cxx_qt::bridge attribute",
+            ));
+        }
 
         // Check that there are items in the module
         if let Some(mut items) = module.content {
@@ -71,10 +83,11 @@ mod tests {
             #[cxx_qt::bridge]
             mod ffi {}
         });
-        let parser = Parser::from(module.clone());
-        assert!(parser.is_ok());
-        let parser = parser.unwrap();
-        assert_eq!(parser.passthrough_module, module);
+        let parser = Parser::from(module).unwrap();
+        let expected_module: ItemMod = tokens_to_syn(quote! {
+            mod ffi {}
+        });
+        assert_eq!(parser.passthrough_module, expected_module);
         assert_eq!(parser.cxx_qt_data.qobjects.len(), 0);
     }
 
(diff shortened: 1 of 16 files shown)


cxx-qt/src/lib.rs
```diff
@@ -32,10 +32,18 @@ use cxx_qt_gen::{extract_qobject, generate_qobject_rs};
...
-    let module = parse_macro_input!(input as ItemMod);
+    let mut module = parse_macro_input!(input as ItemMod);
+
+    // Build the #[cxx_qt::bridge(namespace = "A::B")] back into our module
+    // this is because the parser phase expects to see this to read the namespace
+    //
+    // As other routes that enter the parser keep the attributes so this is then consistent
+    let args_input = format!("#[cxx_qt::bridge{}] mod dummy;", args);
+    let attrs = syn::parse_str::<ItemMod>(&args_input).unwrap().attrs;
+    module.attrs = attrs.into_iter().chain(module.attrs.into_iter()).collect();
```
Be-ing, line 46: This seems weird... why not parse the `(namespace = "foobar")` now even if we're not doing anything with it yet? Seems odd to replace it with a dummy instead.

cxx-qt-gen/src/parser/mod.rs
```diff
@@ -165,4 +183,17 @@ mod tests {
...
+    fn test_parser_from_error_no_attribute() {
+        let module: ItemMod = tokens_to_syn(quote! {
+            mod ffi {
+                extern "Rust" {
+                    fn test();
+                }
+            }
+        });
+        let parser = Parser::from(module);
+        assert!(parser.is_err());
```
Be-ing, line 197: Should this test more specifically what kind of error it is?

cxx-qt/src/lib.rs
ahayzen-kdab: Because as stated in the comment other routes don't enter the extract_* or parser phase with the attribute removed and have the whole attribute. Eg the -build phase parses the ItemMod to extract methods and parse phase with the cxx_qt::bridge attributes still intact. So by simply reconstructing it we can emulate this from the macro and create a consistent interface. (and the code i used here was recommended by the syn/CXX author for a similar usecase)

cxx-qt-gen/src/parser/mod.rs
ahayzen-kdab: We can't easily as the syn Error's are just a span and a string, so just checking for is_err is enough for now.

Be-ing: okay

cxx-qt/src/lib.rs
Be-ing: So the TokenStream input to the macro doesn't receive the attribute from rustc? Or does the `parse_macro_input` macro remove it?

ahayzen-kdab: So if you look at how `proc_macro_attribute` works https://doc.rust-lang.org/reference/procedural-macros.html#attribute-macros.

Their example code
```
#[proc_macro_attribute]
pub fn show_streams(attr: TokenStream, item: TokenStream) -> TokenStream {
    println!("attr: \"{}\"", attr.to_string());
    println!("item: \"{}\"", item.to_string());
    item
}
```
would do the following
```
// Example: Attribute with input
#[show_streams(bar)]
fn invoke2() {}
// out: attr: "bar"
// out: item: "fn invoke2() {}"
```
Here you can see how if we have `#[cxx_qt::bridge(namespace = "a")]` we would get just `namespace = "a"` in the first `TokenStream` argument and then anything below/after the attribute in the second `TokenStream`. So we basically don't receive the attribute in the macro, we only partially get it.

Hence to create a consistent interface with other uses where we want to read the `cxx_qt::bridge` macro at the parser phase (eg cxx-qt-build gives the whole ItemMod block it finds to the cxx-qt-gen methods), we need to reconstruct the macro and insert it into the attribute list before passing the macro generated ItemMod onwards.

ahayzen-kdab: The format line is also wrong here and fixed in the next pull request, it should be `format!("#[cxx_qt::bridge({})] mod dummy;", args);` not `format!("#[cxx_qt::bridge{}] mod dummy;", args);`

ahayzen-kdab: > The format line is also wrong here and fixed in the next pull request, it should be `format!("#[cxx_qt::bridge({})] mod dummy;", args);` not `format!("#[cxx_qt::bridge{}] mod dummy;", args);`

Fixed :-)

Be-ing: Huh, I guess that makes sense because generally macros wouldn't have any reason to process their own name. But the other code paths don't start from a macro invocation, so we have this odd edge case. I think this could be explained better in the comment. How about:

```
    // Macros do not typically need to do anything with their own attribute name, so rustc does not include that in the `args` or `input` TokenStreams.
    // However, other code paths that use the parser do not enter from a macro invocation, so they rely on parsing the `cxx_qt::bridge` attribute to identify where to start parsing.
    // To keep the inputs to the parser consistent for all code paths, add the attribute to the module before giving it to the parser.
```

ahayzen-kdab: Replaced the comment with that :-)

Be-ing (approved):

PR #196, opened 2022-08-08
cxx-qt-gen: split qobject.rs into qobject and cxxqtdata

ahayzen-kdab:


cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -44,187 +38,33 @@ pub struct ParsedQObject {
...
-}
-
-impl ParsedCxxQtData {
-    /// Find the QObjects within the module and adds the keys into the qobjects HashMap
-    pub fn find_qobject_keys(&mut self, items: &[Item]) -> Result<()> {
+impl ParsedQObject {
+    /// Extract either invokables or methods from [syn::ImplItem]'s from each Impl block
+    ///
+    /// These will have come from a impl cxx_qt::QObject<T> block
+    pub fn parse_impl_items(&mut self, items: &[ImplItem]) -> Result<()> {
```
ahayzen-kdab, line 45: Think it makes sense to only have ImplItem here. The way it is split is determining which QObject the item is referring to happens in cxxqtdata, then any simple one liners just edit the qobject. Any that have more parsing to do but already know which qobject it is for then have their method in the qobject.rs.  I guess eventually this will be both invokables and properties parsing will be in qobject.rs

cxx-qt-gen/src/parser/cxxqtdata.rs
```diff
@@ -0,0 +1,456 @@
...
+    /// Any `use` statements end up in the CXX-Qt generated module
+    pub uses: Vec<Item>,
+}
+
+impl ParsedCxxQtData {
+    /// Find the QObjects within the module and adds the keys into the qobjects HashMap
+    pub fn find_qobject_keys(&mut self, items: &[Item]) -> Result<()> {
+        for item in items {
+            match item {
+                Item::Struct(s) => {
```
Be-ing, line 32: `if let`

cxx-qt-gen/src/parser/cxxqtdata.rs
```diff
@@ -0,0 +1,456 @@
...
+                    // TODO: instead find the cxx_qt::qobject macro
+                    // and support multiple QObjects in one block
+                    if s.ident == "RustObj" {
+                        // TODO: for now we only support one qobject per block
+                        if !self.qobjects.is_empty() {
+                            return Err(Error::new(
+                                s.span(),
+                                "Only on RustObj struct is supported per mod",
```
Be-ing, line 40: `Only one`

cxx-qt-gen/src/parser/qobject.rs
ahayzen-kdab: Makes sense, any complexity once we know the target qobject can be in an impl QObject. So far these are simple though other than invokables (later properties will likely be complex too).

cxx-qt-gen/src/parser/cxxqtdata.rs
ahayzen-kdab: Done

cxx-qt-gen/src/parser/cxxqtdata.rs
ahayzen-kdab: Done

Be-ing:
This needs a rebase, otherwise looks good to me.

ahayzen-kdab:
> This needs a rebase, otherwise looks good to me.

Done

Be-ing (approved):

PR #195, opened 2022-08-08
cxx-qt-gen: parse invokables and methods into separate lists

ahayzen-kdab:

PR #198, opened 2022-08-09
cxx-qt-gen: use cxx_qt::qobject struct as the QObject name

ahayzen-kdab:
Requires #195 #196 #197


README.md
```diff
@@ -85,10 +85,11 @@ mod my_object {
         }
     }
 
+    #[cxx_qt::qobject]
     #[derive(Default)]
-    pub struct RustObj;
+    pub struct MyObject;
```
ahayzen-kdab, line 90: Add docs to the book how the struct name becomes the QObject name

cxx-qt-build/src/lib.rs
```diff
@@ -61,7 +61,10 @@ impl GeneratedCpp {
         let mut cxx_qt = None;
         // TODO: later change how the resultant filename is chosen, can we match the input file like
         // CXX does?
-        let mut module_ident: String = "".to_owned();
+        //
+        // For CXX-Qt generating one header per QObject likely makes sense, but what happens with CXX data?
+        // for now this uses the module ident
+        let mut file_ident: String = "".to_owned();
```
ahayzen-kdab, line 67: We might need this to be at least one CXX-Qt header per module block (or per file) and not per QObject. Otherwise where does the include line go in a block like this.
```rust
#[cxx_qt::bridge] 
mod ffi {
  #[cxx_qt::qobject(base = "QAbstractListModel")]
  struct MyObject;

  #[cxx_qt::qobject]
  struct AnotherObject;

  extern "C++" {
    include!("<QAbstractListModel>");
  }
}
```

README.md
ahayzen-kdab: Done

cxx-qt-build/src/lib.rs
ahayzen-kdab: So the problem if we use the rust file name is, how do we generate `include!("cxx-qt-gen/include/my_object.cxxqt.h");` in the Rust macro ? as we are inside another macro how can we find the file name ?

ahayzen-kdab: Wonder if we have `cxx_qt_include!("<QAbstractListModel>")`, any `cxx_qt_include` can go into the .cxxqt.h and any `include` can go into the .cxx.h

ahayzen-kdab: We could use the nightly feature to get the source file of the macro https://doc.rust-lang.org/proc_macro/struct.SourceFile.html

ahayzen-kdab: Moved the source file changes into #200. For now this pull request uses the QObject ident name as the generated name.

Be-ing (approved):
The magic struct name bugged me a bit before. I like explicitly marking this with an attribute better.

PR #199, opened 2022-08-09
cxx-qt-gen: rearrange includes so that we use CXX include's

ahayzen-kdab:
This then allows us later to have include!(<QAbstractListModel>)
and this will be included into our .cxxqt.h via the .cxx.h

### DIFF
--- a/cxx-qt-gen/test_outputs/handlers.h
+++ b/cxx-qt-gen/test_outputs/handlers.h
@@ -1,14 +1,15 @@
 #pragma once
 
+#include <memory>
 #include <mutex>
 
-#include "cxx-qt-lib/include/qt_types.h"
+namespace cxx_qt::my_object {
+class MyObject;
+} // namespace cxx_qt::my_object
 
-#include <QtCore/QString>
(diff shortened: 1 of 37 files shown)


cxx-qt-gen/test_outputs/handlers.h
```diff
@@ -1,10 +1,13 @@
...
+#include <memory>
 #include <mutex>
 
-#include "cxx-qt-lib/include/qt_types.h"
+namespace cxx_qt::my_object {
+class MyObject;
+} // namespace cxx_qt::my_object
 
-#include <QtCore/QString>
+#include "cxx-qt-gen/include/my_object.cxx.h"
```
ahayzen-kdab, line 10: Do we still need to forward declare the Rust side of the MyObject? as that's included from the .cxx.h?

ahayzen-kdab: We don't, removed :-)

LeonMatthesKDAB (approved):
LGTM

PR #200, opened 2022-08-09
WIP: cxx-qt-build: use filename for rust name 

ahayzen-kdab:
This uses nightly feature to find the source file for the macro.
https://github.com/rust-lang/rust/issues/54725

Requires #198


Be-ing:
What is the reason for this change?

ahayzen-kdab:
> What is the reason for this change?

We need to figure out how to choose the C++ file name from the CXX-Qt blocks, as the old code was taking the module ident, the latest code is taking the QObject name. But this won't work when we support multiple QObject's in one block.

Ideally (i think) we would copy CXX and have the input Rust file name be the output C++ name too, but with the macros this creates fun for us as we need to write in the generated rust code the line `include!("cxx-qt-gen/include/{file_name}.cxxqt.h")`. The last commit on this branch was just showing how this could be done with nightly APIs (ugh).

But we need to think further about the best way of doing this, so for now it's WIP.

ahayzen-kdab:
This was done via #856

PR #201, opened 2022-08-10
cxx-qt-gen: CXX-Qt module is private, then keep attrs on CXX module

ahayzen-kdab:


cxx-qt-gen/src/gen_rs.rs
```diff
@@ -933,8 +931,7 @@ pub fn generate_qobject_rs(obj: &QObject) -> Result<TokenStream, TokenStream> {
         #cxx_block
 
         pub use self::#cxx_qt_mod_ident::*;
```
ahayzen-kdab, line 933: this could be just `pub use self::cxx_qt_t::T;`? so we only expose the QObject's?

ahayzen-kdab: But the other public items in the module are going to either go away or move to a private internals mod, so `*` may be simplest anyway?

ahayzen-kdab:
Waiting for the other one to pass and land first, then rebase this...

Be-ing (approved):

PR #202, opened 2022-08-10
Remove CppObj

LeonMatthesKDAB:
This will move all invokables onto MyObjectQt and remove the need for
the CppObj.

### DIFF
--- a/cxx-qt-gen/src/gen_rs.rs
+++ b/cxx-qt-gen/src/gen_rs.rs
@@ -111,6 +111,82 @@ impl RustType for QtTypes {
     }
 }
 
+fn generate_invokable_cxx_declaration(obj: &QObject, i: &Invokable) -> TokenStream {
+    let rust_class_name = &obj.original_rust_struct.ident;
+    let cpp_class_name = format_ident!("{}Qt", rust_class_name);
+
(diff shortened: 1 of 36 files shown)


Be-ing:
Is this still a WIP? If not, please change the title of the PR.

cxx-qt-gen/src/gen_rs.rs
```diff
@@ -111,6 +111,82 @@ impl RustType for QtTypes {
...
+    );
+    let parameters = &i.parameters;
+    let mutablility = if i.mutable {
+        Some(quote! { mut })
+    } else {
+        None
+    };
+
+    // TODO: invokables need to also become freestanding functions that
+    // take as input a reference to both the Rs class and the CppObject
```
ahayzen-kdab, line 131: not sure which usecase this is? Can't you just use the rust/rust_mut methods from the C++ pointer? but for future branches.

ahayzen-kdab: Ah lol i see it's an old comment that has moved, so yup fine for now as this code will be ripped out anyway.

ahayzen-kdab (approved):
Looks good, lets land this as a first step so we can move forward and parallelise tasks

PR #204, opened 2022-08-11
cxx-qt-gen: add a writer phase for the C++ side

ahayzen-kdab:


cxx-qt-gen/src/syntax/cpp.rs
```diff
@@ -0,0 +1,10 @@
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+/// A generic C++ header and source pairing
+pub struct CppPair {
```
Be-ing, line 7:
```suggestion
pub struct CppSourceCodePair {
```

Be-ing: `CppPair` could be confused with C++'s `std::pair`

cxx-qt-gen/src/generator/cpp/mod.rs
```diff
@@ -0,0 +1,28 @@
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use crate::syntax::cpp::CppPair;
+
+/// Representation of the generated C++ code for a QObject
+pub struct GeneratedCpp {
```
Be-ing, line 9: This isn't the generated C++ code, this is the data used to generate the C++ code, so I don't think this name fits. What could be a better name? CppData?

ahayzen-kdab: It's generated from the perspective of, from the input of Rust code, this is the abstract representation of the generated C++ code. And only abstract from the point of view of "these things are methods" and "these things are sources". The actual strings inside them are full generated. The writer phase just takes this "generated" C++ code and performs a layout on it with parts that are static.

I'm not sure that CppData is a better name, as is this an input or output? is it a syntax CppData input or a parsed CppData abstract presentation or a generated CppData or a written out CppData header? :thinking: GeneratedAbstractCpp or GeneratedCppParts or GeneratedCppBlocks or something seem better? I think `GeneratedCppBlocks` is the best so far...

cxx-qt-gen/src/syntax/cpp.rs
ahayzen-kdab: Done

cxx-qt-gen/src/generator/cpp/mod.rs
ahayzen-kdab: Renamed to GeneratedCppBlocks

cxx-qt-gen/src/syntax/cpp.rs
```diff
@@ -0,0 +1,10 @@
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+/// A generic C++ header and source pairing
+pub struct CppSourceCodePair {
```
Be-ing, line 7:
```suggestion
pub struct CppFragmentPair {
```
Sorry for bikeshedding this again :P This is kinda the same thing as `CppObject` in gen_cpp.rs, but it's not quite the same and easy to confuse them. Alternatively, we could have on struct for both of these use cases, but I think differentiating them is better.

Be-ing: Come to think of it, why is this in the `syntax` submodule?

ahayzen-kdab: Renamed to `CppFragmentPair` and moved into the generator module. I was thinking before it might be used in more places but it isn't so it can go there for now :-).

Be-ing (approved):

PR #212, opened 2022-08-12
cxx-qt-gen: inject the CXX definition for common methods in writer

ahayzen-kdab:
Requires #206

### DIFF
--- a/cxx-qt-gen/src/writer/rust/mod.rs
+++ b/cxx-qt-gen/src/writer/rust/mod.rs
@@ -6,6 +6,46 @@
 use crate::generator::rust::GeneratedRustBlocks;
 use proc_macro2::TokenStream;
 use quote::{format_ident, quote, ToTokens};
+use syn::Ident;
+
+/// Return common blocks for CXX bridge which the C++ writer adds as well
+fn cxx_common_blocks(
+    cpp_struct_ident: &Ident,
+    rust_struct_ident: &Ident,
+    namespace_internals: &String,
+) -> Vec<TokenStream> {
+    vec![
+        quote! {
+            unsafe extern "C++" {
+                include!("cxx-qt-lib/include/convert.h");
+
+                #[cxx_name = "unsafeRust"]
+                fn rust(self: &#cpp_struct_ident) -> &#rust_struct_ident;
+
+                #[rust_name = "new_cpp_object"]
+                #[namespace = #namespace_internals]
+                fn newCppObject() -> UniquePtr<#cpp_struct_ident>;
+            }
+        },
+        quote! {
+            extern "C++" {
+                #[cxx_name = "unsafeRustMut"]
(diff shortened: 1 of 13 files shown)


ahayzen-kdab:
Lets wait for the namespace one to land first as it's more important for other branches to rebase onto.

cxx-qt-gen/src/writer/rust/mod.rs
```diff
@@ -18,9 +66,18 @@ pub fn write_rust(generated: &GeneratedRustBlocks) -> TokenStream {
...
 
     // Retrieve the module contents and namespace
-    let cxx_mod = &generated.cxx_mod;
+    let mut cxx_mod = generated.cxx_mod.clone();
     let cxx_qt_mod_contents = &generated.cxx_qt_mod_contents;
     let namespace = &generated.namespace;
+    let namespace_internals = &generated.namespace_internals;
+
+    // Inject the common methods into the bridge which we need
+    write_cxx_common_methods(
```
LeonMatthesKDAB, line 75: I feel like this shouldn't be handled by passing a mutable reference to the cxx_mod.
It leads to code duplication in terms of `content.push(syn::parse2(...)).expect("...")`.

Maybe refactor write_cxx_common_methods to:
`cxx_common_blocks` that returns a `Vec<TokenStream>`.
Then just loop over that:
``` rust
let cxx_mod_items = cxx_mod.content.as_mut().expect("").1;
for block in cxx_common_blocks(cpp_struct_ident, rust_struct_ident, namespace_internals){ 
    cxx_mod_items.push(syn::parse2(block).expect("Could not build extern block!"));
}
```

cxx-qt-gen/src/writer/rust/mod.rs
```diff
@@ -6,6 +6,54 @@
...
+/// Add common methods into the CXX bridge which the C++ writer adds as well
+fn write_cxx_common_methods(
+    content: &mut Vec<Item>,
+    cpp_struct_ident: &Ident,
+    rust_struct_ident: &Ident,
+    namespace_internals: &String,
+) {
+    content.push(
+        syn::parse2(quote! {
+            unsafe extern "C++" {
```
LeonMatthesKDAB, line 20: Should this also contain the convert include?
``` rust
include!("cxx-qt-lib/include/convert.h");
```

LeonMatthesKDAB (changes requested):
Going in the right direction, just a bit of nitpicking.

cxx-qt-gen/src/writer/rust/mod.rs
ahayzen-kdab: Done

cxx-qt-gen/src/writer/rust/mod.rs
ahayzen-kdab: Done

Be-ing (approved):

PR #213, opened 2022-08-12
WIP: cxx-qt-gen: add checks phase

ahayzen-kdab:


ahayzen-kdab:
This needs more thought around how it can figure out the names, probably will be combined with the generator phase refactor. As it needs to know how the case conversion is done, the names of generated methods, wrappers etc.

ahayzen-kdab:
> Is this still relevant?

Potentially, when improving the error messages from the macro expansion in the next cycles, we need to consider if there are errors we need to detect and emit vs leaving them up to the compiler. A checks phase like CXX was the idea for that ...

PR #214, opened 2022-08-12
cxx-qt-gen: support base class for the qobject macro

ahayzen-kdab:


ahayzen-kdab:
Need documentation in the book and probably needs #212 so that we can easily always include `QtCore/QObject` ?

ahayzen-kdab:
> Need documentation in the book and probably needs #212 so that we can easily always include `QtCore/QObject` ?

Done

book/src/qobject/macro.md
```diff
@@ -31,4 +31,8 @@ Then when registering the type you use the type with your namespace as usual.
 {{#include ../../../examples/qml_with_threaded_logic/src/lib.rs:book_namespace_register}}
 ```
 
-Note: this might change in the future to allow for defining the base class or options when exporting to QML and could become namespaced to `#[cxx_qt::qobject(base = "QAbstractListModel")]` ( [https://github.com/KDAB/cxx-qt/issues/22](https://github.com/KDAB/cxx-qt/issues/22) ).
+You can also specify the base class by using `#[cxx_qt::qobject(base = "QStringListModel")]`, note that you need to use CXX to include any headers.
```
Be-ing, line 34:
```suggestion
You can also specify the base class by using `#[cxx_qt::qobject(base = "QStringListModel")]`. Note that you need to use CXX to include the header that the base class is declared in.
```

cxx-qt-gen/src/parser/cxxqtdata.rs
```diff
@@ -148,13 +149,18 @@ impl ParsedCxxQtData {
...
             if let Some(qobject) = self.qobjects.get_mut(&s.ident) {
+                // Find if there is any base class
+                if let Some(base) = attribute_tokens_to_map::<Ident, LitStr>(&s.attrs[index])?
+                    .get(&quote::format_ident!("base"))
+                {
+                    qobject.base_class = Some(base.value());
+                }
+
+                // Remove the macro from the struct
+                let mut s = s.clone();
```
Be-ing, line 161:
```suggestion
                let mut struct = s.clone();
```
Single letter variable names are a pet peeve of mine.

cxx-qt-gen/src/writer/cpp/header.rs
```diff
@@ -71,6 +71,8 @@ pub fn write_cpp_header(generated: &GeneratedCppBlocks) -> String {
           {members}
         }};
 
+        static_assert(std::is_base_of<QObject, {ident}>::value, "QObject must be the base of the class");
```
Be-ing, line 74:
```suggestion
        static_assert(std::is_base_of<QObject, {ident}>::value, "{ident} must inherit from QObject");
```
I think this wording makes it a bit clearer that the class doesn't need to *directly* inherit from QObject.

Be-ing:
This seems like a nice way to handle C++ inheritance with Rust.

cxx-qt-gen/src/parser/cxxqtdata.rs
ahayzen-kdab: Renaming to `item_struct` as `struct` is reserved.

book/src/qobject/macro.md
ahayzen-kdab: Done

cxx-qt-gen/src/writer/cpp/header.rs
ahayzen-kdab: Done

Issue #157, opened 2022-07-14
Tracking issue for new API changes for 0.4


ahayzen-kdab:
# Design

There are discussions open discussing the change of design of API so that we can become closer to a superset of CXX and allow to mix and match between. Some of the latest design points are in #156

# Main Steps

And while doing these consider any general refactoring that can occur :-) 

- [x] Rename `make_qobject` to `cxxqt::bridge` #158 
- [ ] ~~Move current definition to be inside an `extern "Qt"` block #159~~
- [x] Remove C++ namespace prefix #160 
- [x] Add namespace support into the macro attribute `cxxqt::bridge(namespace = "rust::object")` #194
- [ ] ~~Change `extract_qobject` / `extract_extern_qt` to respond with a vector and also have the passthrough contents for `extern "C++" / "Rust"` etc~~
- [x] Remove `handle_property_changed` as it won't be used anymore #151 
- [x] Remove QQmlExtensionPlugin helpers for now #152
- [x] Rearrange the generation so that the mod ffi is removed and we move the Qt code to another mod (see the discussion) #162 
- [x] Pass through extern blocks for now so we can get to the next stage #164
- [x] Require the user to define the types they use in the C++ extern, eg `type QPoint = cxx_qt_lib::QPoint` #167
- [x] Change invokable definition to be onto `impl Qt<T>` block #178
- [x] Rename types exposed to C++ to be MyObject and MyObjectRust #179
- [x] Add rust getter method to go from C++ to Rust #181 
- [x] Change opaque helper types to be UniquePtr and have the wrapping occur on the C++ side #165 #166 
- [x] Namespace internal methods so that we don't collide if QObject are in the same namespace #210 
- [x] Combine the Data and RustObj structs (once this has happened we can try https://github.com/ahayzen-KDAB/cxx-qt/tree/cxx-qt-libs-tidy-derives again) #235 
- [x] Add cxx_qt::qobject macro and use RustObj as the struct for the QObject for now #197 #198
- [x] Add the CppObj methods onto the Cpp obj as a method?
- [x] Put the `impl Qt<T>` block onto the Cpp object and update the wrappers to be namespaced free functions #202 
- [x] Add the macros for cxx_qt:::signals etc #180 
- [x] Remove pattern matching and let compiler failures handle typing issues on the C++ side #239 
- [x] Remove pattern matching and let compiler failures handle typing issues on the Rust side #254 
[...]

PR #216, opened 2022-08-16
cxx-qt-gen: add support for multiple QObjects in writer/cpp

ahayzen-kdab:


ahayzen-kdab:
This assumes that one module block will result in one C++ header file (which i believe will always be true?).

Our other debates are whether the module or file name should be used for the resultant files, so then that would determine if multiple module blocks result in multiple C++ header files or a single header file. However for now we only support one module per file anyway (and also one QObject per module, but that may change sooner)

ahayzen-kdab:
For now this can be draft as the Rust side needs figuring out in the CXX bridge to avoid collisions.

ahayzen-kdab:
See in #157 under "Support multiple QObject's being defined in one block" why the Rust side of this is blocked.

LeonMatthesKDAB (approved):
LGTM

PR #217, opened 2022-08-16
cxx-qt-gen: add a naming module to the generator phase

ahayzen-kdab:
This will handle taking the Parser components and then generating
the C++ or Rust idents from them.

### DIFF
--- a/cxx-qt-gen/src/gen_cpp.rs
+++ b/cxx-qt-gen/src/gen_cpp.rs
@@ -4,13 +4,13 @@
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
 use clang_format::{ClangFormatStyle, CLANG_FORMAT_STYLE};
-use convert_case::{Case, Casing};
 use indoc::formatdoc;
 use proc_macro2::TokenStream;
 use syn::Ident;
 
 use crate::extract::{Invokable, Parameter, ParameterType, Property, QObject, QtTypes, Signal};
 use crate::generator::cpp::{fragment::CppFragmentPair, GeneratedCppBlocks};
+use crate::generator::{naming, naming::property::QPropertyName};
 use crate::writer::cpp::write_cpp;
 
 /// A trait which we implement on QtTypes allowing retrieval of attributes of the enum value.
@@ -143,7 +143,6 @@ impl CppType for QtTypes {
 }
 
 /// Describes a C++ parameter, which is a name combined with a type
-#[derive(Debug)]
 struct CppParameter<'a> {
     /// The ident of the parameter
     ident: String,
@@ -152,7 +151,6 @@ struct CppParameter<'a> {
 }
 
 /// Describes a C++ invokable with header and source parts
-#[derive(Debug)]
 struct CppInvokable {
     /// The header definition of the invokable
     header: String,
@@ -160,7 +158,6 @@ struct CppInvokable {
     source: String,
 }
 /// Describes a C++ signal with header and source parts
-#[derive(Debug)]
 struct CppSignal {
     /// Any public methods that are defined by the signal
     header_public: Vec<String>,
@@ -170,7 +167,6 @@ struct CppSignal {
     source: String,
 }
 /// Describes a C++ property with header and source parts
-#[derive(Debug)]
 struct CppProperty {
     /// Any members that are required for the property
     header_members: Vec<String>,
@@ -187,7 +183,6 @@ struct CppProperty {
 }
 
 /// Describes a C++ header and source files of a C++ class
-#[derive(Debug)]
 pub struct CppObject {
     /// The header of the C++ class
     pub header: String,
@@ -267,7 +262,7 @@ fn generate_invokables_cpp(
         let body = format!(
             "m_rustObj->{ident}({parameter_names})",
             // If we are a pointer to a CppObj then we need a wrapper
-            ident = invokable.ident_wrapper.cpp_ident,
+            ident = invokable.ident_wrapper.cpp,
             parameter_names = (vec!["*this".to_owned()]
                 .into_iter()
                 .chain(parameters.names.into_iter()))
@@ -289,7 +284,7 @@ fn generate_invokables_cpp(
             // we are passing *this across for cpp objects in rust.
             header: format!(
                 "Q_INVOKABLE {return_ident} {ident}({parameter_types});",
-                ident = invokable.ident.cpp_ident,
+                ident = invokable.ident.cpp,
                 parameter_types = parameter_arg_line,
                 return_ident = return_ident,
             ),
@@ -315,7 +310,7 @@ fn generate_invokables_cpp(
                 } else {
                     body
                 },
-                ident = invokable.ident.cpp_ident.to_string(),
+                ident = invokable.ident.cpp.to_string(),
                 parameter_types = parameter_arg_line,
                 struct_ident = struct_ident.to_string(),
                 return_ident = return_ident,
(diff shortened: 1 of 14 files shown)


ahayzen-kdab:
This is a first phase of moving the naming of Rust and C++ methods into a common single place.

cxx-qt-gen/src/generator/naming/qobject.rs
```diff
@@ -0,0 +1,51 @@
...
+use crate::generator::naming::CombinedIdent;
+use crate::parser::qobject::ParsedQObject;
+use convert_case::{Case, Casing};
+use quote::format_ident;
+use syn::Ident;
+
+/// Names for parts of a Q_OBJECT
+pub struct QObjectName {
+    /// The name of the stem of the file
+    pub cxx_stem: Ident,
```
ahayzen, line 14: Needs more thought, should probably come from the module not the QObject at least

cxx-qt-gen/src/generator/naming/qobject.rs
```diff
@@ -0,0 +1,68 @@
...
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+use crate::generator::naming::CombinedIdent;
+use crate::parser::qobject::ParsedQObject;
+use quote::format_ident;
+use syn::Ident;
+
+/// Names for parts of a Q_OBJECT
+pub struct QObjectName {
```
ahayzen-kdab, line 11: If #215 lands first, then the TCxxQtThread name should be added to the naming here.

ahayzen-kdab:
(force pushed as a merge commit was added)

cxx-qt-gen/src/gen_cpp.rs
```diff
@@ -393,16 +389,24 @@ fn generate_properties_cpp(
...
-        let ident_changed = property.notify.as_ref().unwrap().cpp_ident.to_string();
+        let ident_getter = naming::property::getter_from_ident(&property.ident)
+            .cpp
+            .to_string();
+        let ident_setter = naming::property::setter_from_ident(&property.ident)
+            .cpp
+            .to_string();
+        let ident_changed = naming::property::notify_from_ident(&property.ident)
+            .cpp
+            .to_string();
```
LeonMatthesKDAB, line 409: These three calls should be combined into something like this:
``` rust
let prop_ident = QPropertyName::from(&property.ident);
```
Then you can just use prop_ident.getter.cpp, etc.

cxx-qt-gen/src/gen_rs.rs
```diff
@@ -214,15 +210,13 @@ pub fn generate_qobject_cxx(obj: &QObject) -> Result<ItemMod, TokenStream> {
...
+        let getter_rust = getter.rust.to_string();
 
-        // Build the C++ method declarations names
-        let getter_str = &property_ident_snake;
-        let getter_cpp = format_ident!("get{}", property_ident_pascal);
-        let setter_str = format!("set_{}", property_ident_snake);
-        let setter_cpp = format_ident!("set{}", property_ident_pascal);
+        let setter = naming::property::setter_from_ident(&property.ident);
+        let setter_cpp = &setter.cpp;
+        let setter_rust = setter.rust.to_string();
```
LeonMatthesKDAB, line 219: Again, this should be a single QPropertyName construction

cxx-qt-gen/src/gen_cpp.rs
LeonMatthesKDAB: Also include the above call to name_from_ident into this.

cxx-qt-gen/src/generator/naming/property.rs
```diff
@@ -0,0 +1,90 @@
...
+            name: name_from_ident(ident),
+            getter: getter_from_ident(ident),
+            setter: setter_from_ident(ident),
+            notify: notify_from_ident(ident),
+        }
+    }
+}
+
+/// For a given ident generate the Rust and C++ getter names
+pub fn getter_from_ident(ident: &Ident) -> CombinedIdent {
```
LeonMatthesKDAB, line 32: These function shouldn't be `pub` imho.
Just create a QPropertyName once, then access the different members.

cxx-qt-gen/src/generator/naming/signals.rs
```diff
@@ -0,0 +1,66 @@
...
+pub fn name_from_ident(ident: &Ident) -> CombinedIdent {
+    CombinedIdent {
+        cpp: format_ident!("{}", ident.to_string().to_case(Case::Camel)),
+        // Note that signal names are in camel case so we need to convert to snake and can't clone
+        rust: format_ident!("{}", ident.to_string().to_case(Case::Snake)),
+    }
+}
+
+/// For a given signal ident generate the Rust and C++ queued name
+pub fn queued_name_from_ident(ident: &Ident) -> CombinedIdent {
```
LeonMatthesKDAB, line 39: doesn't need to be pub.

cxx-qt-gen/src/generator/naming/invokable.rs
```diff
@@ -0,0 +1,66 @@
...
+        let ident = &invokable.sig.ident;
+        Self {
+            name: name_from_ident(ident),
+            wrapper: wrapper_from_ident(ident),
+        }
+    }
+}
+
+/// For a given ident generate the Rust and C++ names
+pub fn name_from_ident(ident: &Ident) -> CombinedIdent {
```
LeonMatthesKDAB, line 27: Doesn't need to be pub

LeonMatthesKDAB (changes requested):
Generally, I don't really like the free functions floating around in the naming module.

Please make them private in favor of constructing the appropriate Name structs and accessing into them.

cxx-qt-gen/src/gen_cpp.rs
ahayzen-kdab: So the general thing is that eventually they will be ... just porting the old code was easier to use the free functions. As eventually it'll be `QPropertyName::from(&property)` as in when property is `ParsedQProperty` and not an `extract::Property` ...

Could add some temporary conversions from the "old" style for now to make these free functions private but wasn't sure if it was worth it.

ahayzen-kdab: Changed properties, i'll sort the others out next week ...

LeonMatthesKDAB: Thanks, even though its more work now, I think its still valuable to do this right away.
Especially if we're reusing old code at some point, this might come back to bite us otherwise :see_no_evil:

ahayzen-kdab:
> Generally, I don't really like the free functions floating around in the naming module.
> 
> Please make them private in favor of constructing the appropriate Name structs and accessing into them.

Think i've fixed them all now...

ahayzen-kdab:
I wonder if `naming` should be within generator or if it should be it's own module ? :thinking: As the writer phase could use it depending what happens with the name mangling of free Rust functions (until CXX has static method support) :thinking:

LeonMatthesKDAB:
> I wonder if `naming` should be within generator or if it should be it's own module ? thinking As the writer phase could use it depending what happens with the name mangling of free Rust functions (until CXX has static method support) thinking

Hm, good point. I think it doesn't have to be part of the generator, it's useful on its own.
Rest looks good now and can be merged. I'll approve once we have made up our minds in which folder to put this mod.

ahayzen-kdab:
> > I wonder if `naming` should be within generator or if it should be it's own module ? thinking As the writer phase could use it depending what happens with the name mangling of free Rust functions (until CXX has static method support) thinking
> 
> Hm, good point. I think it doesn't have to be part of the generator, it's useful on its own. Rest looks good now and can be merged. I'll approve once we have made up our minds in which folder to put this mod.

Right, I'm not sure yet, it could be it's own module could be separate. Lets land this branch and then we can move it later easily if required.

LeonMatthesKDAB (approved):
Approved, might be moved out of the generator at a later date.

PR #218, opened 2022-08-22
cxx-qt-build: build generated C++ with cc from build.rs

Be-ing:


.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -155,6 +155,7 @@ jobs:
       run: >-
         sudo apt-get update &&
         sudo apt-get install -y
+        gcc-12
```
ahayzen-kdab, line 158: does this mean we require specifically gcc 12 ? why not just gcc / build-essential ?

cxx-qt-build/src/lib.rs
```diff
@@ -55,7 +54,7 @@ pub struct GeneratedCpp {
 
 impl GeneratedCpp {
     /// Generate QObject and cxx header/source C++ file contents
-    pub fn new(rust_file_path: &impl AsRef<std::path::Path>) -> Self {
+    pub fn new(rust_file_path: &impl AsRef<Path>) -> Self {
```
ahayzen-kdab, line 57: Just curious what the advantage of `&impl AsRef<Path>` over `&Path` is when `Path` is already a slice of `PathBuf` ?

cxx-qt-build/src/lib.rs
```diff
@@ -130,126 +129,97 @@ impl GeneratedCpp {
...
+    generated_file_paths
 }
 
 /// Describes a cxx Qt builder which helps parse and generate sources for cxx-qt
 #[derive(Default)]
 pub struct CxxQtBuilder {
     cpp_format: Option<ClangFormatStyle>,
     rust_sources: Vec<&'static str>,
-    qt_enabled: bool,
+    pub cc_builder: cc::Build,
```
ahayzen-kdab, line 222: Should this be a public field? or a getter ? 

But hmm looking further down the include() calls break the builder pattern. Why not have `include(mut self, ...)` as a method on `CxxQtBuilder` so that you can do

```rust
CxxQtBuilder::new()
        .cpp_format(ClangFormatStyle::Mozilla)
        .file("src/lib.rs")
        .include("include")
        .build();
```

instead of 

```rust
    let mut cxxqt_builder = CxxQtBuilder::new()
        .cpp_format(ClangFormatStyle::Mozilla)
        .file("src/lib.rs");
    cxxqt_builder.cc_builder.include("include");
    cxxqt_builder.build();
```

cxx-qt-build/src/lib.rs
```diff
@@ -268,43 +238,74 @@ impl CxxQtBuilder {
...
-        // TODO: later use the module::object to turn into module/object.h
+        self.cc_builder.cpp(true);
+        // MSVC
+        self.cc_builder.flag_if_supported("/std:c++17");
+        self.cc_builder.flag_if_supported("/Zc:__cplusplus");
+        self.cc_builder.flag_if_supported("/permissive-");
+        // GCC + Clang
+        self.cc_builder.flag_if_supported("-std=c++17");
+
+        // TODO: let user specify additional Qt modules to link
```
ahayzen-kdab, line 265: Should we do this now in this branch or the next one ?

Seems it'd just be like `qt_modules(mut self, &[&str])` then store those as a vec in the CxxQtBuilder, then use that list in this build method. It would default to [core,gui] and if the downstream developer writes their own it can replace them? (so this then allows them to clear the list too for non-Qt builds?)

cxx-qt-gen/src/syntax/qtitem.rs
```diff
@@ -10,6 +10,7 @@ use syn::parse::{Parse, ParseStream};
 use syn::{Attribute, Item, ItemMod, Result, Token, Visibility};
 
 #[derive(Clone, PartialEq, Eq)]
+#[allow(clippy::large_enum_variant)]
```
ahayzen-kdab, line 13: why is this suddenly needed? i thought we fixed the 1.63 clippy lints already?

cxx-qt-lib-headers/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+# SPDX-FileCopyrightText: 2021 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
```
ahayzen-kdab, line 1: 2022 ?

cxx-qt-lib-headers/src/lib.rs
```diff
@@ -0,0 +1,30 @@
+// SPDX-FileCopyrightText: 2021 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Be Wilson <be.wilson@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! This crate is a hack so build.rs for cxx-qt-lib and cxx-qt-build both have access to cxx-qt-lib's C++ headers.
```
ahayzen-kdab, line 6: this sounds fun :-)

cxx-qt-lib-headers/src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+pub static HEADERS: [(&str, &str); 3] = [
+    (include_str!("../include/convert.h"), "convert.h"),
+    (include_str!("../include/qt_types.h"), "qt_types.h"),
+    (
+        include_str!("../include/update_requester.h"),
+        "update_requester.h",
+    ),
+];
+
+pub fn write_headers(directory: &impl AsRef<Path>) {
```
ahayzen-kdab, line 22: again, still missing (probably something obvious) why not just `&Path` then `directory.display()` etc rather than needing to do the `.as_ref()` ?

.github/workflows/github-cxx-qt-tests.yml
Be-ing: yes, read the commit message

Cargo.toml
```diff
@@ -0,0 +1,24 @@
+# SPDX-FileCopyrightText: 2021 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
```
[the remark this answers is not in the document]
Be-ing: oops, silly copypasta error

cxx-qt-build/src/lib.rs
Be-ing: `&impl AsRef<Path>` allows passing `String` and `&str` so the caller doesn't need to explicitly create a `Path` or `PathBuf`

cxx-qt-build/src/lib.rs
Be-ing: yes, this just wasn't a priority before getting CI working on every OS

cxx-qt-gen/src/syntax/qtitem.rs
Be-ing: I have no idea why clippy decided to complain about this now because nothing in this crate changed. I tried rerunning CI for the last commit in the main branch to double check and it passed, so :shrug:

cxx-qt-lib-headers/src/lib.rs
Be-ing: lots of fun :)
image

cxx-qt-lib/build.rs
```diff
@@ -81,6 +42,11 @@ fn main() {
         builder.file(cpp_file);
         println!("cargo:rerun-if-changed={}", cpp_file);
     }
+
+    let out_dir = std::env::var("OUT_DIR").unwrap();
```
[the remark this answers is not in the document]
Be-ing: comment added

cxx-qt-build/src/lib.rs
Be-ing: We could make wrapper functions for the cc::Builder, but then it would be up to us to track any API changes to cc::Builder. So there is a trade off between maintenance burden for us and keeping strictly to a builder pattern API. I'd prefer the former.

ahayzen-kdab: OK what about a closure in the builder pattern to access and mutate the inner cc_builder, like this ? This then keeps the builder pattern, but exposes the cc_builder object so we don't have a larger maintenance burden ?

```rust
CxxQtBuilder::new()
        .cpp_format(ClangFormatStyle::Mozilla)
        .file("src/lib.rs")
        .cc_builder(|cc| cc.include("include"))
        .build();
```

Be-ing: clever. I like it :)

Be-ing: Implemented. Now the diffs for the build.rs scripts are trivial :)

.github/workflows/github-cxx-qt-tests.yml
ahayzen-kdab: OK, i see, maybe we should have a comment above this stating that isn't an issue on ubuntu with GCC 11. But not that we require GCC 12+ etc.

cxx-qt-build/src/lib.rs
ahayzen-kdab: OK, so for public API that makes sense i guess.

examples/qml_minimal/CMakeLists.txt
```diff
@@ -22,47 +16,55 @@ set(CMAKE_AUTORCC ON)
...
-    CPP_SOURCES
-    ${CMAKE_CURRENT_SOURCE_DIR}/src/main.cpp
-)
+find_package(QT NAMES Qt6 Qt5 COMPONENTS Core Gui Qml QuickControls2 QmlImportScanner REQUIRED)
+find_package(Qt${QT_VERSION_MAJOR} COMPONENTS Core Gui Qml QuickControls2 QmlImportScanner REQUIRED)
+# The path to the qmake executable path needs to be passed to the Rust library's build script to ensure
+# it uses the same installation of Qt as CMake.
+get_target_property(QMAKE Qt${QT_VERSION_MAJOR}::qmake IMPORTED_LOCATION)
+
+find_package(Corrosion)
```
ahayzen-kdab, line 28: As on the next line if the package isn't found we then fetch it, i would prefer this not to warn (as it's a false warning). So i think put `find_package(Corrosion QUIET)` instead?

cxx-qt-build/src/lib.rs
```diff
@@ -258,7 +261,9 @@ impl CxxQtBuilder {
         Self {
             cpp_format: None,
             rust_sources: vec![],
-            qt_enabled: true,
+            qobject_headers: vec![],
+            qt_modules: vec!["Core".to_owned(), "Gui".to_owned()],
```
ahayzen-kdab, line 265: If they can't have duplicates, then this should be a set rather than a vector ? like a std::collections::HashSet ?

cxx-qt-build/src/lib.rs
```diff
@@ -268,43 +273,115 @@ impl CxxQtBuilder {
...
     /// Currently the path should be relative to CARGO_MANIFEST_DIR
     pub fn file(mut self, rust_source: &'static str) -> Self {
         self.rust_sources.push(rust_source);
+        println!("cargo:rerun-if-changed={}", rust_source);
+        self
+    }
+
+    /// Link additional [Qt modules](https://doc.qt.io/qt-6/qtmodules.html).
+    /// Specify their names without the `Qt` prefix, for example `"Widgets"`.
+    /// Do *not* specify "Core" or "Gui"; these are always linked.
```
ahayzen-kdab, line 287: Once qt_modules is a set then it doesn't matter if they specify the same ones twice.

cxx-qt-gen/src/syntax/qtitem.rs
ahayzen-kdab: Can we try to remove this ? As locally it passes without it. And i'd prefer the project to have default clippy settings where possible.

cxx-qt-lib-headers/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+# SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+# SPDX-FileContributor: Be Wilson <be.wilson@kdab.com>
+#
+# SPDX-License-Identifier: MIT OR Apache-2.0
+[package]
```
ahayzen-kdab, line 5: Can you add more metadata here for when we publish, eg copy from cxx-qt-lib the license, repository, description, authors etc.

cxx-qt-lib-headers/src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+static HEADERS: [(&str, &str); 3] = [
+    (include_str!("../include/convert.h"), "convert.h"),
+    (include_str!("../include/qt_types.h"), "qt_types.h"),
+    (
+        include_str!("../include/update_requester.h"),
+        "update_requester.h",
+    ),
+];
+
+pub fn write_headers(directory: &impl AsRef<Path>) {
```
ahayzen-kdab, line 22: documentation string ?

cxx-qt-build/src/lib.rs
Be-ing: done

cxx-qt-lib-headers/Cargo.toml
Be-ing: no, that isn't applicable to virtual manifests; there is no `[package]` https://doc.rust-lang.org/cargo/reference/workspaces.html

cxx-qt-gen/src/syntax/qtitem.rs
Be-ing: clippy fails when I comment this out locally

.github/workflows/github-cxx-qt-tests.yml
Be-ing: Well, it is an issue with GCC 11 on Ubuntu. GCC 11 on Ubuntu only works if lld (or likely mold as well) is used, but not with GNU ld.

cxx-qt-build/src/lib.rs
```diff
@@ -268,43 +277,112 @@ impl CxxQtBuilder {
...
+        println!("cargo:rerun-if-changed={}", rust_source);
         self
     }
 
-    // TODO: support globs with files("src/**/*.rs")
+    /// Link additional [Qt modules](https://doc.qt.io/qt-6/qtmodules.html).
+    /// Specify their names without the `Qt` prefix, for example `"Widgets"`.
+    /// The Core and Gui modules are linked automatically; there is no need to specify them.
+    pub fn qt_modules(mut self, modules: &[&str]) -> Self {
+        for module in modules {
```
ahayzen-kdab, line 293: seems this can also be

```rust
        self.qt_modules.extend(modules.iter().cloned().map(String::from));
```
not sure if it's better or worse :-)

Be-ing: sure

cxx-qt-gen/src/syntax/qtitem.rs
ahayzen-kdab: Do you have extra lints enabled or something ? As i have `clippy 0.1.63 (4b91a6e 2022-08-08)` and it doesn't...

cxx-qt-lib-headers/Cargo.toml
ahayzen-kdab: Hmm but it does have a `[package]` ? so `cxx-qt-lib-headers` is not a crate that would go to crates.io ?

Be-ing: oh sorry I thought you were commenting on the workspace manifest

.github/workflows/github-cxx-qt-tests.yml
ahayzen-kdab: Right, so just put a comment above, so when we upgrade to Ubuntu 24.04 we know we can probably drop it for gcc.

cxx-qt-lib-headers/Cargo.toml
Be-ing: added

ahayzen-kdab (approved):
LGTM :-)

PR #221, opened 2022-08-25
add example application built entirely with Cargo

Be-ing:
:crab: 

image


ahayzen-kdab (changes requested):
Looks good, but for me it fails to link to Qt ...

ahayzen-kdab (approved):
LGTM, reported #223 and #224 as issue to look at later.

LeonMatthesKDAB:
As a follow-up:
Do we want to replace qml_minimal with this?
Then we don't have to maintain what's basically the same code twice?
Or could we even have qml_minimal with two separate build scripts, one for building with CMake and one without?

ahayzen-kdab:
> As a follow-up: Do we want to replace qml_minimal with this? Then we don't have to maintain what's basically the same code twice? Or could we even have qml_minimal with two separate build scripts, one for building with CMake and one without?

I think we probably do as they are different use cases. One is showing the minimum to integrate a Rust QML module into a CMake C++ project. And the other is showing the minimum to create a from scratch Rust QML module building with cargo. And as the cargo build has some issues still the qml_minimal one is a good example for the book etc.

LeonMatthesKDAB:
Right, I'm just wondering whether we can reduce the code duplication for the actual `src` code.
Could we symlink e.g. `src/cxxqt_object.rs` and `main.qml` to the qml_minimal ones?

Be-ing:
Symlinking the duplicated files is implemented in #226

PR #222, opened 2022-08-26
cxx-qt-lib: remove derive Copy and use Clone instead

ahayzen-kdab:


tests/qt_types_standalone/src/lib.rs
```diff
@@ -388,7 +388,7 @@ fn read_qdate(d: &QDate) -> bool {
 }
 
 fn copy_qdate(d: &QDate) -> QDate {
```
Be-ing, line 390: rename these to `clone_X` as well?

ahayzen-kdab: Can do they are just test method names...

ahayzen-kdab: Done

Be-ing (approved):

PR #227, opened 2022-08-26
move Rust crates to their own subdirectories

Be-ing:

PR #229, opened 2022-08-26
move published crates to a new `crates` directory

Be-ing:
depends on #227


LeonMatthesKDAB:
Is it possible that you forgot to `git add` the crates folder?
I think its missing atm :thinking:

ahayzen-kdab (approved):
LGTM :-)

PR #231, opened 2022-08-26
Update README to remove mentions of CppObj.

LeonMatthesKDAB:


Be-ing (approved):
It would be nice if we could compile this code in README.md in a test to ensure it keeps working.

PR #232, opened 2022-08-29
update book for new build process

Be-ing:
based on #227


book/src/getting-started/5-cmake-integration.md
```diff
@@ -26,58 +23,71 @@ edition = "2021"
...
 If this fails for any reason, take a look at the [`examples/qml_minimal`](https://github.com/KDAB/cxx-qt/tree/main/examples/qml_minimal) folder, which contains the complete example code.
 
 This should now configure and compile our project.
 If this was successful, you can now run our little project.
 ```shell
-$ ./qml_minimal
+$ build/qml_minimal
 ```
 
 You should now see the two Labels that display the state of our `MyObject`, as well as the two buttons to call our two Rust functions.
```
ahayzen-kdab, line 93: To the "Success" and "Further reading" part below, maybe add link to the `6-cargo-executable` example?

Be-ing: I don't linking is necessary because it's clear from the table of contents and pressing next. But I could add a sentence or two to transition to the next chapter.

ahayzen-kdab: Maybe just a statement somewhere saying, how 5 describing using CMake. But it is possible not just use Cargo and 6 describes this?

ahayzen-kdab (approved):
LGTM :-)

PR #233, opened 2022-08-30
README: add table comparing to other Qt Rust bindings

Be-ing:


LeonMatthesKDAB:
Good addition.
In the interest of a fair comparison, do we want to add a `Requires C++`, or maybe `Rust only` column?
Because I think for many Rust devs that s something they really want to know and CXX-Qt just doesn't provide that.

Be-ing:
Hmm, while technically yes, cxx-qt does currently require writing some C++, it's just a bit of boilerplate that I think someone who doesn't really know what they're doing with C++ could write without too much trouble. So I'm not sure we should add that to the table.

README.md
```diff
@@ -120,6 +120,20 @@ mod ffi {
 }
 ```
 
+## Comparison to other Rust Qt bindings
+
+| project |  integrate into C++ codebase | safe Rust | maintained* | binding mechanism |
+|-------- | ---------------------------- | --------- | ----------- | ----------------- |
+| cxx-qt  |  ✔                           | ✔        |  ✔          | [cxx](https://cxx.rs) plus additional code generation to implement QObject subclasses in Rust and bind them to C++ |
```
ahayzen-kdab, line 127: We don't state where CXX-Qt is *not* suited ?

README.md
```diff
@@ -120,6 +120,20 @@ mod ffi {
...
+| [qmetaobject](https://github.com/woboq/qmetaobject-rs/) | ✗ | ✔ | ✔ | [cpp](https://github.com/mystor/rust-cpp) macro to write C++ inline in Rust, plus Rust macros to create QObject subclasses from Rust structs |
```
ahayzen-kdab, line 128: `qmetaobject` can sortof integrate with C++ by creating a `QQmlExtensionPlugin`, although I don't think there is a way to directly use the classes created by Rust like CXX-Qt can..

README.md
```diff
@@ -120,6 +120,20 @@ mod ffi {
...
+| [Rust Qt binding generator](https://invent.kde.org/sdk/rust-qt-binding-generator) | ✔ | ✔ | ✗ | generates Rust traits and C++ bindings from JSON description of QObject subclass |
+| [rust-qt](https://rust-qt.github.io/) | ✗ | ✗ | ✗ | [ritual](https://rust-qt.github.io/ritual/) to generate unsafe Rust bindings from C++ headers |
+| [qml-rust](https://github.com/White-Oak/qml-rust) | ✗ | ✔ | ✗ | [DOtherSide](https://github.com/filcuc/DOtherSide) C wrapper for QML C++ classes |
+| [qmlrs](https://github.com/flanfly/qmlrs) | ✗ | ✔ | ✗ | own C++ library to bind QQmlApplicationEngine |
+| [qmlrsng](https://github.com/nbigaouette/qmlrsng) | ✗ | ✔ | ✗ | [libqmlbind](https://github.com/seanchas116/libqmlbind) with [bindgen](https://rust-lang.github.io/rust-bindgen/) |
+| [rust-qml](https://github.com/florianjacob/rust-qml) | ✗ | ✔ | ✗ | [libqmlbind](https://github.com/seanchas116/libqmlbind) |
+
+*maintained: repository has had nontrivial commits within last year as of August 2022
```
ahayzen-kdab, line 136: One of those unfortunate things that will make the README look out of date unless we constantly update the date :thinking:

README.md
```diff
@@ -120,6 +120,20 @@ mod ffi {
 }
 ```
 
+## Comparison to other Rust Qt bindings
```
ahayzen-kdab, line 123: BTW we had come up with this while making a presentation ...

A list of use cases where existing solutions may be suitable
```
I want to use arbitrary Qt API unsafely from Rust
* Ritual
I want to attempt to remove C++ and write everything in Rust
* qmetaobject-rs
I have a small data boundary between Rust and Qt and don’t mind writing binding definitions by hand
* rust-qt-binding-generator
```

And then a list of use cases where CXX-Qt is suited
```
CXX-Qg
* Step-by-step ports to Rust
* Projects with a large amount of QML/C++ knowledge
* Clear data boundary between Rust and Qt
* Integrate C++ and Rust ecosystems
```

README.md
Be-ing: QWidgets, but kinda... it should still be feasible to use cxx-qt to implement Qt models in Rust within a QWidgets application even if we don't have bindings for QWidgets in cxx-qt-lib

README.md
Be-ing: yeah... my intention with this is to have some straightforward criterion to classify maintained versus unmaintained

README.md
Be-ing: > I have a small data boundary between Rust and Qt and don’t mind writing binding definitions by hand
* rust-qt-binding-generator

I would recommend cxx-qt for this use case over an unmaintained project.

README.md
ahayzen-kdab: Qt Widgets should be fine, as CXX-Qt is just trying to add enough syntax to CXX so that you can describe a QObject from Rust. Eg we allow you to define properties, signals, invokables (for QML), slots, base classes etc. And by using the new syntax you should be able to describe existing widgets API and objects that you need to expose across. A few more common types could be added to cxx-qt-lib (eg QByteArray) but I don't think we would need to add eg QPushButton.

But this comes to the area where CXX-Qt is weaker, in CXX-Qt you mainly create a split between your UI layer and your business logic layer. Then you should use models or QObject's from CXX-Qt to describe state. And finally then use the existing QWidgets/C++ or QML/JS API for describing the UI. Some of the other bindings however allow you to write the UI layer in "Rust" as well (although they tend to just be a wrapped C++ API).

README.md
ahayzen-kdab: Thing is what is "maintained" as you state `rust-qt-binding-generator` as unmaintained, whereas they claim it as "complete". This also creates an arbitrary maintained period for us where CXX-Qt might appear to be unmaintained when it is infact complete.

Be-ing: You could say the ritual bindings are "complete" too, but they don't work with Qt6. They don't even build when Qt6 is installed. So I do think "maintained" is a useful column to have in the table.

README.md
Be-ing: Come to think of it, I wouldn't really call that integrating with C++. That's integrating with QML.

ahayzen-kdab (approved):
Looks good, lets start with this.

PR #235, opened 2022-09-01
cxx-qt-gen: remove the Data struct and move properties to Rust

ahayzen-kdab:


cxx-qt-gen/src/gen_cpp.rs
```diff
@@ -346,6 +346,8 @@ fn generate_properties_cpp(
         let ident_getter = property.getter.as_ref().unwrap().cpp_ident.to_string();
         let ident_setter = property.setter.as_ref().unwrap().cpp_ident.to_string();
         let ident_changed = property.notify.as_ref().unwrap().cpp_ident.to_string();
+        // TODO: share in naming diff
+        let ident_emit = format!("emit{}", ident_changed.to_case(Case::Pascal));
```
ahayzen-kdab, line 350: if the naming branch lands first, we need to change this to use it.

examples/qml_minimal/src/cxxqt_object.rs
```diff
@@ -15,36 +15,37 @@ mod ffi {
     unsafe extern "C++" {
         include!("cxx-qt-lib/include/qt_types.h");
         type QString = cxx_qt_lib::QString;
+        type QPoint = cxx_qt_lib::QPoint;
     }
 
     // ANCHOR: book_data_struct
-    pub struct Data {
+    // TODO: remove book links
```
ahayzen-kdab, line 22: still need to update the book

examples/qml_minimal/src/cxxqt_object.rs
```diff
@@ -15,36 +15,37 @@ mod ffi {
...
+    // ANCHOR_END: book_data_struct
+
+    // ANCHOR: book_rustobj_struct
+    #[cxx_qt::qobject]
+    pub struct MyObject {
+        #[qproperty]
         number: i32,
+        #[qproperty]
         string: UniquePtr<QString>,
```
ahayzen-kdab, line 31: Having an opaque type C++ type stored inside a UniquePtr in the struct then causes the build to fail as `Opaque types are currently required to be [Sized](https://doc.rust-lang.org/std/marker/trait.Sized.html) and [Unpin](https://doc.rust-lang.org/std/marker/trait.Unpin.html).` https://cxx.rs/extern-rust.html#opaque-rust-types and the QString is opaque so uses PhantonPinned which doesn't implement Unpin ;-/

ahayzen-kdab: Using `string: Box<UniquePtr<QString>>` looks like it might work :thinking:

ahayzen-kdab: It does, but how do we express this to the developer. Or do they need to write `Box<UniquePtr<T>>` when they want to have T as a property type ? And how do we know that this type needs special conversions to T for C++ ? And what happens to the getters/setters on the Rust side? Do they return `Box<UniquePtr<T>>` ? etc

Then do we have a special type such as `Owned<T>` or something?

ahayzen-kdab: Solved with https://github.com/dtolnay/cxx/pull/1091

cxx-qt/src/lib.rs
```diff
@@ -118,13 +116,11 @@ pub fn qobject(_args: TokenStream, _input: TokenStream) -> TokenStream {
 /// #[cxx_qt::bridge]
 /// mod my_object {
 ///     #[derive(Default)]
-///     struct Data {
+///     struct RustObj {
```
LeonMatthesKDAB, line 119: This is missing a `#[cxx_qt::qobject]` attribute

cxx-qt/src/lib.rs
```diff
@@ -16,13 +16,11 @@ use cxx_qt_gen::{extract_qobject, generate_qobject_rs};
 /// #[cxx_qt::bridge(namespace = "cxx_qt::my_object")]
 /// mod my_object {
 ///     #[derive(Default)]
-///     struct Data {
+///     struct RustObj {
```
LeonMatthesKDAB, line 19: This is missing a #[cxx_qt::qobject] attribute

LeonMatthesKDAB (commented):
Noticed a few bits missing

cxx-qt/src/lib.rs
ahayzen-kdab: Done

cxx-qt/src/lib.rs
ahayzen-kdab: Done

LeonMatthesKDAB (approved):
LGTM

PR #239, opened 2022-09-05
cxx-qt-gen: remove gen_cpp and use new phases for C++ side

ahayzen-kdab:
Depends on #248


cxx-qt-gen/src/generator/cpp/signal.rs
```diff
@@ -0,0 +1,114 @@
...
+
+        // Generate the Q_SIGNAL
+        let parameter_types = signal
+            .parameters
+            .iter()
+            .map(|parameter| {
+                Ok(format!(
+                    "{cpp_ty} {ident}",
+                    ident = parameter.ident,
+                    cpp_ty = if let Some(cpp_ty) = &parameter.cpp_type {
```
ahayzen-kdab, line 39: can this be common to the types.rs ?

crates/cxx-qt-gen/src/generator/cpp/signal.rs
```diff
@@ -0,0 +1,173 @@
...
+
+        generated.signals.push(format!(
+            "void {ident}({parameters});",
+            ident = signal_ident,
+            parameters = parameter_types
+        ));
+
+        // Generate the queued emitters
+        //
+        // TODO: perform this in one loop
```
ahayzen-kdab, line 53: should be able to move all of these into one loop over the parameters.

crates/cxx-qt-gen/src/generator/cpp/fragment.rs
```diff
@@ -8,3 +8,8 @@ pub struct CppFragmentPair {
     pub header: String,
     pub source: String,
 }
+
+pub struct CppNamedType {
+    pub ident: String,
+    pub ty: String,
```
LeonMatthesKDAB, line 14: Should this `ty` be a CxxType?

LeonMatthesKDAB: Also, regarding naming, CppNamedType or CxxNamedType?
We should choose one for our internal APIs.
Atm, we're always using Cpp, so maybe change CxxType to CppType?

ahayzen-kdab: Regarding teh Cpp vs Cxx naming i thought that too. But maybe it's better to do a mass rename at the end as we already have a mixture of cpp and cxx module names etc etc?

crates/cxx-qt-gen/src/generator/cpp/invokable.rs
```diff
@@ -0,0 +1,214 @@
...
+
+                    // If the name of the argument is self then ignore,
+                    // as this is likely the self: Pin<T>
+                    if ident == "self" {
+                        Ok(None)
+                    } else {
+                        Ok(Some(CppNamedType {
+                            ident: ident.to_string(),
+                            ty: format!(
+                                "{ty} {ident}",
```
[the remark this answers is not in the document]
ahayzen-kdab: Removed :-)

crates/cxx-qt-gen/src/generator/cpp/fragment.rs
ahayzen-kdab: Changed to be CxxType :-)

ahayzen-kdab: And renamed things to CppType so we use Cpp for now.

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
```diff
@@ -0,0 +1,184 @@
...
+use syn::Result;
+
+mod emitter;
+mod getter;
+mod meta;
+mod setter;
+mod signal;
+
+pub fn generate_cpp_properties(
+    generated: &mut GeneratedCppQObjectBlocks,
```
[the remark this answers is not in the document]
ahayzen-kdab: As discussed changed :-)

LeonMatthesKDAB (approved):
LGTM

PR #240, opened 2022-09-05
Prepare for cpp generator

ahayzen-kdab:


cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -19,12 +27,15 @@ pub struct ParsedQObject {
     //
     // TODO: this should not be optional
     pub qobject_struct: Option<ItemStruct>,
+    /// The namespace of the QObject, if one isn't specified for the QObject
```
Be-ing, line 30:
```suggestion
    /// The namespace of the QObject. If one isn't specified for the QObject,
```

Be-ing:
I'm unclear what the purpose of these new `cpp_type` and `return_cpp_type` attributes are. In every case, they are used for the type inside a UniquePtr. I'm unclear if they're required when using UniquePtr? If so, can't we automate this? Requiring downstream code to handle this seems like asking for trouble.

ahayzen-kdab:
> I'm unclear what the purpose of these new `cpp_type` and `return_cpp_type` attributes are. In every case, they are used for the type inside a UniquePtr. I'm unclear if they're required when using UniquePtr? If so, can't we automate this? Requiring downstream code to handle this seems like asking for trouble.

So as we move away from pattern matching to support arbitrary types in the properties, invokables, and signals. We have problems with CXX and opaque types, when going from Rust -> C++ we either need to respond with a reference or a UniquePtr or a SharedPtr to the opaque type.

The old generation had special pattern matching on opaque Qt types to do this behind the scenes but we don't want this as we want to support custom types as well. We can't do it for all UniquePtr's in the long term as we also want to support having UniquePtr as it's own type. Then on the Qt side, we need by value for the property and invokable returns.

So the current solution likely for 0.4 is to have an attribute which specifies the C++ type that should be used, this then uses our templated conversion methods to go from UniquePtr<T> -> T etc. (It could also be used by developers to go from Rust type X to C++ type Y).

Later we may be able to submit something upstream to CXX to help with invokables, but properties are still a problem as they are a field that needs to be a UniquePtr on the Rust side but not on the other side (so we need a way of knowing what the C++ signature should be anyway). This is something I have to discuss next week to come up with the best long term solution. As CXX have already rejected some possible solutions, so we may have to create some middleware for us. Furthermore this is also linked to other issues around using CXX with existing Qt API where the Qt API returns opaque types, so we need to find a general solution long term to the use cases.

tl;dr; for now with 0.4 we'll likely have these attributes for opaque types, long term we may be able to improve the situation but will require some thought.

cxx-qt-gen/src/parser/qobject.rs
ahayzen-kdab: Done

Be-ing:
Oof, that's ugly but it makes sense. Can we add `std::is_convertible` or other static assertions to verify that the types are actually convertible? I tried using gibberish nonexistent types for both `cpp_type` and `return_cpp_type` and in both cases it compiled.

Be-ing:
>  CXX have already rejected some possible solutions, so we may have to create some middleware for us. Furthermore this is also linked to other issues around using CXX with existing Qt API where the Qt API returns opaque types, so we need to find a general solution long term to the use cases.

Can you link to those upstream discussions?

Be-ing:
Hmm, the static asserts don't seem to be working. I can still compile specifying nonexistent types with `cpp_type` and `return_cpp_type`.

ahayzen-kdab:
> Hmm, the static asserts don't seem to be working. I can still compile specifying nonexistent types with `cpp_type` and `return_cpp_type`.

Yup because this branch doesn't result in the cpp_type or return_cpp_type ending up in the generated C++ yet, you'd need to try with #239 for that.

README.md
```diff
@@ -68,7 +68,7 @@ mod ffi {
     pub struct MyObject {
         #[qproperty]
         pub number: i32,
-        #[qproperty]
+        #[qproperty(cpp_type = "QString")]
```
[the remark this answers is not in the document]
ahayzen-kdab: Done :-)

README.md
```diff
@@ -99,10 +99,10 @@ mod ffi {
         pub fn reset(mut self: Pin<&mut Self>) {
             let data: DataSerde = serde_json::from_str(DEFAULT_STR).unwrap();
             self.as_mut().set_number(data.number);
-            self.as_mut().set_string(&QString::from_str(&data.string));
+            self.as_mut().set_string(QString::from_str(&data.string));
         }
 
-        #[qinvokable]
+        #[qinvokable(return_cxx_type = "QString")]
```
LeonMatthesKDAB, line 105: Do we also have a way to also declare the parameter C++ types? :thinking:

crates/cxx-qt-gen/test_outputs/naming.cpp
```diff
@@ -34,19 +34,20 @@ MyObject::qtThread() const
                                                m_rustObjMutex);
 }
 
-qint32
+const qint32&
```
LeonMatthesKDAB, line 37: Later we may want to differentiate between trivial and non-trivial types, because a qint32 doesn't really benefit from being returned by reference.

LeonMatthesKDAB (approved):
LGTM

README.md
ahayzen-kdab: So far we don't need it, as you can do references eg `&QVariant`. Maybe later we might need something.

crates/cxx-qt-gen/test_outputs/naming.cpp
ahayzen-kdab: Right, the fun of trying to work out which are trivial is the main issue, maybe we can have an include list to match them or something. But for now keeping the generation consistent was simplest for the next stages...

PR #244, opened 2022-09-13
README: rewrite introduction

Be-ing:


README.md
```diff
@@ -9,19 +9,19 @@ SPDX-License-Identifier: MIT OR Apache-2.0
...
-the source for a crate which provides helper functions to be used in a `build.rs` file.
-
-Initially the projects in the examples folder will also serve as a template for new projects should use CXX-Qt.
-In future we might improve upon this with a custom CMake module for instance.
+CXX-Qt is a set of Rust crates for creating bidirectional Rust ⇄ C++ bindings with [Qt](https://www.qt.io/).
+It can be used to integrate Rust into C++ applications using CMake or used to build Rust applications with Cargo.
+CXX-Qt provides tools for implementing [QObject](https://doc.qt.io/qt-6/object.html) subclasses in Rust which can
+be used from C++, QML, and JavaScript. It consists of two parts:
+
+1. cxx-qt-lib, a library of Rust bindings to common QtCore and QtGui classes made with [CXX](https://cxx.rs/)
```
ahayzen-kdab, line 17: personally prefer to use bullet points rather than numbers points if they aren't being referred to

README.md
```diff
@@ -9,19 +9,19 @@ SPDX-License-Identifier: MIT OR Apache-2.0
...
+
+2. cxx-qt + cxx-qt-build, a pair of Rust & C++ code generators which are a superset of CXX plus additional attributes
```
ahayzen-kdab, line 19: `and` is used elsewhere but `&` is used here, i generally prefer `and`

README.md
```diff
@@ -9,19 +9,19 @@ SPDX-License-Identifier: MIT OR Apache-2.0
...
+to interface with Qt's [signals & slots](https://doc.qt.io/qt-6/signalsandslots.html) and [property system](https://doc.qt.io/qt-6/properties.html).
+The cxx-qt crate implements a macro for Rust code generation. cxx-qt-build is used in [Cargo build scripts](https://doc.rust-lang.org/cargo/reference/build-scripts.html)
```
ahayzen-kdab, line 21: we have been styling as CXX-Qt in text, but if this is referring to the crate name, maybe put it in a code block like `cxx-qt`

README.md
Be-ing: I like using `&` for closely coupled pairs to help avoid getting lost in the words.

README.md
Be-ing: Eh, I don't think that would help the readability :shrug:

ahayzen-kdab (approved):
Looks good, could maybe mention cxx-qt-gen as well, but then maybe the developer doesn't need to know about that as it's an implementation detail of cxx-qt and cxx-qt-build.

Be-ing:
Yeah, I only mentioned the crates downstream users will use directly. The rest are implementation details.

PR #245, opened 2022-09-13
remove connections to QQmlApplicationEngine::objectCreated

Be-ing:
I don't know why these were there. The applications run fine without them.

### DIFF
--- a/examples/cargo_without_cmake/src/cpp/run.cpp
+++ b/examples/cargo_without_cmake/src/cpp/run.cpp
@@ -40,15 +40,6 @@ run_cpp(int argc, char* argv[])
   QQmlApplicationEngine engine;
 
   const QUrl url(QStringLiteral("qrc:/main.qml"));
-  QObject::connect(
-    &engine,
-    &QQmlApplicationEngine::objectCreated,
-    &app,
-    [url](QObject* obj, const QUrl& objUrl) {
-      if (!obj && url == objUrl)
-        QCoreApplication::exit(-1);
-    },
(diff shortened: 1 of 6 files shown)


Be-ing:
@LeonMatthesKDAB added these in #7. I don't see any discussion on that old PR about them.

ahayzen-kdab:
These come from the Qt Creator template for QML applications, I believe it is related to if loading the QML was unsuccessful then it provides an exit code etc.

LeonMatthesKDAB:
I think I just copied it from the existing code in the [`basic_cxx_qt_qml`](https://github.com/LeonMatthesKDAB/cxx-qt/blob/a23c005b7a704774c4df4aa337ca269346e98393/examples/basic_cxx_qt_qml/src/main.cpp) example that already existed back then.

The docs on this signal are here: https://doc.qt.io/qt-6/qqmlapplicationengine.html#objectCreated
It says that NULL is emitted if there was an error when loading a QML object.
So I'd keep the connection, as it provides us with a reproducable exit in that case.

Be-ing:
It would help to add a comment explaining what this does. I'm sure this looks quite strange to people not very familiar with C++.

ahayzen-kdab:
> It would help to add a comment explaining what this does. I'm sure this looks quite strange to people not very familiar with C++.

Right a code comment would be useful (a separate PR would be good) :-) As a Qt developer though it comes from the Qt Creator template so they are probably used to it, even if they don't know what it does! :see_no_evil:

PR #250, opened 2022-09-15
cxx-qt-gen: add support for multiple qobjects in rust writer

ahayzen-kdab:
Requires #249 #251

### DIFF
--- a/crates/cxx-qt-gen/src/writer/rust/mod.rs
+++ b/crates/cxx-qt-gen/src/writer/rust/mod.rs
@@ -3,11 +3,11 @@
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
-use crate::generator::rust::GeneratedRustBlocks;
+use crate::generator::rust::{qobject::GeneratedRustQObjectBlocks, GeneratedRustBlocks};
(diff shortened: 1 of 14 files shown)


crates/cxx-qt-gen/src/writer/rust/mod.rs
```diff
@@ -76,33 +72,70 @@ fn cxx_common_blocks(
...
+
+        // Inject the common blocks into the bridge which we need
+        let cxx_mod_items = &mut cxx_mod.content.as_mut().expect("").1;
+        cxx_mod_items.extend_from_slice(&qobject.cxx_mod_contents);
+        for block in cxx_common_blocks(
+            cpp_struct_ident,
+            rust_struct_ident,
+            cxx_qt_thread_ident,
+            namespace_internals,
+        ) {
```
Be-ing, line 112: Instead of extracting all these struct properties to pass to this function, how about refactoring `cxx_common_blocks` to take the `qobject`?

ahayzen-kdab: Refactored that whole area, i think it looks better now :-)

Be-ing (approved):

PR #252, opened 2022-09-15
README: remove inline example

Be-ing:
We do not have the code in this example built and tested, so it is cumbersome to keep up-to-date. It also makes the README rather crowded. Instead, refer readers to the book and examples.


ahayzen-kdab (approved):
LGTM :-)

PR #254, opened 2022-09-16
cxx-qt-gen: remove gen_rust and use new phases for Rust side

ahayzen-kdab:
Requires #255


crates/cxx-qt-gen/src/lib.rs
```diff
@@ -214,4 +249,46 @@ mod tests {
...
+            syn::parse_str(include_str!("../test_inputs/types_qt_invokable.rs")).unwrap(),
+        )
+        .unwrap();
+        let generated = GeneratedRustBlocks::from(&parser).unwrap();
+        let rust = format_rs_source(&write_rust(&generated).to_string());
+
+        let expected_output =
+            format_rs_source(include_str!("../test_outputs/types_qt_invokable.rs"));
+        assert_str_eq!(rust, expected_output);
+    }
```
Be-ing, line 293: This test code is redundant. It could be refactored into a common function that takes the filename as an argument.

crates/cxx-qt-gen/src/generator/rust/fragment.rs
```diff
@@ -0,0 +1,29 @@
...
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use proc_macro2::TokenStream;
+use syn::{Item, Result};
+
+/// A generic Rust CXX bridge definition and source pairing
+pub struct RustFragmentPair {
+    pub cxx: Vec<TokenStream>,
+    pub source: Vec<TokenStream>,
+}
```
Be-ing, line 13:
```suggestion
/// A generic Rust CXX bridge definition and the corresponding implementation
pub struct RustFragmentPair {
    pub cxx_bridge: Vec<TokenStream>,
    pub implementation: Vec<TokenStream>,
}
```

I did not understand this on first read and had to see some example usages to understand what these fields indicated.

crates/cxx-qt-gen/src/generator/rust/qobject.rs
```diff
@@ -18,9 +45,140 @@ pub struct GeneratedRustQObject {
...
+            namespace_internals: namespace_idents.internal,
+            rust_struct_ident: qobject_idents.rust_struct.rust.clone(),
+            blocks: GeneratedRustQObjectBlocks {
+                cxx_mod_contents: vec![],
+                cxx_qt_mod_contents: qobject.others.clone(),
+            },
+        };
+
+        // Add the QObject class definitions
+        generated.blocks.append(&mut generate(&qobject_idents)?);
```
Be-ing, line 65:
```suggestion
        generated.blocks.append(&mut generate_qobject_definitions(&qobject_idents)?);
```

A more specific function name would make the comment unnecessary

crates/cxx-qt-gen/src/generator/rust/qobject.rs
```diff
@@ -18,9 +45,140 @@ pub struct GeneratedRustQObject {
...
+                .map(|method| {
+                    let cpp_class_name_rust = &qobject_idents.cpp_class.rust;
+                    syn::parse2(quote! {
+                        impl #cpp_class_name_rust {
+                            #method
+                        }
+                    })
+                })
+                .collect::<Result<Vec<Item>>>()?,
+        );
```
Be-ing, line 96: This is the only code generation directly within this function. The rest of the function calls smaller functions that generate a specific part of the code. How about factoring this out to its own function?

Be-ing:
REWRITE ALL THE CODE

crates/cxx-qt-gen/src/lib.rs
ahayzen-kdab: Done

crates/cxx-qt-gen/src/generator/rust/mod.rs
```diff
@@ -18,5 +25,120 @@ pub struct GeneratedRustBlocks {
...
+            mod ffi {
+                #[cxx_qt::qobject]
+                struct MyObject;
+            }
+        });
+        let parser = Parser::from(module).unwrap();
+
+        let rust = GeneratedRustBlocks::from(&parser).unwrap();
+        assert_eq!(rust.cxx_mod.content.unwrap().1.len(), 0);
+        assert_eq!(rust.cxx_mod_contents.len(), 1);
```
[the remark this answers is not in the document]
ahayzen-kdab: So `cxx_mod` is the module (like `mod ffi { blah }` and `cxx_mod_contents` is a vector of contents that are added to the module later.

Not sure how best to name them as passthrough is something different and not this as cxx_mod_contents are generated.

Note that ItemMod has a field call `content` hence where `contents` has come from.

crates/cxx-qt-gen/src/generator/rust/fragment.rs
ahayzen-kdab: Done

crates/cxx-qt-gen/src/generator/rust/qobject.rs
ahayzen-kdab: Done

crates/cxx-qt-gen/src/generator/rust/qobject.rs
ahayzen-kdab: Done

ahayzen-kdab: Do we want this in the qobject.rs file? or in the invokables.rs file? or in it's own methods.rs file?

crates/cxx-qt-gen/src/generator/rust/fragment.rs
Be-ing: I still think `source` is too vague for the member name and would prefer renaming it to `implementation`. Whether you want to change `cxx` to `cxx_bridge`, I don't really care.

ahayzen-kdab: Oops missed that in the diff, changing....

crates/cxx-qt-gen/src/generator/rust/qobject.rs
Be-ing: A separate file for a single, small function seems a bit overkill, but sure if you want :shrug:

ahayzen-kdab: I think lets just leave it as a separate small function for now, if this file starts becoming larger we can split out.

crates/cxx-qt-gen/src/generator/rust/fragment.rs
ahayzen-kdab: Done

crates/cxx-qt-gen/src/generator/rust/mod.rs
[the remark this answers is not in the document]
ahayzen-kdab: It could inject it in, it does cause some code duplication, and then changes the ordering of components a bit - I can't remember if this was the reason it doesn't work as it then affects the include order.... let me try it again...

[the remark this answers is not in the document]
Be-ing: In `write_rust`, there is:
```rust
    // Inject the CXX blocks
    if let Some((_, items)) = &mut cxx_mod.content {
        items.extend(cxx_mod_contents.into_iter());
    } else {
        cxx_mod.content = Some((syn::token::Brace::default(), cxx_mod_contents));
    }
```

Could that be moved to `GeneratedRustBlocks::from` to get rid of the confusingly similar fields?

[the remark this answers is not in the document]
ahayzen-kdab: That block needs to stay there are the write phase also adds blocks in.

Copying that block into the `GeneratedRustBlocks::from` allows removal of the `cxx_mod_contents`, but this code would still need to be in `write_rust` as well as it also needs to inject blocks into the `cxx_mod`.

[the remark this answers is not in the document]
ahayzen-kdab: This was originally why i split it out into a vec that is passed along, as then there is only one place that needs to inject the blocks into `cxx_mod`. As both the generator and write phase have global blocks to add.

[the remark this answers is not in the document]
Be-ing: okay

Be-ing (approved):

PR #256, opened 2022-09-16
cxx-qt-gen: reduce generation tests as we don't pattern match anymore

ahayzen-kdab:
Now that we don't pattern match on types anymore, we can reduce
the amount of tests we have for the expected output for a given
input. Now we just have tests for properties, invokables, signals,
and general passthrough or naming.

Requires #254


crates/cxx-qt-gen/test_outputs/passthrough_and_naming.rs
```diff
@@ -1,5 +1,64 @@
...
+        };
+    }
+
+    macro macro2() {
+        0
+    }
+
+    mod m {}
+
+    static BIKE: Event = Event::MyEvent;
```
Be-ing, line 47: A strange name to test this, but okay! :bicyclist:

Be-ing (approved):

PR #258, opened 2022-09-19
cxx-qt-gen: change qobject struct to not be optional in Parser

ahayzen-kdab:
The qobject_struct is not optional so change the parser to remove the Option<T>

### DIFF
--- a/crates/cxx-qt-gen/src/parser/qobject.rs
+++ b/crates/cxx-qt-gen/src/parser/qobject.rs
@@ -22,14 +22,11 @@ use syn::{FnArg, Pat, PatIdent, PatType};
 ///
 /// This has initial splitting of [syn::Item]'s into relevant blocks, other phases will
 /// then mutate these [syn::Item]'s for generation purposes.
-#[derive(Default)]
 pub struct ParsedQObject {
     /// The base class of the struct
     pub base_class: Option<String>,
     /// QObject struct that stores the invokables for the QObject
-    //
-    // TODO: this should not be optional
-    pub qobject_struct: Option<ItemStruct>,
+    pub qobject_struct: ItemStruct,
     /// The namespace of the QObject. If one isn't specified for the QObject,
     /// this will be the same as the module
     pub namespace: String,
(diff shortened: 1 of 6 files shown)


crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -147,22 +174,86 @@ impl ParsedQObject {
...
         }
 
-        Ok(())
+        Ok(properties)
     }
 }
 
 #[cfg(test)]
-mod tests {
+pub mod tests {
```
Be-ing, line 182: Why make the test module public?

crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -126,7 +152,8 @@ impl ParsedQObject {
     }
 
     /// Extract all the properties from [syn::Fields] from a [syn::ItemStruct]
-    pub fn parse_struct_fields(&mut self, fields: &mut Fields) -> Result<()> {
+    pub fn parse_struct_fields(fields: &mut Fields) -> Result<Vec<ParsedQProperty>> {
```
Be-ing, line 155: How about making this private and changing test_parse_struct_fields_valid to use the `properties` field of the QObject returned from `from_struct`?

crates/cxx-qt-gen/src/parser/qobject.rs
ahayzen-kdab: Because the `create_parsed_qobject` is reused elsewhere IIRC.

crates/cxx-qt-gen/src/parser/qobject.rs
ahayzen-kdab: This can just be made private anyway and the tests could still reach it. But i think it makes sense to only unit test public methods anyway, so i'll change it to the from_struct...

ahayzen-kdab: Done

crates/cxx-qt-gen/src/parser/qobject.rs
ahayzen-kdab: Eventually i want to split out all the test helpers into a tree somewhere, so that tests can create examples of any component and have the expected output of any component. But for now we have some helpers in the test module for creating a basic version of the component and use it elsewhere.

Be-ing (approved):

Be-ing (approved):

PR #259, opened 2022-09-22
WIP: cxx-qt-gen: swap Rust type names around

ahayzen-kdab:
Requires #254


ahayzen-kdab:
We are likely going to use #276 instead.

PR #261, opened 2022-09-26
cxx-qt-lib: Qt opaque types as trivial

ahayzen-kdab:
Requires #254


examples/qml_features/rust/src/serialisation.rs
```diff
@@ -36,8 +36,8 @@ mod ffi {
     pub struct Serialisation {
         #[qproperty]
         pub number: i32,
-        #[qproperty(cxx_type = "QString")]
-        pub string: UniquePtr<QString>,
+        #[qproperty]
+        pub string: QString,
     }
```
Be-ing, line 41: Stretch goal (after this PR): could we get rid of DataSerde by implementing Serialize/Deserialize on QString? :thinking:

crates/cxx-qt-lib/src/types/qstring.rs
```diff
@@ -3,52 +3,105 @@
...
+    /// Destroys the string.
+    fn drop(&mut self) {
+        ffi::qstring_drop(self)
     }
 }
+
+impl From<&str> for QString {
+    /// Constructs a QString from a Rust string
+    ///
+    /// Note that this performs a UTF conversion
```
Be-ing, line 86: `Note that this converts from UTF-8 to UTF-16`

crates/cxx-qt-lib/src/types/qstring.rs
```diff
@@ -3,52 +3,105 @@
...
+}
+
+impl From<&String> for QString {
+    /// Constructs a QString from a Rust string
+    ///
+    /// Note that this performs a UTF conversion
+    fn from(str: &String) -> Self {
+        ffi::qstring_init_from_rust_string(str)
+    }
+}
```
Be-ing, line 99: Could this redundant impl be avoided by using `impl From<AsRef<str>>`?

crates/cxx-qt-lib-headers/include/qt_types.h
```diff
@@ -31,12 +31,30 @@ struct rust::IsRelocatable<QColor> : std::true_type
...
 static_assert(alignof(QColor) <= alignof(std::size_t[2]),
               "unexpectedly large QColor alignment");
 static_assert(sizeof(QColor) == sizeof(std::size_t[2]),
               "unexpected QColor size");
 
+#if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))
+static_assert(alignof(QString) <= alignof(std::size_t[2]),
+              "unexpectedly large QString alignment");
+static_assert(sizeof(QString) == sizeof(std::size_t[2]),
+              "unexpected QString size");
```
Be-ing, line 63: This is incorrect:
```
warning: In file included from /home/runner/cxx-qt/build/cargo/build/x86_64-unknown-linux-gnu/release/build/cxx-qt-lib-5d4bdfe03d4a5e41/out/cxxbridge/sources/cxx-qt-lib/src/types/qcolor.rs.cc:1:
warning: /home/runner/cxx-qt/build/cargo/build/x86_64-unknown-linux-gnu/release/build/cxx-qt-lib-5d4bdfe03d4a5e41/out/cxx-qt-lib/include/qt_types.h:62:31: error: static assertion failed: unexpected QString size
warning:    62 | static_assert(sizeof(QString) == sizeof(std::size_t[2]),
warning:       |               ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
warning: In file included from /home/runner/cxx-qt/build/cargo/build/x86_64-unknown-linux-gnu/release/build/cxx-qt-lib-5d4bdfe03d4a5e41/out/cxxbridge/sources/cxx-qt-lib/src/types/qdate.rs.cc:1:
warning: /home/runner/cxx-qt/build/cargo/build/x86_64-unknown-linux-gnu/release/build/cxx-qt-lib-5d4bdfe03d4a5e41/out/cxx-qt-lib/include/qt_types.h:62:31: error: static assertion failed: unexpected QString size
warning:    62 | static_assert(sizeof(QString) == sizeof(std::size_t[2]),
warning:       |               ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
warning: In file included from /home/runner/cxx-qt/build/cargo/build/x86_64-unknown-linux-gnu/release/build/cxx-qt-lib-5d4bdfe03d4a5e41/out/cxxbridge/sources/cxx-qt-lib/src/types/qdatetime.rs.cc:1:
warning: /home/runner/cxx-qt/build/cargo/build/x86_64-unknown-linux-gnu/release/build/cxx-qt-lib-5d4bdfe03d4a5e41/out/cxx-qt-lib/include/qt_types.h:62:31: error: static assertion failed: unexpected QString size
warning:    62 | static_assert(sizeof(QString) == sizeof(std::size_t[2]),
warning:       |               ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
```

How did you determine the sizes of these Qt types?

crates/cxx-qt-lib-headers/include/qt_types.h
```diff
@@ -43,6 +44,12 @@ struct rust::IsRelocatable<QString> : std::true_type
 };
 static_assert(QTypeInfo<QString>::isRelocatable);
 
+// Ensure that trivially destructible is correct
+// If this is false then we need to implement Drop rather than derive
+static_assert(std::is_trivially_destructible<QColor>::value);
```
Be-ing, line 49: qIsRelocatable checks for this already: https://codebrowser.dev/qt6/qtbase/src/corelib/global/qtypeinfo.h.html#qIsRelocatable

examples/qml_features/rust/src/serialisation.rs
ahayzen-kdab: Right in theory we can now impl traits directly for an T easier now :-)

crates/cxx-qt-lib-headers/include/qt_types.h
ahayzen-kdab: I haven't yet for Qt 6, it was a blind guess :-D

crates/cxx-qt-lib/src/types/qstring.rs
ahayzen-kdab: Done

crates/cxx-qt-lib-headers/include/qt_types.h
ahayzen-kdab: Yes but isn't this the default inferred one rather than the explicit one that can be overridden by the macros?

Eg `static_assert(qIsRelocatable<QString>());` fails but `static_assert(QTypeInfo<QString>::isRelocatable);` passes

crates/cxx-qt-lib/src/types/qstring.rs
ahayzen-kdab: Ah seems it doesn't work when i compile it further, it then wants AsRef to be dyn and requires the T to be Sized

crates/cxx-qt-lib-headers/include/qt_types.h
ahayzen-kdab: Fixed, Qt 5 has one pointer as a member, Qt 6 has one member which contains two pointers and a size_t

crates/cxx-qt-lib/src/types/qcolor.rs
```diff
@@ -3,69 +3,69 @@
...
 }
 
-/// The QColorCpp class provides colors based on RGB, HSV or CMYK values.
+/// The QColor class provides colors based on RGB, HSV or CMYK values.
 ///
 /// Note that we only expose RGB methods for now.
-///
-/// Note that this is the C++ representation and QColor should be used in Rust.
-pub type QColor = ffi::QColor;
+#[derive(Clone)]
```
ahayzen-kdab, line 41: If the C++ type doesn't have an explicit copy constructor are we OK to use derive(Clone) here ? And are we OK to skip the impl Drop if there is a default deconstructor ?

Be-ing: I guess we'll find out :)

ahayzen-kdab: valgrind is happy so far :-)

crates/cxx-qt-lib-headers/include/qt_types.h
```diff
@@ -24,6 +25,100 @@
...
+
+static_assert(alignof(QUrl) <= alignof(std::size_t),
+              "unexpectedly large QUrl alignment");
+static_assert(sizeof(QUrl) == sizeof(std::size_t), "unexpected QUrl size");
+
+// The layout has changed between Qt 5 and Qt 6
+//
+// Qt5 QString has one member, which contains three uints (but they are
+// optimised to a size of 8) and a union Qt6 QString has one member, which
+// contains three pointers and a union (pointer largest)
```
Be-ing, line 109:
```suggestion
// Qt5 QVariant has one member, which contains three uints (but they are
// optimised to a size of 8) and a union. Qt6 QVariant has one member, which
// contains three pointers and a union (pointer largest)
```

Be-ing:
Does Qt guarantee that the sizes of these types will remain the same for all Qt versions within the same major version?

crates/cxx-qt-lib-headers/include/qt_types.h
```diff
@@ -24,6 +25,100 @@
...
+static_assert(QTypeInfo<QUrl>::isRelocatable);
+
+template<>
+struct rust::IsRelocatable<QVariant> : std::true_type
+{
+};
+static_assert(QTypeInfo<QVariant>::isRelocatable);
+
+// Ensure that trivially copyable is correct
+// If this is false then we need to implement Clone rather than derive
```
Be-ing, line 60:
```suggestion
// If this is false then we need to manually implement Clone rather than derive
```

crates/cxx-qt-lib-headers/include/qt_types.h
```diff
@@ -24,6 +25,100 @@
...
+static_assert(sizeof(QString) == sizeof(std::size_t[3]),
+              "unexpected QString size");
+#else
+static_assert(alignof(QString) <= alignof(std::size_t),
+              "unexpectedly large QString alignment");
+static_assert(sizeof(QString) == sizeof(std::size_t),
+              "unexpected QString size");
+#endif
+
+static_assert(alignof(QUrl) <= alignof(std::size_t),
```
Be-ing, line 101: It would be helpful to have comments documenting where these values came from for each type like you've written for QString. Even better would be links to Qt implementations on https://codebrowser.dev/

ahayzen-kdab:
> Does Qt guarantee that the sizes of these types will remain the same for all Qt versions within the same major version?

I was hoping that the Qt binary compatibility would help for this case ? As if the size of the structs/classes changed in a minor or patch release wouldn't this break binary compatibility ?

"Major releases may break backwards binary and source compatibility, although source compatibility may be maintained.

Minor releases are backwards binary and source compatible.

Patch releases are both backwards and forwards binary and source compatible. "

https://wiki.qt.io/Qt-Version-Compatibility
https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B#The_Do.27s_and_Don.27ts

crates/cxx-qt-lib-headers/include/qt_types.h
ahayzen-kdab: Done

crates/cxx-qt-lib-headers/include/qt_types.h
ahayzen-kdab: Done

crates/cxx-qt-lib-headers/include/qt_types.h
LeonMatthesKDAB: I think @ahayzen-kdab is correct here, qIsRelocatable only checks for this if it isn't overwritten elsewhere, which is still possible.

crates/cxx-qt-lib/src/types/qcolor.rs
LeonMatthesKDAB: https://en.cppreference.com/w/cpp/types/is_trivially_copyable
^^^^ See the notes section ^^^

Anything that is trivially copyable should be okay to derive(Clone), as Rust just does a memory-copy in the case of `[usize;X]`.

On that note, why are we currently checking for `is_trivially_copy_constructible` and `is_trivially_copy_assignable`, if we could just check for `is_trivially_copyable`?

crates/cxx-qt-lib/src/types/qdatetime.rs
```diff
@@ -11,73 +13,93 @@ mod ffi {
...
+        #[doc(hidden)]
+        #[rust_name = "qdatetime_init_from_qdatetime"]
+        fn qdatetimeInitFromQDateTime(datetime: &QDateTime) -> QDateTime;
 
         // Note that Qt 5 takes const-ref and Qt 6 takes by-value
         // for QDateTime::setDate and QDateTime::setTime
         //
-        // So we need our own methods otherwise CXX can't match the method types
-        #[namespace = "rust::cxxqtlib1"]
+        // We want by-value so for Qt 5 we create a proxy
```
LeonMatthesKDAB, line 43:
```suggestion
        // We want by-value, as that is Rust-idiomatic, so for Qt 5 we create a proxy
```

crates/cxx-qt-lib/src/types/qstring.rs
```diff
@@ -3,52 +3,111 @@
...
+        ffi::qstring_init_from_qstring(self)
     }
+}
 
-    pub fn from_str(str: &str) -> cxx::UniquePtr<Self> {
-        ffi::qstring_init_from_rust_string(str)
+impl Default for QString {
+    /// Constructs an invalid color. An invalid color is a color that is not properly set up for the underlying window system.
+    ///
+    /// The alpha value of an invalid color is unspecified.
```
LeonMatthesKDAB, line 67:
```suggestion
    /// Constructs a null QString. Null QStrings are also empty.
```

crates/cxx-qt-lib/src/types/qstring.rs
```diff
@@ -3,52 +3,111 @@
...
+    fn default() -> Self {
+        ffi::qstring_init_default()
     }
 }
 
 impl std::fmt::Display for QString {
+    /// Convert the QString to a Rust string
+    ///
+    /// Note that this converts from UTF-16 to UTF-8
```
LeonMatthesKDAB, line 76: Why was the QString -> String conversion done in `impl Display for QString` again?
Shouldn't we rather add a `Into<String>` implementation for QString and use that here?

examples/qml_features/rust/src/lib.rs
```diff
@@ -24,15 +24,15 @@ mod ffi {
     pub struct MyObject {
         #[qproperty]
         number: i32,
-        #[qproperty(cxx_type = "QString")]
-        string: UniquePtr<QString>,
+        #[qproperty]
+        string: QString,
```
LeonMatthesKDAB, line 28: :partying_face:

LeonMatthesKDAB (commented):
I found a few more minor things to fix and still have a few questions.
Otherwise, great PR! The API is getting much better every day :partying_face:

crates/cxx-qt-lib/src/types/qdatetime.rs
ahayzen-kdab: Done

crates/cxx-qt-lib/src/types/qstring.rs
ahayzen-kdab: Haha, can you tell where i copied this from? :') Done

crates/cxx-qt-lib/src/types/qstring.rs
ahayzen-kdab: This allows for a `to_string()` method on the `QString` which feels quite natural? Eg `qstring.to_string()`

ahayzen-kdab: Oh you mean in the impl why add a Into<String> and reuse that here... good idea i'll add it :-)

ahayzen-kdab: Done

LeonMatthesKDAB (approved):
LGTM

PR #264, opened 2022-09-28
cargo_without_cmake: set linker from build.rs instead of CMake

Be-ing:
This allows running `cargo run -p qml-minimal-no-cmake` without the user needing to edit ~/.cargo/config.toml or set the RUSTFLAGS environment variable.


examples/cargo_without_cmake/build.rs
```diff
@@ -23,5 +29,43 @@ fn main() {
...
+
+            // mold is faster, but specifing mold with -fuse-ld requires GCC >= 12 or Clang.
+            // Unfortunately cargo does not provide a means to set the linker driver via build scripts,
+            // so linking would fail trying to use -fuse-ld=mold with GCC < 12 even if clang is installed.
+            // So, prefer lld to mold for robustness on the widest range of systems.
+            // mold can still be used by manually specifying it in ~/.cargo/config.toml or the RUSTFLAGS environment variable.
+            if command_version_output("lld").is_ok() {
+                println!("cargo:rustc-link-arg=-fuse-ld=lld");
+                return;
+            }
```
ahayzen-kdab, line 61: any reason for not doing and if else if else ?

```
if lld {
...
} else if mold {
...
} else {
...
}
```

Be-ing: Nope, that's simpler :)

examples/cargo_without_cmake/build.rs
```diff
@@ -23,5 +29,40 @@ fn main() {
...
         .build();
+
+    // This fails to link with GNU ld, which is the default on most Linux distributions, so use lld or mold instead.
+    #[cfg(unix)]
+    {
+        let flags = env::var("CARGO_ENCODED_RUSTFLAGS").unwrap();
+        // Don't override custom flags
+        if !flags.contains("-fuse-ld") {
+            let linker_is_gnu = str::from_utf8(
+                &command_version_output("ld")
```
ahayzen-kdab, line 40: as discussed lets put a comment here about how on macOS `ld` is actually `ldd` hence why this check is looking at the version output. As this wasn't obvious to me why this check was doing this.

ahayzen-kdab (approved):
LGTM :-)

PR #265, opened 2022-09-28
qt-build: use println!("cargo:warning") instead of eprintln!

Be-ing:
so the warnings show up with `cargo build -vv`
https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargo-warning


ahayzen-kdab (approved):
LGTM :-)

PR #266, opened 2022-09-28
Assert qt type layouts

LeonMatthesKDAB:
Also splits up the qt_types.h / qt_types.cpp into individual files for the different types.

### DIFF
--- a/crates/cxx-qt-lib-headers/include/qcolor.h
+++ b/crates/cxx-qt-lib-headers/include/qcolor.h
@@ -11,11 +11,14 @@
 
 #include "rust/cxx.h"
 
+// QColor still had copy & move constructors in Qt 5 but they were basically
+// trivial.
+#if (QT_VERSION < QT_VERSION_CHECK(6,0,0))
 template<>
 struct rust::IsRelocatable<QColor> : std::true_type
 {
 };
-static_assert(QTypeInfo<QColor>::isRelocatable);
+#endif
 
 namespace rust {
 namespace cxxqtlib1 {
--- a/crates/cxx-qt-lib-headers/include/qcolor.h
+++ b/crates/cxx-qt-lib-headers/include/qcolor.h
@@ -13,7 +13,7 @@
 
 // QColor still had copy & move constructors in Qt 5 but they were basically
 // trivial.
-#if (QT_VERSION < QT_VERSION_CHECK(6,0,0))
+#if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
 template<>
 struct rust::IsRelocatable<QColor> : std::true_type
 {
--- a/crates/cxx-qt-lib/src/types/assertion_utils.h
+++ b/crates/cxx-qt-lib/src/types/assertion_utils.h
@@ -1,3 +1,10 @@
+// clang-format off
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// clang-format on
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+// SPDX-FileContributor: Leon Matthes <leon.matthes@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
 #pragma once
 
 #define assert_alignment_and_size(TYPE, ALIGNMENT, SIZE)                       \
--- a/crates/cxx-qt-lib/src/types/assertion_utils.h
+++ b/crates/cxx-qt-lib/src/types/assertion_utils.h
@@ -1,0 +1,5 @@
+#pragma once
+
+#define assert_alignment_and_size(TYPE,ALIGNMENT,SIZE) \
+  static_assert(alignof(TYPE) <= (ALIGNMENT), "unexpectedly large " #TYPE " alignment!"); \
+  static_assert(sizeof(TYPE) == (SIZE), "unexpected " #TYPE " size!");
--- a/crates/cxx-qt-lib/src/types/assertion_utils.h
+++ b/crates/cxx-qt-lib/src/types/assertion_utils.h
@@ -1,5 +1,6 @@
 #pragma once
 
-#define assert_alignment_and_size(TYPE,ALIGNMENT,SIZE) \
-  static_assert(alignof(TYPE) <= (ALIGNMENT), "unexpectedly large " #TYPE " alignment!"); \
+#define assert_alignment_and_size(TYPE, ALIGNMENT, SIZE)                       \
+  static_assert(alignof(TYPE) <= (ALIGNMENT),                                  \
+                "unexpectedly large " #TYPE " alignment!");                    \
   static_assert(sizeof(TYPE) == (SIZE), "unexpected " #TYPE " size!");
--- a/crates/cxx-qt-lib/src/types/qdate.cpp
+++ b/crates/cxx-qt-lib/src/types/qdate.cpp
@@ -12,15 +12,13 @@
 #include <cstdint>
 
 // QDate has a single 64-Bit integer member
-// https://codebrowser.dev/qt5/qtbase/src/corelib/time/qdatetime.h.html#QDate::jd
+// https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/time/qdatetime.h?h=v5.15.6-lts-lgpl#n176
 //
-// https://codebrowser.dev/qt6/qtbase/src/corelib/time/qdatetime.h.html#QDate::jd
+// https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/time/qdatetime.h?h=v6.2.4#n147
 assert_alignment_and_size(QDate, alignof(std::int64_t), sizeof(std::int64_t));
 
-static_assert(std::is_trivially_copy_assignable<QDate>::value);
-static_assert(std::is_trivially_copy_constructible<QDate>::value);
-
-static_assert(std::is_trivially_destructible<QDate>::value);
+static_assert(std::is_trivially_copyable<QDate>::value,
+              "QDate must be trivially copyable!");
 
 namespace rust {
 namespace cxxqtlib1 {
--- a/crates/cxx-qt-lib/src/types/qdate.cpp
+++ b/crates/cxx-qt-lib/src/types/qdate.cpp
@@ -7,6 +7,21 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 #include "cxx-qt-lib/include/qdate.h"
 
+#include "assertion_utils.h"
+
+#include <cstdint>
+
+// QDate has a single 64-Bit integer member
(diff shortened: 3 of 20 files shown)


crates/cxx-qt-lib-headers/include/qcolor.h
```diff
@@ -0,0 +1,34 @@
...
+
+#include <memory>
+
+#include "qt_types.h"
+
+template<>
+struct rust::IsRelocatable<QColor> : std::true_type
+{
+};
+static_assert(QTypeInfo<QColor>::isRelocatable);
```
ahayzen-kdab, line 20: does the assert need to be in the header? or could this also be in the source? if we are trying to avoid multiple static_asserts at compile time?

Then just leave the template here ?

LeonMatthesKDAB: We could also just remove this static_assert, as it's kind of pointless to check something we're literally defining to be true in the line above :thinking:

ahayzen-kdab: The point of the check is to check that Qt has defined that it is relocatable. The line above is telling Rust it is relocatable. So we still need the static_assert.

As what happens if suddenly QColor is not relocatable and we marked it wrong?

examples/qml_features/rust/src/mock_qt_types.rs
```diff
@@ -10,18 +10,32 @@ mod ffi {
 
     unsafe extern "C++" {
         include!("cxx-qt-lib/include/qt_types.h");
+        include!("cxx-qt-lib/include/qcolor.h");
```
Be-ing, line 13: Hmm, do we want to make downstreams include a header for each type? We could #include each of these in qt_types.h.

ahayzen-kdab: Eventually we'll have a macro to include them all, eg `cxx_qt_lib::prelude!()` or something.

And qt_types will go away with this approach. Then we end up with smaller headers for each type you use. Which may also help later if we split up QtCore and QtGui support in cxx-qt-lib etc.

crates/cxx-qt-lib-headers/include/qcolor.h
LeonMatthesKDAB: Ooops, you're right, I misread, will move these into the CPP files then.

examples/qml_features/rust/src/mock_qt_types.rs
LeonMatthesKDAB: Also the advantage of using a macro is that that can possibly also take care of all the
`type QType = cxx_qt_lib::types::QType` boilerplate.

ahayzen-kdab: Yup, i'm hoping to remove the whole blocks. We could also later have like `cxx_qt_lib::qtcore::prelude!()` if we split out QtCore/QtGui to be separate in cxx-qt-lib and have separate features.

The only "fun" part is whether CXX/our macro actually causes the macro expansion to happen early enough? As if needs to happen after our macro but before the CXX one. Worst case we might need the expansion to actually happen via cxx-qt-gen ? Eg if it spots a macro called cxx_qt_lib::prelude!() then replace it for this block.

crates/cxx-qt-lib/src/types/assertion_utils.h
```diff
@@ -0,0 +1,5 @@
+#pragma once
```
ahayzen-kdab, line 1: Missing a license header :-)

crates/cxx-qt-lib/src/types/assertion_utils.h
```diff
@@ -0,0 +1,5 @@
...
+
+#define assert_alignment_and_size(TYPE,ALIGNMENT,SIZE) \
```
ahayzen-kdab, line 3: clang-format needs to run on this file

crates/cxx-qt-lib/src/types/qdate.cpp
```diff
@@ -7,6 +7,21 @@
...
+
+#include <cstdint>
+
+// QDate has a single 64-Bit integer member
+// https://codebrowser.dev/qt5/qtbase/src/corelib/time/qdatetime.h.html#QDate::jd
+//
+// https://codebrowser.dev/qt6/qtbase/src/corelib/time/qdatetime.h.html#QDate::jd
+assert_alignment_and_size(QDate, alignof(std::int64_t), sizeof(std::int64_t))
+
+  static_assert(std::is_trivially_copy_assignable<QDate>::value);
```
ahayzen-kdab, line 20: why is this indented?

LeonMatthesKDAB: ah, forgot the semicolon at the end of the previous line.
That confuses Clang-format.

Fixed now

crates/cxx-qt-lib/src/types/qrectf.cpp
```diff
@@ -7,6 +7,14 @@
...
 #include "cxx-qt-lib/include/qrectf.h"
 
+#include "assertion_utils.h"
+
+// QRectF has 4 double members
+// https://codebrowser.dev/qt5/qtbase/src/corelib/tools/qrect.h.html#QRectF::xp
+//
+// https://codebrowser.dev/qt6/qtbase/src/corelib/tools/qrect.h.html#QRectF::xp
+assert_alignment_and_size(QRectF, alignof(double), sizeof(double[4]));
+
```
ahayzen-kdab, line 17: why do we not check that QRectF is_trivially_copyable like QRect ?

examples/qml_features/rust/src/mock_qt_types.rs
LeonMatthesKDAB: I fear the same, I'm pretty sure the cxx_qt::bridge macro runs first, unfortunately :thinking:

crates/cxx-qt-lib/src/types/qrectf.cpp
LeonMatthesKDAB: Fixed

crates/cxx-qt-lib/src/types/qpoint.cpp
```diff
@@ -7,6 +7,21 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 #include "cxx-qt-lib/include/qpoint.h"
 
+#include "assertion_utils.h"
+
+#include <cstdint>
+
+// QPoint has "int" members - xp and yp
+// Rust represents these as 32-bit integer types.
+// https://codebrowser.dev/qt5/qtbase/src/corelib/tools/qpoint.h.html#QPoint::xp
```
ahayzen-kdab, line 16: i was using `code.qt.io` rather than `codebrowser.dev` as the canonical source (as codebrowser.dev just moved and we don't know if it'll move/change again), and pinned to either `v5.15.6-lts-lgpl` or `v6.2.4` so that the code doesn't move/change.

crates/cxx-qt-lib/src/types/assertion_utils.h
LeonMatthesKDAB: Fixed

crates/cxx-qt-lib/src/types/qpoint.cpp
LeonMatthesKDAB: The new links do have the advantage of not pinning to a specific line though.
So these links should be stable :thinking:

ahayzen-kdab: But the tag will never change, v6.2.4 will always have a member called xp. Whereas your link will fail if they change QPoint to be `xpos` as it's not pinned to a specific Qt release

LeonMatthesKDAB: Updated them to perma-links

examples/qml_features/rust/src/mock_qt_types.rs
ahayzen-kdab: Something to battle latter, but i think the worst case of using cxx-qt-gen to do the expansion isn't terrible, We can have a tokenstream in cxx_qt_lib_headers or something that it can use, so that cxx-qt-gen doesn't depend directly on cxx-qt-lib...

crates/cxx-qt-lib-headers/include/qcolor.h
```diff
@@ -11,11 +11,14 @@
 
 #include "rust/cxx.h"
 
+// QColor still had copy & move constructors in Qt 5 but they were basically
+// trivial.
+#if (QT_VERSION < QT_VERSION_CHECK(6,0,0))
```
ahayzen-kdab, line 16: clang-format is complaining about this line.

crates/cxx-qt-lib/src/types/qdate.cpp
```diff
@@ -7,6 +7,21 @@
...
+// QDate has a single 64-Bit integer member
+// https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/time/qdatetime.h?h=v5.15.6-lts-lgpl#n176
+//
+// https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/time/qdatetime.h?h=v6.2.4#n147
+assert_alignment_and_size(QDate, alignof(std::int64_t), sizeof(std::int64_t));
+
+static_assert(std::is_trivially_copy_assignable<QDate>::value);
+static_assert(std::is_trivially_copy_constructible<QDate>::value);
+
+static_assert(std::is_trivially_destructible<QDate>::value);
```
ahayzen-kdab, line 23: should we be checking for is_trivially_destructible in other trivial types? eg QPoint etc ?

LeonMatthesKDAB: is_trivially_copyable already asserts this.
See: https://en.cppreference.com/w/cpp/language/classes#Trivially_copyable_class

ahayzen-kdab: Aha

ahayzen-kdab: So then can QDate just be `is_trivially_copyable` ?

crates/cxx-qt-lib/src/types/qdate.cpp
```diff
@@ -7,6 +7,21 @@
...
+
+#include <cstdint>
+
+// QDate has a single 64-Bit integer member
+// https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/time/qdatetime.h?h=v5.15.6-lts-lgpl#n176
+//
+// https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/time/qdatetime.h?h=v6.2.4#n147
+assert_alignment_and_size(QDate, alignof(std::int64_t), sizeof(std::int64_t));
+
+static_assert(std::is_trivially_copy_assignable<QDate>::value);
```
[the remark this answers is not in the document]
LeonMatthesKDAB: Fixed.

ahayzen-kdab (approved):
LGTM :-)

ahayzen-kdab (approved):
LGTM :-)

PR #267, opened 2022-09-28
cxx-qt-gen: add prop_mut and rename get_prop to prop

ahayzen-kdab:


crates/cxx-qt-gen/test_outputs/properties.rs
```diff
@@ -113,17 +113,23 @@ mod cxx_qt_ffi {
...
 
     impl MyObjectQt {
-        pub fn get_primitive(&self) -> &i32 {
+        pub fn primitive(&self) -> &i32 {
             &self.rust().primitive
         }
     }
 
+    impl MyObjectQt {
+        pub unsafe fn primitive_mut<'a>(mut self: Pin<&'a mut Self>) -> Pin<&'a mut i32> {
```
ahayzen-kdab, line 128: Should we return `&mut T` or `Pin<&mut T>` here?

LeonMatthesKDAB: For Rust-native types, probably a `&mut T`.
If you want to pin the type yourself, you can use a `Pin<Box<T>>` or similar

ahayzen-kdab: OK, done :-)

crates/cxx-qt-gen/src/generator/rust/property/getter.rs
```diff
@@ -43,6 +44,13 @@ pub fn generate(
                     }
                 }
             },
+            quote! {
+                impl #cpp_class_name_rust {
+                    pub unsafe fn #getter_mutable_rust<'a>(mut self: Pin<&'a mut Self>) -> &'a mut #ty {
+                        unsafe { &mut self.rust_mut().get_unchecked_mut().#ident }
```
Be-ing, line 50: Do you need the `unsafe` block inside the function when the whole function is marked `unsafe`?

ahayzen-kdab: Seems you can, i thought it'd just mark the function unsafe to call, but seems also the contents is unsafe.

Be-ing (approved):

PR #268, opened 2022-09-28
cxx-qt-gen: add prop/prop_mut/set_prop for Rust fields as well

ahayzen-kdab:
Note that the prop_mut is not unsafe as there is no signal that should be emitted.

Requires #267

### DIFF
--- a/examples/demo_threading/rust/src/lib.rs
+++ b/examples/demo_threading/rust/src/lib.rs
@@ -299,28 +299,18 @@ mod ffi {
 
                             qt_thread
                                 .queue(|mut qobject_energy_usage| {
-                                    // TODO: for now we use the unsafe rust_mut() API
-                                    // later there will be getters and setters for the properties
-                                    unsafe {
-                                        // Process the new data from the background thread
-                                        if let Some(data) = qobject_energy_usage
+                                    // Process the new data from the background thread
+                                    if let Some(data) =
+                                        qobject_energy_usage.as_mut().qt_rx_mut().try_iter().last()
+                                    {
+                                        // Here we have constructed a new Data struct so can consume its values.
+                                        // For other uses, we could have passed an Enum across the channel
+                                        // and then process the required action here
+                                        qobject_energy_usage
                                             .as_mut()
-                                            .rust_mut()
-                                            .qt_rx
-                                            .try_iter()
-                                            .last()
-                                        {
-                                            // Here we have constructed a new Data struct so can consume it's values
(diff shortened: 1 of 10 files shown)


examples/demo_threading/rust/src/lib.rs
```diff
@@ -299,28 +299,18 @@ mod ffi {
...
-                                    // TODO: for now we use the unsafe rust_mut() API
-                                    // later there will be getters and setters for the properties
-                                    unsafe {
-                                        // Process the new data from the background thread
-                                        if let Some(data) = qobject_energy_usage
+                                    // Process the new data from the background thread
+                                    if let Some(data) =
+                                        qobject_energy_usage.as_mut().qt_rx_mut().try_iter().last()
+                                    {
+                                        // Here we have constructed a new Data struct so can consume it's values
```
Be-ing, line 306:
```suggestion
                                        // Here we have constructed a new Data struct so can consume its values
```

examples/demo_threading/rust/src/lib.rs
```diff
@@ -299,28 +299,18 @@ mod ffi {
...
+                                        // for other uses we could have passed an Enum across the channel
```
Be-ing, line 307:
```suggestion
                                        // For other uses, we could have passed an Enum across the channel
```

examples/demo_threading/rust/src/lib.rs
ahayzen-kdab: Done

examples/demo_threading/rust/src/lib.rs
ahayzen-kdab: Done

examples/demo_threading/rust/src/lib.rs
```diff
@@ -299,28 +299,18 @@ mod ffi {
...
-                                    // later there will be getters and setters for the properties
-                                    unsafe {
-                                        // Process the new data from the background thread
-                                        if let Some(data) = qobject_energy_usage
+                                    // Process the new data from the background thread
+                                    if let Some(data) =
+                                        qobject_energy_usage.as_mut().qt_rx_mut().try_iter().last()
+                                    {
+                                        // Here we have constructed a new Data struct so can consume its values.
+                                        // Foror other uses, we could have passed an Enum across the channel
```
Be-ing, line 307:
```suggestion
                                        // For other uses, we could have passed an Enum across the channel
```

still typos :)

Be-ing (approved):

PR #269, opened 2022-09-28
Split up C++ code for each wrapped Qt class

LeonMatthesKDAB:


ahayzen-kdab (dismissed):
LGTM :-)

ahayzen-kdab:
So looks like CI is failing due

- Linux - just the book would need a rerun
- macOS / Windows - likely missing an include as they are complaining about std::size_t missing

ahayzen-kdab (approved):
LGTM :-)

PR #271, opened 2022-09-28
qml_minimal: explicitly define colors

Be-ing:
before with Qt6:
image

after:
after


examples/qml_minimal/qml/main.qml
```diff
@@ -18,6 +18,7 @@ Window {
     title: qsTr("Hello World")
     visible: true
     width: 640
+    color: "white"
```
ahayzen-kdab, line 21: As discussed, this only happens when building with cargo. Is some env var missing or it not linking to something?

PR #274, opened 2022-09-29
Change doctests to qtest

ahayzen-kdab:
This is part of a larger effort to refactor the tests and examples.


tests/basic_cxx_only/CMakeLists.txt
```diff
@@ -27,9 +34,12 @@ target_link_libraries(${CRATE} INTERFACE
     Qt${QT_VERSION_MAJOR}::Gui
 )
 
-add_executable(${APP_NAME} cpp/main.cpp)
+add_executable(${APP_NAME}
+    cpp/cxx_test.h
```
Be-ing, line 38: You don't need to list headers as sources, but it doesn't hurt I guess.

tests/basic_cxx_only/cpp/main.cpp
```diff
@@ -18,9 +17,19 @@ get_cpp_number()
...
-  CHECK(get_numbers_sum() == 202);
-}
+  Q_OBJECT
+
+private Q_SLOTS:
+  // Clean cxx allows basic interaction between C++ and Rust
+  void test_cxx_interaction()
+  {
+    QCOMPARE(get_numbers_sum(), 102);
+    hidden_num = 200;
```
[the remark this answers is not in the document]
ahayzen-kdab: So this is an old test that i've just ported. What it does is ....
* Calls `get_numbers_sum()` (Rust)
  * Which calls `get_cpp_number()` (C++) and returns + 2
    * `get_cpp_number()` returns `hidden_num`
  * So we should get 102 as `hidden_num` starts as 100
* Changes `hidden_num` to 200
* Calls `get_numbers_sum()` (Rust)
  * Which calls `get_cpp_number()` (C++) and returns + 2
    * `get_cpp_number()` returns `hidden_num`
  * So we should get 202 as `hidden_num` is now 200

It looks like `hidden_sum` is unused, but it's checking that the basic CXX bridge works as we expect back and forth across the bridge.

Eventually I want to merge this small test with basic_cxx_qt, as it doesn't warrant it's own project folder anymore.

But the point of this PR was to port the tests to QTest, so that change can be in later changes.

tests/basic_cxx_qt/cpp/main.cpp
```diff
@@ -5,225 +5,222 @@
...
+    QCOMPARE(stringSpy.count(), 0);
+    data.setString(QStringLiteral("Hello"));
+    QCOMPARE(stringSpy.count(), 0);
+    QCoreApplication::processEvents();
+    QCOMPARE(stringSpy.count(), 1);
+    QCOMPARE(data.getString(), QStringLiteral("Hello"));
+
+    // Check that initial value of the deserialised data
+    QCOMPARE(data.asJsonStr(),
+             QStringLiteral("{\"number\":16,\"string\":\"Hello\"}"));
```
Be-ing, line 99: For bonus points you could check if Qt's JSON parser parses this information correctly.

tests/basic_cxx_only/cpp/main.cpp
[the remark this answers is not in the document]
Be-ing: okay

tests/basic_cxx_qt/cpp/main.cpp
```diff
@@ -5,225 +5,222 @@
...
+    types.setUint32(4);
+    QCOMPARE(uint32Spy.count(), 0);
+    QCoreApplication::processEvents();
+    QCOMPARE(uint32Spy.count(), 1);
+    QCOMPARE(types.getUint32(), 4);
+  }
+};
+
+QTEST_MAIN(CxxQtTest)
+#include "main.moc"
```
Be-ing, line 226: Is this line required?

ahayzen-kdab: Yes otherwise automoc fails. It's the common way of doing it for source only files like a main.cpp.

tests/basic_cxx_qt/cpp/main.cpp
ahayzen-kdab: Right, in the future we could make these tests more meaningful. I was thinking about sharing the Rust code here with qml_features. Then have a QML page that does something with that data, then have C++ tests here that can use the same Rust interface etc.

tests/basic_cxx_qt/cpp/main.cpp
ahayzen-kdab: Eg this is the output
```
[build] "SRC:/tests/basic_cxx_qt/cpp/main.cpp"
[build] contains a "Q_OBJECT" macro, but does not include "main.moc"!
[build] Consider to
[build]   - add #include "main.moc"
[build]   - enable SKIP_AUTOMOC for this file
```

Then if you remove the `Q_OBJECT` macro, none of the tests run as QTest expects private Q_SLOTS to be tests. But as it's not then a Q_OBJECT it doesn't know there are slots.

ahayzen-kdab: And this is actually what they do in the QTest tutorial as well :-) https://doc.qt.io/qt-6/qttestlib-tutorial1-example.html#preparing-the-stand-alone-executable

Be-ing (approved):

PR #275, opened 2022-09-30
cxx-qt-gen: use recursive mutex for rust object

ahayzen-kdab:
This then allows for methods to trigger signals which are linked on the same thread to invokables.
eg beginInsertRows triggers rowCount().

This is safe because to emit a signal you need a mutable self, which then allows other methods to be called with similar rules in Rust around ownership. And we still protect against other threads from calling the RustObj at the same time.

We also still have queued signals as the "safe" route for now, to ensure that slots are called "later". Obvious C++ code can be written to call on a different thread and cause a deadlock from a method in the base class being called.


LeonMatthesKDAB (approved):
LGTM

PR #276, opened 2022-09-30
cxx-qt-gen: use impl qobject::T to define the C++ context

ahayzen-kdab:
This then creates a natural way to add methods to the C++ context.

It also means later with nested objects they can be referred to by crates::module::qobject::MyObject.

And any Rust traits or methods are natural as they are impl A for T, with traits for the C++ object being natural with impl A for qobject::T

### DIFF
--- a/book/src/qobject/qobject_struct.md
+++ b/book/src/qobject/qobject_struct.md
@@ -21,7 +21,7 @@ The `#[cxx_qt::qobject]` marked struct allows you to define the following items
 
 ## Invokables
 
-A `impl cxx_qt::QObject<T>` is used to define invokables, the `impl cxx_qt::QObject<T>` defines that the methods are implemented onto the C++ QObject `T`.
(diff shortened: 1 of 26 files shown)


Be-ing (approved):
I liked the `impl cxx_qt::QObject<T>` but I understand the issues with it. `impl qobject::T` feels backwards; I would prefer `impl T::qobject`. Maybe there would be a way to do that with associated types, but I also understand the issues that would bring up as discussed in https://github.com/KDAB/cxx-qt/discussions/237.

Be-ing (dismissed):

Be-ing:
As a follow-up, it would be good to have a test using a QObject type from one cxx-qt module in another cxx-qt module.

ahayzen-kdab:
> As a follow-up, it would be good to have a test using a QObject type from one cxx-qt module in another cxx-qt module.

This is something to figure out in 0.5 #299 as we might need more API we'll see (eg types to define ownership). But in theory the API is now close.

Be-ing (approved):

PR #282, opened 2022-10-03
cxx-qt-gen: add C++ specifiers support for final, virtual, override

ahayzen-kdab:


crates/cxx-qt-gen/src/syntax/attribute.rs
```diff
@@ -30,22 +31,26 @@ impl Parse for AttributeList {
 }
 
 /// Representation of a key and value in an attribute, eg attribute(key = value)
```
LeonMatthesKDAB, line 33:
```suggestion
/// Representation of a key and an optional value in an attribute, eg `attribute(key = value)` or `attribute(key)`
```

LeonMatthesKDAB (changes requested):
I'd like to see this tested as a complete chain from test_inputs/invokables.rs to test_outputs/invokables.rs

crates/cxx-qt-gen/src/syntax/attribute.rs
ahayzen-kdab: Done

LeonMatthesKDAB (approved):
LGTM

PR #283, opened 2022-10-03
cxx-qt-gen: use the module name or attribute for the CXX file names

ahayzen-kdab:
So that we can support multiple QObjects later in one module block, use the module name or cxx_stem attribute as the output file name.

Later if the Rust API is ever stable we could use the inspection APIs on Span to read the filename of the macro. This would mean we would be the same as CXX but this requires nightly.


LeonMatthesKDAB:
How would you feel about renaming `cxx_stem` to `cxx_file_stem` or even `cxx_filename_stem`?
This might be a non-native-speaker problem, but I didn't understand right away why that was the name of the attribute and how that refers to the filename.

ahayzen-kdab:
> How would you feel about renaming `cxx_stem` to `cxx_file_stem` or even `cxx_filename_stem`? This might be a non-native-speaker problem, but I didn't understand right away why that was the name of the attribute and how that refers to the filename.

We could go for `cxx_file_stem`

ahayzen-kdab:
Done :-)

LeonMatthesKDAB (dismissed):
LGTM

LeonMatthesKDAB (approved):

PR #312, opened 2022-10-06
cxx-qt-gen: strip lifetimes from CXX definitions for signals

ahayzen-kdab:
As the lifetime will be the lifetime of the enum. Then we don't need this in the CXX method as it can be inferred.


LeonMatthesKDAB (approved):
LGTM

PR #314, opened 2022-10-14
rename qt-build crate to qt-build-utils

Be-ing:
qt-build could be misinterpreted as something to build Qt


Cargo.toml
```diff
@@ -10,6 +10,7 @@ members = [
     "crates/cxx-qt-gen",
     "crates/cxx-qt-lib",
     "crates/cxx-qt-lib-headers",
+    "crates/qt-build-utils",
```
Be-ing, line 13: qt-build wasn't in the workspace before :sweat_smile:

LeonMatthesKDAB:
Good idea, I've even thought that myself when I first read the crate name.

Do we maybe even want to go with cxxqt-build-utils?
To be consistent with our other crates.
I know you can technically use this crate without the rest of cxx-qt, but do we want to actively advertise this?

ahayzen-kdab:
> Good idea, I've even thought that myself when I first read the crate name.
> 
> Do we maybe even want to go with cxxqt-build-utils? To be consistent with our other crates. I know you can technically use this crate without the rest of cxx-qt, but do we want to actively advertise this?

So far we think so, later this might even move to it's own repo, as it's useful for just generally build / using Qt with Rust not just for CXX-Qt.

LeonMatthesKDAB:
> > Good idea, I've even thought that myself when I first read the crate name.
> > Do we maybe even want to go with cxxqt-build-utils? To be consistent with our other crates. I know you can technically use this crate without the rest of cxx-qt, but do we want to actively advertise this?
> 
> So far we think so, later this might even move to it's own repo, as it's useful for just generally build / using Qt with Rust not just for CXX-Qt.

In that case I'll approve.

LeonMatthesKDAB (approved):
LGTM

Be-ing:
Yes after we release it I want to try refactoring qttypes and Whisperfish's build scripts to use it.

Issue #316, opened 2022-10-14
Use same types as CXX for C++ pattern matching


ahayzen-kdab:
In `generator/cpp/types.` there are methods which copy CXX to match `CxxString` -> `::std::string` etc. 

#328 covers ensuring that we have all the types that CXX supports.

This issue is for tracking if we can remove the difference between CXX-Qt and CXX types, eg can we use `::std::int32_t` instead of `qint32`.

- [x] Can we use `::std::int32_t` instead of `qint32` ? #317

PR #317, opened 2022-10-14
WIP: cxx-qt-gen: use std types for integers instead of qint

ahayzen-kdab:
This doesn't work with Qt 5 it becomes confused with the types and thinks they are empty.

Related to #316


crates/cxx-qt-lib-headers/include/std_types.h
```diff
@@ -0,0 +1,23 @@
...
+
+#pragma once
+
+#include <QtGlobal>
+
+// For versions less than Qt 6 we need to manually register the std numerics
+#if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
+#include <QtCore/QMetaType>
+
+static const int i8 = qRegisterMetaType<::std::int8_t>("::std::int8_t");
```
ahayzen-kdab, line 16: Seems that https://doc.qt.io/qt-6/qmetatype.html#registerConverter-1 could also be used to convert to qt numbers?

ahayzen-kdab: Appears that `qRegisterMetaType` is enough

PR #318, opened 2022-10-14
cxx-qt-gen: rename cxx_qt::signals to cxx_qt::qsignals

ahayzen-kdab:
This then matches the Qt Q_SIGNALS closer and is consistent with qproperty and qinvokable etc.


LeonMatthesKDAB (approved):
LGTM

PR #322, opened 2022-10-17
changelog: add a CHANGELOG.md file for the releases 

ahayzen-kdab:


Be-ing:
I'd prefer to call this CHANGELOG.md, but the content looks good to me.

ahayzen-kdab:
Note that their file automatically creates links to the diffs by just putting `[0.1.0]` and `[Unreleased]` ( https://github.com/olivierlacan/keep-a-changelog/blob/main/CHANGELOG.md ) ... I assume this will work for us too once it is merged, as my fork doesn't have any tags / releases.

ahayzen-kdab:
> Note that their file automatically creates links to the diffs by just putting `[0.1.0]` and `[Unreleased]` ( https://github.com/olivierlacan/keep-a-changelog/blob/main/CHANGELOG.md ) ... I assume this will work for us too once it is merged, as my fork doesn't have any tags / releases.

aha, i found it there are actually some links at the bottom of the markdown hidden from view.

Be-ing (approved):

PR #323, opened 2022-10-17
cmake: rework Qt version selection

yuja:
It only applies to vcpkg builds in our cmake, and the original name conflicts
with the one used in Qt6 cmake.

https://code.qt.io/cgit/qt/qtdeclarative.git/tree/src/qml/Qt6QmlMacros.cmake?h=v6.3.1#n2087

Tested on Debian sid with `s/NAMES Qt6 Qt5/NAMES Qt6/`:
```
  CMake Error at /usr/lib/x86_64-linux-gnu/cmake/Qt6Qml/Qt6QmlMacros.cmake:2092 (message):
    qt_generate_foreign_qml_types() is only available in Qt 6.
  Call Stack (most recent call first):
    /usr/lib/x86_64-linux-gnu/cmake/Qt6Qml/Qt6QmlConfig.cmake:125 (include)
    /usr/lib/x86_64-linux-gnu/cmake/Qt6/Qt6Config.cmake:223 (find_package)
    examples/cargo_without_cmake/CMakeLists.txt:15 (find_package)
```


Be-ing:
Hello and thanks for this PR. There's a closely related bug which I recently noticed and was going to solve soon. My local builds which I thought were using Qt6 were actually linking Qt5, so using the way `QT_DEFAULT_MAJOR_VERSION` is being used is currently broken. I was planning on taking care of this soon, but if you want to do that first, that would be welcome.

Instead of using `QT_DEFAULT_MAJOR_VERSION`, I think we need to first only search for Qt6 with `find_package`, then if it's not available, fallback to Qt5. A new `QT5` CMake option would need to be added to skip the search for Qt6 when both are installed, which could also be used to tell vcpkg to download Qt5 instead of Qt6. I think that would be better than using a new CMake variable only to control what vcpkg does.

ahayzen-kdab:
Maybe I'm slightly confused what's happening, but if i had a normal Qt application that supports Qt 5 and Qt 6, how would I tell CMake which one to use? And why can we not copy that mechanism to decide which one to use with vcpkg ? Or is the problem that we want the vcpkg to be independent ?

Instead of `QT_DEFAULT_MAJOR_VERSION` being an `option` should we instead be reading it so we don't conflict ? As if i set the version then it should use that version. Or is the problem that the system might have set this to one version but then you might have a vcpkg locally with a different one?

Seems that https://doc.qt.io/qt-6/cmake-qt5-and-qt6-compatibility.html has some ideas and looks similar to what we do in some places.

CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ endif()
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
-option(QT_DEFAULT_MAJOR_VERSION "Major version of Qt to use (6 or 5)" 6)
+set(VCPKG_QT_DEFAULT_MAJOR_VERSION 6 CACHE STRING "Major version of Qt to use (6 or 5)")
```
ahayzen-kdab, line 17: Is there any reason why this is not an `option` and instead a `set` ? 

Also if this is now a VCPKG only option, should it be defined inside the `if(VCPKG)` below ?

yuja:
> Instead of using QT_DEFAULT_MAJOR_VERSION, I think we need to first only search for Qt6 with find_package, then if it's not available, fallback to Qt5.

Yes. It appears the order of `NAMES Qt6 Qt5` lookup is arbitrary (or maybe lexicographical?),
and I think we do need a cmake variable to select which Qt version to use. I've added `USE_QT5/6` options to my personal project:

https://hg.sr.ht/~yuja/hgv/browse/CMakeLists.txt?rev=tip

If you have any suggestion about naming, bool vs version, etc., I can implement Qt5/6 version switch based on that (though I'm not a cmake expert.)

> Seems that https://doc.qt.io/qt-6/cmake-qt5-and-qt6-compatibility.html has some ideas and looks similar to what we do in some places.

Yeah, but afaik it only works if cmake directories are isolated like the ones installed by Qt Installer. On Debian for example, both Qt5 and Qt6 are installed to /usr/lib/x86_64-linux-gnu/cmake, so we need to choose `find_package(Qt5)` xor `find_package(Qt6)`. That's my understanding.

CMakeLists.txt
yuja: If I understand cmake correctly, `option(NAME ...)` is equivalent to `set(NAME ... CACHE BOOL ...)`, and `5|6` isn't bool.

> Also if this is now a VCPKG only option, should it be defined inside the if(VCPKG) below ?

I didn't try that because I'm not sure if config knob can be conditionally defined. Anyway, it's probably better to have a common user variable to select Qt version.

Be-ing:
> Yeah, but afaik it only works if cmake directories are isolated like the ones installed by Qt Installer. On Debian for example, both Qt5 and Qt6 are installed to /usr/lib/x86_64-linux-gnu/cmake, so we need to choose find_package(Qt5) xor find_package(Qt6). That's my understanding.

Oh, interesting. Using `QT_DEFAULT_MAJOR_VERSION` as the Qt documentation suggests also doesn't work for me on Fedora 36 when I have both Qt5 and Qt6 -devel packages installed.

Be-ing:
I think we should use something like this example in the docs, but with an added `if(QT5)` guard around the first check for Qt6: https://doc.qt.io/qt-6/cmake-qt5-and-qt6-compatibility.html#versionless-targets

We could also switch to versionless targets then, which would cleanup the CMake code a bit.

yuja:
> I think we should use something like this example in the docs, but with an added if(QT5) guard around the first check for Qt6: https://doc.qt.io/qt-6/cmake-qt5-and-qt6-compatibility.html#versionless-targets

I'll rework this PR in that direction (but without versionless target for the moment), thanks.

> I already tried versionless targets but wasn't able to get the path to qmake from Qt::qmake because that target isn't created with the versionless targets.

I use versionless `Qt::qmake` in other project, and it appears working.
```
find_package(Qt5 COMPONENTS Core)
get_target_property(QMAKE Qt::qmake IMPORTED_LOCATION)
#get_target_property(QMAKE Qt::qmake LOCATION)
message(WARNING "XXX ${QMAKE}")
```

Be-ing:
> I use versionless Qt::qmake in other project, and it appears working.

Oh, nice. Yes, if that works, I'd prefer versionless targets.

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -198,7 +198,7 @@ jobs:
     - name: "Configure"
       run: >-
         cmake
-        -D QT_DEFAULT_MAJOR_VERSION=${{ matrix.qt_version }}
+        -D USE_QT5=${{ matrix.qt_version == 5 }}
```
Be-ing, line 201: oh neat, I didn't know you could use conditional expressions in this way in GitHub Actions

Be-ing (approved):

Be-ing:
Thank you!

Note that the documentation currently deployed at https://kdab.github.io/cxx-qt/book/ is for the 0.3 release and we've change a lot since then. We're aiming to make a 0.4 release in the next couple weeks but have [a few loose ends to finish first](https://github.com/KDAB/cxx-qt/milestone/3), among them is thoroughly updating the book. The first chapter of the book has been kept up to date and you can view that by running `mdbook serve` in the `book` directory of this repository, but the rest needs an overhaul.

Please don't hesitate to open issues or discussions here on this GitHub repository if you have questions.

PR #325, opened 2022-10-18
cmake: add CMakePresets with defaults relevant for building CXX-Qt

ahayzen-kdab:
This allows us to have presets for building with Qt 5, specifying recommended generators, caching tools, and showing output on failure of tests.

- [x] Add preset for vcpkg
- [x] Add preset for sscache (?)
- [x] Add preset for ninja (?)
- [ ] Potentially use presets for CI too
- [x] Add RUSTC_WRAPPER to environment for sscache presets
- [x] Should we have all the combinations or just the simplest + "best" ?

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,314 @@
+{
+    "version": 5,
+    "configurePresets": [
+        {
+            "name": "common",
+            "hidden": true,
+            "binaryDir": "${sourceDir}/build/${presetName}"
+        },
+        {
+            "name": "build-debug",
+            "hidden": true,
+            "cacheVariables": {
+                "CMAKE_BUILD_TYPE": "Debug"
+            }
+        },
+        {
+            "name": "build-release",
+            "hidden": true,
+            "cacheVariables": {
+                "CMAKE_BUILD_TYPE": "Release"
+            }
+        },
+        {
+            "name": "ninja",
+            "hidden": true,
+            "generator": "Ninja"
+        },
+        {
+            "name": "sscache",
+            "hidden": true,
+            "cacheVariables": {
+                "CMAKE_CXX_COMPILER_LAUNCHER": "sccache"
+            },
+            "environment": {
+                "RUSTC_WRAPPER": "sccache"
+            }
+        },
+        {
+            "name": "qt5",
+            "hidden": true,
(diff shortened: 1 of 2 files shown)


CMakePresets.json
```diff
@@ -0,0 +1,63 @@
+{
+    "version": 5,
+    "configurePresets": [
+        {
+            "name": "common",
+            "hidden": true,
+            "cacheVariables": {
+                "CMAKE_CXX_COMPILER_LAUNCHER": "sccache"
+            }
```
Be-ing, line 9: Won't this break if the user doesn't have sccache installed?

ahayzen-kdab: Right, and if you are missing ninja. I could add more presets for the different options like other projects do...

ahayzen-kdab: Added different presets for different combinations.

ahayzen-kdab:
Seems Windows might need the `--config Release` etc at build time ? So might need `buildPresets` for all the combinations too :sweat_smile:

Be-ing:
> vscode... setting the options manually is kinda a pain.

That sounds like a VSCode problem. :)

I don't have any use for these because I build in a shell, so I don't really mind what you do with these. @LeonMatthesKDAB, thoughts?

ahayzen-kdab:
> > vscode... setting the options manually is kinda a pain.
> 
> That sounds like a VSCode problem. :)

Hehe :-) I can keep them as CMakeUserPresets just didn't know if they were useful upstream. Also CI could use them and when building from the CLI they could be used rather than multiple flags.

Or the matrix could be reduced so there are only some options :-)

CMakePresets.json
```diff
@@ -0,0 +1,314 @@
...
+            "name": "build-debug",
+            "configuration": "Debug",
+            "hidden": true
+        },
+        {
+            "name": "build-release",
+            "configuration": "Release",
+            "hidden": true
+        },
+        {
```
ahayzen-kdab, line 225: If we do use this we likely also want to set the parallel / execution jobs to something larger than 1

```
"execution": {
    "jobs": 2
},
```

ahayzen-kdab:
Going to close this for now as there are too many combinations to push into the repo.

PR #326, opened 2022-10-19
cxx-qt-gen: wrap CxxQtThread::queue() to support Send + 'static closure

yuja:
The idea is to pass a `(callback fn, Send + 'static closure)` pair in to
`QMetaObject::invokeMethod()`, and invoke the callback fn with that closure
as an argument. The callback fn just invokes the passed closure. This way,
we can queue Rust closure back to QObject thread.

    thread::spawn(move || {
        ...
        let response = String::new();
        qt_thread.queue(move |qobject| qobject.emit(Finished { response }));
    });

This should be simpler and less error-prone than sending result back via
a separate channel.

### DIFF
--- a/book/src/concepts/threading.md
+++ b/book/src/concepts/threading.md
@@ -27,8 +27,8 @@ To achieve safe multi-threading on the Rust side we use an [`CxxQtThread<T>`](..
 A `CxxQtThread<T>` represents a reference to the Qt thread that the QObject of type `T` lives in.
 When a new Rust thread is started (e.g. in an invokable) the `CxxQtThread<T>` can be moved into the thread to later update the QObject in a thread safe manner.
 
-When the Rust thread needs to update a value in the QObject it can then queue a function pointer to the thread.
-This function pointer will be executed on the thread the QObject lives in.
+When the Rust thread needs to update a value in the QObject it can then queue a closure to the thread.
+This closure will be executed on the thread the QObject lives in.
 Updating the QObject is then thread-safe.
 
 Below is a complete Rust example of a multi-threaded object.
--- a/crates/cxx-qt-gen/src/generator/naming/qobject.rs
+++ b/crates/cxx-qt-gen/src/generator/naming/qobject.rs
@@ -15,6 +15,8 @@ pub struct QObjectName {
     pub rust_struct: CombinedIdent,
     /// The name of the CxxQtThread
     pub cxx_qt_thread_class: Ident,
+    /// The name of the Rust closure wrapper to be passed in to CxxQtThread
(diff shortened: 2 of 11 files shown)


ahayzen-kdab (changes requested):
So firstly, this is awesome and what we tried to get working originally but couldn't figure out how to get it working with CXX. The closure -> Box hack is genius ! :partying_face: 

It looks good for merging we just have one suggestion around the lambda, with some extra move's and making the lambda mutable (because we want to move captures https://stackoverflow.com/questions/8640393/move-capture-in-lambda ) we were able to avoid the shared_ptr / making copyable. Are you able to apply this diff to your commits?

```diff
diff --git a/crates/cxx-qt-lib-headers/include/cxxqt_thread.h b/crates/cxx-qt-lib-headers/include/cxxqt_thread.h
index 8e4222a2..e477bf37 100644
--- a/crates/cxx-qt-lib-headers/include/cxxqt_thread.h
+++ b/crates/cxx-qt-lib-headers/include/cxxqt_thread.h
@@ -61,14 +61,13 @@ public:
     auto lambda = [obj = std::move(obj),
                    rustObjMutex = std::move(rustObjMutex),
                    func = std::move(func),
-                   // invokeMethod() requires lambda to be copyable
-                   arg = std::make_shared<rust::Box<A>>(std::move(arg))]() {
+                   arg = std::move(arg)]() mutable {
       // Ensure that we can read the pointer and it's not being written to
       const auto guard = std::shared_lock(obj->mutex);
       if (obj->ptr) {
         // Ensure that the rustObj is locked
         const std::lock_guard<std::recursive_mutex> guardRustObj(*rustObjMutex);
-        func(*obj->ptr, std::move(*arg));
+        func(*obj->ptr, std::move(arg));
       } else {
         qWarning()
           << "Could not call the function pointer as object has been destroyed";
@@ -76,7 +75,8 @@ public:
     };
 
     // Add the lambda to the queue
-    if (!QMetaObject::invokeMethod(m_obj->ptr, lambda, Qt::QueuedConnection)) {
+    if (!QMetaObject::invokeMethod(
+          m_obj->ptr, std::move(lambda), Qt::QueuedConnection)) {
       throw std::runtime_error(
         "Cannot queue function pointer as invokeMethod on object failed");
     }
```

yuja:
> with some extra move's and making the lambda mutable (because we want to move captures https://stackoverflow.com/questions/8640393/move-capture-in-lambda ) we were able to avoid the shared_ptr / making copyable.

Pushed new version. 

I didn't know mutable lambda things. Many thanks!

ahayzen-kdab (approved):
Thanks awesome :partying_face: LGTM!

PR #329, opened 2022-10-19
WIP: qml-features: use friend to allow protected methods to be reached

ahayzen-kdab:


ahayzen-kdab:
This needs API discussions of how this would be triggered, and would go into the generated C++

Likely something like
```rust
#[cxx_qt::bridge]
mod ffi {
    unsafe extern "C++" {
        #[cxx_qt::protected]
        #[cxx_name = "beginResetModel"]
        fn begin_reset_model(self: Pin<&mut MyObjectQt>);
        #[cxx_qt::protected]
        #[cxx_name = "endResetModel"]
        fn end_reset_model(self: Pin<&mut MyObjectQt>);
    }
}
```

LeonMatthesKDAB:
This PR is abandoned now I think.
And no user seems to have cared about the leaking of protected functions yet :)

PR #335, opened 2022-10-26
Bump pkg-config-rs dependency to 0.3.26

### LINKED ISSUE #304: Remove any branches from cargo.toml's
Remove any branches from cargo.toml's (eg qt-build) so that we can publish to crates.io.

LeonMatthesKDAB:
Fixes: https://github.com/rust-lang/pkg-config-rs/pull/134

Closes #304

### DIFF
--- a/crates/qt-build-utils/Cargo.toml
+++ b/crates/qt-build-utils/Cargo.toml
@@ -14,7 +14,7 @@ repository = "https://github.com/KDAB/cxx-qt/"
 
 [dependencies]
 versions = "4.1.0"
-pkg-config = { git = "https://github.com/Be-ing/pkg-config-rs.git", branch = "library_pub" }
+pkg-config = "0.3.26"
 regex = "1.6.0"
 lazy_static = "1.0"
 thiserror = "1.0"


crates/qt-build-utils/Cargo.toml
```diff
@@ -14,7 +14,7 @@ repository = "https://github.com/KDAB/cxx-qt/"
 
 [dependencies]
 versions = "4.1.0"
-pkg-config = { git = "https://github.com/Be-ing/pkg-config-rs.git", branch = "library_pub" }
```
ahayzen-kdab, line 17: this won't work as it still requires vendoring of some private methods, but now that upstream has stabilised what we are going to use we can implement this.

LeonMatthesKDAB: Ah, alright, I thought the changes linked in the pkg-config-rs PR were enough for our purposes :thinking:

Be-ing: I will work on decoupling the required code from pkg-config's structs this week.

Be-ing:
Superceded by #337

crates/qt-build-utils/Cargo.toml
Be-ing: For context: I initially tried to use pkg-config, but [Qt6 doesn't ship .pc files for static builds yet](https://bugreports.qt.io/browse/QTBUG-86080). Also, using pkg-config would require somehow getting the pkg-config executable on Windows. Instead, we're parsing the .prl files which are used by QMake.

PR #337, opened 2022-10-26
qt-build-utils: vendor code from pkg-config

Be-ing:
instead of using my branch hacking some private pkg-config functions to be public. Publication to crates.io requires only using crates on crates.io, not Git URLs.

This also removes the dependency on regex which is overkill for the string processing required.

Fixes https://github.com/KDAB/cxx-qt/issues/304


ahayzen-kdab (approved):
LGTM :-) And looks like mingw would use libxxx.dll not xxx.dll so that has potentially fixed something for that untested usecase :-)

PR #343, opened 2022-10-28
book: port existing Dia image to GraphViz (dot)

ahayzen-kdab:


book/src/images/overview_abstract.dot
```diff
@@ -0,0 +1,53 @@
...
+        }
+    }
+
+    // Connect bridge to our fake nodes
+    subgraph {
+        edge [dir = none;];
+
+        bridge -> center;
+        center -> left [label = "Macro Expansion";];
+        center -> right [label = "Code Generation";];
```
LeonMatthesKDAB, line 44: This is the wrong way round ;)

This does bring up the point though:
Currently we're the opposite of the CXX overview diagram (C++ on the left, Rust on the Right).
So I suggest flipping the nodes, instead of the arrow labels.
Can you move "Rust implementation" to the left and "QObject subclass" to the right?

LeonMatthesKDAB (changes requested):

book/src/images/overview_abstract.dot
ahayzen-kdab: Sure :-)

LeonMatthesKDAB (approved):
LGTM

PR #348, opened 2022-11-01
parser: allow multiple QObjects in a single block and add tests

### LINKED ISSUE #293: Support multiple QObject's being defined in one block
Most items are mangled or namespaced now, the macro parsing and generation supports multiple QObjects.

So once #283 has landed, is there anything preventing this being enabled? and is there anything that breaks with it?

ahayzen-kdab:
Closes #293

### DIFF
--- a/crates/cxx-qt-gen/src/parser/cxxqtdata.rs
+++ b/crates/cxx-qt-gen/src/parser/cxxqtdata.rs
@@ -8,13 +8,16 @@ use crate::syntax::{
     attribute::{attribute_find_path, attribute_tokens_to_ident},
     path::path_to_single_ident,
 };
-use std::collections::HashMap;
+use std::collections::BTreeMap;
 use syn::{spanned::Spanned, Error, Ident, Item, ItemEnum, ItemImpl, Result, Type, TypePath};
 
 #[derive(Default)]
 pub struct ParsedCxxQtData {
     /// Map of the QObjects defined in the module that will be used for code generation
-    pub qobjects: HashMap<Ident, ParsedQObject>,
+    //
+    // We have to use a BTreeMap here, instead of a HashMap, to keep the order of QObjects stable.
+    // Otherwise, the output order would be different, depending on the environment, which makes it hard to test/debug.
+    pub qobjects: BTreeMap<Ident, ParsedQObject>,
     /// The namespace of the CXX-Qt module
     pub namespace: String,
     /// Any `use` statements end up in the CXX-Qt generated module
@@ -22,7 +25,7 @@ pub struct ParsedCxxQtData {
 }
 
 impl ParsedCxxQtData {
-    /// Find the QObjects within the module and add into the qobjects HashMap
+    /// Find the QObjects within the module and add into the qobjects BTreeMap
     pub fn find_qobject_structs(&mut self, items: &[Item]) -> Result<()> {
         for item in items {
             if let Item::Struct(qobject_struct) = item {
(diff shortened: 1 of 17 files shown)


examples/qml_features/rust/src/multiple_qobjects.rs
```diff
@@ -0,0 +1,61 @@
...
+        pub fn increment(mut self: Pin<&mut Self>) {
+            let new_value = self.as_ref().counter() + 1;
+            self.as_mut().set_counter(new_value);
+
+            self.emit(FirstSignals::DataChanged);
+        }
+    }
+
+    #[cxx_qt::qobject]
+    pub struct SecondObject {
```
LeonMatthesKDAB, line 36: I think in this example the SecondObject should at least have slightly different behavior from the FirstObject.
Otherwise the example makes it seem like there are just two instances of the same class.

For example, have the page show two "Click FirstObject" and "Click SecondObject" buttons.
If you click the FirstObject, the button color cycles.
If you click the SecondObject, it moves the button to the right.
Or something silly like that.

crates/cxx-qt-gen/src/parser/cxxqtdata.rs
```diff
@@ -8,35 +8,27 @@ use crate::syntax::{
...
 impl ParsedCxxQtData {
-    /// Find the QObjects within the module and add into the qobjects HashMap
+    /// Find the QObjects within the module and add into the qobjects BTreeMap
     pub fn find_qobject_structs(&mut self, items: &[Item]) -> Result<()> {
         for item in items {
             if let Item::Struct(qobject_struct) = item {
                 if let Some(index) =
                     attribute_find_path(&qobject_struct.attrs, &["cxx_qt", "qobject"])
                 {
-                    // TODO: for now we only support one qobject per block
```
LeonMatthesKDAB, line 32: Nice to see all it took was to remove a few lines :D

LeonMatthesKDAB (changes requested):
I think the example could be improved a bit, as having two classes with the exact same implementation doesn't really show of the feature well.

crates/cxx-qt-gen/src/parser/cxxqtdata.rs
```diff
@@ -8,35 +8,27 @@ use crate::syntax::{
...
 };
-use std::collections::HashMap;
+use std::collections::BTreeMap;
 use syn::{spanned::Spanned, Error, Ident, Item, ItemEnum, ItemImpl, Result, Type, TypePath};
 
 #[derive(Default)]
 pub struct ParsedCxxQtData {
     /// Map of the QObjects defined in the module that will be used for code generation
-    pub qobjects: HashMap<Ident, ParsedQObject>,
+    pub qobjects: BTreeMap<Ident, ParsedQObject>,
```
LeonMatthesKDAB, line 17: There's no functional reason this changed, it's only for performance, right?

crates/cxx-qt-gen/src/parser/cxxqtdata.rs
ahayzen-kdab: Right, most of the other stages were tested to be working, we just didn't know if there were any other missing parts :-)

examples/qml_features/rust/src/multiple_qobjects.rs
LeonMatthesKDAB: Your fix should make this obvious enough :)

ahayzen-kdab: Changed so they have a unique property and behaviour. But note that i chose having some of the same propertties/signals/invokables to test that our namespacing works correctly and doesn't collide.

crates/cxx-qt-gen/src/parser/cxxqtdata.rs
LeonMatthesKDAB: Ah, that's a good point :thinking: 
Can you add a comment explaining this 
```suggestion
    // We have to use a BTreeMap here, instead of a HashMap, to keep the order of QObjects stable.
    // Otherwise, the output order would be different, depending on the environment, which makes it hard to test/debug.
    pub qobjects: BTreeMap<Ident, ParsedQObject>,
```

ahayzen-kdab: Done

LeonMatthesKDAB (approved):
LGTM

Issue #347, opened 2022-11-01
Explore how to represent container types


ahayzen-kdab:
# QtCore

- [x] QHash (key-value) #354 
- [x] QMap (key-value) #378 
- [x] QSet #350 
- [x] QVector (QList in Qt6) #383 
- [x] Qt5List #387 
- [x] ~QLinkedList ? (Qt5 only?)~

# Problems

Container types are hard to represent via extern FFI or CXX blocks due to the following problems

- extern "C" doesn't support generics in it's type https://internals.rust-lang.org/t/generic-externs-for-high-level-ffi/15046/
- CXX doesn't support generics in it's type https://github.com/dtolnay/cxx/issues/683
- Key-value's are particularly tricky as we can't even add type specific methods like for vector `T::push_back`

# CXX

For representing `std::vector<T>`, CXX has a `CxxVector<T>` this has PhantomData inside it https://github.dev/dtolnay/cxx/blob/c7060d40361c4d955fc64adf35f67b28b5c75299/src/cxx_vector.rs#L24

Then a trait `VectorElement` https://github.dev/dtolnay/cxx/blob/c7060d40361c4d955fc64adf35f67b28b5c75299/src/cxx_vector.rs#L333 is implemented for all supported types `T`. In methods such as `len` then `T::__vector_size(self)` is used. https://github.dev/dtolnay/cxx/blob/c7060d40361c4d955fc64adf35f67b28b5c75299/src/cxx_vector.rs#L45

A macro is used to generate the trait implementation for each type which then, eg for `__vector_size`, uses an inline `extern "C"` block to create a method such as `cxxbridge1$std$vector$i32$size` which then has the type `__vector_size(&CxxVector<i32>) -> usize;` https://github.dev/dtolnay/cxx/blob/c7060d40361c4d955fc64adf35f67b28b5c75299/src/cxx_vector.rs#L401

There is then also C++ generation in the write phase which creates the matching C++ method with the name and type then using `return self.size();`  https://github.dev/dtolnay/cxx/blob/c7060d40361c4d955fc64adf35f67b28b5c75299/gen/src/write.rs#L1886

## Questions

This allows you to represent something with a single type in Rust, but has the following problems
- Does not work with a key-value store as how would the `T::__map_size(self)` work, as it needs to have both K and V ? 
- CXX's `VectorElement` is undocumented, so it's not documented that developers could implement this on their own types, we should have a stable trait for this
[...]

PR #350, opened 2022-11-04
cxx-qt-lib: build a QSet<T>

ahayzen-kdab:
Related to #347

### DIFF
--- a/crates/cxx-qt-lib-headers/include/qset.h
+++ b/crates/cxx-qt-lib-headers/include/qset.h
@@ -22,38 +22,74 @@ struct rust::IsRelocatable<QSet<T>> : std::true_type
 {
 };
 
-// Note on the Rust side we rename the methods to "len" etc
-// so we need to put each set in it's own namespace otherwise the generated
-// CXX code collides.
-#define CXX_QT_QSET_METHODS(typeName, name)                                    \
-  using QSet_##name = QSet<typeName>;                                          \
-                                                                               \
-  namespace rust {                                                             \
-  namespace cxxqtlib1 {                                                        \
-  namespace qset_##name                                                        \
-  {                                                                            \
-    QSet_##name qset_clone_##name(const QSet_##name& s) noexcept;              \
-    QSet_##name qset_default_##name() noexcept;                                \
-    void qset_drop_##name(QSet_##name& s) noexcept;                            \
-    const typeName& qset_get_unchecked_##name(const QSet_##name& s,            \
-                                              ::std::size_t pos) noexcept;     \
-    void qset_insert_##name(QSet_##name& s, const typeName& value) noexcept;   \
-    ::std::size_t qset_len_##name(const QSet_##name& s) noexcept;              \
-  }                                                                            \
-  }                                                                            \
-  }
+namespace rust {
+namespace cxxqtlib1 {
 
-CXX_QT_QSET_METHODS(bool, bool);
-CXX_QT_QSET_METHODS(float, f32);
-CXX_QT_QSET_METHODS(double, f64);
-CXX_QT_QSET_METHODS(::qint8, i8);
-CXX_QT_QSET_METHODS(::qint16, i16);
-CXX_QT_QSET_METHODS(::qint32, i32);
-CXX_QT_QSET_METHODS(::QDate, QDate);
-CXX_QT_QSET_METHODS(::QDateTime, QDateTime);
-CXX_QT_QSET_METHODS(::QString, QString);
-CXX_QT_QSET_METHODS(::QTime, QTime);
-CXX_QT_QSET_METHODS(::QUrl, QUrl);
-CXX_QT_QSET_METHODS(::quint8, u8);
-CXX_QT_QSET_METHODS(::quint16, u16);
-CXX_QT_QSET_METHODS(::quint32, u32);
+template<typename T>
+QSet<T>
+qset_clone(const QSet<T>& s) noexcept
+{
+  return QSet(s);
+}
+
+template<typename T>
+QSet<T>
+qset_default() noexcept
+{
+  return QSet<T>();
+}
+
+template<typename T>
+void
+qset_drop(QSet<T>& s) noexcept
+{
+  s.~QSet<T>();
+}
+
+template<typename T>
+const T&
+qset_get_unchecked(const QSet<T>& s, ::std::size_t pos) noexcept
+{
+  Q_ASSERT(pos < static_cast<::std::size_t>(s.size()));
+  auto it = s.cbegin();
+  std::advance(it, pos);
+  return *it;
+}
+
+template<typename T>
+void
+qset_insert(QSet<T>& s, const T& value) noexcept
+{
+  s.insert(value);
+}
+
+template<typename T>
+::std::size_t
+qset_len(const QSet<T>& s) noexcept
+{
+  // In Qt 5 the type was int now it is qsizetype, so we need to ensure the type
+  // is the same for CXX
+#if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))
+  return s.size();
+#else
+  return static_cast<::std ::size_t>(s.size());
+#endif
+}
+
+}
+}
+
+using QSet_bool = QSet<bool>;
+using QSet_f32 = QSet<float>;
+using QSet_f64 = QSet<double>;
+using QSet_i8 = QSet<::qint8>;
+using QSet_i16 = QSet<::qint16>;
+using QSet_i32 = QSet<::qint32>;
+using QSet_QDate = QSet<::QDate>;
+using QSet_QDateTime = QSet<::QDateTime>;
+using QSet_QString = QSet<::QString>;
+using QSet_QTime = QSet<::QTime>;
+using QSet_QUrl = QSet<::QUrl>;
+using QSet_u8 = QSet<::quint8>;
+using QSet_u16 = QSet<::quint16>;
+using QSet_u32 = QSet<::quint32>;
--- a/crates/cxx-qt-lib-headers/include/qset.h
+++ b/crates/cxx-qt-lib-headers/include/qset.h
@@ -22,7 +22,7 @@ struct rust::IsRelocatable<QSet<T>> : std::true_type
 {
 };
 
-// Note on the Rust side we rename the methods to "clear" or "len" etc
+// Note on the Rust side we rename the methods to "len" etc
 // so we need to put each set in it's own namespace otherwise the generated
 // CXX code collides.
 #define CXX_QT_QSET_METHODS(typeName, name)                                    \
@@ -32,17 +32,13 @@ struct rust::IsRelocatable<QSet<T>> : std::true_type
   namespace cxxqtlib1 {                                                        \
   namespace qset_##name                                                        \
   {                                                                            \
-    void qset_clear_##name(QSet_##name& s) noexcept;                           \
     QSet_##name qset_clone_##name(const QSet_##name& s) noexcept;              \
-    bool qset_contains_##name(const QSet_##name& s,                            \
-                              const typeName& value) noexcept;                 \
     QSet_##name qset_default_##name() noexcept;                                \
     void qset_drop_##name(QSet_##name& s) noexcept;                            \
     const typeName& qset_get_unchecked_##name(const QSet_##name& s,            \
                                               ::std::size_t pos) noexcept;     \
     void qset_insert_##name(QSet_##name& s, const typeName& value) noexcept;   \
-    std::size_t qset_len_##name(const QSet_##name& s) noexcept;                \
-    bool qset_remove_##name(QSet_##name& s, const typeName& value) noexcept;   \
+    ::std::size_t qset_len_##name(const QSet_##name& s) noexcept;              \
   }                                                                            \
   }                                                                            \
   }
--- a/crates/cxx-qt-lib-headers/include/qset.h
+++ b/crates/cxx-qt-lib-headers/include/qset.h
@@ -1,0 +1,63 @@
+// clang-format off
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// clang-format on
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+#pragma once
+
+#include <QtCore/QSet>
+
+#include <QtCore/QDate>
+#include <QtCore/QDateTime>
+#include <QtCore/QString>
+#include <QtCore/QTime>
+#include <QtCore/QUrl>
+
+#include "rust/cxx.h"
+
+// This has static asserts in the cpp file to ensure this is valid.
+template<typename T>
+struct rust::IsRelocatable<QSet<T>> : std::true_type
+{
+};
+
+// Note on the Rust side we rename the methods to "clear" or "len" etc
+// so we need to put each set in it's own namespace otherwise the generated
+// CXX code collides.
+#define CXX_QT_QSET_METHODS(typeName, name)                                    \
+  using QSet_##name = QSet<typeName>;                                          \
+                                                                               \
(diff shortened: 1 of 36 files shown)


examples/qml_features/rust/src/custom_base_class.rs
```diff
@@ -53,6 +60,12 @@ mod ffi {
         fn end_reset_model(self: Pin<&mut CustomBaseClassQt>);
     }
 
+    // TODO: have a way to generate the type specific code
```
ahayzen-kdab, line 63: this has problems as you hit `only traits defined in the current crate can be implemented for types defined outside of the crate` and the same for primitive types like `i32`.

When trying to for `impl QSetElement for i32` and `unsafe impl ExternType for QSetI32` (`QSet<i32>`)

Structs can be used to create intermediate types and then deref to get back, but doesn't quite work with the CXX bridge yet...

examples/qml_features/rust/src/custom_base_class.rs
```diff
@@ -53,6 +53,25 @@ mod ffi {
...
+    //
+    // TODO: can this also define that QSetI32 is a type alias?
+    //
+    // impl cxx_qt_lib::QSet<i32> {}
+    //
+    // TODO: generate the following
+    unsafe extern "C++" {
+        // TODO: the include won't be required if the C++ is generated
+        include!("cxx-qt-lib/qset.h");
+        type QSetI32 = super::private::QSetI32;
```
LeonMatthesKDAB, line 66: Hm, we could generate this part from:
``` rust
type QSetI32 = cxx_qt_lib::QSet<i32>;
```
As well as the typedef `using QSetI32 = QSet<i32>;` in the C++ side.

However, then we'd have to again start scanning `extern "C++"` blocks, which we want to avoid...

LeonMatthesKDAB: Hm, it should also be doable to support a templated extern type :thinking: 
https://github.com/dtolnay/cxx/issues/251

And more specifically:
https://github.com/dtolnay/cxx/pull/524/files#diff-a9b6bb47cc68e30ec36cb376284aaa6fdee44f9fc71308fc043df611e760f7a1

examples/qml_features/rust/src/custom_base_class.rs
ahayzen-kdab: Solved, with some templating and trait fun :-)

examples/qml_features/rust/src/custom_base_class.rs
ahayzen-kdab: Right this works :-)

crates/cxx-qt-lib-headers/include/qset.h
```diff
@@ -0,0 +1,100 @@
...
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+#pragma once
+
+#include <QtCore/QSet>
+
+#include <QtCore/QDate>
+#include <QtCore/QDateTime>
+// #include <QtCore/QPoint>
```
ahayzen-kdab, line 13: Only `qHash`able types are supported by QSet due to it's internal implementation.

As seen in https://doc.qt.io/qt-6/qhash.html#related-non-members QPoint, QPointF, QRect, QRectF, QSize, QSizeF, QVariant, and QColor are missing.

So should we not implement these ? or should we create a qHash function for them ?

crates/cxx-qt-lib/src/types/qset/mod.rs
```diff
@@ -0,0 +1,426 @@
...
+mod qset_qdatetime;
+mod qset_qstring;
+mod qset_qtime;
+mod qset_qurl;
+mod qset_u16;
+mod qset_u32;
+mod qset_u8;
+
+// TODO: do we need packed?
+#[repr(C, packed)]
```
ahayzen-kdab, line 24: do we need packed here?

crates/cxx-qt-lib/src/types/qset/mod.rs
```diff
@@ -0,0 +1,426 @@
...
+    type TypeId;
+
+    fn clear(set: &mut QSet<Self>);
+    fn clone(set: &QSet<Self>) -> QSet<Self>;
+    fn contains(set: &QSet<Self>, value: Self) -> bool;
+    fn default() -> QSet<Self>;
+    fn drop(set: &mut QSet<Self>);
+    // TODO: we are passing as a const ref internally to C++,
+    // so do we need to have ManuallyDrop?
+    fn insert(set: &mut QSet<Self>, value: Self)
```
ahayzen-kdab, line 123: do we need ManuallyDrop if we are passing as a const-ref internally, so we can safely let Rust drop the item?

crates/cxx-qt-lib-headers/include/qset.h
```diff
@@ -0,0 +1,100 @@
...
+
+namespace rust {
+namespace cxxqtlib1 {
+namespace qset {
+
+// Note on the Rust side we rename the methods to "clear" or "len" etc
+// so we need to put each set in it's own namespace otherwise the generated
+// CXX code collides.
+#define CXX_QT_QSET_METHODS(typeName, name, lowerName)                         \
+  namespace lowerName {                                                             \
```
ahayzen-kdab, line 63: the namespace name can't be `bool` as it conflicts with `bool`, but then are names like `u8` or `boolean` any better? or should we call them something like `cxx_bool` or instead drop the nesting and have `qset_bool` (this seems the best solution?)

crates/cxx-qt-lib-headers/include/qset.h
ahayzen-kdab: Removed these for now

crates/cxx-qt-lib-headers/include/qset.h
ahayzen-kdab: Changed to `qset_T`

crates/cxx-qt-lib/src/types/qset/mod.rs
ahayzen-kdab: Removed.

crates/cxx-qt-lib-headers/include/qset.h
ahayzen-kdab: Might even need to be behind a feature as we have a few differing issues

- If we implement `qHash` for the types, does this conflict with any `qHash` implementations the developer has made?
- If we require the developer to implement `qHash` then they can't `impl QSetElement for QPoint` as both are outside their crate (they would have to create `MyPoint` and do it that way)
- So maybe we need to have a `assume-qhashed-types` feature which assumes the developer has made a `qHash` method (and therefore all Qt types in cxx-qt-lib are hashable) and is opt-in

crates/cxx-qt-lib/src/types/qset/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+mod qset_u32;
+mod qset_u8;
+
+/// The QSet class is a template class that provides a hash-table-based set.
+///
+/// Note that this means that T needs to have a global `qHash()` function.
+///
+/// To use QSet with a custom type, implement the [`QSetElement`] trait for T.
+#[repr(C)]
+pub struct QSet<T>
```
ahayzen-kdab, line 30: need to impl an iter on QSet too

ahayzen-kdab: There is Iter now :-)

crates/cxx-qt-lib/src/types/qset/generate.sh
```diff
@@ -0,0 +1,110 @@
...
+
+#[cxx::bridge]
+pub mod ffi {
+    unsafe extern "C++" {
+        include!("cxx-qt-lib/qset.h");
+        type QSet_$1 = crate::QSet<$1>;
+    }
+
+    #[namespace = "rust::cxxqtlib1::qset_$1"]
+    unsafe extern "C++" {
```
LeonMatthesKDAB, line 28: At the moment this is C-style FFI again.
We can simplify this with CXX.
Just use:
``` rust
unsafe extern "C++" {
    #[rust_name="cxx_clear"]
    fn clear(self: &mut QSet_$1);
}
```

Which will generate the clear method right onto QSet.

Example patch:
[0001-Move-clear-to-using-CXX-style-ffi.patch.txt](https://github.com/KDAB/cxx-qt/files/10076572/0001-Move-clear-to-using-CXX-style-ffi.patch.txt)

We might also be able to use template methods for the remaining functions, like qset_clone...

LeonMatthesKDAB (changes requested):
We could simplify this using CXX instead of C-style FFI.

LeonMatthesKDAB:
Initial version of the templated idea:
[0002-Use-template-methods-for-clone-etc.patch.txt](https://github.com/KDAB/cxx-qt/files/10077031/0002-Use-template-methods-for-clone-etc.patch.txt)

crates/cxx-qt-lib/src/types/qset/generate.sh
ahayzen-kdab: Done

crates/cxx-qt-lib/build.rs
```diff
@@ -44,11 +58,28 @@ fn main() {
...
 
     let mut builder = cxx_build::bridges(
-        bridges
+        rust_bridges
             .iter()
             .map(|bridge| format!("src/types/{}.rs", bridge)),
     );
-    for bridge in bridges {
+
+    let cpp_bridges = [
```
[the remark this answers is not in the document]
ahayzen-kdab: Done

crates/cxx-qt-lib-headers/include/qset.h
```diff
@@ -0,0 +1,62 @@
...
+#include <QtCore/QDate>
+#include <QtCore/QDateTime>
+#include <QtCore/QString>
+#include <QtCore/QTime>
+#include <QtCore/QUrl>
+
+#include "rust/cxx.h"
+
+template<typename T>
+struct rust::IsRelocatable<QSet<T>> : std::true_type
```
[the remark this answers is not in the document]
ahayzen-kdab: Done

ahayzen-kdab:
> Initial version of the templated idea: [0002-Use-template-methods-for-clone-etc.patch.txt](https://github.com/KDAB/cxx-qt/files/10077031/0002-Use-template-methods-for-clone-etc.patch.txt)

Done

Be-ing:
It would help to incorporate this into one of the examples to see how the API works in practice.

crates/cxx-qt-lib/src/types/qset/generate.sh
```diff
@@ -0,0 +1,164 @@
+#!/usr/bin/env bash
```
Be-ing, line 1: :see_no_evil: Can we use a macro for this instead?

crates/cxx-qt-lib/src/types/qset/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+        T::insert(self, value);
+    }
+}
+
+unsafe impl<T> ExternType for QSet<T>
+where
+    T: ExternType + QSetElement,
+{
+    type Id = T::TypeId;
+    type Kind = T::Kind;
```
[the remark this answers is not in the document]
LeonMatthesKDAB: This should always be trivial.
`QSet<X>` is always a trivial type, no matter what X is.

tests/qt_types_standalone/rust/src/qset.rs
```diff
@@ -0,0 +1,49 @@
...
+    let vec = s.iter().cloned().collect::<Vec<i32>>();
+
+    s.contains(&1)
+        && !s.contains(&2)
+        && s.contains(&3)
+        && s.len() == 2
+        && vec.contains(&1)
+        && !vec.contains(&2)
+        && vec.contains(&3)
+        && vec.len() == 2
```
LeonMatthesKDAB, line 44:
```suggestion
    // Sort the vec as a set iterator does not have a guaranteed order
    vec.sort();
    vec == vec![1,3]
```

LeonMatthesKDAB (changes requested):
The overall code is good, I just have a few more nitpicks, then ready for merge IMHO.

crates/cxx-qt-lib/src/types/qset/generate.sh
LeonMatthesKDAB: Also I think generating a CXX bridge from a macro doesn't work, as the C++ code generation can't do macro expansion I think :thinking:

ahayzen-kdab: Right, it'd be a lot of work in the macro to make it trigger CXX inside the macro and then with the concat_ident not being stable :shrug: this was the simplest way. As it's only a tool to help the developer update the files.

crates/cxx-qt-lib/src/types/qset/mod.rs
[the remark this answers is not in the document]
ahayzen-kdab: Done

crates/cxx-qt-lib/build.rs
```diff
@@ -44,11 +58,28 @@ fn main() {
...
+        "qrectf",
+        "qset",
+        "qsize",
+        "qsizef",
+        "qstring",
+        "qtime",
+        "qurl",
+        "qvariant",
+    ];
+    for bridge in cpp_files {
```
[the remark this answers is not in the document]
ahayzen-kdab: Done

crates/cxx-qt-lib/src/types/qset.cpp
```diff
@@ -0,0 +1,37 @@
+// clang-format off
```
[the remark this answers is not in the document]
ahayzen-kdab: Done

crates/cxx-qt-lib-headers/include/qset.h
```diff
@@ -0,0 +1,95 @@
...
+template<typename T>
+QSet<T>
+qset_clone(const QSet<T>& s) noexcept
+{
+  return QSet(s);
+}
+
+template<typename T>
+QSet<T>
+qset_default() noexcept
```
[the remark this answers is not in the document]
ahayzen-kdab: :-D

crates/cxx-qt-lib/src/types/qset.cpp
```diff
@@ -0,0 +1,37 @@
...
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// clang-format on
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+#include "cxx-qt-lib/qset.h"
+
+#include "assertion_utils.h"
+
+#define CXX_QT_QSET_ASSERTS(typeName, name)                                    \
```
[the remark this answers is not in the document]
ahayzen-kdab: :tada:

tests/qt_types_standalone/rust/src/qset.rs
ahayzen-kdab: Done

ahayzen-kdab:
> It would help to incorporate this into one of the examples to see how the API works in practice.

Added a `containers` section to the `qml_features` so we can have QSet/QHash/QVector etc all here later.

Also note that you can see how these are used in qt_types_standalone as well.

LeonMatthesKDAB (approved):
LGTM

crates/cxx-qt-lib/src/types/qset/generate.sh
ahayzen-kdab: Discussed with @Be-ing another option we explored was potentially using the `build.rs` to generate the files, but this would likely confuse IDEs, look odd to git, and could be tricky if it wrote to the OUT_DIR (how would you get the module path right?) or odd if it writes to the source.

For now we'll stick to this platform specific bash script helper, hopefully the `concat_ident!` stuff becomes stable and we can go that route eventually if we could get CXX to expand as well ...

Issue #328, opened 2022-10-19
Support all types that CXX supports for C++ generation side


ahayzen-kdab:
In the C++ generation we have to map Rust types to C++ types, here not all types have been mapped.

Some of the missing types

- [x] Pointers (eg `* const T`) needs to handle TypePtr (related #367 #299 )
- [x] Slice to rust::Slice #389
- [x] Function pointer to Fn #389
- [x] Array type #389
- [x] Result<T> ? Split into #404 for later as it's more complicated
- [x] Do we handle &str correctly? #389 
- [x] i64,u64 ( These may help as then we can use std types #317 #36 )
- [x] Pin<T> #353 
- [x] Any others :-) 

https://github.com/KDAB/cxx-qt/blob/main/crates/cxx-qt-gen/src/generator/cpp/types.rs

LeonMatthesKDAB:
Could we get this exposed by upstream?
E.g. could CXX somehow tell us how to transform this stuff...

ahayzen-kdab:
Maybe, but they have already parsed the types into their own enum via syn. We would have to expose some API to get from our syn types to the theirs etc.

ahayzen-kdab:
Could be worth a discussion upstream though if something like the `cxx-gen` crate could expose something like this. It would need to be able to go from a Rust `syn::Type` to a C++ `String`. We would also need to know if the type means that we need to mark the CXX bridge method as unsafe (eg `fn is_unsafe(ty: &syn::Type) -> bool`).

ahayzen-kdab:
autocxx also has similar things in their code

  * Registering of the types https://github.com/google/autocxx/blob/6cca9b29c1b2a64a6ee7d442f35d6fbfbc76cc42/engine/src/known_types.rs#L363
  * Type -> C++ conversion https://github.com/google/autocxx/blob/6cca9b29c1b2a64a6ee7d442f35d6fbfbc76cc42/engine/src/conversion/codegen_cpp/type_to_cpp.rs#L67 (note this goes from a `syn::Type` to a `String` so is the same as what we want.

ahayzen-kdab:
Split out `Result` as that is more complicated #404 .

PR #353, opened 2022-11-10
cxx-qt-gen: generate C++ types for Pin<T> correctly

ahayzen-kdab:
Related to #328 

This adds missing support in our generation for having `Pin<T>` as a type.


Be-ing (dismissed):

ahayzen-kdab:
Really would be nice if Github had a way to remove that update with merge commit option, as it's so easy to hit. And considering there is another checkbox that claims to prevent merge commits going into `main`, when it doesn't actually block that :shrug:

Issue #332, opened 2022-10-24
Explore if using Q_SLOT and Q_SIGNAL could simplify C++ generation


ahayzen-kdab:
At the moment we have `slots` and `signals` in `GeneratedCppQObjectBlocks` as we put each of these into blocks in the generated C++ code, eg

```cpp
public Q_SLOTS:
  ...

public Q_SIGNALS:
  ...
```

Explore if moving to just tagging each method with `Q_SLOT` and `Q_SIGNAL` (like `Q_INVOKABLE`) would simplfy the code as we wouldn't need these separate lists and they could all go into the `methods` list.

eg
```cpp
public:
   Q_INVOKABLE void invokable();
   Q_SLOT void mySlot();
   Q_SIGNAL void mySignal();
```

Would there be any disadvantage to doing this ?

https://doc.qt.io/qt-6/qobject.html#Q_SIGNAL
https://doc.qt.io/qt-6/qobject.html#Q_SLOT

ahayzen-kdab:
Solved by #356

PR #356, opened 2022-11-15
#332 add `Q_SIGNAL` and `Q_SLOT` to methods definitions

przempore:
## Issue
#332

Using `Q_SIGNAS` and `Q_SLOT` to tag `QObject` methods.

## Question:
What's your opinion about having those methods grouped together?
e.g.
1
```
    public:
      int count() const;
      bool toggle() const;
      Q_INVOKABLE void invokable();
      void cppMethod();

    public:
      Q_SLOT void setCount(int count);
      Q_SLOT void setToggle(bool toggle);

    public:
      Q_SIGNAL void countChanged();
      Q_SIGNAL void toggleChanged();
```
2
or having one group like `Q_INVOKABLE` is expected:
```
    public:
      int count() const;
      bool toggle() const;
      Q_INVOKABLE void invokable();
      void cppMethod();
      Q_SLOT void setCount(int count);
      Q_SLOT void setToggle(bool toggle);
      Q_SIGNAL void countChanged();
      Q_SIGNAL void toggleChanged();
```

This PR provides 1 solution.

### DIFF
--- a/crates/cxx-qt-build/src/lib.rs
+++ b/crates/cxx-qt-build/src/lib.rs
@@ -15,7 +15,7 @@ use std::{
 };
 
 use cxx_qt_gen::{
-    parse_qt_file, write_cpp, write_rust, CppFragmentPair, CxxQtItem, GeneratedCppBlocks,
+    parse_qt_file, write_cpp, write_rust, CppFragment, CxxQtItem, GeneratedCppBlocks,
     GeneratedRustBlocks, Parser,
 };
 
@@ -34,7 +34,7 @@ struct GeneratedCppFilePaths {
 }
 
 struct GeneratedCpp {
-    cxx_qt: Option<CppFragmentPair>,
+    cxx_qt: Option<CppFragment>,
     cxx: cxx_gen::GeneratedCode,
     file_ident: String,
 }
@@ -140,8 +140,13 @@ impl GeneratedCpp {
             ));
             let mut header =
                 File::create(&header_path).expect("Could not create cxx-qt header file");
+            let header_generated = match cxx_qt_generated {
+                CppFragment::Pair { header, source: _ } => header,
+                CppFragment::Header(header) => header,
+                CppFragment::Source(_) => panic!("Unexpected call for source fragment."),
+            };
             header
-                .write_all(cxx_qt_generated.header.as_bytes())
+                .write_all(header_generated.as_bytes())
                 .expect("Could not write cxx-qt header file");
             cpp_file_paths.qobject_header = Some(header_path);
 
@@ -151,7 +156,12 @@ impl GeneratedCpp {
                 self.file_ident
             ));
             let mut cpp = File::create(&cpp_path).expect("Could not create cxx-qt source file");
-            cpp.write_all(cxx_qt_generated.source.as_bytes())
+            let source_generated = match cxx_qt_generated {
+                CppFragment::Pair { header: _, source } => source,
+                CppFragment::Header(_) => panic!("Unexpected call for header fragment."),
+                CppFragment::Source(source) => source,
+            };
+            cpp.write_all(source_generated.as_bytes())
                 .expect("Could not write cxx-qt source file");
             cpp_file_paths.qobject = Some(cpp_path);
         }
@@ -161,7 +171,7 @@ impl GeneratedCpp {
             header_directory.display(),
             self.file_ident
         ));
-        let mut header = File::create(&header_path).expect("Could not create cxx header file");
+        let mut header = File::create(header_path).expect("Could not create cxx header file");
         header
             .write_all(&self.cxx.header)
             .expect("Could not write cxx header file");
@@ -186,15 +196,15 @@ fn generate_cxxqt_cpp_files(
     header_dir: impl AsRef<Path>,
 ) -> Vec<GeneratedCppFilePaths> {
     let manifest_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
-    let cpp_directory = format!("{}/cxx-qt-gen/src", env::var("OUT_DIR").unwrap());
 
     let mut generated_file_paths: Vec<GeneratedCppFilePaths> = Vec::with_capacity(rs_source.len());
     for rs_path in rs_source {
+        let cpp_directory = format!("{}/cxx-qt-gen/src", env::var("OUT_DIR").unwrap());
         let path = format!("{}/{}", manifest_dir, rs_path.display());
         println!("cargo:rerun-if-changed={}", path);
 
         let generated_code = GeneratedCpp::new(&path);
-        generated_file_paths.push(generated_code.write_to_directories(&cpp_directory, &header_dir));
+        generated_file_paths.push(generated_code.write_to_directories(cpp_directory, &header_dir));
     }
 
     generated_file_paths
@@ -357,21 +367,21 @@ impl CxxQtBuilder {
         self.cc_builder.include(&header_root);
         let generated_header_dir = format!("{}/cxx-qt-gen", header_root);
 
-        cxx_qt_lib_headers::write_headers(&format!("{}/cxx-qt-lib", header_root));
+        cxx_qt_lib_headers::write_headers(format!("{}/cxx-qt-lib", header_root));
 
         // Write cxx header
-        std::fs::create_dir_all(&format!("{}/rust", header_root))
+        std::fs::create_dir_all(format!("{}/rust", header_root))
             .expect("Could not create cxx header directory");
         let h_path = format!("{}/rust/cxx.h", header_root);
         // Wrap the File in a block scope so the file is closed before the compiler is run.
         // Otherwise MSVC fails to open cxx.h because the process for this build script already has it open.
         {
-            let mut header = File::create(&h_path).expect("Could not create cxx.h");
+            let mut header = File::create(h_path).expect("Could not create cxx.h");
             write!(header, "{}", cxx_gen::HEADER).expect("Could not write cxx.h");
         }
 
         // Generate files
-        for files in generate_cxxqt_cpp_files(&self.rust_sources, &generated_header_dir) {
+        for files in generate_cxxqt_cpp_files(&self.rust_sources, generated_header_dir) {
             self.cc_builder.file(files.plain_cpp);
             if let (Some(qobject), Some(qobject_header)) = (files.qobject, files.qobject_header) {
                 self.cc_builder.file(&qobject);
--- a/crates/cxx-qt-gen/src/generator/cpp/signal.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/signal.rs
@@ -6,7 +6,7 @@
 use crate::{
     generator::{
         cpp::{
-            fragment::CppFragmentPair, qobject::GeneratedCppQObjectBlocks, types::CppType,
+            fragment::CppFragment, qobject::GeneratedCppQObjectBlocks, types::CppType,
             CXX_QT_CONVERT,
         },
         naming::{qobject::QObjectName, signals::QSignalName},
@@ -57,14 +57,14 @@ pub fn generate_cpp_signals(
         let signal_ident = idents.name.cpp.to_string();
 
         // Generate the Q_SIGNAL
-        generated.signals.push(format!(
-            "void {ident}({parameters});",
+        generated.methods.push(CppFragment::Header(format!(
+            "Q_SIGNAL void {ident}({parameters});",
             ident = signal_ident,
             parameters = parameter_types.join(", "),
-        ));
+        )));
 
         // Generate the emitters
-        generated.methods.push(CppFragmentPair {
+        generated.methods.push(CppFragment::Pair {
             header: format!(
                 "void {ident}({parameters});",
                 ident = emit_ident,
@@ -122,19 +122,28 @@ mod tests {
 
         let generated = generate_cpp_signals(&signals, &qobject_idents).unwrap();
 
-        assert_eq!(generated.signals.len(), 1);
+        assert_eq!(generated.methods.len(), 2);
+        let header = if let CppFragment::Header(header) = &generated.methods[0] {
+            header
+        } else {
+            panic!("Expected header")
+        };
         assert_str_eq!(
-            generated.signals[0],
-            "void dataChanged(qint32 trivial, QColor opaque);"
+            header,
+            "Q_SIGNAL void dataChanged(qint32 trivial, QColor opaque);"
         );
 
-        assert_eq!(generated.methods.len(), 1);
+        let (header, source) = if let CppFragment::Pair { header, source } = &generated.methods[1] {
+            (header, source)
+        } else {
+            panic!("Expected Pair")
+        };
         assert_str_eq!(
-            generated.methods[0].header,
+            header,
             "void emitDataChanged(qint32 trivial, ::std::unique_ptr<QColor> opaque);"
         );
         assert_str_eq!(
-            generated.methods[0].source,
+            source,
             indoc! {r#"
             void
             MyObject::emitDataChanged(qint32 trivial, ::std::unique_ptr<QColor> opaque)
--- a/crates/cxx-qt-gen/src/writer/cpp/header.rs
+++ b/crates/cxx-qt-gen/src/writer/cpp/header.rs
@@ -3,13 +3,17 @@
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
-use crate::generator::cpp::{fragment::CppFragmentPair, GeneratedCppBlocks};
+use crate::generator::cpp::{fragment::CppFragment, GeneratedCppBlocks};
 use crate::writer::cpp::namespace_pair;
 use indoc::formatdoc;
 
-/// Extract the header from a given CppFragmentPair
-fn pair_as_header(pair: &CppFragmentPair) -> &str {
-    &pair.header
+/// Extract the header from a given CppFragment
+fn pair_as_header(pair: &CppFragment) -> Option<&str> {
+    match pair {
+        CppFragment::Pair { header, source: _ } => Some(header),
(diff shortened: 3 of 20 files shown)


Be-ing:
Hello and welcome. I don't particularly mind either way because this only affects generated code which users shouldn't need to be looking at... so I'm curious, why are you interested in changing this?

przempore:
Hello! 

End user won't be affected, just a question about the tast ;)
>why are you interested in changing this?
wanted to maybe help a bit with the project, found `good first issue` thought I may start with it?

Be-ing:
>  found good first issue thought I may start with it?

Ah, I see @ahayzen-kdab suggested this in #332. I think the intention was to also get rid of the separate `public` blocks so all the methods are listed together? @ahayzen-kdab can you confirm that?

ahayzen-kdab:
Hey, yes the intention in #332 was to potentially change it to the following, i don't believe there would be any downside from a Qt point of view of doing this (?)
```cpp
    public:
      int count() const;
      bool toggle() const;
      Q_INVOKABLE void invokable();
      void cppMethod();
      Q_SLOT void setCount(int count);
      Q_SLOT void setToggle(bool toggle);
      Q_SIGNAL void countChanged();
      Q_SIGNAL void toggleChanged();
```

As this also then means in the Rust code for `GeneratedCppQObjectBlocks`, the `slots` and `signals` fields would be removed and all of them would be groups into the `methods` field.
This would then result in less code in the generator/writer phase and in the tests etc :-)

crates/cxx-qt-gen/src/generator/cpp/signal.rs
```diff
@@ -58,7 +58,7 @@ pub fn generate_cpp_signals(
 
         // Generate the Q_SIGNAL
         generated.signals.push(format!(
```
ahayzen-kdab, line 60: so, eg, i would expect this to just add to the `generated.methods`.

crates/cxx-qt-gen/src/writer/cpp/header.rs
```diff
@@ -104,8 +104,8 @@ fn qobjects_header(generated: &GeneratedCppBlocks) -> Vec<String> {
         base_class = qobject.base_class,
         metaobjects = qobject.blocks.metaobjects.join("\n  "),
         methods = create_block("public", &qobject.blocks.methods.iter().map(pair_as_header).collect::<Vec<&str>>()),
-        slots = create_block("public Q_SLOTS", &qobject.blocks.slots.iter().map(pair_as_header).collect::<Vec<&str>>()),
-        signals = create_block("Q_SIGNALS", &qobject.blocks.signals.iter().map(AsRef::as_ref).collect::<Vec<&str>>()),
+        slots = create_block("public", &qobject.blocks.slots.iter().map(pair_as_header).collect::<Vec<&str>>()),
```
ahayzen-kdab, line 107: and, eg, these would be removed as it would come from the `methods` line above etc

ahayzen-kdab:
Let me know if you have more questions, it should simplify the code in quite a few places, I can check it out tomorrow :-) Note that CI is failing due to Rust and C++ formatting ;-)

przempore:
Great! Thx! So I have another question, as `Q_SIGNAL`s [have no body](https://github.com/KDAB/cxx-qt/blob/173f936bb1d1b626da5cb7564c809d3db96d7637/crates/cxx-qt-gen/src/generator/cpp/property/signal.rs#L8), and methods in `GeneratedCppQObjectBlocks` expects `CppFragmentPair` do you think making `source` an `Option` would be a good idea?
``` rust
pub struct CppFragmentPair {
    pub header: String,
    pub source: Option<String>,
}
```
or better would be introducing a new field in `GeneratedCppQObjectBlocks` something like `inlines`(?) where it would be possible to put inline methods (if ever will be wanted) or explicitly `signals` but generate them directly below methods in [`qobjects_header` ](https://github.com/KDAB/cxx-qt/blob/173f936bb1d1b626da5cb7564c809d3db96d7637/crates/cxx-qt-gen/src/writer/cpp/header.rs#L81) which gives
``` rust
fn qobjects_header(generated: &GeneratedCppBlocks) -> Vec<String> {
   // ...
    generated.qobjects.iter().map(|qobject| {
        // ...
        {methods}
        {signals}
```

The best would be to have everything in `methods`, because _why there are `signals` but no `slots`?_

ahayzen-kdab:
@przempore Hmm so the options seem to be

- Put an empty string in the source
- Add Option<String> to the source
- Have a separate field like `inlines` which is a Vec<String>
- Something completely different by using an enum

I think calling them signals as you say would be odd.

Calling them `inlines` is interesting and as you say maybe useful in the future, but then these signals are not actually inline methods they are just header declarations so that would be the "wrong" name anyway.

This leaves using an empty string or Option<String>, empty strings could cause fun with our tests with extra empty lines.

The Option<String> route looks possible there is a `pair_as_source` method in the writer that would need to return an Option<String> as well then the map / collect part would need to filter out any none items etc.

Another option could be to use an enum

```rust
enum CppFragment {
  Pair { header: String, source: String },
  Header(String),
  Source(String),  // <- probably don't need this one right now, but could be added later
}
```

This would then mean most of the existing places would change to just `CppFragment::Pair { ... }`. Then the writer would need to just change from `fn pair_as_source(pair: &CppFragmentPair) -> String {` to `fn pair_as_source(pair: &CppFragment) -> Option<String> {` and the map that uses it filter out none's etc.

How does this sound? I think the enum might be the best long term and probably wouldn't be too hard to change right now ?

Feel free to do this in a separate commit in the branch if you want to split the two stages or do them together in one commit :-)

LeonMatthes:
Random chime into this.

The concept of "inline" methods (i.e. with definition in header) might be needed in future anyway, as we may jeed to generate templated methods to solve some of our inheritancee issues.

So +1 for that feature.

ahayzen:
> I'm sorry for spamming with commits. Do you have a policy for pull requests? All commits can be later squashed 🙄

The amount of commits doesn't generally matter, they can be squashed or merged as is if they have meaning.

But we don't normally accept merge commits, so use rebase etc to ensure there are none in your chain :-)

Be-ing:
The cxx-qt-gen tests are failing to build.

ahayzen-kdab:
You probably already know this ... but FYI if you do a `git rebase origin/main` (or `git rebase -i origin/main` for an interactive one) this currently still removes the merge commit cleanly from your chain. And if we land more things into `main` you'll need to do this to ensure this branch is up to date instead of using a merge :-)

crates/cxx-qt-build/src/lib.rs
```diff
@@ -186,15 +196,15 @@ fn generate_cxxqt_cpp_files(
     header_dir: impl AsRef<Path>,
 ) -> Vec<GeneratedCppFilePaths> {
     let manifest_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
-    let cpp_directory = format!("{}/cxx-qt-gen/src", env::var("OUT_DIR").unwrap());
 
     let mut generated_file_paths: Vec<GeneratedCppFilePaths> = Vec::with_capacity(rs_source.len());
     for rs_path in rs_source {
+        let cpp_directory = format!("{}/cxx-qt-gen/src", env::var("OUT_DIR").unwrap());
```
ahayzen-kdab, line 202: what was the error that meant this needed to be moved?

ahayzen-kdab: Maybe when changing from reference to value for the others this caused this one to fail as it's used multiple times.

So then is this one better being a reference below so that it can be built once outside the loop ?

crates/cxx-qt-build/src/lib.rs
```diff
@@ -186,15 +196,15 @@ fn generate_cxxqt_cpp_files(
...
         let path = format!("{}/{}", manifest_dir, rs_path.display());
         println!("cargo:rerun-if-changed={}", path);
 
         let generated_code = GeneratedCpp::new(&path);
-        generated_file_paths.push(generated_code.write_to_directories(&cpp_directory, &header_dir));
+        generated_file_paths.push(generated_code.write_to_directories(cpp_directory, &header_dir));
```
ahayzen-kdab, line 207: eg maybe keep this one as a reference and move the `cpp_directory` out of the loop

crates/cxx-qt-build/src/lib.rs
przempore: > what was the error that meant this needed to be moved?

```
   Compiling cxx-qt-build v0.4.1 (/home/przemek/Projects/rust/cxx-qt/crates/cxx-qt-build)
error: the borrowed expression implements the required traits
   --> crates/cxx-qt-build/src/lib.rs:207:71
    |
207 |         generated_file_paths.push(generated_code.write_to_directories(&cpp_directory, &header_dir));
    |                                                                       ^^^^^^^^^^^^^^ help: change this to: `cpp_directory`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow
    = note: `-D clippy::needless-borrow` implied by `-D warnings`

error: could not compile `cxx-qt-build` due to previous error
```

but removal of `&` forced me to move the line inside the loop.
`AsRef::<Path>::as_ref(&cpp_directory)` helped. Is that pretty enough?
I guess using only `&` was ambiguous? Or?

ahayzen-kdab: Hmm I just thought that the following would make more sense, as it constructs the cpp_directory once rather than for every rs_path then hands it a reference. But if it really doesn't work then guess we have to build it in each loop. And i wondered why clippy didn't complain about this before with the old code.

```rust
let cpp_directory = format!("{}/cxx-qt-gen/src", env::var("OUT_DIR").unwrap());

for rs_path in rs_source {
   ...
   generated_file_paths.push(generated_code.write_to_directories(&cpp_directory, &header_dir));
}
```

ahayzen-kdab: Maybe creating it inside the loop is better and passing by value, i was just more curious what was happening :-)

I guess that the borrow is dereferenced straight away anyway which then causes the copy ? So it thinks it is no worse to have it inside the loop and be by value ?

In that case probably just move it back to doing cpp_directory inside the loop like you had with by-value, sorry.

ahayzen-kdab:
Then while reverting that last commit, if you rebase against main to remove the merge commit this is probably ready for review :-)

przempore:
> Then while reverting that last commit, if you rebase against main to remove the merge commit this is probably ready for review :-)

```
❯ git pull
warning: skipped previously applied commit 409c1cf
warning: skipped previously applied commit f6320cf
warning: skipped previously applied commit 8fe0ca7
warning: skipped previously applied commit 39c4b1a
warning: skipped previously applied commit 438d924
warning: skipped previously applied commit 7385267
warning: skipped previously applied commit 000d729
warning: skipped previously applied commit 595b617
warning: skipped previously applied commit 796b497
warning: skipped previously applied commit eec92ea
warning: skipped previously applied commit fb0a342
warning: skipped previously applied commit 0d596bf
warning: skipped previously applied commit 5aedfa5
warning: skipped previously applied commit 58c52d2
hint: use --reapply-cherry-picks to include skipped commits
hint: Disable this message with "git config advice.skippedCherryPicks false"
Successfully rebased and updated refs/heads/signal_slot_method_tagging.

cxx-qt on  signal_slot_method_tagging [$?] is 📦 v0.4.1 via △ v3.25.0 via 🦀 v1.67.0-nightly
❯ git push origin
Everything up-to-date
```

crates/cxx-qt-build/src/lib.rs
przempore: > Hmm I just thought that the following would make more sense, as it constructs the cpp_directory once rather than for every rs_path then hands it a reference. But if it really doesn't work then guess we have to build it in each loop. And i wondered why clippy didn't complain about this before with the old code.

That's also makes more sense to me, as the variable will be the same always. Does Rust compiler optimize that such of things?

ahayzen-kdab:
> > Then while reverting that last commit, if you rebase against main to remove the merge commit this is probably ready for review :-)
> 
> ```
> ❯ git pull
> warning: skipped previously applied commit 409c1cf
> warning: skipped previously applied commit f6320cf
> warning: skipped previously applied commit 8fe0ca7
> warning: skipped previously applied commit 39c4b1a
> warning: skipped previously applied commit 438d924
> warning: skipped previously applied commit 7385267
> warning: skipped previously applied commit 000d729
> warning: skipped previously applied commit 595b617
> warning: skipped previously applied commit 796b497
> warning: skipped previously applied commit eec92ea
> warning: skipped previously applied commit fb0a342
> warning: skipped previously applied commit 0d596bf
> warning: skipped previously applied commit 5aedfa5
> warning: skipped previously applied commit 58c52d2
> hint: use --reapply-cherry-picks to include skipped commits
> hint: Disable this message with "git config advice.skippedCherryPicks false"
> Successfully rebased and updated refs/heads/signal_slot_method_tagging.
> 
> cxx-qt on  signal_slot_method_tagging [$?] is 📦 v0.4.1 via △ v3.25.0 via 🦀 v1.67.0-nightly
> ❯ git push origin
> Everything up-to-date
> ```

So I wouldn't use `git pull` it'll create these odd commit chains with merge commits etc. Instead try to do the following...

- Ensure you have a remote pointing to the upstream (`git remote add upstream git@github.com:KDAB/cxx-qt.git`)
- Fetch the latest code in the remote (`git fetch upstream`)
- Now start interactively rebasing your commits onto the latest upstream commit (`git rebase -i upstream/main`)
- This will open in your normal editor (eg nano/vim) a view that lists each commit on a single line, ones with `pick` will be selected, you can choose `fixup` or `squash` to combine commits with the one above them, or `drop` will remove the commit. Note that it will also remove your merge commits
- When you save and close the editor git will then try to replay each commit, if there are conflicts you may need to resolve them and then do what `git status` says, usually resolve the conflict, some `git add`'s then a `git rebase --continue` (also if everything goes horribly wrong in the rebase you can do `git rebase --abort`)
- Then finally you need to push the changes back to your branch and overwrite the existing branch so `git push --force-with-lease`

There are probably also guides somewhere that can help you through doing an interactive rebase in git.

Otherwise if you can't get this to work, instead ensure that you have enabled "Allow edits and access to secrets by maintainers" on the pull request and one of us can write to your branch to do it assuming you are happy. I'm on holiday tomorrow but will check back on Thursday :-)

przempore:
> > > Then while reverting that last commit, if you rebase against main to remove the merge commit this is probably ready for review :-)
> > 
> > 
> > ```
> > ❯ git pull
> > warning: skipped previously applied commit 409c1cf
> > warning: skipped previously applied commit f6320cf
> > warning: skipped previously applied commit 8fe0ca7
> > warning: skipped previously applied commit 39c4b1a
> > warning: skipped previously applied commit 438d924
> > warning: skipped previously applied commit 7385267
> > warning: skipped previously applied commit 000d729
> > warning: skipped previously applied commit 595b617
> > warning: skipped previously applied commit 796b497
> > warning: skipped previously applied commit eec92ea
> > warning: skipped previously applied commit fb0a342
> > warning: skipped previously applied commit 0d596bf
> > warning: skipped previously applied commit 5aedfa5
> > warning: skipped previously applied commit 58c52d2
> > hint: use --reapply-cherry-picks to include skipped commits
> > hint: Disable this message with "git config advice.skippedCherryPicks false"
> > Successfully rebased and updated refs/heads/signal_slot_method_tagging.
> > 
> > cxx-qt on  signal_slot_method_tagging [$?] is 📦 v0.4.1 via △ v3.25.0 via 🦀 v1.67.0-nightly
> > ❯ git push origin
> > Everything up-to-date
> > ```
> 
> So I wouldn't use `git pull` it'll create these odd commit chains with merge commits etc. Instead try to do the following...
> 
>     * Ensure you have a remote pointing to the upstream (`git remote add upstream git@github.com:KDAB/cxx-qt.git`)
> 
>     * Fetch the latest code in the remote (`git fetch upstream`)
> 
>     * Now start interactively rebasing your commits onto the latest upstream commit (`git rebase -i upstream/main`)
> 
>     * This will open in your normal editor (eg nano/vim) a view that lists each commit on a single line, ones with `pick` will be selected, you can choose `fixup` or `squash` to combine commits with the one above them, or `drop` will remove the commit. Note that it will also remove your merge commits
> 
>     * When you save and close the editor git will then try to replay each commit, if there are conflicts you may need to resolve them and then do what `git status` says, usually resolve the conflict, some `git add`'s then a `git rebase --continue` (also if everything goes horribly wrong in the rebase you can do `git rebase --abort`)
> 
>     * Then finally you need to push the changes back to your branch and overwrite the existing branch so `git push --force-with-lease`
> 
> 
> There are probably also guides somewhere that can help you through doing an interactive rebase in git.
> 
> Otherwise if you can't get this to work, instead ensure that you have enabled "Allow edits and access to secrets by maintainers" on the pull request and one of us can write to your branch to do it assuming you are happy. I'm on holiday tomorrow but will check back on Thursday :-)

Thank you very much for so extensive description! I've _focefully_ pushed after rebasing but it looks like there is no changes between that. Hope you had good day off ;)

ahayzen-kdab:
> 
> Thank you very much for so extensive description! I've _focefully_ pushed after rebasing but it looks like there is no changes between that. Hope you had good day off ;)

Right you've done it correctly, so because there are no new commits in `main` there is no change in terms of the diff of code when you have force pushed. But if you look at the commits that github presents we have gone from multiple commits with a merge commit in the middle, to a much cleaner single commit :-) (this is now what we want so that we can merge it into `main` :-) ).

Note that if new commits were added to the `main` branch by another pull request you should follow the same procedure again, so fetch upstream, rebase ontop of main, resolve any conflicts, and force push. Also note that if you wanted to modify your own changes you can use `git commit --amend` to modify the commit and then force push. This all keeps the commit history nice and clean :-)

ahayzen-kdab (approved):
LGTM! Lets go with this any further changes can be made separately. Thanks for the contribution @przempore :-)

Issue #224, opened 2022-08-26
rpath is not set for Qt library built in cargo_without_cmake 


ahayzen-kdab:
When using the `cargo_without_cmake` example if `qmake` is in your `PATH` but the libraries are not in normal LD search locations. Then CMake works by using rpath's to point to the Qt libraries (?).

This is not possible with `cargo` yet due to https://github.com/rust-lang/cargo/issues/5077

For now the workaround is to add the Qt libraries path to your `LD_LIBRARY_PATH` environment variable.

Be-ing:
Maybe rustc already supports this? https://doc.rust-lang.org/rustc/codegen-options/index.html#rpath

PR #358, opened 2022-11-17
support linking Qt as macOS framework

Be-ing:
The CI changes are only for testing this branch; I'll remove those before merging so we don't have 2 more CI jobs to wait on.

### DIFF
--- a/crates/qt-build-utils/src/lib.rs
+++ b/crates/qt-build-utils/src/lib.rs
@@ -233,7 +233,8 @@ impl QtBuild {
         let lib_path = self.qmake_query("QT_INSTALL_LIBS");
         println!("cargo:rustc-link-search={}", lib_path);
 
-        let prefix = match env::var("TARGET") {
+        let target = env::var("TARGET");
+        let prefix = match &target {
             Ok(target) => {
                 if target.contains("msvc") {
                     ""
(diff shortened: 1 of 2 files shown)


ahayzen-kdab:
> The CI changes are only for testing this branch; I'll remove those before merging so we don't have 2 more CI jobs to wait on.

How about we switch macOS to use homebrew and the frameworks stuff ?

Then vcpkg is still tested with Windows builds, but it increases our testing coverage without additional CI time?

We then are able to test
- Qt 5 vs Qt 6
- gcc vs clang vs msvc
- linux packages (ubuntu debs) vs vcpkg vs macOS homebrew frameworks

Be-ing:
vcpkg on macOS is our only test for linking Qt statically, so I want to keep that.

ahayzen-kdab:
> vcpkg on macOS is our only test for linking Qt statically, so I want to keep that.

ah good point, Windows is still dynamic with vcpkg ? So maybe it would be another running we could have ...

So maybe in the future we might want a MinGW and a homebrew runner. :thinking:

Be-ing:
> Windows is still dynamic with vcpkg ?

For now, yes, though I do want to change this to deal with #224.

ahayzen-kdab:
> > Windows is still dynamic with vcpkg ?
> 
> For now, yes, though I do want to change that to deal with #224.

Right, so if Windows moved to static vcpkg and homebrew works. Then we could potentially move the macOS runners over.

crates/qt-build-utils/src/lib.rs
```diff
@@ -233,7 +233,8 @@ impl QtBuild {
         let lib_path = self.qmake_query("QT_INSTALL_LIBS");
         println!("cargo:rustc-link-search={}", lib_path);
 
-        let prefix = match env::var("TARGET") {
+        let target = env::var("TARGET");
+        let prefix = match &target {
```
ahayzen-kdab, line 237: make sure you add a "fixed" changelog entry after the `### Added` section

```
### Fixed

- Fixed using Qt with macOS frameworks from homebrew
```

crates/qt-build-utils/src/lib.rs
```diff
@@ -245,11 +246,37 @@ impl QtBuild {
...
+                    }
+                }
+                Err(_) => false,
+            };
+
+            let (link_lib, prl_path) = if framework {
+                (
+                    format!("framework=Qt{}", qt_module),
+                    format!(
+                        "{}/Qt{}.framework/Resources/Qt{}.prl",
```
ahayzen-kdab, line 264: so in the Qt code it searches the Resources folder first then searches the framework folder for backwards compatibility. This would be for < 5.14 support.

As we only support 5.15+ we can ignore this I guess but something to note and could be fixed separately in qt-build-utils.

ahayzen-kdab (approved):
LGTM :-)

PR #361, opened 2022-11-18
Fix qml minimal statically linked

ahayzen-kdab:


examples/qml_minimal/CMakeLists.txt
```diff
@@ -85,6 +85,9 @@ add_executable(${APP_NAME} cpp/main.cpp qml/qml.qrc)
 target_link_libraries(${APP_NAME} PRIVATE ${CRATE})
 # ANCHOR_END: book_cmake_executable
 
+# If we are using a statically linked Qt then we need to import any qml plugins
```
ahayzen-kdab, line 88: should probably mention this in the book as well ....

ahayzen-kdab: Added to the linking part of the example, the code comment is enough to explain it.

Be-ing (approved):

PR #362, opened 2022-11-23
cargo_without_cmake: Fix '-fuse-ld=/usr/bin/mold' error

LeonMatthesKDAB:
For some reason upgrading corrosion in commit
224baad9c59e085bcc0b24d82be126073ebef1a0 broke the build for me with the linker error:
```
note: c++: error: unrecognized command-line option '-fuse-ld=/usr/bin/mold'
```

Something about setting up the cargo-only project using corrosion is causing issues.
This was fixed by simply using a custom target.

### DIFF
--- a/examples/cargo_without_cmake/CMakeLists.txt
+++ b/examples/cargo_without_cmake/CMakeLists.txt
@@ -19,20 +19,6 @@ if(NOT Qt6_FOUND)
 endif()
 get_target_property(QMAKE Qt::qmake IMPORTED_LOCATION)
 
-# Use Corrosion to create a target for the crate that can be built without CMake.
-# This allows it to share built dependencies with the other crates in the workspace.
-find_package(Corrosion QUIET)
-if(NOT Corrosion_FOUND)
-    include(FetchContent)
-    FetchContent_Declare(
-        Corrosion
-        GIT_REPOSITORY https://github.com/corrosion-rs/corrosion.git
-        GIT_TAG v0.3.0
-    )
-
-    FetchContent_MakeAvailable(Corrosion)
-endif()
-
-set(CRATE qml-minimal-no-cmake)
-corrosion_import_crate(MANIFEST_PATH Cargo.toml CRATES ${CRATE})
-corrosion_set_env_vars(${CRATE} "QMAKE=${QMAKE}")
+add_custom_target(qml-minimal-no-cmake ALL
+    cargo build -p qml-minimal-no-cmake
+    COMMENT "Building qml-minimal-no-cmake using cargo")


examples/cargo_without_cmake/CMakeLists.txt
```diff
@@ -19,20 +19,6 @@ if(NOT Qt6_FOUND)
...
-    )
-
-    FetchContent_MakeAvailable(Corrosion)
-endif()
-
-set(CRATE qml-minimal-no-cmake)
-corrosion_import_crate(MANIFEST_PATH Cargo.toml CRATES ${CRATE})
-corrosion_set_env_vars(${CRATE} "QMAKE=${QMAKE}")
+add_custom_target(qml-minimal-no-cmake ALL
+    cargo build -p qml-minimal-no-cmake
```
ahayzen-kdab, line 23: right as we discussed this needs QMAKE env var set to find Qt

Be-ing (changes requested):
Let's investigate the issue with Corrosion instead of getting rid of it. Simply using `add_custom_target` will cause a rebuild of all dependencies that isn't shared with the other examples.

LeonMatthesKDAB:
> Let's investigate the issue with Corrosion instead of getting rid of it. Simply using `add_custom_target` will cause a rebuild of all dependencies that isn't shared with the other examples.

Hm, you're right, it does build cxx-qt-lib twice :/
I'll try bisecting corrosion to figure out where the problem originates

LeonMatthesKDAB:
@Be-ing This seems to be the commit that caused it:
https://github.com/corrosion-rs/corrosion/commit/b0b16b989a4359b7a3f578f142ff2d58c675081d

Do you have any idea why this would cause such an error :thinking: 
We're sort off out of ideas atm.

Be-ing:
Hmm, the main branch is working on my machine and on CI... so what's different on your machine?

LeonMatthesKDAB:
I'm guessing CMake is picking mold as my default linker (I have mold installed through dnf on Fedora), so it probably installed itself as default.
But I don't know what else is different otherwise.

LeonMatthesKDAB:
So I've narrowed it down to line `356` which has changed from:
``` cmake
corrosion_add_target_rustflags("${target_name}" "-Cdefault-linker-libraries=yes")
```
to:
``` cmake
corrosion_add_target_local_rustflags("${target_name}" "-Cdefault-linker-libraries=yes")
```
It seems that telling rustc to link to system libraries for downstream projects will fix the issue.
So most likely there's some downstream library that is depending on linking to system libraries for some reason.
But I have now idea how this is even connected to the original error message  of `-fuse-ld=/usr/bin/mold` :/

Be-ing:
> note: c++: error: unrecognized command-line option '-fuse-ld=/usr/bin/mold'

c++ is an alias for gcc, which does not accept paths to the linker executable, only a predefined allowlist. I was able to reproduce this by using GCC with CMake and adding this to ~/.cargo/config.toml

```
rustflags = ["-C", "link-arg=-fuse-ld=/usr/bin/mold"]
```

You can fix this by changing it to

```
rustflags = ["-C", "link-arg=-fuse-ld=mold"]
```

or telling CMake to use Clang instead of GCC by setting the environment variables `CC=clang CXX=clang++`.

This isn't an issue if you don't have anything in ~/.cargo/config.toml because [build.rs does not use a path to the linker](https://github.com/KDAB/cxx-qt/blob/73f7da527420214c49035f1cc653ac7d8668a0d7/examples/cargo_without_cmake/build.rs#L64).

LeonMatthesKDAB:
> c++ is an alias for gcc, which does not accept paths to the linker executable, only a predefined allowlist. I was able to reproduce this by using GCC with CMake and adding this to ~/.cargo/config.toml

> rustflags = ["-C", "link-arg=-fuse-ld=/usr/bin/mold"]

> You can fix this by changing it to

> rustflags = ["-C", "link-arg=-fuse-ld=mold"]

Oh my god, thank you! I figured out that `/user/bin/mold` isn't valid in that context, but I never expected it to come from `~/.cargo/config.toml`!
I'm surprised it worked up to this point as well :thinking: 

@Be-ing You're a build-system wizard :pray:

Issue #330, opened 2022-10-20
Consider how to support defining protected methods in the CXX bridge


ahayzen-kdab:
We need to use `friend` as shown in #329 then this needs some API to determine when to generate eg

```rust
#[cxx_qt::bridge]
mod ffi {
    unsafe extern "C++" {
        #[cxx_qt::protected]
        #[cxx_name = "beginResetModel"]
        fn begin_reset_model(self: Pin<&mut MyObjectQt>);
        #[cxx_qt::protected]
        #[cxx_name = "endResetModel"]
        fn end_reset_model(self: Pin<&mut MyObjectQt>);
    }
}
```

ahayzen-kdab:
Like #/328 this is another task that could be helped by having CXX API. As we need to generate `void cxxbridge1$CustomBaseClass$begin_reset_model(::CustomBaseClass& self) noexcept;` from `fn begin_reset_model(self: Pin<&mut MyObjectQt>);`.

If there was a CXX method to go from the Rust signature to the C++ signature this would help us here.

LeonMatthesKDAB:
## New API suggestion:
``` rust
#[cxx_qt::bridge]
pub mod ffi {

    #[cxx_qt::qobject(base = "QAbstractListModel")]
    pub struct MyObject {}

    impl qobject::MyObject {
        cxx_qt::protected! {
            #[cxx_name="beginResetModel"]
            fn begin_reset_model(self: Pin<&mut Self>);
        }

        #[invokable]
        pub fn my_fun(self: Pin<&mut Self>) {
            self.beginResetModel();
        }
    }
}
```
cxx_qt::protected! would be a new syntax, that needs to be within an `impl qobject::MyObject` block.
That way, we make it syntactically clear that methods defined here should only be available from within qobject::MyObject.
Also, we could assert that a correct `self` type exists, and that it only uses either:
- `self: Pin<&mut Self>`
- `&self` 

`cxx_qt::protected!` would then generate the appropriate `extern "C++"` block, as well as the required "friend" methods.
We should be careful here to, if possible, limit access to the generated `extern "C++"` methods, such that they can only be used from the `qobject` module, or ideally, only from `qobject::MyObject` itself.

ahayzen-kdab:
This looks interesting, can we also use it for public methods ? As this then removes the need for the developer to know about `TQt`.

Eg as discussed

```rust
impl qobject::T {
    cxx_qt::inherit! {
        #[protected] 
        fn x(&self)
[...]

PR #363, opened 2022-11-24
Implement cxx_qt::inherit! to allow inheriting methods from super class

LeonMatthesKDAB:
Initial implementation for #330.
Does not include the `friend` implementation for the wrapper method yet, so doesn't complete #330.


crates/cxx-qt-gen/test_inputs/inheritance.rs
```diff
@@ -0,0 +1,34 @@
...
+        data: Vec<i32>,
+    }
+
+    impl qobject::MyObject {
+        #[qinvokable(cxx_override)]
+        pub fn data(&self, _index: &QModelIndex, _role: i32) -> QVariant {
+            QVariant::default()
+        }
+
+        cxx_qt::inherit! {
```
ahayzen-kdab, line 22: the `passthrough_and_naming` test did have a custom base class already to test that. So you could reuse that one to test the inherits too to avoid creating more tests for us to maintain :-) (maybe it should be called something different)

crates/cxx-qt-gen/src/parser/invokable.rs
```diff
@@ -28,3 +33,55 @@ pub struct ParsedQInvokable {
     /// Any specifiers that declared on the invokable
     pub specifiers: HashSet<ParsedQInvokableSpecifiers>,
 }
+
+impl ParsedQInvokable {
```
ahayzen-kdab, line 37: this part has moved from `parse_impl_items` ? maybe split this into a separate commit so we can land this separately while the rest is being prepared, then this also reduces the noise in this commit.

crates/cxx-qt-gen/src/parser/parameter.rs
```diff
@@ -14,3 +14,48 @@ pub struct ParsedFunctionParameter {
     /// The name of the C++ type if one has been specified
     pub cxx_type: Option<String>,
 }
+
+impl ParsedFunctionParameter {
```
ahayzen-kdab, line 18: same here, can we split this into another commit then merge separately ?

crates/cxx-qt-gen/src/syntax/implitemmethod.rs
```diff
@@ -4,16 +4,16 @@
...
     TypePath, TypeReference,
 };
 
 /// Return whether the first parameter of a method is a mutable self argument
 //
 // Note that self: Box<Self> is parsed as FnArg::Typed not FnArg::Receiver so will be false
 // but we don't use this case with CXX, so this can be ignored.
-pub fn is_method_mutable(method: &ImplItemMethod) -> bool {
-    match method.sig.inputs.first() {
+pub fn is_method_mutable(signature: &Signature) -> bool {
```
ahayzen-kdab, line 15: also feels like all of this could be separated out as well if we are separating things :-)

crates/cxx-qt-gen/test_inputs/inheritance.rs
LeonMatthesKDAB: Right, I was hesitant to use that, as I'm not at all testing `passthrough_and_naming`.
I would rather remove the `#[qinvokable(cxx_inherit)]`, etc. tests from `passthrough_and_naming`, as that test-case seems to become quite large by now.

crates/cxx-qt-gen/src/generator/naming/invokable.rs
```diff
@@ -23,20 +23,12 @@ impl From<&ImplItemMethod> for QInvokableName {
     fn from(method: &ImplItemMethod) -> Self {
         let ident = &method.sig.ident;
         Self {
-            name: name_from_ident(ident),
+            name: ident.clone().into(),
```
ahayzen-kdab, line 26: wonder if going from `Ident` -> `CombinedIdent` like this is a good idea, as not all places may have a straight camel case conversion for the C++ side. What if we want to start prefixing all invokables or somewhere else uses this by accident then things change :thinking:

crates/cxx-qt-gen/src/parser/invokable.rs
LeonMatthesKDAB: Right, ideally I'd want to split a few of the refactorings I did into it's own PR. Will see if I can find the time before my QML training next week :thinking:

crates/cxx-qt-gen/src/generator/naming/invokable.rs
LeonMatthesKDAB: Right, I was somewhat unsure about this myself.
Can revert.
Might especially be problematic for class names, which don't need conversion :thinking:

crates/cxx-qt-gen/test_inputs/inheritance.rs
ahayzen-kdab: right passthrough_and_naming has ended up testing "everything else"

- namespaces
- file stem
- correctly ignoring/passthrough CXX blocks
- base class
- invokable and property names to camel case work etc
- two qobjects in the same mod don't collide

So maybe an inheritance one is OK and "base class" could be removed from passthrough and naming :sweat_smile:

crates/cxx-qt-gen/src/generator/naming/invokable.rs
ahayzen-kdab: I think it's better to be explicit here otherwise we could start getting lost again like the old code with one magic thing that affects unrelated places.

crates/cxx-qt-gen/test_inputs/inheritance.rs
LeonMatthesKDAB: I agree, we can remove that in a separate PR I'd say.

crates/cxx-qt-gen/src/parser/invokable.rs
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/parser/parameter.rs
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/generator/naming/invokable.rs
LeonMatthesKDAB: Moved into associated function instead.

crates/cxx-qt-gen/src/syntax/implitemmethod.rs
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/generator/cpp/inherit.rs
```diff
@@ -0,0 +1,48 @@
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Leon Matthes <leon.matthes@kdab.com>
+
+use indoc::formatdoc;
+
+// SPDX-License-Identifier: MIT OR Apache-2.0
```
ahayzen-kdab, line 6: seems this has been moved down somehow?

examples/qml_features/cpp/qabstractlistmodelcxx.h
```diff
@@ -10,6 +10,13 @@
 
 #include "rust/cxx.h"
 
+template<class T>
```
ahayzen-kdab, line 13: we've been using `typename` elsewhere instead of `class`

examples/qml_features/cpp/qabstractlistmodelcxx.h
```diff
@@ -10,6 +10,13 @@
...
+T
+constructQModelIndex()
```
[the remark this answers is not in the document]
ahayzen-kdab: #391 implements QModelIndex as a cxx-qt-lib type so would make this part obsolete.

crates/cxx-qt-gen/src/generator/cpp/inherit.rs
LeonMatthesKDAB: Ah, rust-analyzers `auto-use` doing it's thing again :see_no_evil:

crates/cxx-qt-gen/src/generator/cpp/inherit.rs
```diff
@@ -0,0 +1,48 @@
...
+    let mut result = GeneratedCppQObjectBlocks::default();
+
+    for method in &qobject.inherited_methods {
+        let return_type = if let ReturnType::Type(_, ty) = &method.method.sig.output {
+            CppType::from(ty, &None)?.as_cxx_ty().to_owned()
+        } else {
+            "void".to_owned()
+        };
+
+        let missing_base_class_error = || {
```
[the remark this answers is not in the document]
LeonMatthesKDAB: Hm, fair point...

crates/cxx-qt-gen/test_outputs/inheritance.h
```diff
@@ -0,0 +1,57 @@
...
+  const MyObjectRust& unsafeRust() const;
+  MyObjectRust& unsafeRustMut();
+  std::unique_ptr<MyObjectCxxQtThread> qtThread() const;
+
+public:
+  Q_INVOKABLE QVariant data(const QModelIndex& _index,
+                            qint32 _role) const override;
+  Q_INVOKABLE bool hasChildren(const QModelIndex& _parent) const override;
+  template<class... Args>
+  void fetchMore_cxxqt_inherit(Args... args)
```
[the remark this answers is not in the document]
LeonMatthesKDAB: I think for now we can stick with exposing `private/protected` methods as `public`, as the method name is obfuscated somewhat.
We should add the `friend` approach in a separate PR, as it is more complicated to generate and brittle.

examples/qml_features/cpp/qabstractlistmodelcxx.h
LeonMatthesKDAB: This file was now be removed anyway :partying_face:

examples/qml_features/cpp/qabstractlistmodelcxx.h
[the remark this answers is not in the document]
LeonMatthesKDAB: Removed :+1:

crates/cxx-qt-gen/src/generator/cpp/inherit.rs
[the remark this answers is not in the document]
LeonMatthesKDAB: Fixed, defaults to `QObject` now.

crates/cxx-qt-gen/src/generator/cpp/inherit.rs
ahayzen-kdab, line 6: why is this below the `indoc::formatdoc` include? it should be grouped with the top

crates/cxx-qt-gen/src/generator/cpp/inherit.rs
```diff
@@ -0,0 +1,48 @@
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
```
ahayzen-kdab, line 1: it's 2023 now :-)

crates/cxx-qt-gen/src/generator/naming/mod.rs
```diff
@@ -17,3 +19,12 @@ pub struct CombinedIdent {
     /// The ident for rust
     pub rust: Ident,
 }
+
+impl CombinedIdent {
+    pub fn from_rust_function(ident: Ident) -> Self {
```
ahayzen-kdab, line 24: could add a test and a doc line?

crates/cxx-qt-gen/src/parser/inherit.rs
```diff
@@ -0,0 +1,75 @@
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
```
ahayzen-kdab, line 1: 2023?

crates/cxx-qt-gen/src/generator/naming/mod.rs
ahayzen-kdab, line 24: could maybe take by reference, then clone internally, this then matches a lot of the other methods?

crates/cxx-qt-gen/src/parser/inherit.rs
```diff
@@ -0,0 +1,75 @@
...
+pub struct ParsedInheritedMethod {
+    /// The original [syn::ForeignItemFn] of the inherited method declaration
+    pub method: ForeignItemFn,
+    /// whether the inherited method is marked as mutable
+    pub mutable: bool,
+    /// the parameters of the method, without the `self` argument
+    pub parameters: Vec<ParsedFunctionParameter>,
+    /// the name of the function in Rust, as well as C++
+    pub ident: CombinedIdent,
+    pub wrapper_ident: Ident,
```
ahayzen-kdab, line 43: doc line?

crates/cxx-qt-gen/src/parser/parameter.rs
```diff
@@ -16,29 +20,48 @@ pub struct ParsedFunctionParameter {
...
+                    Ok(())
+                } else {
+                    Err(Error::new_spanned(type_pattern, missing_self_arg))
+                }
+            }
+            Some(FnArg::Receiver(Receiver {
+                reference: Some(_), // `self` needs to be by-ref, by-value is not supported.
+                mutability: None,   // Mutable self references are not supported.
+                ..
+            })) => Ok(()), /*Okay, found a &self reference*/
```
ahayzen-kdab, line 43: why is this `/* ... */` whereas the others are `//` ?

crates/cxx-qt-gen/src/syntax/types.rs
```diff
@@ -0,0 +1,33 @@
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Leon Matthes <leon.matthes@kdab.com>
+
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use crate::syntax::path::*;
+use syn::{GenericArgument, PathArguments, Type, TypePath, TypeReference};
+
+pub fn is_pin_of_self(ty: &Type) -> bool {
```
ahayzen-kdab, line 9: add tests

crates/cxx-qt-gen/src/syntax/types.rs
```diff
@@ -0,0 +1,33 @@
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
```
ahayzen-kdab, line 1: 2023 :-)

crates/cxx-qt-gen/test_inputs/inheritance.rs.license
```diff
@@ -0,0 +1,5 @@
+SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
```
ahayzen-kdab, line 1: 2023 :-)

crates/cxx-qt-gen/test_outputs/inheritance.cpp.license
```diff
@@ -0,0 +1,5 @@
+SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
```
ahayzen-kdab, line 1: 2023 :-)

crates/cxx-qt-gen/test_outputs/inheritance.h.license
```diff
@@ -0,0 +1,5 @@
+SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
```
ahayzen-kdab, line 1: 2023 :-)

examples/qml_features/CMakeLists.txt
```diff
@@ -54,7 +54,6 @@ add_executable(${APP_NAME}
     cpp/custom_object.h
     cpp/custom_object.cpp
     cpp/main.cpp
-    cpp/qabstractlistmodelcxx.h
```
ahayzen-kdab, line 57: :tada:

crates/cxx-qt-gen/test_outputs/inheritance.rs.license
```diff
@@ -0,0 +1,5 @@
+SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
```
ahayzen-kdab, line 1: 2023 :-)

examples/qml_features/rust/src/custom_base_class.rs
```diff
@@ -94,14 +80,26 @@ mod ffi {
...
+            self.as_mut()
+                .begin_remove_rows(&QModelIndex::default(), index, index);
             self.as_mut().vector_mut().remove(index as usize);
             self.as_mut().end_remove_rows();
         }
     }
 
     // QAbstractListModel implementation
     impl qobject::CustomBaseClass {
+        cxx_qt::inherit! {
```
ahayzen-kdab, line 92: Add an entry into the CHANGELOG about this :-)

crates/cxx-qt-gen/src/generator/cpp/inherit.rs
LeonMatthesKDAB: Fixed

crates/cxx-qt-gen/src/generator/cpp/inherit.rs
LeonMatthesKDAB: Nah, that can't be right ;D

crates/cxx-qt-gen/src/generator/naming/mod.rs
LeonMatthesKDAB: That's not very rust-y though.
We definitely need to own the ident, so if you can pass it by-value, you don't need to copy.

crates/cxx-qt-gen/src/parser/inherit.rs
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/parser/parameter.rs
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/generator/naming/mod.rs
ahayzen-kdab: I'd have to check but places where it was being used meant that the Ident was used multiple times, so passing by ref made sense back then. Maybe it should be changed in another commit...

LeonMatthesKDAB (commented):
TODO: Functions generated by cxx_qt::inherit! should be `unsafe` by default, similar to how they behave in CXX!!!
Currently you can call unsafe C++ functions without having `unsafe` anywhere.

examples/qml_features/rust/src/custom_base_class.rs
LeonMatthesKDAB: Done.
Also added documentation to the cxx-qt crate + book

crates/cxx-qt-gen/src/generator/naming/mod.rs
LeonMatthesKDAB: Right, from how we call it, it does make sense to have it by-ref.
However, Rust default is to take by-value if you want to own. So we should stick with that, even if that means a few more `clone` calls.
Afterall, Rust favors explicitness at the call-site.

ahayzen-kdab:
What did we think about using the attribute ?

```rust
#[cxx_qt::inherit]
extern "C++" {
    unsafe fn begin_insert_rows(self: Pin<&mut MyObject>, parent: &QModelIndex, first: i32, last: i32);
}
```

This then allows
- same syntax as CXX
- valid Rust syntax rather than a function macro
- multiple QObjects in the same block

We just then need to scan for the self param and then rewrite the MyObject -> MyObjectQt etc

LeonMatthesKDAB:
> What did we think about using the attribute ?
> 
> ```rust
> #[cxx_qt::inherit]
> extern "C++" {
>     unsafe fn begin_insert_rows(self: Pin<&mut MyObject>, parent: &QModelIndex, first: i32, last: i32);
> }
> ```
> 
> This then allows
> 
>     * same syntax as CXX
> 
>     * valid Rust syntax rather than a function macro
> 
>     * multiple QObjects in the same block
> 
> We just then need to scan for the self param and then rewrite the MyObject -> MyObjectQt etc

The reason I'm hesitant to implement it using the attribute is because it raises the question why we don't use the argueably better API:
``` rust
extern "C++" {
    #[cxx_qt::inherit]
    unsafe fn begin_insert_rows(self: Pin<&mut MyObject>, parent: &QModelIndex, first: i32, last: i32);
}
```
Which of course the answer to is: We don't want to parse CXX blocks. (Even though technically for something like this we should be able 

I also prefer that the function-style inherit macro is within the `impl` block of the qobject::MyObject.
In my mind this is the right place to put it.
Rust usually has all member declarations inside an `impl` block on the corresponding type. So CXX is already the exception here.

Sidenote: If we use the attribute route, the self type should probably not be `Pin<&mut MyObject>`, but `Pin<&mut qobject::MyObject>`, as the MyObject is a Rust-only type, so can't inherit something from C++.

In the end it kind off doesn't matter whether we use the attribute or function-style route.
We get enough information to implement the feature with either.
So if you guys prefer the attribute route, let me know and I'll change the implementation.

ahayzen-kdab:
> > I also prefer that the function-style inherit macro is within the impl block of the qobject::MyObject.
> > In my mind this is the right place to put it.
> > Rust usually has all member declarations inside an impl block on the corresponding type. So CXX is already the exception here.
> 
> I think I'd prefer it outside the `impl` block because the implementation isn't really in Rust, it's creating a binding for an existing C++ function. But I don't have a strong opinion on this. I'm more excited to get rid of the C++ boilerplate grinning

Same i prefer a block outside the `impl` as this is similar to a normal `extern "C++"` CXX block, except that the bindings need to be generated slightly differently.

> Sidenote: If we use the attribute route, the self type should probably not be Pin<&mut MyObject>, but Pin<&mut qobject::MyObject>, as the MyObject is a Rust-only type, so can't inherit something from C++.

Yup, think that should feel natural enough :thinking: And as long as it's documented then it should be fine.

examples/qml_features/rust/src/custom_base_class.rs
```diff
@@ -137,14 +127,57 @@ mod ffi {
...
+            unsafe {
+                self.as_mut()
+                    .begin_remove_rows(&QModelIndex::default(), index, index);
+                self.as_mut().vector_mut().remove(index as usize);
+                self.as_mut().end_remove_rows();
+            }
         }
     }
 
+    // ANCHOR: book_inherit_qalm_impl_unsafe
```
Be-ing, line 139:
```suggestion
    // ANCHOR: book_inherit_qalm_impl_unsafe
    // Create Rust bindings for C++ functions of the base class (QAbstractItemModel)
```

book/src/concepts/inheritance.md
```diff
@@ -0,0 +1,76 @@
...
+
+    impl qobject::CustomBaseClass {
+
+{{#include ../../../examples/qml_features/rust/src/custom_base_class.rs:book_inherit_data}}
+    }
+```
+[Full example](https://github.com/KDAB/cxx-qt/blob/main/examples/qml_features/rust/src/custom_base_class.rs)
+
+Note that if a method is overridden using `cxx_override` the base class version of the method can be accessed by using `#[cxx_qt::inherit]` in combination with the `#[cxx_name]` attribute.
+In this case the base class version of the function must get a different name, as Rust can't natively express the concept of calling a base class method.
```
Be-ing, line 64:
```suggestion
In this case the base class version of the function must get a different name, as Rust can't have two functions with the same name on one type.
```

crates/cxx-qt-gen/test_outputs/inheritance.h
```diff
@@ -0,0 +1,57 @@
...
+  Q_INVOKABLE QVariant data(QModelIndex const& _index,
+                            ::std::int32_t _role) const override;
+  Q_INVOKABLE bool hasChildren(QModelIndex const& _parent) const override;
+  template<class... Args>
+  bool hasChildren_cxxqt_inherit(Args... args) const
+  {
+    return QAbstractItemModel::hasChildren(args...);
+  }
+  template<class... Args>
+  void fetchMore_cxxqt_inherit(Args... args)
```
ahayzen-kdab, line 37: should these be camelcase instead of snake ?

crates/cxx-qt-gen/src/parser/inherit.rs
```diff
@@ -0,0 +1,156 @@
...
+                    attribute.span(),
+                    "Unsupported attribute in #[cxx_qt::inherit]",
+                ));
+            }
+
+            let name = attribute_tokens_to_value::<LitStr>(attribute)?;
+
+            ident.cpp = format_ident!("{}", name.value());
+        }
+        let wrapper_ident = format_ident!("{}_cxxqt_inherit", &ident.cpp);
```
ahayzen-kdab, line 143: Looks like this part should be camelcase?

Also wonder if this `wrapper_ident` should be in the parser phase or in the generator phase ? Then have a struct in naming to generate the wrapper from a given method ident ?

crates/cxx-qt-gen/src/generator/cpp/inherit.rs
```diff
@@ -0,0 +1,48 @@
...
+        mutability = if method.mutable { "" } else { "const" },
+        func_ident = method.ident.cpp,
+        wrapper_ident = method.wrapper_ident,
+        return_type = return_type,
+        base_class = base_class
+        }));
+    }
+
+    Ok(result)
+}
```
ahayzen-kdab, line 48: add tests to this module

crates/cxx-qt-gen/src/generator/naming/invokable.rs
```diff
@@ -23,20 +23,12 @@ impl From<&ImplItemMethod> for QInvokableName {
     fn from(method: &ImplItemMethod) -> Self {
         let ident = &method.sig.ident;
         Self {
-            name: name_from_ident(ident),
+            name: CombinedIdent::from_rust_function(ident.clone()),
```
ahayzen-kdab, line 26: we seem to have a style difference here.

Before we had module + function name -> CombinedIdent. Whereas this one is using a from method on the struct. We should ensure these are consistent.

At the moment they were split so that each of the naming implementations are in their own module so they are easy to find and we don't end up with a huge file again.

Maybe have another commit that does the `from_nnn` route but puts the implementation of each of them in the module files so that they can be found easily?

Eg have `from_rust_function` in a methods file. And keep the others in their own files but change the `nnn_from_ident() -> CombinedIdent` to become `CombinedIdent::from_nnn`. Although some of these are private methods currently and i wonder if they would be optimised to being inlined currently.
```rust
impl CombinedIdent {
  pub fn from_nnn(T) -> Self ...
}
```

crates/cxx-qt-gen/src/generator/rust/qobject.rs
```diff
@@ -143,6 +148,52 @@ pub fn generate_methods(
     Ok(blocks)
 }
 
+fn generate_inherited_methods(
```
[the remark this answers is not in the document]
ahayzen-kdab: (and like invokables, properties, signals etc)

crates/cxx-qt-gen/src/parser/cxxqtdata.rs
```diff
@@ -716,4 +779,44 @@ mod tests {
...
+            unsafe extern "C++" {
+                fn test(self: &qobject::MyObject);
+
+                fn with_args(self: &qobject::MyObject, arg: i32);
+            }
+        });
+        let safe_block: Item = tokens_to_syn(quote! {
+            #[cxx_qt::inherit]
+            extern "C++" {
+                #[cxx_name="withRename"]
```
ahayzen-kdab, line 798: wonder if we need to protect against having a `#[namespace = "N"]` here, or does this already fail ?

As inherit doesn't support a namespace as they come from the qobject instead.

crates/cxx-qt-gen/src/parser/inherit.rs
```diff
@@ -0,0 +1,156 @@
...
+            method,
+            qobject_ident,
+            mutable,
+            parameters,
+            ident,
+            wrapper_ident,
+            safe,
+        })
+    }
+}
```
ahayzen-kdab, line 156: add tests in here

crates/cxx-qt-gen/src/parser/parameter.rs
```diff
@@ -16,29 +20,67 @@ pub struct ParsedFunctionParameter {
...
+                ..
+            })) => Ok(()), // Okay, found a &self reference
+            Some(arg) => Err(Error::new_spanned(arg, missing_self_arg)),
+            None => Err(Error::new_spanned(signature, "Missing 'self' receiver!")),
+        }?;
+
+        Self::parse_remaining(iter)
     }
 
     pub fn parse(type_pattern: &PatType) -> Result<Self> {
```
ahayzen-kdab, line 86: do we need more tests for the new methods here?

crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -4,6 +4,7 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
 use crate::parser::{
+    inherit::*,
```
ahayzen-kdab, line 7: why not just include `ParsedInheritedMethod` ?

crates/cxx-qt-gen/src/syntax/types.rs
```diff
@@ -0,0 +1,179 @@
+// SPDX-FileCopyrightText: 2023 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Leon Matthes <leon.matthes@kdab.com>
+
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use crate::syntax::path::*;
```
ahayzen-kdab, line 6: is there any advantage / disadvantage to using `*` vs more explicit includes like we are doing elsewhere?

crates/cxx-qt-gen/test_outputs/inheritance.h
```diff
@@ -0,0 +1,57 @@
...
+  ~MyObject();
+  MyObjectRust const& unsafeRust() const;
+  MyObjectRust& unsafeRustMut();
+  ::std::unique_ptr<MyObjectCxxQtThread> qtThread() const;
+
+public:
+  Q_INVOKABLE QVariant data(QModelIndex const& _index,
+                            ::std::int32_t _role) const override;
+  Q_INVOKABLE bool hasChildren(QModelIndex const& _parent) const override;
+  template<class... Args>
```
ahayzen-kdab, line 31: can you open a separate bug about considering if we want to use `friend` as discussed before in the future. Or if just making them public like this is also "fine". And in that bug put a link to #329 so we don't forget the syntax for that :sweat_smile:

crates/cxx-qt-gen/src/syntax/types.rs
LeonMatthesKDAB: This is a habit I picked up at some point from a personal project.
Generally it means you're avoiding a lot of the bloat in "use" declarations.
They're kinda boilerplate for the most part.
So I like using wildcards to not have to list 20 different things I'm pulling from the same lib.

That being said, effective Rust recommends not using them for compatibility reasons: https://www.lurklurk.org/effective-rust/wildcard.html
So I'll switch that over to explicit imports.

crates/cxx-qt-gen/test_outputs/inheritance.h
LeonMatthesKDAB: Fair point, will update.

crates/cxx-qt-gen/src/parser/cxxqtdata.rs
LeonMatthesKDAB: `ParsedInheritedMethod::parse_safe` already checks this.
Only the `cxx_name` attribute is allowed, all others result in a compiler error.

crates/cxx-qt-gen/src/parser/parameter.rs
LeonMatthesKDAB: It's not really much new here, but still can't hurt.
I'll add some.

crates/cxx-qt-gen/test_outputs/inheritance.h
LeonMatthesKDAB: #452

crates/cxx-qt-gen/src/parser/inherit.rs
LeonMatthesKDAB: Actually I think it makes most sense to just make this a function call on ParsedInheritedMethod instead.
Kind off doesn't make sense to have this as a data member.

crates/cxx-qt-gen/src/generator/rust/qobject.rs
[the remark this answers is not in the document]
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/parser/qobject.rs
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/syntax/types.rs
LeonMatthesKDAB: Done

crates/cxx-qt-gen/test_outputs/inheritance.h
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/generator/cpp/inherit.rs
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/parser/inherit.rs
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/parser/parameter.rs
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/generator/naming/invokable.rs
LeonMatthesKDAB: I've moved this onto CombinedIdent now.
Whilst this results into somewhat longer method names, I think it's more expressive, as it's clearer whats constructed, and also clearer what the input "ident" should be.
Those changes are in a separate commit, so would be easy to undo if you prefer the previous style.

crates/cxx-qt-gen/src/generator/naming/mod.rs
LeonMatthesKDAB: I've changed this in the newest commit for the other CombinedIdent constructors as well.
Let me know if you're okay with that.

crates/cxx-qt-gen/src/generator/naming/functions.rs
```diff
@@ -0,0 +1,42 @@
+// SPDX-FileCopyrightText: 2023 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Leon Matthes <leon.matthes@kdab.com>
+
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use convert_case::{Case, Casing};
+use quote::format_ident;
+use syn::Ident;
+
+use super::*;
```
ahayzen-kdab, line 10: should this just be `super::CombinedIdent` ?

crates/cxx-qt-gen/src/generator/naming/invokable.rs
ahayzen-kdab: I like this way so far, as it keeps the implementations in separate files but still is rusty with T::from_nnn(...)

crates/cxx-qt-gen/src/parser/inherit.rs
```diff
@@ -0,0 +1,272 @@
...
+            mutable,
+            parameters,
+            ident,
+            safe,
+        })
+    }
+
+    /// the name of the wrapper function in C++
+    pub fn wrapper_ident(&self) -> Ident {
+        format_ident!("{}CxxqtInherit", self.ident.cpp)
```
ahayzen-kdab, line 157: elsewhere we use `CxxQt` not `Cxxqt` ? (eg the threading stuff)

book/src/concepts/inheritance.md
```diff
@@ -0,0 +1,76 @@
...
+SPDX-FileCopyrightText: 2023 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+SPDX-FileContributor: Leon Matthes <leon.matthes@kdab.com>
+
+SPDX-License-Identifier: MIT OR Apache-2.0
+-->
+
+# Inheritance
+
+Some Qt APIs require you to override certain methods from an abstract base class.
+Specifically [QAbstractItemModel](https://doc.qt.io/qt-6/qabstractitemmodel.html).
```
Be-ing, line 11:
```suggestion
Some Qt APIs require you to override certain methods from an abstract base class, for example [QAbstractItemModel](https://doc.qt.io/qt-6/qabstractitemmodel.html).
```

book/src/concepts/inheritance.md
```diff
@@ -0,0 +1,76 @@
...
+```
+[Full example](https://github.com/KDAB/cxx-qt/blob/main/examples/qml_features/rust/src/custom_base_class.rs)
+
+This code implements a QAbstractListModel subclass.
+For this, access to the `beginResetModel` and related methods is required.
+See [the Qt docs](https://doc.qt.io/qt-6/qabstractlistmodel.html) for more details on the specific subclassing requirements.
+These methods are then made accessible by using `#[cxx_qt::inherit]`.
+
+Methods can be declared inside `#[cxx_qt::inherit]` in `extern "C++"` blocks similar to CXX, with the same restrictions regarding which types can be used.
+Additionally, the `self` type must be either `self: Pin<&mut qobject::T>` or `self: &qobject::T`. Where `qobject::T` may refer to any valid `#[cxx_qt::qobject]` in the `#[cxx_qt::bridge]`
```
Be-ing, line 36:
```suggestion
Additionally, the `self` type must be either `self: Pin<&mut qobject::T>` or `self: &qobject::T`, where `qobject::T` must refer to a QObject marked with `#[cxx_qt::qobject]` in the `#[cxx_qt::bridge]`
```

book/src/concepts/inheritance.md
```diff
@@ -0,0 +1,76 @@
...
+
+    impl qobject::CustomBaseClass {
+
+{{#include ../../../examples/qml_features/rust/src/custom_base_class.rs:book_inherit_data}}
+    }
+```
+[Full example](https://github.com/KDAB/cxx-qt/blob/main/examples/qml_features/rust/src/custom_base_class.rs)
+
+Note that if a method is overridden using `cxx_override` the base class version of the method can be accessed by using `#[cxx_qt::inherit]` in combination with the `#[cxx_name]` attribute.
+In this case the base class version of the function must get a different name, as Rust can't have two functions with the same name on one type.
```
Be-ing, line 64:
```suggestion
Note that if a method is overridden using `cxx_override`, the base class version of the method can be accessed by using `#[cxx_qt::inherit]` in combination with the `#[cxx_name]` attribute.
In this case the base class version of the function must get a different name because Rust can't have two functions with the same name on one type.
```

book/src/concepts/inheritance.md
```diff
@@ -0,0 +1,76 @@
...
+    impl qobject::CustomBaseClass {
+{{#include ../../../examples/qml_features/rust/src/custom_base_class.rs:book_inherit_clear}}
+    }
+```
+[Full example](https://github.com/KDAB/cxx-qt/blob/main/examples/qml_features/rust/src/custom_base_class.rs)
+
+This code implements a QAbstractListModel subclass.
+For this, access to the `beginResetModel` and related methods is required.
+See [the Qt docs](https://doc.qt.io/qt-6/qabstractlistmodel.html) for more details on the specific subclassing requirements.
+These methods are then made accessible by using `#[cxx_qt::inherit]`.
```
Be-ing, line 33:
```suggestion
For this, the `clear` method implemented in Rust needs to call `beginResetModel` and related methods from the base class, which are made accessible by using `#[cxx_qt::inherit]`.
See [the Qt docs](https://doc.qt.io/qt-6/qabstractlistmodel.html) for more details on the specific subclassing requirements.
```

book/src/concepts/inheritance.md
```diff
@@ -0,0 +1,76 @@
...
+{{#include ../../../examples/qml_features/rust/src/custom_base_class.rs:book_inherit_qalm}}
+
+    impl qobject::CustomBaseClass {
+
+{{#include ../../../examples/qml_features/rust/src/custom_base_class.rs:book_inherit_data}}
+    }
+```
+[Full example](https://github.com/KDAB/cxx-qt/blob/main/examples/qml_features/rust/src/custom_base_class.rs)
+
+Note that if a method is overridden using `cxx_override` the base class version of the method can be accessed by using `#[cxx_qt::inherit]` in combination with the `#[cxx_name]` attribute.
```
Be-ing, line 63:
```suggestion
When a method is overridden using `cxx_override`, the base class version of the method can be accessed by using `#[cxx_qt::inherit]` in combination with the `#[cxx_name]` attribute.
```

LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/generator/naming/functions.rs
LeonMatthesKDAB: darn it, this again :see_no_evil:

crates/cxx-qt-gen/src/parser/inherit.rs
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/generator/naming/functions.rs
LeonMatthesKDAB: Done

ahayzen-kdab (approved):
LGTM, lets merge this and any further changes can be after!

PR #368, opened 2022-11-29
ParsedQInvokable, ParsedFunctionParameter Refactor

LeonMatthesKDAB:
Extract their parsing into their own `parse` methods.

### DIFF
--- a/crates/cxx-qt-gen/src/parser/invokable.rs
+++ b/crates/cxx-qt-gen/src/parser/invokable.rs
@@ -3,9 +3,12 @@
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
-use crate::parser::parameter::ParsedFunctionParameter;
+use crate::{
+    parser::parameter::ParsedFunctionParameter,
+    syntax::{attribute::*, implitemmethod::is_method_mutable},
+};
 use std::collections::HashSet;
-use syn::ImplItemMethod;
+use syn::{Ident, ImplItemMethod, LitStr, Result};
 
 /// Describes a C++ specifier for the Q_INVOKABLE
 #[derive(Eq, Hash, PartialEq)]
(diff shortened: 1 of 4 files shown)


LeonMatthesKDAB:
As discussed in #363 I tried to extract the refactorings into this PR so they can already be merged

crates/cxx-qt-gen/src/parser/invokable.rs
```diff
@@ -28,3 +31,55 @@ pub struct ParsedQInvokable {
     /// Any specifiers that declared on the invokable
     pub specifiers: HashSet<ParsedQInvokableSpecifiers>,
 }
+
+impl ParsedQInvokable {
+    pub fn parse(method: &ImplItemMethod) -> Result<Self> {
```
ahayzen-kdab, line 36: should we pass in the index ?

crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -87,91 +83,30 @@ impl ParsedQObject {
         })
     }
 
+    fn parse_impl_method(&mut self, method: &ImplItemMethod) -> Result<()> {
+        // Determine if this method is an invokable
+        if attribute_find_path(&method.attrs, &["qinvokable"]).is_some() {
```
ahayzen-kdab, line 88: we know the index of the `qinvokable` here should we do `if let Some(index) = ...` then pass in the index? Like we do for `ParsedQObject::from_struct(&ItemStruct, attr_index: usize)` ?

crates/cxx-qt-gen/src/parser/invokable.rs
ahayzen-kdab: As discussed lets move the stuff from cxxqtdata.rs into here. So then this can take a whole syn item, we'll do the same with signals etc.

ahayzen-kdab: Lets do this here for now and consider a larger refactor later, as for example signals still need to ask the qobject btreemap if the T exists, so maybe with further thought something cleaner could be made.

I have wondered before about having our own syn tree that is fully parsed by syn instead.

ahayzen-kdab (approved):
LGTM :-)

LeonMatthes:
@ahayzen-kdab Please reapprove 😊

ahayzen-kdab (approved):
LGTM :-)

PR #377, opened 2022-12-08
cxx-qt-lib: use C++ templates for common methods

### LINKED ISSUE #376: cxx-qt-lib: Can we use C++ templates to reduce the amount of C++ code
### Discussed in https://github.com/KDAB/cxx-qt/discussions/370

<div type='discussions-op-text'>

<sup>Originally posted by **LeonMatthes** December  6, 2022</sup>
The most common C++ functions we need for cxx-qt-lib are:
- Various constructors (default, copy, etc.)
- Destructor

So could we just add a "cxx-qt-lib/common-utils.h" that defines templates for these.
e.g. for the destructor:
``` cpp
template<class T>
void destruct(T &t) {
    t.~T();
}
```

We've already found that CXX can deal with this pretty well, as long as it can infer all template type arguments.</div>

ahayzen-kdab:
Closes #376

### DIFF
--- a/crates/cxx-qt-lib-headers/include/common.h
+++ b/crates/cxx-qt-lib-headers/include/common.h
@@ -2,9 +2,27 @@
 // SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
 // clang-format on
 // SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
-// SPDX-FileContributor: Leon Matthes <leon.matthes@kdab.com>
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
+
 #pragma once
 
-#include <QPoint>
+namespace rust {
+namespace cxxqtlib1 {
+
+template<typename T, typename... Args>
+T
+construct(Args... args)
+{
+  return T(args...);
+}
+
+template<typename T>
+void
+drop(T& value)
+{
+  value.~T();
+}
+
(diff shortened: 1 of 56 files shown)


crates/cxx-qt-lib-headers/include/common.h
```diff
@@ -0,0 +1,35 @@
...
+template<typename T>
+T
+initDefault()
+{
+  return T();
+}
+
+template<typename T>
+T
+clone(const T& value)
```
LeonMatthesKDAB, line 29: This is where we could try getting creative:
How about a `construct` method, that can take any argument?
e.g.:
```
template<typename... Args>
T construct(Args.. args) {
    return T(args...);
}
```

Then even stuff like `qDateInit`, `qdatetimeInitFromDateAndTime` could all boil down to this one template

LeonMatthesKDAB: We wouldn't even need the `initDefault` method then, just `construct` without arguments would be equivalent.

LeonMatthesKDAB (changes requested):
Going in a great direction. I think we can improve even more here though :thinking:

crates/cxx-qt-lib-headers/include/common.h
ahayzen-kdab: Done

LeonMatthesKDAB (approved):
LGTM

Be-ing:
+277 −724 :+1:

PR #381, opened 2022-12-13
qset: use rust::isize for qsizetype

ahayzen-kdab:


crates/cxx-qt-lib/src/types/qset/mod.rs
```diff
@@ -220,7 +220,7 @@ macro_rules! impl_qset_element {
             }
 
             unsafe fn get_unchecked(set: &QSet<Self>, pos: usize) -> &Self {
-                $module::get_unchecked(set, pos)
+                $module::get_unchecked(set, pos as isize)
```
ahayzen-kdab, line 223: where should we perform the usize -> isize conversion? As Qt uses signed types.

Should we expose len() to rust as isize?

But the Iter trait still needs usize for other parts?

crates/cxx-qt-lib-headers/include/qset.h
```diff
@@ -25,11 +25,16 @@ struct rust::IsRelocatable<QSet<T>> : std::true_type
...
 template<typename T>
 const T&
-qsetGetUnchecked(const QSet<T>& s, ::std::size_t pos) noexcept
+qsetGetUnchecked(const QSet<T>& s, ::rust::isize pos) noexcept
 {
-  Q_ASSERT(pos < static_cast<::std::size_t>(s.size()));
+  Q_ASSERT(pos < qsetLen(s));
+  Q_ASSERT(pos >= 0);
   auto it = s.cbegin();
   std::advance(it, pos);
```
ahayzen-kdab, line 39: Qt uses effectively `return s.cbegin()[pos];` internally, is that more performant / better than using `std::advance` ?

crates/cxx-qt-lib/src/types/qset/mod.rs
ahayzen-kdab: Right Qt is weird with using a signed type here :-)

I think changing the get_unchecked to isize makes sense. The only thing i wondered was if the public `QSet::len(&self)` should return a `usize` ? As it might seem odd to a Rust developer that the size is signed.

However when we move onto QVector/QList etc they have index_of/append/remove/at/len which are all potentially `isize`. So do we want to be consistent in using `isize` like Qt or convert some to `usize`? :thinking:

crates/cxx-qt-lib-headers/include/qset.h
ahayzen-kdab: Looks like `std::advance` is safe / more correct as depending how the container is organised simply using an index offset might not be the Nth item, whereas `std::advance` can take this into account - although it's slower potentially.

crates/cxx-qt-lib/src/types/qset/mod.rs
ahayzen-kdab: Right, i've gone with using `isize` everywhere except in the iterator internals.

PR #389, opened 2022-12-14
cxx-qt-gen: add support for more CXX types

ahayzen-kdab:
Related to #328


ahayzen-kdab:
Add changelog entry for fixes when done.

PR #398, opened 2022-12-20
cxx-qt-lib-headers: don't use static to avoid fixed size requirement

ahayzen-kdab:
This also allows us to add #[cfg(..)] to the headers later.


ahayzen-kdab:
Actually might be simpler to have if's in the C++ headers and always include them, otherwise we need to set the rustc thing in multiple places.

But this still simplifies things as we don't need to increment the number.

crates/cxx-qt-lib-headers/src/lib.rs
```diff
@@ -10,36 +10,34 @@
...
+    for (file_contents, file_name) in [
+        (include_str!("../include/common.h"), "common.h"),
+        (include_str!("../include/convert.h"), "convert.h"),
+        (include_str!("../include/cxxqt_thread.h"), "cxxqt_thread.h"),
+        (include_str!("../include/qcolor.h"), "qcolor.h"),
+        (include_str!("../include/qdate.h"), "qdate.h"),
+        (include_str!("../include/qdatetime.h"), "qdatetime.h"),
+        (include_str!("../include/qhash.h"), "qhash.h"),
+        (include_str!("../include/qmodelindex.h"), "qmodelindex.h"),
+        (include_str!("../include/qpersistentmodelindex.h"), "qpersistentmodelindex.h"),
```
ahayzen-kdab, line 26: Appears format checks aren't running on this cargo project, this line is too long...

ahayzen-kdab: Fixed

ahayzen-kdab:
> Already included in #387?

It was separated out to try and land earlier... there are conflicts here too to resolve :-)

Issue #291, opened 2022-10-03
Decide on more useful types to add from QtCore/QtGui


ahayzen-kdab:
Decide on more useful types to add from QtCore/QtGui (eg QByteArray, QModelIndex, QVector2d, QHash, QVector etc)

QtCore - https://doc.qt.io/qt-6/qtcore-module.html

- [x] QByteArray #392 
- [x] ~QHash~  -> #347
- [X] ~QLatin1String~ - in Qt 6 it's just a view and should you use just QByteArray and QString mainly, do we need this?
- [x] ~QMap~ -> #347
- [x] QModelIndex #391 
- [x] QPersistentModelIndex #393 
- [x] QStringList -> #408 
- [x] ~QSet~  -> #347
- [x] ~QVector~  -> #347
- [x] QCoreApplication -> #412
- [x] QMargins -> #419
- [x] QTimeZone -> #416 
- [x] Global enums ? -> #419 
- [x] ~QObject~ - for now lets skip as it's simple enough to define `type MyQObject` in CXX
- ... (we should consider which other common types are useful to maintain)
- enums from the global namespace ? if they are used for overrides in proxies etc

QtGui - https://doc.qt.io/qt-6/qtgui-module.html

- [X] ~QAction~ - has a lot of other Qt classes like fonts and icons, lets skip for now
- [x] ~QMatrix4x4 ? (related to QVectorND)~
- [x] ~QTransform ? (related to QVectorND)~
- [x] QVector2D -> #403
- [x] QVector3D -> #403
- [x] QVector4D -> #403 
- [x] QGuiApplication -> #412 
- ... (we should consider which other common types are useful to maintain)

QtQML - https://doc.qt.io/qt-6/qtqml-module.html

- [x] QQmlApplicationEngine -> #412
- [x] QQmlEngine -> #412 
- ... (we should consider which other common types are useful to maintain)
- [x] ~QQmlComponent (?)~

We need to come up with some rules as to which classes we should bridge in cxx-qt-lib, the following should be considered
- Maintenance of the bridge
- Practical use from Rust (eg if you have a clear UI / model split, UI details should not be in Rust?)
- Usage in applications (eg don't wrap classes that are hardly used)
- Enable deriving from other classes (eg QModelIndex and QHash are needed for listmodels)
- Can we easily bridge templated classes ? this was an issue with the CxxQtThread (eg https://github.com/dtolnay/cxx/issues/683 )

ahayzen-kdab:
[...]

PR #403, opened 2023-01-09
WIP: cxx-qt-lib: add support for QVector2D, QVector3D, QVector4D

ahayzen-kdab:
Related to #291


crates/cxx-qt-lib/src/types/qpointf.rs
```diff
@@ -61,6 +61,28 @@ impl Default for QPointF {
     }
 }
 
+impl From<crate::QVector2D> for QPointF {
```
ahayzen-kdab, line 64: does it make more sense for these to be in the QVector2D rust file?

ahayzen-kdab: Done

PR #405, opened 2023-01-09
cxx-qt-lib: use a feature for Gui parts of cxx-qt-lib

### LINKED ISSUE #290: Add features to cxx-qt-lib to be able to split QtCore and QtGui
Add features to cxx-qt-lib to be able to split QtCore and QtGui.

And split the folders in src/qtcore/T.rs rather than src/types/T.rs.

This then means if your project only links to QtCore you could still use cxx-qt-lib.

ahayzen-kdab:
Closes #290


ahayzen-kdab:
An example of testing the qt_gui feature is below, eventually there will be a `qt_qml` feature that users may need to enable.

```diff
commit 804f1e82bb41179b67f0c24821326cdc20f80ef0
Author: Andrew Hayzen <andrew.hayzen@kdab.com>
Date:   Mon Jan 9 21:04:46 2023 +0000

    WIP: example of using Qt Gui feature

diff --git a/crates/cxx-qt-lib/Cargo.toml b/crates/cxx-qt-lib/Cargo.toml
index 8de837ff..9a82d64c 100644
--- a/crates/cxx-qt-lib/Cargo.toml
+++ b/crates/cxx-qt-lib/Cargo.toml
@@ -24,5 +24,5 @@ cxx-qt-lib-headers.workspace = true
 qt-build-utils.workspace = true
 
 [features]
-default = ["qt_gui"]
+default = []
 qt_gui = ["cxx-qt-lib-headers/qt_gui"]
diff --git a/examples/qml_features/rust/Cargo.toml b/examples/qml_features/rust/Cargo.toml
index 7282ec16..51a850d0 100644
--- a/examples/qml_features/rust/Cargo.toml
+++ b/examples/qml_features/rust/Cargo.toml
@@ -16,7 +16,7 @@ crate-type = ["staticlib"]
 [dependencies]
 cxx.workspace = true
 cxx-qt.workspace = true
-cxx-qt-lib.workspace = true
+cxx-qt-lib = { workspace = true, features = ["qt_gui"] }
 serde.workspace = true
 serde_json.workspace = true
 
diff --git a/examples/qml_features/rust/build.rs b/examples/qml_features/rust/build.rs
index d782af46..8b19d285 100644
--- a/examples/qml_features/rust/build.rs
+++ b/examples/qml_features/rust/build.rs
@@ -9,6 +9,7 @@ use cxx_qt_build::CxxQtBuilder;
 
 fn main() {
     CxxQtBuilder::new()
+        .qt_modules(&["Gui"])
         .cc_builder(|cc| {
             cc.include("../cpp");
         })
```

crates/cxx-qt-lib/build.rs
```diff
@@ -154,6 +162,10 @@ fn main() {
     cxx_qt_lib_headers::write_headers(format!("{}/cxx-qt-lib", out_dir));
     builder.include(out_dir);
 
+    // Enable Qt Gui in C++ if the feature is enabled
+    #[cfg(feature = "qt_gui")]
```
ahayzen-kdab, line 166: Do we need this in `cxx-qt-build` as that also has `cxx-qt-lib-headers`, but is that only for `common.h` etc ?

Maybe this is the point to split out `cxx-qt-gen` headers from `cxx-qt-lib` ones.

crates/cxx-qt-lib/src/types/mod.rs
```diff
@@ -6,7 +6,9 @@
 mod qbytearray;
 pub use qbytearray::QByteArray;
 
+#[cfg(feature = "qt_gui")]
 mod qcolor;
```
ahayzen-kdab, line 10: could also change the folder / module structure to be
```
cxx-qt-lib/
  - core/
    - ...
  - gui/
    - ...
```

And then have `cxx_qt_lib::core::QPoint` and `cxx_qt_lib::gui::QColor` ?

ahayzen-kdab: Done

crates/cxx-qt-build/src/lib.rs
```diff
@@ -256,7 +256,6 @@ impl CxxQtBuilder {
     pub fn new() -> Self {
         let mut qt_modules = HashSet::new();
         qt_modules.insert("Core".to_owned());
-        qt_modules.insert("Gui".to_owned());
```
ahayzen-kdab, line 259: cxx-qt-lib defaults toe enabling qt_gui, but now cxx-qt-build doesn't.

Should we also have a feature on cxx-qt-build that then enables the cxx-qt-lib feature? Or we should default to Core + Gui but have a way to replace/clear the modules ? Maybe `qt_modules(...)` should replace rather than extend ?

crates/cxx-qt-lib/build.rs
```diff
@@ -4,10 +4,14 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
 fn main() {
-    let qt_modules = vec!["Core", "Gui"]
-        .iter()
-        .map(|m| String::from(*m))
-        .collect();
+    let qt_modules = vec![
+        "Core",
+        #[cfg(feature = "qt_gui")]
```
ahayzen-kdab, line 9: This might need to use env vars instead? https://doc.rust-lang.org/cargo/reference/features.html#build-scripts

ahayzen-kdab: Done

crates/cxx-qt-lib/build.rs
ahayzen-kdab: Build uses whether the qt_module is enabled.

crates/cxx-qt-build/src/lib.rs
ahayzen-kdab: Lets only enable QtCore for now, then opt in to further modules. This is then similar to CMake and is more explicit.

ahayzen-kdab: Done

LeonMatthesKDAB:
Adding GUI as a feature flag is definitely the right thing to do in my book.
The question is whether to add this to the default features.

I think as most people are going to want to have access to things like QColor, let's make `qt_gui` a default feature.

ahayzen-kdab:
> Adding GUI as a feature flag is definitely the right thing to do in my book. The question is whether to add this to the default features.
> 
> I think as most people are going to want to have access to things like QColor, let's make `qt_gui` a default feature.

How should developers be able to opt out of QtGui in CxxQtBuilder ? as at the moment the `qt_modules(..)` method extends the array. Whereas if we default to Core + Gui, we need a way to remove Gui ? I've wondered before about `qt_modules` replacing the array instead :thinking: Or any other ideas ?

ahayzen-kdab:
> > How should developers be able to opt out of QtGui in CxxQtBuilder ? as at the moment the `qt_modules(..)` method extends the array. Whereas if we default to Core + Gui, we need a way to remove Gui ? I've wondered before about `qt_modules` replacing the array instead thinking Or any other ideas ?
> 
> Hm, good question thinking So basically we have two places where we need to know whether you want to build with gui support:
> 
>     * cxx-qt-lib
> 
>     * your build script (i.e. cxx-qt-build).
> 
> 
> Does it make sense to have the two be separate (i.e. is there a use-case where I want cxx-qt-lib _without_ GUI, but my own program _with_ GUI). I don't think so... thinking
> 
> So can we somehow link the two?

I don't think we can link the two, when you have a feature Cargo only tells the build script if the current package has a feature enabled? https://doc.rust-lang.org/cargo/reference/features.html#build-scripts

cxx-qt-build doesn't depend directly on cxx-qt-lib so can't have it's own feature.

Then also you need to enable them in CMake if you are using a CMake build.

So you have three places
- CMake
- build script of your app (CxxQtBuilder)
- cxx-qt-lib

I wonder if we default `qt_modules` to Core + Gui, but then change the qt_modules method to replace rather than extend. As then if you wanted only Core you could do `qt_modules(&["Core"])` and if you want to enable Qml you can do `qt_modules(&["Core", "Gui", "Qml"])` .. or we could add another method like `qt_modules_extend(&["Qml"])` ?

LeonMatthesKDAB:
Hm, so if we look at this:
https://github.com/rust-lang/cargo/issues/10489

It seems cargo will perform a union of features for dependencies.
So a basic idea:
* Make qt_gui a feature of cxx-qt-build instead of lib
* disable qt_gui in the build dependency of cxx-qt-lib.
* Have cxx-qt-build somehow signal to the further compilation of cxx-qt-lib whether qt_gui feature is enabled.

and because `qt_gui` would be a default feature of cxx-qt-build, it would default-enable itself for cxx-qt-lib as well.
At least if all of this works with build-dependencies as well :thinking:

ahayzen-kdab:
> Hm, so if we look at this: [rust-lang/cargo#10489](https://github.com/rust-lang/cargo/issues/10489)
> 
> It seems cargo will perform a union of features for dependencies. So a basic idea:
> 
>     * Make qt_gui a feature of cxx-qt-build instead of lib
> 
>     * disable qt_gui in the build dependency of cxx-qt-lib.
> 
>     * Have cxx-qt-build somehow signal to the further compilation of cxx-qt-lib whether qt_gui feature is enabled.
> 
> 
> and because `qt_gui` would be a default feature of cxx-qt-build, it would default-enable itself for cxx-qt-lib as well. At least if all of this works with build-dependencies as well thinking

Hmm so if i have a feature called `qt_gui` on `cxx-qt-build` it'll enable it for other crates too? :thinking: 

Although "Features are unique to the package that defines them. Enabling a feature on a package does not enable a feature of the same name on other packages." https://doc.rust-lang.org/cargo/reference/features.html#feature-unification says it won't. It only unifies features on the same crate. So cxx-qt-build depends on cxx-qt-lib-headers but not cxx-qt-lib. So not sure how cxx-qt-build could indicate to cxx-qt-lib that `qt_gui` needs to be enabled.

ahayzen-kdab:
> I like the idea of using Cargo features to replace the list of Qt libraries to link in the build script, but I'm not sure how that can work without cxx-qt-build depending on cxx-qt-lib... thinking

Right, for now I think the best solution is to default to a preferred set (this would be enable core, gui, qml etc, but not enable widgets - eg optimise for a QML app). Then allow developers to opt-out via setting the default-features off and overriding the qt_modules in CxxQtBuilder :-)

ahayzen-kdab:
> So, looking at the diff of this PR, I'm not really happy with how the enabled Qt libraries are listed once in Cargo.toml for cxx-qt-lib and again in build.rs. I think it would be less bad to move the build script part from build.rs to the Cargo.toml line for cxx-qt-build. Yes, it would be ugly to have to duplicate the list of enabled features for both cxx-qt-lib and cxx-qt-build, but at least those lists could be adjacent to each other in Cargo.toml instead of split between two files.

OK so you are thinking to have ...

- cxx-qt-build features for qt_gui, qt_qml etc (if enabled sets them as default modules?)
- cxx-qt-lib features for qt_gui, qt_qml etc

Then a developer in their app when they depend on cxx-qt-build and/or cxx-qt-lib they set the features in their cargo.toml next to each other (although we'll likely default to qt_gui + qt_qml enabled?).

In theory a normal app then doesn't need to touch the qt_modules as the features are set and this enables the modules - but there is still the escape hatch if they need to manually add / replace the qt_modules.

I think this sounds like a good idea, i'll try doing it in a further commit to see what it looks like :-)

ahayzen-kdab:
Done :-)

crates/cxx-qt-build/src/lib.rs
```diff
@@ -304,8 +305,15 @@ impl CxxQtBuilder {
...
-    /// The Core and Gui modules are linked automatically; there is no need to specify them.
-    pub fn qt_modules(mut self, modules: &[&str]) -> Self {
+    /// The Core and any feature enabled modules are linked automatically; there is no need to specify them.
+    pub fn qt_module(mut self, module: &str) -> Self {
+        self.qt_modules.insert(module.to_owned());
+        self
+    }
+
+    /// Replace the list of [Qt modules](https://doc.qt.io/qt-6/qtmodules.html) to link.
+    pub fn set_qt_modules(mut self, modules: &[&str]) -> Self {
```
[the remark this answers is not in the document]
ahayzen-kdab: It was so that you could opt out of Gui before as some projects might want to just link Core.

This is now sortof possible with the feature in cxx-qt-build, except if you put your core-only lib in the same workspace as your gui-enabling lib :thinking: 

So i think i'll move this back to `qt_module` for now.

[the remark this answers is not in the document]
ahayzen-kdab: Done

examples/cargo_without_cmake/build.rs
```diff
@@ -10,7 +10,8 @@ fn main() {
     CxxQtBuilder::new()
         // Link Qt's Qml and Network libraries. Qt Core and Gui are always
         // linked, so there is no need to specify them here.
```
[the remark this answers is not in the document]
ahayzen-kdab: done

PR #411, opened 2023-01-11
qvariant: rename value to value_or_default

ahayzen-kdab:


crates/cxx-qt-lib/src/types/qvariant/mod.rs
```diff
@@ -105,7 +105,10 @@ impl QVariant {
     /// Returns the stored value converted to the template type T
     ///
     /// If the value cannot be converted, a default-constructed value will be returned.
-    pub fn value<T: QVariantValue>(&self) -> T {
+    ///
+    /// Note that this calls the `QVariant::value` method, without performance loss.
+    /// Whereas `try_value` first calls `QVariant::canConvert`.
+    pub fn value_or_default<T: QVariantValue>(&self) -> T {
         T::value(self)
```
ahayzen-kdab, line 112: should the trait be renamed as well? Or should that match the Qt API ? :thinking:

ahayzen-kdab: Renamed otherwise the Rust code looks weird.

crates/cxx-qt-lib/src/types/qvariant/mod.rs
```diff
@@ -96,7 +96,7 @@ impl QVariant {
     /// if QVariant::canConvert is true otherwise returns None
     pub fn try_value<T: QVariantValue>(&self) -> Option<T> {
```
LeonMatthesKDAB, line 97: This function should now be named `value`. Should be the default to use for the Rust API.

LeonMatthesKDAB (changes requested):

crates/cxx-qt-lib/src/types/qvariant/mod.rs
ahayzen-kdab: Done

LeonMatthesKDAB (approved):
LGTM

PR #413, opened 2023-01-13
Implement std::cmp::PartialEq for all Qt types

vimpostor:
This also allows us to trigger `changed()` signals only when the value really changed.

After this we can also implement std::cmp::PartialOrd for selected types, but it does not make sense for all types (e.g. not for `QByteArray`)

See #53

### DIFF
--- a/crates/cxx-qt-gen/src/generator/rust/property/mod.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/property/mod.rs
@@ -161,6 +161,9 @@ mod tests {
             tokens_to_syn::<syn::Item>(quote! {
                 impl MyObjectQt {
                     pub fn set_trivial_property(mut self: Pin<&mut Self>, value: i32) {
+                        if self.rust().trivial_property == value {
+                            return;
+                        }
                         unsafe {
                             self.as_mut().rust_mut().trivial_property = value;
                         }
@@ -249,6 +252,9 @@ mod tests {
             tokens_to_syn::<syn::Item>(quote! {
                 impl MyObjectQt {
                     pub fn set_opaque_property(mut self: Pin<&mut Self>, value: UniquePtr<QColor>) {
+                        if self.rust().opaque_property == value {
+                            return;
+                        }
                         unsafe {
                             self.as_mut().rust_mut().opaque_property = value;
                         }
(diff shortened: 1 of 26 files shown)


crates/cxx-qt-gen/src/generator/rust/property/mod.rs
```diff
@@ -161,6 +161,9 @@ mod tests {
             tokens_to_syn::<syn::Item>(quote! {
                 impl MyObjectQt {
                     pub fn set_trivial_property(mut self: Pin<&mut Self>, value: i32) {
+                        if self.rust().trivial_property == value {
```
ahayzen-kdab, line 164: do we get a sane error if the type does not implement `PartialEq` ? eg if you try to run the `qml_features` and you didn't have `PartialEq` on one of the property types like `QString` what kind of error message do you get?

And i assume do pointer types already have PartialEq ? eg like `*mut T` ?

As i didn't know if somewhere we need to put a `T: PartialEq` or if the error message is sane enough anyway?

ahayzen-kdab: (i also wasn't sure where this `T: PartialEq` would go, so if the current error message is sane that's fine :-)

vimpostor: This is the error message you get when `PartialEq` is not implemented:

```
error[E0369]: binary operation `==` cannot be applied to type `cxx_qt_lib::QString`
  --> examples/cargo_without_cmake/src/cxxqt_object.rs:11:1
   |
11 | #[cxx_qt::bridge]
   | ^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in the attribute macro `cxx_qt::bridge` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0369`.
```

I think this error message is reasonable enough, right?

ahayzen-kdab: Sane enough :-)

examples/qml_features/rust/src/signals.rs
```diff
@@ -25,7 +25,10 @@ pub mod ffi {
     // ANCHOR: book_signals_struct
     #[cxx_qt::qobject]
     #[derive(Default)]
-    pub struct RustSignals;
+    pub struct RustSignals {
```
ahayzen-kdab, line 28: hmm this example was supposed to be showing signals not properties :thinking: 

Maybe you can reuse the properties.rs example for the unit test. Eg just have an additional test in `tst_properties.qml` which creates a `RustProperties` then call `disconnect` multiple times and see if the properties like `connected` change or something :thinking:

vimpostor: Oh I didn't realize that this file would be embedded in the docs. In that case I shall remove the property there, even though the test technically tested behaviour of signals.

ahayzen-kdab (approved):
Awesome :partying_face:  LGTM :-)

Issue #55, opened 2022-01-12
Add member functions to basic Qt types in Rust


LeonMatthesKDAB:
- [x] Add missing member functions to types #416
- [ ] Consider if we should adapt the Rust API to be "safe" / less Qt-like (eg QRectF set_x)

Original Description
Types like QPoint have a lot of member functions, as well as operators (i.e. +,-,/,*) that make them a lot more useful. We should give Rust access to these as well.

ahayzen-kdab:
Note that this is related to #53

LeonMatthesKDAB:
Note: QRect's interface should propably *not* mimic Qt directly. set_x and set_y will change the width and height, which is not at all obvious/expected behavior. So we should improve on Qt in this regard, and come up with our own, safer interface.

ahayzen-kdab:
Also ensure that we are using CXX where we can. 

Eg
```rust
#[rust_name = "qurl_to_qstring"]
fn qurlToQString(url: &QUrl) -> QString;
```

could probably just be

```rust
#[rust_name = "to_string"]
fn toString(self: &QUrl) -> QString;
```

ahayzen-kdab:
Also note that `QDateTime` it would be useful to link up the static methods like `QDateTime::currentDateTime()`, otherwise might have static methods that could be useful too.

ahayzen-kdab:
Closing via #416.

PR #416, opened 2023-01-17
cxx-qt-lib: add member functions

ahayzen-kdab:
Related to #55 

Requires #412 

Requires #419


tests/qt_types_standalone/rust/src/qtimezone.rs
```diff
@@ -0,0 +1,32 @@
...
+
+    extern "Rust" {
+        fn construct_qtimezone() -> QTimeZone;
+        fn read_qtimezone(t: &QTimeZone) -> bool;
+        fn clone_qtimezone(t: &QTimeZone) -> QTimeZone;
+    }
+}
+
+fn construct_qtimezone() -> QTimeZone {
+    QTimeZone::from_iana(&QByteArray::from("Europe/London"))
```
ahayzen-kdab, line 23: valgrind is failing somewhere here...

ahayzen-kdab: Appears to be related to the inner QSharedDataPointer<T> being set by a non default QTimeZone (when default it is nullptr), then if that QTimeZone is given to C++ it leaks.

ahayzen-kdab: If a non default QTimeZone is constructed in Rust and given to C++, if C++ then calls `tz.~QTimeZone()` then the memory is freed correctly :thinking:

crates/cxx-qt-lib/src/core/mod.rs
```diff
@@ -48,9 +54,15 @@ pub use qstring::QString;
 mod qstringlist;
 pub use qstringlist::QStringList;
 
+mod qt;
+pub use qt::{AspectRatioMode, CaseSensitivity, DateFormat, SplitBehaviorFlags, TimeSpec};
```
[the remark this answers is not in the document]
ahayzen-kdab: Moved into #419

crates/cxx-qt-lib/src/core/mod.rs
```diff
@@ -48,9 +54,15 @@ pub use qstring::QString;
...
+
 mod qtime;
 pub use qtime::QTime;
 
+mod qtimezone;
+pub use qtimezone::QTimeZone;
+
 mod qpoint;
 pub use qpoint::QPoint;
```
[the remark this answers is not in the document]
ahayzen-kdab: Done

tests/qt_types_standalone/rust/src/qtimezone.rs
ahayzen-kdab: Moved into #420

crates/cxx-qt-lib/src/core/qbytearray.cpp
```diff
@@ -161,5 +161,29 @@ qbytearrayResize(QByteArray& byteArray, ::rust::isize size)
 #endif
 }
 
+QByteArray
+qbytearraySimplified(const QByteArray& byteArray)
```
LeonMatthesKDAB, line 165: Why are these needed? Can't this just be declared directly with CXX?

ahayzen-kdab: See the comment above them in the header
```
// If Q_COMPILER_REF_QUALIFIERS is set the definition of these is
// T method() const& which CXX doesn't bind it.
```

Basically they are defined as `QByteArray QByteArray::simplified() const&` i wouldn't find a way to bind to a `const&` method in CXX as others are just `const`

ahayzen-kdab: But i could be missing something obvious :-)

LeonMatthesKDAB:
```cpp
#if !defined(Q_CLANG_QDOC)
    [[nodiscard]] QByteArraytoLower() const &
    { return toLower_helper(*this); }
    [[nodiscard]] QByteArray toLower() &&
    { return toLower_helper(*this); }
    [[nodiscard]] QByteArray toUpper() const &
    { return toUpper_helper(*this); }
    [[nodiscard]] QByteArray toUpper() &&
    { return toUpper_helper(*this); }
    // ...
#else
    [[nodiscard]] QByteArray toLower() const;
    [[nodiscard]] QByteArray toUpper() const;
    [[nodiscard]] QByteArray trimmed() const;
    [[nodiscard]] QByteArray simplified() const;
#endif
```
Ooooof, it so they're disabling the `const&` feature just for the docs, which is why I didn't see that when browsing the Qt documentation...

This is probably done for optimization purposes I'd guess :thinking: 

But then it makes sense that we need this.
Something that upstream would need to address, but probably not worth the effort.

crates/cxx-qt-lib/src/core/qstring.rs
```diff
@@ -10,10 +10,77 @@ use std::mem::MaybeUninit;
...
+    unsafe extern "C++" {
+        include!("cxx-qt-lib/qbytearray.h");
+        type QByteArray = crate::QByteArray;
+        include!("cxx-qt-lib/qstring.h");
         type QString = super::QString;
+        include!("cxx-qt-lib/qstringlist.h");
+        type QStringList = crate::QStringList;
+
+        /// Appends the string str onto the end of this string.
+        fn append<'a>(self: &'a mut QString, str: &QString) -> &'a mut QString;
```
ahayzen-kdab, line 29: ensure if #448 lands first that the QString Ops::Add is updated to use this

ahayzen-kdab: Done

crates/cxx-qt-lib/src/core/qbytearray.cpp
LeonMatthesKDAB: Nothing we can fix right now.

crates/cxx-qt-lib/src/core/qurl.cpp
```diff
@@ -34,17 +34,187 @@ qurlInitFromString(::rust::Str string)
...
+qurlFileName(const QUrl& url)
+{
+  return url.fileName();
+}
+
+QString
+qurlFragment(const QUrl& url)
+{
+  return url.fragment();
+}
```
LeonMatthesKDAB, line 60: It's interesting that CXX can't wrap these, as the ComponentFormattingOptions do have a default value for each function.
So I was expecting C++ to generate two versions of this function, but apparently that's not the case and the compiler inserts the appropriate argument at the callsite.
Fascinating.

crates/cxx-qt-lib/src/core/qurl.rs
```diff
@@ -6,16 +6,159 @@ use cxx::{type_id, ExternType};
...
+        /// Sets the port of the URL to port.
+        ///
+        /// port must be between 0 and 65535 inclusive. Setting the port to -1 indicates that the port is unspecified.
+        #[rust_name = "set_port"]
+        fn setPort(self: &mut QUrl, port: i32);
+
+        /// Returns the path of this URL formatted as a local file path.
+        /// The path returned will use forward slashes, even if it was originally created from one with backslashes.
+        #[rust_name = "to_local_file"]
+        fn toLocalFile(self: &QUrl) -> QString;
```
LeonMatthesKDAB, line 86: Could you rename this to `to_local_file_or_default`, then in the actual `to_local_file` implementation, check that it actually is a local file with `is_local_file` and use `Option<Self>` as the result value?
I think that would be the rust-y way to approach this, like with `value` on QVariant.

Similar to what you have done for `fragment`, etc.

crates/cxx-qt-lib/src/core/qurl.rs
```diff
@@ -63,21 +196,190 @@ pub struct QUrl {
...
+    }
+
+    /// Returns the query string of the URL if there's a query string
+    pub fn query(&self) -> Option<ffi::QString> {
+        if self.has_fragment() {
+            Some(self.fragment_or_default())
+        } else {
+            None
+        }
+    }
```
LeonMatthesKDAB, line 277:
``` diff
- This should use `has_query` and `query_or_default`!
```

crates/cxx-qt-lib/src/core/qurl.rs
```diff
@@ -63,21 +196,190 @@ pub struct QUrl {
...
+        ffi::qurl_host(self)
+    }
+
+    /// Returns the current whitelist of top-level domains that are allowed to have non-ASCII characters in their compositions.
+    pub fn idn_whitelist() -> ffi::QStringList {
+        ffi::qurl_idn_whitelist()
+    }
+
+    /// Returns the password of the URL if it is defined; otherwise an empty string is returned.
+    pub fn password_or_default(&self) -> ffi::QString {
```
LeonMatthesKDAB, line 261: Missing `fn password(&self) -> Option<ffi::QString>`

crates/cxx-qt-lib/src/core/qurl.rs
```diff
@@ -63,21 +196,190 @@ pub struct QUrl {
...
         ffi::qurl_to_qstring(self)
     }
+
+    /// Returns the user info of the URL, or an empty string if the user info is undefined.
+    pub fn user_info_or_default(&self) -> ffi::QString {
+        ffi::qurl_user_info(self)
+    }
+
+    /// Returns the user name of the URL if it is defined; otherwise an empty string is returned.
+    pub fn user_name_or_default(&self) -> ffi::QString {
```
LeonMatthesKDAB, line 380: Also missing `fn user_name(&self) -> Option<ffi::QString>`

crates/cxx-qt-lib/src/core/qurl.rs
```diff
@@ -6,16 +6,159 @@ use cxx::{type_id, ExternType};
...
+        /// Returns the result of the merge of this URL with relative. This URL is used as a base to convert relative to an absolute URL.
+        fn resolved(self: &QUrl, relative: &QUrl) -> QUrl;
+
+        /// Returns the scheme of the URL. If an empty string is returned,
+        /// this means the scheme is undefined and the URL is then relative.
+        ///
+        /// The scheme can only contain US-ASCII letters or digits,
+        /// which means it cannot contain any character that would otherwise require encoding
+        /// Additionally, schemes are always returned in lowercase form.
+        fn scheme(self: &QUrl) -> QString;
```
LeonMatthesKDAB, line 75: Maybe also wrap with `scheme_or_default`?

LeonMatthesKDAB (commented):
One actual issue on the `QUrl::query` implementaiton, the rest is mostly nitpicking or opinion-based.

You could consider using a Rust macro for all these `_or_default` methods...

crates/cxx-qt-lib/src/gui/qcolor.rs
```diff
@@ -81,6 +475,24 @@ impl Default for QColor {
...
+        ffi::qcolor_init_from_rust_string(value)
+    }
+}
+
+impl From<&ffi::QString> for QColor {
+    fn from(value: &ffi::QString) -> Self {
+        ffi::qcolor_init_from_qstring(value)
+    }
+}
+
```
LeonMatthesKDAB, line 495: These should probably all be TryFrom implementations and use an invalid check

crates/cxx-qt-lib/src/core/qdate.rs
```diff
@@ -8,26 +8,100 @@ use std::fmt;
...
-        /// Returns the day of the month for this date.
-        fn day(self: &QDate) -> i32;
 
         /// Sets this to represent the date, in the Gregorian calendar, with the given year, month and day numbers.
         /// Returns true if the resulting date is valid, otherwise it sets this to represent an invalid date and returns false.
         #[rust_name = "set_date"]
         fn setDate(self: &mut QDate, y: i32, m: i32, d: i32) -> bool;
+
+        /// Returns the time as a string. The format parameter determines the format of the string.
+        #[rust_name = "format_enum"]
```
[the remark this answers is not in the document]
ahayzen-kdab: So `to_string` would conflict with the Rust trait `ToString` https://doc.rust-lang.org/stable/std/string/trait.ToString.html as this method takes a format enum as well ? That's why i picked a different name, as usually clippy doesn't like that.

Suggestions welcome here.

crates/cxx-qt-lib/src/core/qurl.rs
ahayzen-kdab: Oops :-)

crates/cxx-qt-lib/src/gui/qcolor.rs
```diff
@@ -64,11 +280,189 @@ pub struct QColor {
...
+    pub fn from_hsv_f(h: f32, s: f32, v: f32, a: f32) -> Self {
+        ffi::qcolor_init_from_hsv_f(h, s, v, a)
+    }
+
     /// Constructs a QColor with the RGB value r, g, b, and the alpha-channel (transparency) value of a.
     ///
     /// The color is left invalid if any of the arguments are invalid.
-    pub fn from_rgba(red: i32, green: i32, blue: i32, alpha: i32) -> Self {
-        ffi::qcolor_init_from_rgba(red, green, blue, alpha)
+    pub fn from_rgb(red: i32, green: i32, blue: i32, alpha: i32) -> Self {
```
LeonMatthesKDAB, line 341: I don't think this should stay named `from_rgba`.
Then add an additional `from_rgb` that uses `255` for alpha.
That's closes to rust-style and to Qt-style I think.
As the `fromRgb` function doesn't require you to provide an alpha component, so it's probably meant to be used without it most of the time anyway.

crates/cxx-qt-lib/src/core/qurl.rs
ahayzen-kdab: So the problem is, `QUrl` doesn't have a `hasPassword()` or `isPassword()` method, the only way the docs say is an empty string. So we would have to do `password_or_default().is_empty() -> None`.

However my thought was, what if you have a password but the password is an empty string ? or the same with a username ? As null and empty are different but the `QUrl` API doesn't discriminate between these...

Again suggestions welcome.

LeonMatthesKDAB (commented):
Again, a bit more minor stuff.
Overall, amazing work!

crates/cxx-qt-lib/src/core/qurl.rs
LeonMatthesKDAB: No worries, just made it red to stand out in the bazillion other comments :wink: 
This one's actually important :D

crates/cxx-qt-lib/src/gui/qcolor.rs
ahayzen-kdab: I'm confused what you mean here, i've changed it to `from_rgb`. But do you mean to have both? or do i have it correct ?

crates/cxx-qt-lib/src/core/qurl.rs
LeonMatthesKDAB: Alright, then stick with `password_or_default` for now.
Didn't check whether there was a `hasX` method available for these.

ahayzen-kdab: tbh i think the `QUrl` API itself in Qt isn't great in this aspect as it doesn't consider null vs empty and is generally slightly inconsistent. I'll see which others i can fix though

crates/cxx-qt-lib/src/core/qdate.rs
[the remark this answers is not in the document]
LeonMatthesKDAB: Yeah, I noticed that when I found the `format` method.
So the name can stay I think. We could consider not having two versions of this in the first place, but that's discussed in another comment: https://github.com/KDAB/cxx-qt/pull/416#discussion_r1124498345

crates/cxx-qt-lib/src/gui/qcolor.rs
LeonMatthesKDAB: Basically have: `from_rgba` - the function we have right now, takes `r, g, b, a`
Then also define:
``` rust
pub fn from_rgb(red: i32, green: i32, blue: i32) -> Self {
    Self::from_rgba(red, green, blue, 255)
}
```

Which is kind off how `fromRgb` behaves on the Qt side.
It has a default value of 255 for alpha, so you can call it like this:
`QColor::fromRgb(1, 2, 3)`

crates/cxx-qt-lib/src/core/qtime.rs
```diff
@@ -78,6 +178,21 @@ impl fmt::Debug for QTime {
     }
 }
 
+impl From<i32> for QTime {
```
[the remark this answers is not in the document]
ahayzen-kdab: done

crates/cxx-qt-lib/src/core/qsizef.rs
```diff
@@ -85,6 +145,22 @@ impl fmt::Display for QSizeF {
...
 }
 
+impl From<&ffi::QSize> for QSizeF {
+    /// Constructs a size with floating point accuracy from the given size.
+    fn from(size: &ffi::QSize) -> Self {
+        ffi::qsizef_from_qsize(size)
+    }
+}
+
+impl From<QSizeF> for ffi::QSize {
```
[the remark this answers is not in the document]
ahayzen-kdab: Yeah, i did it here as it uses methods from the CXX ffi block in this module.

crates/cxx-qt-lib/src/core/qstring.cpp
```diff
@@ -54,10 +54,136 @@ qstringToRustString(const QString& string)
...
+}
+
+QByteArray
+qstringToUtf8(const QString& string)
+{
+  return string.toUtf8();
+}
+
+QString
+qstringTrimmed(const QString& string)
```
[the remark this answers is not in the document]
ahayzen-kdab: This comment is in the C++ header like the QByteArray

crates/cxx-qt-lib/src/core/qurl.cpp
ahayzen-kdab: The bigger problem was that `ComponentFormattingOptions` is a QFlags and not an Enum, how do we represent flags with CXX ? Otherwise we could add `ComponentFormattingOptions`.

And yes CXX gets confused with defaults as well, another possible thing to look at. Even if you could specify the default in the Rust definition of the bridge.

crates/cxx-qt-lib/src/core/qurl.rs
ahayzen-kdab: Done

crates/cxx-qt-lib/src/core/qurl.rs
ahayzen-kdab: done

crates/cxx-qt-lib/src/core/qurl.rs
ahayzen-kdab: Done and used the is_empty on the scheme to add a Option<T>

crates/cxx-qt-lib/src/gui/qcolor.rs
ahayzen-kdab: done

crates/cxx-qt-lib/src/core/qdate.rs
```diff
@@ -72,11 +146,74 @@ impl fmt::Debug for QDate {
...
+    pub fn format(&self, format: &ffi::QString) -> ffi::QString {
+        ffi::qdate_to_format(self, format)
+    }
+
+    /// Converts the Julian day jd to a QDate.
+    pub fn from_julian_day(jd: i64) -> Self {
+        Self { jd }
+    }
+
+    /// Returns the QTime represented by the string, using the format given, or an invalid time if the string cannot be parsed.
```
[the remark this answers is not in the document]
ahayzen-kdab: So the `from_string` allows for custom formats https://doc.qt.io/qt-6/qdate.html#fromString-4 eg `QDate::fromString("1MM12car2003", "d'MM'MMcaryyyy");`. Whereas `from_string_enum` is limited to a few known formats https://doc.qt.io/qt-6/qt.html#DateFormat-enum

crates/cxx-qt-lib/src/core/qtime.rs
```diff
@@ -53,10 +120,43 @@ pub struct QTime {
...
+    }
+
+    /// Returns a new QTime instance with the time set to the number of msecs
+    /// since the start of the day, i.e. since 00:00:00.
+    pub fn from_msecs_since_start_of_day(msecs: i32) -> Self {
+        ffi::qtime_from_msecs_since_start_of_day(msecs)
+    }
+
+    /// Returns the QTime represented by the string, using the format given, or an invalid time if the string cannot be parsed.
+    pub fn from_string(string: &ffi::QString, format: &ffi::QString) -> Self {
```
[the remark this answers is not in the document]
ahayzen-kdab: Changed them both to Option<Self>

crates/cxx-qt-lib/src/gui/qcolor.rs
ahayzen-kdab: Hmm but then should we have a `cmyka` and a `hsva` and a `hsla`  ? I ended up copying the Qt naming of dropping the `a` as not required. Then it's unfortunate that rust can't have default values (?) eg we can't do `a = 255` ?

crates/cxx-qt-lib/src/core/qbytearray.cpp
```diff
@@ -161,5 +161,29 @@ qbytearrayResize(QByteArray& byteArray, ::rust::isize size)
...
+{
+  return byteArray.simplified();
+}
+
+QByteArray
+qbytearrayToLower(const QByteArray& byteArray)
+{
+  return byteArray.toLower();
+}
```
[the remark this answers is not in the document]
ahayzen-kdab: Split into #463

crates/cxx-qt-lib/src/core/qurl.rs
ahayzen-kdab: Hmm https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.1 does mention about an empty password `unless the data after the colon is the empty string (indicating no password).`

crates/cxx-qt-lib/src/core/qurl.rs
ahayzen-kdab: So i've not found yet in the spec ( https://datatracker.ietf.org/doc/html/rfc3986 ) if the user name can be empty or not...

LeonMatthesKDAB:
```
The user information, if present, is followed by a
   commercial at-sign ("@") that delimits it from the host.
```
It says "if present" about the user info.
To me that suggests the user may be non-existent.
Not sure whether that means the user name should be "empty" though..
So question is whether:
`https://@test.com` or `https://:@test.com` is a valid URL? :thinking:

crates/cxx-qt-lib/src/gui/qcolor.rs
LeonMatthesKDAB: Hm, right, I think adding `cmyka`, `hsva`, `hsla`, etc. as well makes sense.

Reasoning:
1. Rust doesn't have optional arguments, the closest thing is this:
    - http://xion.io/post/code/rust-optional-args.html
    - However, it still forces users to write `None` often and is discouraged by Rust maintainers: https://www.reddit.com/r/rust/comments/556c0g/comment/d88m93j/?utm_source=share&utm_medium=web2x&context=3
2. It's a lot more explicit. Having `rgb` take 4 paramter is in itself pretty strange. We're just kind off used to knowing that the alpha value belongs there as well.
3. In Qt most calls to Qt::fromRgb probably only take 3 values, the alpha value being defaulted. So if we want to stick to that same API, `from_rgb` and friends need to accept 3 values only. Providing an alpha value is already the special case (as can be seen in our tests as well).

crates/cxx-qt-lib/src/core/qdate.rs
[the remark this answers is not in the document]
LeonMatthesKDAB: Ah, alright, I thought it was using the string to select a DateFormat. My bad.

Then I guess this can stay this way, with the Option<Self> return value.
Still not a great fan of the `_enum` suffix, but I don't have any better ideas either...

crates/cxx-qt-lib/src/core/qurl.rs
LeonMatthesKDAB: Then it seems there's no way to differentiate between an empty and an invalid password by Qts API, so let's just stick with `password_or_default`.

crates/cxx-qt-lib/src/core/qurl.rs
LeonMatthesKDAB: I'm trying different URL validators (mostly based on regexes) and they don't really come to a consensus.
Most accept URLs in the format `https://:@test.com` though :thinking: 
So I don't think we can assume empty usernames are invalid...
So like with password, let's stick with the _or_default variant for name.

crates/cxx-qt-lib/src/gui/qcolor.rs
ahayzen-kdab: Done adding two methods.

LeonMatthesKDAB (approved):
LGTM :partying_face:

PR #418, opened 2023-01-23
Implement {Partial,}Ord for selected Qt types

vimpostor:
This completes the third part of #53.


crates/cxx-qt-lib/src/lib.rs
```diff
@@ -8,8 +8,19 @@
...
+
 pub use crate::core::*;
 
+fn get_ordering(ord: i8) -> Option<Ordering> {
+    match ord {
+        -1 => Some(Ordering::Less),
+        0 => Some(Ordering::Equal),
+        1 => Some(Ordering::Greater),
+        _ => None,
+    }
```
vimpostor, line 21: Is there a way to register `std::cmp::Ordering` for C++ so that we can avoid this conversion?
I know **cxx** has `extern "Rust"`, but I wasn't able to get it working to register this type from the STL.

ahayzen-kdab: Maybe use a Shared Enum https://cxx.rs/shared.html then impl From for it do that you can go .into() from the CXX / C++ enum to `std::cmp::Ordering` ?

ahayzen-kdab (commented):
Looks good so far, see the discussion, i wonder if you can use a shared enum them just have an From for it to jump to Ordering.

Then what about container types ? Eg looks like QList<T> would support some kind of order ? (note use the `generate.sh` file in each of the folders)

crates/cxx-qt-lib/src/lib.rs
vimpostor: But if I understand correctly shared enums only work for user-defined enums. I don't see any way to register the already existing `std::cmp::Ordering` or am I missing something?

ahayzen-kdab: Right, i can't see a sane way to register an existing Rust enum to C++, it'd have to be an opaque type.

So you would have to create a second shared enum in the bridge

```rust
enum CXXOrdering { Less, Equal, Greater }
```

Then do like `impl From<CXXOrdering> for Order { ... }`.

However this would need to be done somewhere common to avoid creating one in each bridge. So maybe just returning a number and using a helper method is fine for now :-)

crates/cxx-qt-lib-headers/include/common.h
```diff
@@ -51,5 +51,12 @@ operatorEq(const T& a, const T& b)
   return a == b;
 }
 
+template<typename T>
+int8_t
```
ahayzen-kdab, line 55: elsewhere we have been using fully qualified types to avoid any namespace fun later, so `::std::int8_t` ?

ahayzen-kdab, line 55: Also add a `#include <cinttypes>` to the top of the file.

crates/cxx-qt-lib/src/core/qpoint.rs
```diff
@@ -45,7 +45,7 @@ mod ffi {
 }
 
 /// The QPoint struct defines a point in the plane using integer precision.
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
```
ahayzen-kdab, line 48: So `QPoint` and some others don't have ordering in C++ ( ? ), so if the C++ side doesn't have ordering I don't think we should derive it on the Rust side ? As what does this ordering mean for QPoint ? is (0, 1) < (1, 0) ? as what is "less" ? bottom left in a Cartesian coordinate system ? 

Then for others like QRect vs QRectF they are stored internally in different ways by Qt, so not sure how this could be consistent with derive.

vimpostor: I agree with you, I have removed the derives for those types.

vimpostor:
> Then what about container types ? Eg looks like QList would support some kind of order ? (note use the `generate.sh` file in each of the folders)

That would be possible, but I wonder if we really want that. For example that would require all value types of `QList<T>` to implement `PartOrd`, what if in the future we want to use the list with a type that doesn't have that?

ahayzen-kdab:
> 

For the QList i think you would have like `impl<T> Ord for QList<T> where T: QListElement + Ord` or something ? Then yes the inner types might need an `unreachable` or something on the macro and then skip the methods in the bash script.

But lets split that into a separate pull request.

ahayzen-kdab (approved):
LGTM :-)

PR #419, opened 2023-01-24
cxx-qt-lib: various fixes before adding methods to types

ahayzen-kdab:
Related to #55


crates/cxx-qt-lib/src/core/mod.rs
```diff
@@ -48,6 +54,9 @@ pub use qstring::QString;
 mod qstringlist;
 pub use qstringlist::QStringList;
 
+mod qt;
+pub use qt::{AspectRatioMode, CaseSensitivity, DateFormat, SplitBehaviorFlags};
```
ahayzen-kdab, line 58: how should we handle these they are normally namespaced to `Qt::AspectRatioMode`?

crates/cxx-qt-lib/src/core/qmarginsf.rs
```diff
@@ -0,0 +1,101 @@
...
+        #[doc(hidden)]
+        #[rust_name = "qmarginsf_to_qstring"]
+        fn toQString(value: &QMarginsF) -> QString;
+    }
+}
+
+/// The QMarginsF class defines the four margins of a rectangle.
+#[derive(Debug, Clone, PartialEq)]
+#[repr(C)]
+pub struct QMarginsF {
```
LeonMatthesKDAB, line 52: Hm, I'd just like to double-check whether it makes sense to expose these members as `pub` :thinking: 

It definitely leads to more rust-y code.
However, these members are not public in Qt's C++ code.
So could a new Qt minor version change the meaning of these members without breaking backwards compatibility?
I think not, because the types must be binary-compatible, but unsure :thinking: 
As the members aren't exposed, they would also have to be constructed by Qt code, which could change whatever each field means.
e.g. change it to `left`, `bottom`, `widht` and `height`. which wouldn't change the binary representation.

So do we want to risk this happening? It's not very likely though I'd guess.

crates/cxx-qt-lib/src/core/mod.rs
LeonMatthesKDAB: I think this approach is fine.
Instead of namespacing with `Qt::AspectRatioMode`, it would be `cxx_qt_lib::AspectRatioMode` or `cxx_qt_lib::qt::AspectRationMode`.
So basically, keep "top-level" Qt types top-level.

crates/cxx-qt-lib/src/core/qpersistentmodelindex.cpp
```diff
@@ -8,8 +8,6 @@
 
 #include "../assertion_utils.h"
 
-// The layout has changed between Qt 5 and Qt 6
```
LeonMatthesKDAB, line 11: Good catch :+1:

crates/cxx-qt-lib/src/core/qmarginsf.rs
ahayzen-kdab: Right, maybe for now we just map the Qt public API across and not the internals ? And do the same for methods that are weird in Qt (eg keep the Qt API for now for things like QRect). Then in a future cycle we can look at exposing the internal fields and improving the methods ?

We also need to be careful that in some types we use derive, but i think for Eq / Ord it should be mainly OK ?

ahayzen-kdab: done

crates/cxx-qt-lib/src/core/qt.rs
```diff
@@ -0,0 +1,48 @@
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
```
LeonMatthesKDAB, line 1: 2023 ;)

crates/cxx-qt-lib/src/core/qset/qset_qpersistentmodelindex.rs
```diff
@@ -0,0 +1,76 @@
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
```
LeonMatthesKDAB, line 1: This should maybe also be updated to 2023 :thinking: 
That does mean we have to adjust the generate.sh though, which will change it in all files, even the ones we didn't touch. :/

ahayzen-kdab: These are effectively "binary" / machine generated files rather than source files, so the `generate.sh` is the one that needs the correct copyright. And worst case the date is a start date so it's probably not worth the effort :thinking:

crates/cxx-qt-lib/src/core/qt.rs
ahayzen-kdab: Done :-)

crates/cxx-qt-lib/src/core/qset/qset_qpersistentmodelindex.rs
LeonMatthesKDAB: Yeah, would assume an earlier date here is better if in doubt :)

crates/cxx-qt-lib/src/core/qmargins.rs
```diff
@@ -0,0 +1,102 @@
...
+        #[doc(hidden)]
+        #[rust_name = "qmargins_to_qstring"]
+        fn toQString(value: &QMargins) -> QString;
+    }
+}
+
+/// The QMargins class defines the four margins of a rectangle.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[repr(C)]
+pub struct QMargins {
```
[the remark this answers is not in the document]
LeonMatthesKDAB: We also don't have QSet/QVector support for these yet.
We should probably decide on whether to add everything for these containers for every type in cxx-qt-lib or only select ones.

QVariant I think we should support for everything. But i.e. `QSet<QMargins>` isn't really a common use-case...

LeonMatthesKDAB (changes requested):
Let's make sure we have support for our generics down.

crates/cxx-qt-lib/src/core/qmargins.rs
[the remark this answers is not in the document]
ahayzen-kdab: So QMargins(F) is not in QVariant in Qt, so i wondered if we should bother adding it too ? Maybe I can add it anyway...

And QMargins(F) is not hashable so shouldn't be in QSet/QHash, but could be added to QList/QVector.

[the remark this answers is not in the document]
ahayzen-kdab: So adding support in QVariant requires that we declare `Q_DECLARE_METATYPE` for `QMargins` which doesn't exist by default, otherwise `canConvert` doesn't work. So should we declare this or skip it like Qt does?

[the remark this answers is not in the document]
LeonMatthesKDAB: If Qt doesn't have it, skip it.
I just didn't know Qt didn't support it by default xD.

Then I think it's also not worth the compile-time tradeoff to support these with QList/QVector if Qt doesn't even bother adding them to QVariant.

LeonMatthesKDAB (approved):
LGTM

PR #420, opened 2023-01-24
WIP: cxx-qt-lib: add QTimeZone and QDateTime methods

ahayzen-kdab:
Requires #416


tests/qt_types_standalone/rust/src/qtimezone.rs
```diff
@@ -0,0 +1,32 @@
...
+
+    extern "Rust" {
+        fn construct_qtimezone() -> QTimeZone;
+        fn read_qtimezone(t: &QTimeZone) -> bool;
+        fn clone_qtimezone(t: &QTimeZone) -> QTimeZone;
+    }
+}
+
+fn construct_qtimezone() -> QTimeZone {
+    QTimeZone::from_iana(&QByteArray::from("Europe/London"))
```
ahayzen-kdab, line 23: valgrind is failing somewhere here...

Appears to be related to the inner QSharedDataPointer being set by a non default QTimeZone (when default it is nullptr), then if that QTimeZone is given to C++ it leaks.

If a non default QTimeZone is constructed in Rust and given to C++, if C++ then calls tz.~QTimeZone() then the memory is freed correctly thinking.

And if QTimeZone is constructed in C++ and std::move'd this does not leak.

ahayzen-kdab: Fixed by changing to an opaque type.

This was due to QTimeZone having a copy constructor but not a move so then "When the move constructor is not implicitly declared or explicitly supplied, expressions that otherwise would have invoked the move constructor may instead invoke a copy constructor. " is hit. Which causes QSharedDataPointer to be incremented and leak.

ahayzen-kdab:
Fixed into #416

PR #423, opened 2023-01-27
Implement {PartialEq,Eq} for container types

vimpostor:
Another one for #53

As discussed, `Ord` probably doesn't make much sense for container types, but `Eq` does (as long as the element types are `Eq` as well).


ahayzen-kdab (approved):
LGTM for the list types, what about the maps ?

vimpostor:
> what about the maps ?

Ah forgot about maps, I can add that in the next PR.

Issue #292, opened 2022-10-03
Decide on useful crates to add as integration features to cxx-qt-lib 


ahayzen-kdab:
Decide on useful crates to add as integration features to cxx-qt-lib.

Any common Rust crates that are either commonly used for parsing data (eg serde) or crates that help with certain types of data (eg colours, or points, rects). We could add any trait implementations to easily go to/from the crate type or implement any traits that parse the data.

These should then be behind feature flags so that `cxx-qt-lib` can be used both with and without serde etc.

- [x] ~`serde` as a feature then impl (de)serialise for types - #425~ - Skipping for 0.5 we can reconsider later if it's worth it
- [x] `rgb` as a feature for QColor - #424 
- [x] `http` ~/ `hyper`~ Uri as a feature for QUrl - #424 
- [x] `url` as a feature for QUrl - #424 
- [x] `time` as a feature Q{Date,Time,DateTime} - #454
- [x] `chorno` as a feature for Q{Date,Time,DateTime} - #454
- [x] `bytes` as a feature for QByteArray (could be lifetime fun here) - #424 
- [ ] ...

#[cfg_attr(feature = "serde_support", derive(Serialize, Deserialize))] allows for conditional derive https://doc.rust-lang.org/reference/conditional-compilation.html This could be used for "trivial" types like QPoint where the struct on the Rust side represents the object correctly.

ahayzen-kdab:
Closed via #454

PR #425, opened 2023-02-01
WIP: cxx-qt-lib: add serde support for types

ahayzen-kdab:
Related to #292

Requires #416


crates/cxx-qt-lib/src/core/qdate.rs
```diff
@@ -48,6 +48,7 @@ mod ffi {
 
 /// The QDate class provides date functions.
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[repr(C)]
 pub struct QDate {
```
ahayzen-kdab, line 53: should we serialise as a number or the d,m,y dictionary?

crates/cxx-qt-lib/src/core/qtime.rs
```diff
@@ -47,6 +47,7 @@ mod ffi {
 
 /// The QTime class provides clock time functions.
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[repr(C)]
 pub struct QTime {
```
ahayzen-kdab, line 52: should we serialise as a number or the h,m,s dictionary?

crates/cxx-qt-lib/src/core/qrect.rs
```diff
@@ -56,6 +56,7 @@ mod ffi {
 
 /// The QRect struct defines a rectangle in the plane using integer precision.
 #[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[repr(C)]
 pub struct QRect {
```
ahayzen-kdab, line 61: should we use the Qt internals here, or add our own x,y,width,height ?

crates/cxx-qt-lib/src/core/qrectf.rs
```diff
@@ -56,6 +56,7 @@ mod ffi {
 
 /// The QRectF struct defines a rectangle in the plane using floating point precision.
 #[derive(Debug, Clone, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[repr(C)]
 pub struct QRectF {
```
ahayzen-kdab, line 61: should we use the Qt internals here, or add our own x,y,width,height ?

crates/cxx-qt-lib/src/core/qbytearray.rs
```diff
@@ -251,3 +251,80 @@ unsafe impl ExternType for QByteArray {
...
+        A: serde::de::SeqAccess<'de>,
+    {
+        // TODO: can we construct a QByteArray in a better way?
+        //
+        // Note there is serialize_bytes and deserialize_bytes but they
+        // don't seem to work consistently? So creating a Vec<u8> appears to be better?
+        let mut bytes = QByteArray::default();
+        bytes.reserve(seq.size_hint().unwrap_or(0) as isize);
+
+        while let Some(value) = seq.next_element::<u8>()? {
```
ahayzen-kdab, line 280: is this the best way to represent QByteArray?

LeonMatthesKDAB:
Side note whilst skimming this:
It seems this brings up the issue of how stable the binary representation of certain Qt classes is (i.e. QPoint).
We should really figure this out first. Does Qt guarantee, by virtue of guaranteeing binary compatibility, that e.g. the layout of QPoint can't change?

If so, we can also keep making members of these classes public, as we don't have to worry about their order changing between minor versions.

ahayzen-kdab:
> Side note whilst skimming this: It seems this brings up the issue of how stable the binary representation of certain Qt classes is (i.e. QPoint). We should really figure this out first. Does Qt guarantee, by virtue of guaranteeing binary compatibility, that e.g. the layout of QPoint can't change?
> 
> If so, we can also keep making members of these classes public, as we don't have to worry about their order changing between minor versions.

Right, I think ones like QPoint are unlikely to change, but as we've also seen while investigating QTimeZone that has changed it's layout in future Qt versions.

Eg with `QSharedDataPointer` https://github.com/qt/qtbase/blob/v6.4.2/src/corelib/time/qtimezone.h#L152 becoming a `Data` (a union with largest size of pointer) https://github.com/qt/qtbase/blob/v6.5.0-beta2/src/corelib/time/qtimezone.h#L242 so i think still the same size ?

crates/cxx-qt-lib/src/core/qdate.rs
ahayzen-kdab: Yes gone with this. Although what format should be used? i've used `{"year":2023,"month":1,"day":1}` for now but is there a standard? like should it be y,m,d ? (and same for QTime, QTimeZone, QSize eg width or w etc).

crates/cxx-qt-lib/src/core/qrect.rs
ahayzen-kdab: Changed to using the public API.

crates/cxx-qt-lib/src/gui/qcolor.rs
```diff
@@ -9,31 +9,247 @@ use std::mem::MaybeUninit;
 
 #[cxx::bridge]
 mod ffi {
```
ahayzen-kdab, line 11: need to add serde support for QColor, either with all the formats as well or converting to one of them?

ahayzen-kdab:
Going to pause for 0.5 and consider if we do want to maintain this. As any change in the serde format would break 1.0 release. And if the developer can easily change it to their own types with To/From do we even need to mantain this?

LeonMatthesKDAB:
Hm, it sounds like a lot to maintain... I honestly doubt it's worth it.

Especially as we don't control the internals of these types, it's hard to specify what each one should take as fields, etc.

Maybe we should just slap some `#[serde(from=...)]` attributes on the most basic types like QString and that's good enough?
E.g. anything that has a "Rust equivalent", we can just reuse their serialization. And anything that Rust doesn't have an equivalent for, we leave up to users?

ahayzen-kdab:
Right, we already have QString serde support so maybe that is enough, and i also note that Qt JSON objects only support some types you need to handle the rest yourself. So maybe this is "enough" for now.

PR #428, opened 2023-02-07
register QML elements at build time

Be-ing:
by specifying `#[cxxqt::qobject(qml_uri = "foo.bar", qml_verison = "1.0")]`

TODO:
 * [x] add tests
 * [x] code documentation
 * [x] user documentation


vimpostor:
> :crab:  Users don't need to write any C++ to write a QML application.

Thanks for shaving this yak, this is really impressive! :crab:

Ironically, cxx-qt is now even better than the official Qt cmake API, as this is also now working for Qt5. :tada: (i.e. at least the fully declarative type declaration, not sure if qmltc and qmlsc will also be possible)

book/src/getting-started/3-qml-gui.md
```diff
@@ -33,4 +33,16 @@ It is again important to emphasize here that `MyObject` is just another QObject
 The only difference being that any invokable functions that are defined are defined in Rust, instead of C++.
 For QML, this doesn't make a difference though.
 
-But enough of that, let's get this project [building and running](./5-cmake-integration.md).
+# Qt resources
+
+To include the `main.qml` file inside the application, use the [Qt resource system](https://doc.qt.io/qt-6/resources.html) by listing it in a `qml.qrc` file in the `qml` folder:
```
vimpostor, line 38: The Qt cmake API `qt_add_qml_module()` has a neat option `QML_FILES` where you can list all QML files and it will automatically generate the qrc files, so that the user doesn't have to create any qrc files manually. I wonder how hard it is to add this feature to cxx-qt as well

Maybe even add all QML files automatically to a qrc file (without requiring the user to list them anywhere), or would there be any downside to doing it fully automatically?

book/src/getting-started/4-cmake-integration.md
```diff
@@ -10,9 +10,27 @@ SPDX-License-Identifier: MIT OR Apache-2.0
...
+
+## Using Rust QObjects in C++
+
+For every `#[cxx_qt::bridge]` that we define in Rust, CXX-Qt will generate a corresponding C++ header file.
+They will always be in the `cxx-qt-gen/` include path and use the snake_case naming convention.
+The name of the header file will be the name of the Rust module of your `#[cxx_qt::bridge]`, followed by `.cxxqt.h`.
+So in our case: `#include cxx-qt-gen/my_object.cxxqt.h`
+
+Including the generated header allows accessing the `MyObject` C++ class, just like any other C++ class.
+Inhert from it, connect signals and slots to it, put it in a QVector, do whatever you want with it.
```
vimpostor, line 32: typo
```diff
-Inhert
+Inherit
```

book/src/getting-started/4-cmake-integration.md
```diff
@@ -44,10 +62,9 @@ We'll then also need to add a script named `build.rs` next to our `Cargo.toml`:
 ```rust,ignore
 {{#include ../../../examples/qml_minimal/rust/build.rs:book_build_rs}}
 ```
-This is what generates the C++ code for our `MyObject` class at compile-time.
-It will output the `cxx-qt-gen/include/my_object.h` file we included earlier in `main.cpp`.
+This is what generates and compiles the C++ code for our `MyObject` class at build time.
 
-Note that all Rust source files that uses the `#[cxx_qt::bridge]` macro need to be included in this script!
+Every Rust source files that uses the `#[cxx_qt::bridge]` macro need to be included in this script.
```
vimpostor, line 67:
```diff
-Every Rust source files that uses the `#[cxx_qt::bridge]` macro need to be included in this script.
+Every Rust source file that uses the `#[cxx_qt::bridge]` macro needs to be included in this script.
```

vimpostor (changes requested):

book/src/getting-started/3-qml-gui.md
vimpostor: Right, it's a minor thing, so let's consider it for a PR later (if ever)

examples/qml_minimal/CMakeLists.txt
```diff
@@ -60,16 +60,21 @@ corrosion_set_env_vars(${CRATE}
     "QMAKE=${QMAKE}"
 )
 
+# Create an INTERFACE library target to link libraries to and add include paths.
+# Linking this to both the application and the tests avoids having to setup
+# the include paths and linked libraries for both of those.
+add_library(${CRATE}_interface INTERFACE)
```
LeonMatthesKDAB, line 66: This actually is very similar to how Rust organizes executable building:
1. create a `lib` target that you link every dependency to (Qt, our Rust code, etc.)
2. Link that `lib` to both the executable, including the main file, as well as any tests.

So we should use that same analogy to explain what you're doing here, as that should be very easy to understand for any Rust developer.

crates/cxx-qt-gen/src/parser/cxxqtdata.rs
```diff
@@ -71,8 +74,62 @@ impl ParsedCxxQtData {
...
+                            qml_metadata = Some(QmlElementMetadata {
+                                uri: qml_uri.value(),
+                                name,
+                                version_major,
+                                version_minor,
+                                uncreatable: qml_uncreatable.is_some(),
+                                singleton: qml_singleton.is_some(),
+                            });
+                        }
+                        (Some(_), None) => panic!("qml_uri specified but no qml_version specified"),
```
LeonMatthesKDAB, line 113: Instead of panicing here, return a syn::Result instead with a syn::Error:
ex.:
```
use syn::Error;
Err(Error::new(
    /*qml_uri attribute span*/,
    "qml_uri specified but no qml_version specified",
));
```
Same with the other cases.

crates/cxx-qt-gen/src/parser/cxxqtdata.rs
```diff
@@ -71,8 +74,62 @@ impl ParsedCxxQtData {
...
+                    )?;
+
+                    // Parse QML metadata
+                    let mut qml_metadata = None;
+                    let qml_uri = attrs_map.get(&quote::format_ident!("qml_uri"));
+                    let qml_version = attrs_map.get(&quote::format_ident!("qml_version"));
+                    let qml_name = attrs_map.get(&quote::format_ident!("qml_name"));
+                    let qml_uncreatable = attrs_map.get(&quote::format_ident!("qml_uncreatable"));
+                    let qml_singleton = attrs_map.get(&quote::format_ident!("qml_singleton"));
+                    match (qml_uri, qml_version) {
```
LeonMatthesKDAB, line 89: Please extract this into a separate function.

LeonMatthesKDAB (changes requested):
A few more cleanups to be done, but overall approach is good! :+1: 
We can finally claim that you don't need C++ anymore :partying_face:

LeonMatthesKDAB:
> I'd like to add a C++ generation test in cxx-qt-gen, but I don't see how to write the output files? Is there a tool for this somewhere that I'm missing?

No, unfortunately not, the easiest way is to start with an empty file, which is gonna give you the complete diff to what is currently being generated.
Then just copy the diff into the file :D

crates/cxx-qt-build/src/lib.rs
```diff
@@ -395,26 +395,82 @@ impl CxxQtBuilder {
...
         // Run moc on C++ headers with Q_OBJECT macro
         for qobject_header in self.qobject_headers {
-            self.cc_builder.file(qtbuild.moc(&qobject_header));
+            let moc_products = qtbuild.moc(&qobject_header.path);
+            self.cc_builder.file(moc_products.cpp);
+            for qml_metadata in qobject_header.qml_metadata {
+                self.cc_builder.define("QT_STATICPLUGIN", None);
+                qml_modules
+                    .entry((
+                        qml_metadata.uri.clone(),
```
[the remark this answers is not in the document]
LeonMatthesKDAB: let's add this in another PR though to keep this one small, agree?

book/src/concepts/register_types.md
```diff
@@ -1,38 +0,0 @@
-<!--
```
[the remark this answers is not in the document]
LeonMatthesKDAB: Right, it might make sense to re-add it if we allow you to specify a default for qml_uri and friends, so that we have a place to explain the different options (qml_elemnt, qml_uri, qml_version, etc.).

But fine to delete for now.

crates/cxx-qt-build/src/lib.rs
```diff
@@ -395,26 +395,82 @@ impl CxxQtBuilder {
...
         }
-
-        // Generate code from .qrc files, but do not compile it. Instead, the user needs to #include them
-        // in a .cpp file. Otherwise, MSVC won't link if the generated C++ is built separately.
         for qrc_file in self.qrc_files {
-            qtbuild.qrc(&qrc_file);
+            cc_builder_whole_archive.file(qtbuild.qrc(&qrc_file));
+            cc_builder_whole_archive_files_added = true;
+        }
+        if cc_builder_whole_archive_files_added {
```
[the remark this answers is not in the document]
LeonMatthesKDAB: We could use the newtype pattern here to ensure the two variables don't get out of sync.

``` rust
struct ConditionalCCBuilder {
    cc_builder: ...,
    added: bool,
}

impl ConditionalCCBuilder {
    fn new(cc_builder: ...) -> Self {
        Self { cc_builder, added: false }    
    }
    fn file(&mut self, path: PathBuf) {
        self.cc_builder.file(path);
        self.added = true; 
    } 
    fn compile(&mut self) {
         if self.added {
            self.cc_builder.compile(); 
        }
    }
}
```

Maybe that's overengineering though xD

LeonMatthesKDAB:
> >    We should probably work on such a tool at some point though, that just updates the test_output files.
> 
> How about adding that in another PR?

Of course, shouldn't be part of this PR.

LeonMatthesKDAB (approved):
LGTM, how about @vimpostor ?

vimpostor (approved):

PR #430, opened 2023-02-08
Qt-build-utils: Use `/get` variables when querying Qt paths

LeonMatthesKDAB:


LeonMatthesKDAB:
Here's the output from the yocto devshell:

[qmake-query.txt](https://github.com/KDAB/cxx-qt/files/10695146/qmake-query.txt)

It's somewhat strange, but seems to follow broadly this logic:
- Variables without a suffix point to where the target sysroot is being assembled or behave like the `/raw` variant if that doesn't exist - I think these are compiled into qmake directly, therefore have to already point to the final output.
- Any variable with `/raw` will point to the final installation directory as seen by the target once built and deployed (i.e. /usr/bin).
- The `/get` versions point to the native sysroot, meaning the yoctosmasher in our case.

So in this case of cross-compilation `/get` seems the right way to go as that points to the executables used for compilation, not the executables that will be deployed.

For comparison, output on my normal Fedora installation:
[qmake-query-fedora.txt](https://github.com/KDAB/cxx-qt/files/10695235/qmake-query-fedora.txt)

This has neither `/raw`, nor `/get` variants, so those shouldn't interfere.

LeonMatthesKDAB:
Unfortunately I don't think that will work.

Even if we chose `QT_INSTALL_BINS`, that would result in:
`QT_INSTALL_BINS:/yocto/imx8-cxx-qt-demo/build/tmp/work/armv8a-fslc-linux/demo-threading/0.1.0.AUTOINC+c645cc12cb-r0/recipe-sysroot/usr/bin`

However, the correct path is:
`QT_HOST_BINS/get:/yocto/imx8-cxx-qt-demo/build/tmp/work/armv8a-fslc-linux/demo-threading/0.1.0.AUTOINC+c645cc12cb-r0/recipe-sysroot-native/usr/bin`

The difference is between `recipe-sysroot` and `recipe-sysroot-native`.

I think your assumption of prefering `HOST` for cross-compilation is absolutetly correct.
However, with Yocto there's a separate level of this.
Yocto doesn't *just* cross-compile, but also creates it's own `sysroot` for each package to be compiled, so everything is compiled in isolation.
Usually `QT_HOST_BINS` would be in `/usr/bin`, but because Yocto builds its own sysroots for every package, it ends up somewhere else.
And because the qmake executable is copied into multiple different compilation sysroots, it can't use the compiled-in `QT_HOST_BINS`, because then you'd have to rebuild qmake for every sysroot that's created.

Which I think is why the `/get` exists in the first place.
It's for exactly this use-case, where qmake is moved into a different `HOST_BIN`, for whatever reason.
Which is why it then has to "get" those variables from somewhere else.

PR #444, opened 2023-02-13
cxx-qt-gen: qml_singleton support

### LINKED ISSUE #25: Have the ability to mark an object as singleton
Add option to mark an option as a singleton in the macro attribute, so make_object(singleton = true) or something.

Once #22 is done this might be `cxx_qt(QObject, singleton = true)` or `cxx_qt(QObject(singleton = true))`. Consider if we can hint the QML versions and name here as well.

ahayzen-kdab:
Closes #25

### DIFF
--- a/crates/cxx-qt-gen/src/generator/cpp/qobject.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/qobject.rs
@@ -37,11 +37,11 @@ impl GeneratedCppQObjectBlocks {
                 "Q_CLASSINFO(\"QML.Element\", \"{}\")",
                 qml_metadata.name
             ));
-            // TODO untested
+
             if qml_metadata.uncreatable {
                 qml_specifiers.push("Q_CLASSINFO(\"QML.Creatable\", \"false\")".to_owned());
             }
-            // TODO untested
+
             if qml_metadata.singleton {
                 qml_specifiers.push("QML_SINGLETON".to_owned());
             }
@@ -142,6 +142,7 @@ mod tests {
         assert_eq!(cpp.cxx_qt_thread_ident, "MyObjectCxxQtThread");
         assert_eq!(cpp.namespace_internals, "cxx_qt_my_object");
         assert_eq!(cpp.base_class, "QObject");
+        assert_eq!(cpp.blocks.metaobjects.len(), 0);
     }
 
     #[test]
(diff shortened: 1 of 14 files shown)


examples/qml_features/rust/src/singleton.rs
```diff
@@ -0,0 +1,28 @@
...
+
+// ANCHOR: book_macro_code
+#[cxx_qt::bridge(cxx_file_stem = "rust_invokables")]
+pub mod ffi {
+    unsafe extern "C++" {
+        include!("cxx-qt-lib/qcolor.h");
+        type QColor = cxx_qt_lib::QColor;
+    }
+
+    #[cxx_qt::qobject(qml_uri = "com.kdab.cxx_qt.demo", qml_version = "1.0", qml_singleton)]
```
ahayzen-kdab, line 14: We get `error: 'QML_SINGLETON' does not name a type`, probably a missing include?

ahayzen-kdab: Yes adding `include!(<QtQml/QQmlEngine>);` makes this work, now we just need it to be injected at the right time

Be-ing: I worked around this for the other C++ macros by putting what they expand to into our generated C++, but the [QML_SINGLETON C++ macro](https://codebrowser.dev/qt6/qtdeclarative/src/qml/qml/qqmlregistration.h.html#_M/QML_SINGLETON) is a bit more complicated. I didn't figure out how to get moc to properly handle an `#include <QQml/qqmlregistration.h>` as noted in the [comment](https://github.com/KDAB/cxx-qt/blob/1a4f5aaac1a7714aaa3e711820bd6af74032a9c4/crates/cxx-qt-gen/src/generator/cpp/qobject.rs#L33).

ahayzen-kdab: Yup, i've got this working locally now, just writing some tests :-) it's simple enough to add `include!(<QtQml/QQmlEngine>);` when we need it.

examples/qml_features/rust/src/singleton.rs
```diff
@@ -0,0 +1,28 @@
+// SPDX-FileCopyrightText: 2023 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+// ANCHOR: book_macro_code
+#[cxx_qt::bridge(cxx_file_stem = "rust_singleton")]
+pub mod ffi {
+    // TODO: automatically inject this when singleton is specified?
```
Be-ing, line 9: yes I think that would be a better solution

examples/qml_features/rust/src/singleton.rs
ahayzen-kdab: Done

examples/qml_features/rust/src/singleton.rs
ahayzen-kdab: Done

examples/qml_features/rust/src/singleton.rs
Be-ing: Interesting. I don't know why `#include <QQml/qmlregistration.h` didn't work because that's where the C++ macros are defined. I'm wondering if we really need to pull in everything from QQmlEngine. Does `#include <QQml/qqml.h>` work?

crates/cxx-qt-gen/src/generator/cpp/qobject.rs
```diff
@@ -37,11 +37,11 @@ impl GeneratedCppQObjectBlocks {
                 "Q_CLASSINFO(\"QML.Element\", \"{}\")",
                 qml_metadata.name
             ));
-            // TODO untested
+
             if qml_metadata.uncreatable {
                 qml_specifiers.push("Q_CLASSINFO(\"QML.Creatable\", \"false\")".to_owned());
             }
```
Be-ing, line 43: It looks like there's only a test for generating the C++ for this. We could use a test actually using it with QML.

examples/qml_features/rust/src/singleton.rs
ahayzen-kdab: You probably need the template from here ? https://codebrowser.dev/qt5/qtdeclarative/src/qml/qml/qqmlengine.cpp.html#1399

And the documentation for `QML_SINGLETON` comes from `QQml/QQmlEngine` so that's the best one to include as it's the documented way rather than assuming it'll be in the right headers.

crates/cxx-qt-gen/src/generator/cpp/qobject.rs
ahayzen-kdab: Look in `examples/qml_features/tests/tst_singleton.qml` :-)

ahayzen-kdab: Oh you are meaning for uncreatable not singleton ... :-)

examples/qml_features/rust/src/singleton.rs
Be-ing: > And the documentation for QML_SINGLETON comes from QQml/QQmlEngine so that's the best one to include as it's the documented way rather than assuming it'll be in the right headers.

Okay.

crates/cxx-qt-gen/src/generator/cpp/qobject.rs
ahayzen-kdab: Hmm need to think of a sane way to test and show uncreatable :thinking:

Be-ing: You could try using a `qml_uncreatable` QObject as a `#[qproperty]` of another QML element.

ahayzen-kdab: Yup, it just then needs to be a raw pointer and the value needs to be created somewhere. Not sure if it's worth showing that or not. I've got a unit test itself checking that trying to create it fails though

Be-ing:
Should we allow specifying the reason the QML element is uncreatable? `#[cxx_qt::qobject(qml_uncreatable = "blahblah can't be created in QML")]`

ahayzen-kdab:
> Should we allow specifying the reason the QML element is uncreatable? `#[cxx_qt::qobject(qml_uncreatable = "blahblah can't be created in QML")]`

Maybe but can be a separate issue and probably not worth it. QML_UNCREATABLE is already such a small usecase.

Be-ing (approved):

Issue #442, opened 2023-02-12
Link library whole archive doesn't work with CMake Ninja generator


ahayzen:
If you use `-G Ninja` then it doesn't understand the `WHOLE_ARCHIVE` workaround ? Is there a way to make this work with Ninja too (seems some use `--whole-archive`?) or do we need to look for an alternate route to `WHOLE_ARCHIVE` ?

```
[cmake] CMake Error:
[cmake]   Error evaluating generator expression:
[cmake] 
[cmake]     $<LINK_LIBRARY:WHOLE_ARCHIVE,qml-features-static>
[cmake] 
[cmake]   Expression did not evaluate to a known generator expression
[cmake] 
[cmake] 
[cmake] CMake Error:
[cmake]   Error evaluating generator expression:
[cmake] 
[cmake]     $<LINK_LIBRARY:WHOLE_ARCHIVE,qml-minimal-static>
[cmake] 
[cmake]   Expression did not evaluate to a known generator expression
[cmake] 
[cmake] 
[cmake] CMake Error:
[cmake]   Error evaluating generator expression:
[cmake] 
[cmake]     $<LINK_LIBRARY:WHOLE_ARCHIVE,cxx-qt-demo-threading-static>
[cmake] 
[cmake]   Expression did not evaluate to a known generator expression
[cmake] 
[cmake] 
[cmake] CMake Error:
[cmake]   Error evaluating generator expression:
[cmake] 
[cmake]     $<LINK_LIBRARY:WHOLE_ARCHIVE,qml-features-static>
[cmake] 
[cmake]   Expression did not evaluate to a known generator expression
[cmake] 
[cmake] 
[cmake] CMake Error:
[cmake]   Error evaluating generator expression:
[cmake] 
[cmake]     $<LINK_LIBRARY:WHOLE_ARCHIVE,qml-minimal-static>
[cmake] 
[cmake]   Expression did not evaluate to a known generator expression
[cmake] 
[cmake] 
[cmake] CMake Error:
[cmake]   Error evaluating generator expression:
[cmake] 
[cmake]     $<LINK_LIBRARY:WHOLE_ARCHIVE,cxx-qt-demo-threading-static>
[cmake] 
[cmake]   Expression did not evaluate to a known generator expression
[cmake] 
[cmake] 
[cmake] CMake Error:
[cmake]   Error evaluating generator expression:
[cmake] 
[...]

PR #445, opened 2023-02-13
set minimum CMake version to 3.24

Be-ing:
required for linking with WHOLE_ARCHIVE
https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html?highlight=whole_archive#genex:LINK_LIBRARY
https://github.com/KDAB/cxx-qt/issues/442

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,8 +5,7 @@
 #
 # SPDX-License-Identifier: MIT OR Apache-2.0
 
-# TODO: figure out what the actual hard minimum is
-cmake_minimum_required(VERSION 3.16)
+cmake_minimum_required(VERSION 3.24)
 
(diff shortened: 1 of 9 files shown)


CMakeLists.txt
```diff
@@ -5,8 +5,7 @@
 #
 # SPDX-License-Identifier: MIT OR Apache-2.0
```
ahayzen-kdab, line 6: the README mentions "CMake v3.16+", maybe this could be changed to `CMake` instead ? And i assume the book doesn't mention the version anywhere?

Be-ing: The book includes code from the qml_minimal CMakeLists.txt which shows the line setting the minimum CMake version.

ahayzen-kdab (approved):
LGTM :-)

Issue #414, opened 2023-01-14
WASM Example


azzamsa:
Hi.

Qt [now supports compiling to WASM](https://doc.qt.io/qt-5/wasm.html). Are we able to do this with CXX-Qt?
If this is true. I think we need to add a WASM example.

Thanks for CXX-Qt.

ahayzen-kdab:
None of us have tried yet, but i think it would be interesting to explore in the future.

If a normal CMake Qt project can be built then it should be possible once the right Rust toolchain is picked. Or maybe there is potentially a route with a cargo build too.

maciek134:
I have a Qt WebAssembly project I'd be interested to integrate Rust into. Starting with a PoC sounds sensible so I'll try to provide an example if nobody beats me to it.

maciek134:
Small update, Qt 5 seems like too much pain - 5.15 uses emscripten 1.39.8, which uses llvm 11, while rust is now at llvm 15. As I understand it such a big difference is asking for trouble (and I get relocation errors when linking anyway).

Qt 6 seems much more promising, since it uses emscripten 3.1.14 with llvm 15 so I'll try that later.

maciek134:
Based on the qml_minimal example, built with Qt 6.4.2, emscripten 3.1.14 and rust 1.67.0 with `wasm32-unknown-emscripten`:
![image](https://user-images.githubusercontent.com/1146309/218358873-9071627e-03b8-4f89-aea5-bd219874b33b.png)
Success, _but_:

I had to use `target_link_options(${APP_NAME}_lib INTERFACE "SHELL:-Wl,--whole-archive $<TARGET_FILE:${CRATE}-static> -Wl,--no-whole-archive")` instead of `"$<LINK_LIBRARY:WHOLE_ARCHIVE,${CRATE}-static>"`, otherwise CMake complained that
> Feature 'WHOLE_ARCHIVE', specified through generator-expression
>  '$<LINK_LIBRARY>' to link target 'example_qml_minimal', is not supported
>  for the 'CXX' link language.

Remove `QmlImportScanner` from Qt components since it's not present there in cross-builds (there is probably a way to include the x86 one but it's almost 4am :)

And (the bad one) I had to comment out the `assert_alignment_and_size` macro to get `cxx-qt-lib` to compile, otherwise I get
```
[...]

PR #446, opened 2023-02-14
cxx-qt-lib: ensure correct sizes for wasm 32bit

ahayzen-kdab:
Related to #414


ahayzen-kdab:
Oops put it on the wrong remote, moved to #447

PR #447, opened 2023-02-14
cxx-qt-lib: ensure correct sizes for wasm 32bit

ahayzen-kdab:
Related to #414

### DIFF
--- a/crates/cxx-qt-lib/src/core/qvariant/qvariant.cpp
+++ b/crates/cxx-qt-lib/src/core/qvariant/qvariant.cpp
@@ -13,24 +13,45 @@
 
 // The layout has changed between Qt 5 and Qt 6
 //
-// Qt5 QVariant has one member, which contains three uints and a union.
-// The three uints are optimised to a reduced size, resulting in a combined size
-// of two pointers.
-// https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/kernel/qvariant.h?h=v5.15.6-lts-lgpl#n491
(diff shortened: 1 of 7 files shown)


crates/cxx-qt-lib/src/core/qvariant/qvariant.cpp
```diff
@@ -13,24 +13,52 @@
...
+  alignof(double),
+  (sizeof(::std::uint16_t /* compiler optimised from ::std::uint32_t */) * 3) +
+    sizeof(double) + 2 /* compiler padding */);
+#else
+assert_alignment_and_size(
+  QVariant,
+  alignof(::std::size_t),
+  (sizeof(::std::uint16_t /* compiler optimised from ::std::uint32_t */) * 3) +
+    sizeof(double) + 2 /* compiler padding */);
+#endif
```
Be-ing, line 60: Why are two asserts needed for QVariant? Also, this could be moved up so two `#if/#else/#endif` preprocessor blocks aren't needed.

ahayzen-kdab: One assert block is Qt 5 and one Qt 6 as they have different layouts.

Then within them there are 32bit and 64bit asserts. For Qt 6 the 32bit size has extra padding. Then for both Qt 5 and Qt 6, the alignment is forced to 8 in 32bit by using a double, whereas in 64bit either the double or pointer cause the alignment to be 8. So this could be merged to just double for the Qt 5 case.

ahayzen-kdab: Done

Be-ing (approved):

PR #448, opened 2023-02-14
Implement std::ops for applicable Qt types

vimpostor:
I used the `derive_more` crate for some types, because it allows to derive these operators for simple types.
Let me know if you want to avoid adding this dependency.

Are there any more types, where an operator could be implemented?

ref: #53

### DIFF
--- a/crates/cxx-qt-lib/src/core/qpoint.rs
+++ b/crates/cxx-qt-lib/src/core/qpoint.rs
@@ -41,6 +41,24 @@ mod ffi {
         #[doc(hidden)]
         #[rust_name = "qpoint_to_qstring"]
         fn toQString(value: &QPoint) -> QString;
+        #[doc(hidden)]
+        #[rust_name = "qpoint_plus"]
+        fn operatorPlus(a: &QPoint, b: &QPoint) -> QPoint;
+        #[doc(hidden)]
+        #[rust_name = "qpoint_minus"]
+        fn operatorMinus(a: &QPoint, b: &QPoint) -> QPoint;
+        #[doc(hidden)]
+        #[rust_name = "qpoint_mul_f32"]
+        fn operatorMul(a: f32, b: &QPoint) -> QPoint;
+        #[doc(hidden)]
+        #[rust_name = "qpoint_mul_f64"]
+        fn operatorMul(a: f64, b: &QPoint) -> QPoint;
+        #[doc(hidden)]
+        #[rust_name = "qpoint_mul_i32"]
+        fn operatorMul(a: i32, b: &QPoint) -> QPoint;
+        #[doc(hidden)]
+        #[rust_name = "qpoint_div"]
+        fn operatorDiv(a: f64, b: &QPoint) -> QPoint;
     }
 }
 
@@ -69,6 +87,48 @@ impl Default for QPoint {
 impl fmt::Display for QPoint {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "{}", ffi::qpoint_to_qstring(self))
+    }
+}
+
+impl std::ops::Add for QPoint {
+    type Output = Self;
+    fn add(self, other: Self) -> Self {
+        ffi::qpoint_plus(&self, &other)
+    }
+}
+
+impl std::ops::Sub for QPoint {
+    type Output = Self;
+    fn sub(self, other: Self) -> Self {
+        ffi::qpoint_minus(&self, &other)
+    }
+}
+
+impl std::ops::Mul<f32> for QPoint {
+    type Output = Self;
+    fn mul(self, rhs: f32) -> Self {
+        ffi::qpoint_mul_f32(rhs, &self)
+    }
+}
+
+impl std::ops::Mul<f64> for QPoint {
+    type Output = Self;
+    fn mul(self, rhs: f64) -> Self {
+        ffi::qpoint_mul_f64(rhs, &self)
+    }
+}
+
+impl std::ops::Mul<i32> for QPoint {
+    type Output = Self;
+    fn mul(self, rhs: i32) -> Self {
+        ffi::qpoint_mul_i32(rhs, &self)
+    }
+}
+
+impl std::ops::Div<f64> for QPoint {
+    type Output = Self;
+    fn div(self, rhs: f64) -> Self {
+        ffi::qpoint_div(rhs, &self)
     }
 }
 
--- a/crates/cxx-qt-lib/src/core/qpointf.rs
+++ b/crates/cxx-qt-lib/src/core/qpointf.rs
@@ -42,6 +42,18 @@ mod ffi {
         #[doc(hidden)]
         #[rust_name = "qpointf_to_qstring"]
         fn toQString(value: &QPointF) -> QString;
+        #[doc(hidden)]
+        #[rust_name = "qpointf_plus"]
+        fn operatorPlus(a: &QPointF, b: &QPointF) -> QPointF;
+        #[doc(hidden)]
+        #[rust_name = "qpointf_minus"]
+        fn operatorMinus(a: &QPointF, b: &QPointF) -> QPointF;
+        #[doc(hidden)]
+        #[rust_name = "qpointf_mul"]
+        fn operatorMul(a: f64, b: &QPointF) -> QPointF;
+        #[doc(hidden)]
+        #[rust_name = "qpointf_div"]
+        fn operatorDiv(a: f64, b: &QPointF) -> QPointF;
     }
 }
 
@@ -70,6 +82,34 @@ impl Default for QPointF {
 impl fmt::Display for QPointF {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "{}", ffi::qpointf_to_qstring(self))
+    }
+}
+
+impl std::ops::Add for QPointF {
+    type Output = Self;
+    fn add(self, other: Self) -> Self {
+        ffi::qpointf_plus(&self, &other)
+    }
+}
+
+impl std::ops::Sub for QPointF {
+    type Output = Self;
+    fn sub(self, other: Self) -> Self {
+        ffi::qpointf_minus(&self, &other)
+    }
+}
+
+impl std::ops::Mul<f64> for QPointF {
+    type Output = Self;
+    fn mul(self, rhs: f64) -> Self {
+        ffi::qpointf_mul(rhs, &self)
+    }
+}
+
+impl std::ops::Div<f64> for QPointF {
+    type Output = Self;
+    fn div(self, rhs: f64) -> Self {
+        ffi::qpointf_div(rhs, &self)
     }
 }
 
--- a/crates/cxx-qt-lib/src/core/qsize.rs
+++ b/crates/cxx-qt-lib/src/core/qsize.rs
@@ -41,6 +41,18 @@ mod ffi {
         #[doc(hidden)]
         #[rust_name = "qsize_to_qstring"]
         fn toQString(value: &QSize) -> QString;
+        #[doc(hidden)]
+        #[rust_name = "qsize_plus"]
+        fn operatorPlus(a: &QSize, b: &QSize) -> QSize;
+        #[doc(hidden)]
+        #[rust_name = "qsize_minus"]
+        fn operatorMinus(a: &QSize, b: &QSize) -> QSize;
+        #[doc(hidden)]
+        #[rust_name = "qsize_mul"]
+        fn operatorMul(a: f64, b: &QSize) -> QSize;
+        #[doc(hidden)]
+        #[rust_name = "qsize_div"]
+        fn operatorDiv(a: f64, b: &QSize) -> QSize;
     }
 }
 
@@ -69,6 +81,34 @@ impl Default for QSize {
 impl fmt::Display for QSize {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "{}", ffi::qsize_to_qstring(self))
+    }
+}
+
+impl std::ops::Add for QSize {
+    type Output = Self;
+    fn add(self, other: Self) -> Self {
+        ffi::qsize_plus(&self, &other)
+    }
+}
+
+impl std::ops::Sub for QSize {
+    type Output = Self;
+    fn sub(self, other: Self) -> Self {
+        ffi::qsize_minus(&self, &other)
+    }
+}
+
+impl std::ops::Mul<f64> for QSize {
+    type Output = Self;
+    fn mul(self, rhs: f64) -> Self {
+        ffi::qsize_mul(rhs, &self)
+    }
+}
+
+impl std::ops::Div<f64> for QSize {
+    type Output = Self;
+    fn div(self, rhs: f64) -> Self {
+        ffi::qsize_div(rhs, &self)
     }
 }
 
--- a/crates/cxx-qt-lib/src/core/qsizef.rs
+++ b/crates/cxx-qt-lib/src/core/qsizef.rs
@@ -42,6 +42,18 @@ mod ffi {
         #[doc(hidden)]
         #[rust_name = "qsizef_to_qstring"]
         fn toQString(value: &QSizeF) -> QString;
+        #[doc(hidden)]
+        #[rust_name = "qsizef_plus"]
+        fn operatorPlus(a: &QSizeF, b: &QSizeF) -> QSizeF;
+        #[doc(hidden)]
+        #[rust_name = "qsizef_minus"]
+        fn operatorMinus(a: &QSizeF, b: &QSizeF) -> QSizeF;
+        #[doc(hidden)]
+        #[rust_name = "qsizef_mul"]
+        fn operatorMul(a: f64, b: &QSizeF) -> QSizeF;
+        #[doc(hidden)]
+        #[rust_name = "qsizef_div"]
+        fn operatorDiv(a: f64, b: &QSizeF) -> QSizeF;
     }
 }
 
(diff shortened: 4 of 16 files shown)


crates/cxx-qt-lib/src/core/qpoint.rs
```diff
@@ -45,7 +46,7 @@ mod ffi {
 }
 
 /// The QPoint struct defines a point in the plane using integer precision.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Add, Sub, Mul)]
 #[repr(C)]
 pub struct QPoint {
```
ahayzen-kdab, line 51: Hmm i was assuming we'd use the C++ operators.

Eg QPoint has += with a QPoint or *= with a factor. Then QRect has += with QMargins etc :thinking: 

And we have the debate in other PRs about if we should rely on the fields or the Qt implementation. Eg for QPoint elsewhere instead of making the field `x` public we have made the method `x()` as we aren't sure if the Qt inner implementation could ever change :thinking:

crates/cxx-qt-lib/src/core/qstring.rs
```diff
@@ -114,6 +118,15 @@ impl fmt::Debug for QString {
     }
 }
 
+impl std::ops::Add for QString {
+    type Output = Self;
+    fn add(self, other: Self) -> Self {
```
ahayzen-kdab, line 123: Hmm assume you can't use `QString &QString::operator+=(const QString &other)` ? 

Guess due to the types of `Add` this is tricky, as is closer to `AddAssign` ?

crates/cxx-qt-lib/src/core/qpoint.rs
LeonMatthesKDAB: CXX is also planning to add a "derive" for C++ operators in future:
https://github.com/dtolnay/cxx/issues/109

Maybe something we should upstream :thinking: 

Unfortunately at the moment something like this doesn't work:
``` rust
    unsafe extern "C++" {
        #[cxx_name = "operator+="]
        fn add(self: &mut QSize, other: &QSize) -> &mut QSize;
    }
```
CXX prevents you from using `operator+=` as the name of the function.

ahayzen-kdab: Right for now creating a template in C++ like we have for constructor/drop etc would be the workaround.

vimpostor: > as we aren't sure if the Qt inner implementation could ever change

Ah I see, I will change it to use an template on the C++ side.

ahayzen-kdab: In the serde one i'm also have a similar debate for simple types like QPoint, should we just use derive and the x, y. Or impl (de)seralise and read from the methods x() and y().

Not relying on the Qt structure but public API is probably more robust and means things don't need to be rewritten when Qt versions change stuff ...

vimpostor:
@ahayzen-kdab I am currently rewriting this with templated operators on the C++ side, but am facing difficulties with the fact that `cxx` [does not support generic extern functions](https://github.com/dtolnay/cxx/issues/683).

Consider for example the type `QPointF`. `std::ops::Add,Sub` are easy to implement, however for `std::ops::Mul` the lack of generics would be really annoying. One might want to multiply a `QPointF` with a scalar of any type, for example with `double`, `float`, `u8`, `i8` etc... Copy pasting the same implementation multiple times becomes infeasible quickly.

Maybe as a compromise we could implement `Add` and `Sub` with the operator on the C++/Qt side, and leave the implementation of `Mul` on the Rust side, where we do have full generics available?
Alternatively, it might be worth investigating how hard it is to implement extern generics support upstream in `cxx`.

ahayzen-kdab:
> @ahayzen-kdab I am currently rewriting this with templated operators on the C++ side, but am facing difficulties with the fact that `cxx` [does not support generic extern functions](https://github.com/dtolnay/cxx/issues/683).
> 
> Consider for example the type `QPointF`. `std::ops::Add,Sub` are easy to implement, however for `std::ops::Mul` the lack of generics would be really annoying. One might want to multiply a `QPointF` with a scalar of any type, for example with `double`, `float`, `u8`, `i8` etc... Copy pasting the same implementation multiple times becomes infeasible quickly.
> 
> Maybe as a compromise we could implement `Add` and `Sub` with the operator on the C++/Qt side, and leave the implementation of `Mul` on the Rust side, using generics? Alternatively, it might be worth investigating how hard it is to implement extern generics support upstream in `cxx`.

@vimpostor  I was assuming you could use a template to do this ? And even potentially have the left side by templated as well :-) As this is how we got around the lack of generics before, or does this not work ?

```cpp
template<typename T>
void
qpointMultiply(QPoint& p, const T& value) noexcept
{
  p *= value;
}
```

vimpostor:
> @vimpostor I was assuming you could use a template to do this ? And even potentially have the left side by templated as well :-) As this is how we got around the lack of generics before, or does this not work ?
> 
> ```c++
> template<typename T>
> void
> qpointMultiply(QPoint& p, const T& value) noexcept
> {
>   p *= value;
> }
> ```

Yes, that's what I have already:
```cxx
template<typename S, typename T>
T
operatorMul(const S scalar, const T& t)
{
    return scalar * t;
}
```

The problem is not the C++ side, but on the Rust side, because cxx does not support generic extern functions (see https://github.com/dtolnay/cxx/issues/683).
This means that we would need to define functions for multiplying e.g. `QPointF` with each scalar data type, e.g. `f32`, `f64`, `u8`, `i8` etc., which would be a lot of boilerplate.

ahayzen-kdab:
> > @vimpostor I was assuming you could use a template to do this ? And even potentially have the left side by templated as well :-) As this is how we got around the lack of generics before, or does this not work ?
> > ```c++
> > template<typename T>
> > void
> > qpointMultiply(QPoint& p, const T& value) noexcept
> > {
> >   p *= value;
> > }
> > ```
> 
> Yes, that's what I have already:
> 
> ```c++
> template<typename S, typename T>
> T
> operatorMul(const S scalar, const T& t)
> {
>     return scalar * t;
> }
> ```
> 
> The problem is not the C++ side, but on the Rust side, because cxx does not support generic extern functions (see [dtolnay/cxx#683](https://github.com/dtolnay/cxx/issues/683)). This means that we would need to define functions for multiplying e.g. `QPointF` with each scalar data type, e.g. `f32`, `f64`, `u32`, `i32` etc., which would be a lot of boilerplate.

Ugh, because then you would need to define `operatorMul(&mut QPoint, i32);`, operatorMul(&mut QPoint, u32);` etc and then the individual impls on the Rust side too (but that part could be a macro).

However what if you only implement the types Qt implements ? Eg for `QPointF` it only has `qreal` (which needs some casting) or `QPoint` only has `f32`, `f64`, `i32` for multiply and only `qreal` for divide.

vimpostor:
> However what if you only implement the types Qt implements ? Eg for `QPointF` it only has `qreal` (which needs some casting) or `QPoint` only has `f32`, `f64`, `i32` for multiply and only `qreal` for divide.

I still think even for 3 types the lack of generics starts getting annoying and is inelegant. Probably worth checking out how hard it is to implement external generic function support in `cxx`.

Anyways, I updated the PR to implement the operators with calls to the Qt operators only (instead of using derive).
Does this look more like what you wanted?

ahayzen-kdab:
> > However what if you only implement the types Qt implements ? Eg for `QPointF` it only has `qreal` (which needs some casting) or `QPoint` only has `f32`, `f64`, `i32` for multiply and only `qreal` for divide.
> 
> I still think even for 3 types the lack of generics starts getting annoying and is inelegant. Probably worth checking out how hard it is to implement external generic function support in `cxx`.
> 
> Anyways, I updated the PR to implement the operators with calls to the Qt operators only (instead of using derive). Does this look more like what you wanted?

I think this looks much closer, obviously generics in CXX would be best but they will take time to get right upstream, and they may also help us with the container types if support can be added.

crates/cxx-qt-lib/src/core/qpoint.rs
```diff
@@ -41,6 +41,21 @@ mod ffi {
...
+        fn operatorMinus(a: &QPoint, b: &QPoint) -> QPoint;
+        #[doc(hidden)]
+        #[rust_name = "qpoint_mul_f32"]
+        fn operatorMul(a: f32, b: &QPoint) -> QPoint;
+        #[doc(hidden)]
+        #[rust_name = "qpoint_mul_f64"]
+        fn operatorMul(a: f64, b: &QPoint) -> QPoint;
+        #[doc(hidden)]
+        #[rust_name = "qpoint_mul_i32"]
+        fn operatorMul(a: i32, b: &QPoint) -> QPoint;
```
ahayzen-kdab, line 58: has divide been missed for QPoint?

crates/cxx-qt-lib/src/core/qpointf.rs
```diff
@@ -42,6 +42,24 @@ mod ffi {
...
+        fn operatorMul(a: f32, b: &QPointF) -> QPointF;
+        #[doc(hidden)]
+        #[rust_name = "qpointf_mul_f64"]
+        fn operatorMul(a: f64, b: &QPointF) -> QPointF;
+        #[doc(hidden)]
+        #[rust_name = "qpointf_mul_i32"]
+        fn operatorMul(a: i32, b: &QPointF) -> QPointF;
+        #[doc(hidden)]
+        #[rust_name = "qpointf_div_f32"]
+        fn operatorDiv(a: f32, b: &QPointF) -> QPointF;
```
ahayzen-kdab, line 62: note we need to be careful with `qreal`, sometimes between Qt 5 and Qt 6 it can be a `qreal` and then a `float` in the other etc or we have fun with 32bit platforms (eg Wasm).

Elsewhere we have a `f32` in rust, `float` in C++, then cast it to a `qreal` for the specific Qt version. So this looks potentially correct anyway :-)

crates/cxx-qt-lib/src/core/qpointf.rs
```diff
@@ -73,6 +91,48 @@ impl fmt::Display for QPointF {
...
+}
+
+impl std::ops::Mul<f64> for QPointF {
+    type Output = Self;
+    fn mul(self, rhs: f64) -> Self {
+        ffi::qpointf_mul_f64(rhs, &self)
+    }
+}
+
+impl std::ops::Mul<i32> for QPointF {
```
ahayzen-kdab, line 122: QPointF only has `QPointF &QPointF::operator*=(qreal factor)` did we want i32 as well ?

crates/cxx-qt-lib/src/core/qsize.rs
```diff
@@ -41,6 +41,15 @@ mod ffi {
...
+        #[doc(hidden)]
+        #[rust_name = "qsize_plus"]
+        fn operatorPlus(a: &QSize, b: &QSize) -> QSize;
+        #[doc(hidden)]
+        #[rust_name = "qsize_minus"]
+        fn operatorMinus(a: &QSize, b: &QSize) -> QSize;
+        #[doc(hidden)]
+        #[rust_name = "qsize_mul"]
+        fn operatorMul(a: f32, b: &QSize) -> QSize;
     }
```
ahayzen-kdab, line 53: are we missing divide intentionally on the non F again ?

crates/cxx-qt-lib/src/core/qsize.rs
```diff
@@ -41,6 +41,15 @@ mod ffi {
...
         fn toQString(value: &QSize) -> QString;
+        #[doc(hidden)]
+        #[rust_name = "qsize_plus"]
+        fn operatorPlus(a: &QSize, b: &QSize) -> QSize;
+        #[doc(hidden)]
+        #[rust_name = "qsize_minus"]
+        fn operatorMinus(a: &QSize, b: &QSize) -> QSize;
+        #[doc(hidden)]
+        #[rust_name = "qsize_mul"]
+        fn operatorMul(a: f32, b: &QSize) -> QSize;
```
ahayzen-kdab, line 52: same qreal questions :thinking:

crates/cxx-qt-lib/src/gui/qvector4d.rs
```diff
@@ -147,6 +159,34 @@ impl std::fmt::Display for QVector4D {
...
+impl std::ops::Div<f32> for QVector4D {
+    type Output = Self;
+    fn div(self, rhs: f32) -> Self {
+        ffi::qvector4d_div(rhs, &self)
+    }
+}
+
 impl From<crate::QVector3D> for QVector4D {
     /// Constructs a 4D vector from the specified 3D vector.
     /// The w coordinate is set to zero.
```
ahayzen-kdab, line 192: my only other questions would then be

- [ ] Do we want Add for QStringList with QString ? However that is more a stream in C++ https://doc.qt.io/qt-6/qstringlist.html#operator-lt-lt there is a QStringList + QStringList though, but just wondered if `stringlist + string` would make any sense in Rust :thinking: 
- [ ] Do we want the Ops in QMargins(F) ?
- [ ] Do we want the Add/Div in QRect(F) with QMargins(F) ?

crates/cxx-qt-lib/src/core/qpointf.rs
```diff
@@ -73,6 +91,48 @@ impl fmt::Display for QPointF {
...
+}
+
+impl std::ops::Mul<f32> for QPointF {
+    type Output = Self;
+    fn mul(self, rhs: f32) -> Self {
+        ffi::qpointf_mul_f32(rhs, &self)
+    }
+}
+
+impl std::ops::Mul<f64> for QPointF {
```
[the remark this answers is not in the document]
ahayzen-kdab: I'll try building this in a Wasm environment when i get a moment on the right machine ...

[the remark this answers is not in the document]
vimpostor: > Which is more correct here and will work across platforms ?

None of them satisfies both concerns.
The one working across platforms is `f32`, as it will implicitly cast to `f64` when qreal is double, so `f32` seems favorable.
But it is only favorable when you don't need the precision of `f64` on the Rust side. Not really sure either what is the best option here.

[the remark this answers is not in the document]
ahayzen-kdab: Looking at https://github.com/KDAB/cxx-qt/commit/e334c2907f776f8e9a729d7a0cc9e3b55b39b8a0 again, it seems a `double` / `qreal` works in Wasm (it does have support for 64bit floating point numbers just `int` is 32bit etc). So using `f64` might be OK here as it is `qreal` (aka `double`) in both Qt 5 and Qt 6. And should work with Wasm ...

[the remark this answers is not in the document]
ahayzen-kdab: (I think the fun comes from when it is `qreal` in Qt 5 and `float` in Qt 6, which is not the case here luckily)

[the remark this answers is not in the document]
ahayzen-kdab: So assuming that is the case, it's probably best to go with `f64` in these places and lets see if it breaks :-)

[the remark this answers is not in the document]
ahayzen-kdab: Right, however i don't think we test / support armhf / armel / ARMv6/v7 at all (aka any ARM 32bit or even x86), so there are likely other issues if that is the case.

We know that arm64 / ARMv8-A / AArch64 works and that Wasm works with some patches though.

crates/cxx-qt-lib/src/gui/qvector4d.rs
vimpostor: Implemented the last two points. I don't think the first one would make much sense, as it is neither Rust-idiomatic (for example neither Rust's `Vec` nor `LinkedList` implement an operator like that) nor does the Qt API implement that plus operator.

crates/cxx-qt-lib/src/core/qpoint.rs
vimpostor: Yes, I didn't see that there was Qt API for this that rounds the results.

crates/cxx-qt-lib/src/core/qpointf.rs
vimpostor: Fixed

crates/cxx-qt-lib/src/core/qpoint.rs
```diff
@@ -41,6 +41,24 @@ mod ffi {
...
+        #[doc(hidden)]
+        #[rust_name = "qpoint_div"]
+        fn operatorDiv(a: f32, b: &QPoint) -> QPoint;
```
ahayzen-kdab, line 61: should this be `f64` as it's a `qreal` (which tends to be a `double`) in C++ ?

crates/cxx-qt-lib/src/core/qsizef.rs
```diff
@@ -42,6 +42,18 @@ mod ffi {
...
         fn toQString(value: &QSizeF) -> QString;
+        #[doc(hidden)]
+        #[rust_name = "qsizef_plus"]
+        fn operatorPlus(a: &QSizeF, b: &QSizeF) -> QSizeF;
+        #[doc(hidden)]
+        #[rust_name = "qsizef_minus"]
+        fn operatorMinus(a: &QSizeF, b: &QSizeF) -> QSizeF;
+        #[doc(hidden)]
+        #[rust_name = "qsizef_mul"]
+        fn operatorMul(a: f32, b: &QSizeF) -> QSizeF;
```
ahayzen-kdab, line 53: why is this `f32` when `QSize` is `f64` and it's `qreal` for both in C++ ?

crates/cxx-qt-lib/src/core/qstring.rs
```diff
@@ -114,6 +118,15 @@ impl fmt::Debug for QString {
...
+        let mut res = ffi::qstring_init_from_qstring(&self);
+        ffi::qstring_append(&mut res, &other);
```
ahayzen-kdab, line 125: note that #416 adds the append method to QString, so i need to remember to update this part in that branch if that lands after...

Also note this was done without an FFI step via the following if you wanted to copy
```rust
/// Appends the string str onto the end of this string.
fn append<'a>(self: &'a mut QString, str: &QString) -> &'a mut QString;
````

crates/cxx-qt-lib/src/core/qsizef.rs
vimpostor: Oops, I missed it.

ahayzen-kdab (approved):
LGTM :-)

crates/cxx-qt-lib/src/core/qstring.rs
ahayzen-kdab: We can leave this one to #416 and not this merge :-)

ahayzen-kdab (approved):
LGTM

PR #451, opened 2023-02-20
Draft: Add support for attributes to #[qproperty]

CarlSchwan:
Make it possible to define if only a getter should be generated or only a setter and/or define special getter and setter.

Syntax is as follow

```rust
struct MyObject {
    #[qproperty]
    prop1: f32, // generate default getter and setter

    #[qproperty(get)]
    prop2: f32, // generate default getter only

    #[qproperty(set)]
    prop3: f32, // generate default setter only

    #[qproperty(get = Self::prop4_fn)]
    prop4: f32, // use prop4_fn as getter
}
```

WIP since the custom getter part doesn't work yet

### DIFF
--- a/crates/cxx-qt-build/src/lib.rs
+++ b/crates/cxx-qt-build/src/lib.rs
@@ -407,7 +407,7 @@ impl CxxQtBuilder {
         // Use a separate cc::Build for the little amount of code that needs to be linked with +whole-archive
         // to avoid bloating the binary.
         let mut cc_builder_whole_archive = cc::Build::new();
-        cc_builder_whole_archive.link_lib_modifier("+whole-archive");
+        //cc_builder_whole_archive.link_lib_modifier("+whole-archive");
         for builder in [&mut self.cc_builder, &mut cc_builder_whole_archive] {
             builder.cpp(true);
             // MSVC
--- a/crates/cxx-qt-gen/src/generator/cpp/property/meta.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/property/meta.rs
@@ -4,7 +4,7 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
 use crate::generator::{cpp::types::CppType, naming::property::QPropertyName};
-use crate::parser::property::{ParsedQProperty, MaybeCustomFn};
+use crate::parser::property::ParsedQProperty;
 
 /// Generate the metaobject line for a given property
 pub fn generate(idents: &QPropertyName, property: &ParsedQProperty, cxx_ty: &CppType) -> String {
--- a/crates/cxx-qt-gen/src/generator/cpp/property/meta.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/property/meta.rs
@@ -4,15 +4,30 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
 use crate::generator::{cpp::types::CppType, naming::property::QPropertyName};
+use crate::parser::property::{ParsedQProperty, MaybeCustomFn};
 
 /// Generate the metaobject line for a given property
-pub fn generate(idents: &QPropertyName, cxx_ty: &CppType) -> String {
+pub fn generate(idents: &QPropertyName, property: &ParsedQProperty, cxx_ty: &CppType) -> String {
+    let getter_setter_not_explicit = property.get.is_none() && property.set.is_none();
+
+    let getter = if getter_setter_not_explicit || property.get.is_some() {
+        format!("READ {ident_getter}", ident_getter = idents.getter.cpp)
+    } else {
+        String::new()
+    };
+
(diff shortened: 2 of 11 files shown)


crates/cxx-qt-gen/src/generator/rust/qobject.rs
```diff
@@ -33,6 +33,7 @@ impl GeneratedRustQObjectBlocks {
     }
 }
 
+#[derive(Debug)]
```
ahayzen-kdab, line 36: any reasoning for these extra derive Debug's ? They could be done in a separate commit if they are useful ...

CarlSchwan: Mostly for debugging, I'll remove then

ahayzen-kdab:
Also note the clippy warnings :-)

crates/cxx-qt-gen/src/generator/cpp/property/meta.rs
```diff
@@ -4,15 +4,30 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
```
ahayzen-kdab, line 4: don't forget to add an entry to the CHANGELOG.md :-)

And update anything in the book and we'll have to find somewhere in the properties example of qml_features to use it

crates/cxx-qt-build/src/lib.rs
```diff
@@ -407,7 +407,7 @@ impl CxxQtBuilder {
         // Use a separate cc::Build for the little amount of code that needs to be linked with +whole-archive
         // to avoid bloating the binary.
         let mut cc_builder_whole_archive = cc::Build::new();
-        cc_builder_whole_archive.link_lib_modifier("+whole-archive");
+        //cc_builder_whole_archive.link_lib_modifier("+whole-archive");
```
CarlSchwan, line 410:
```suggestion
```

remove before merging

LeonMatthesKDAB:
@CarlSchwan good to see this change progressing nicely :+1: 

Quick chime-in on the `#[qproperty(get=Self::x)]` API.
I've just noticed that the `Self::` part might be redundant, or even misleading in the worst-case, so it would be simpler to just leave it out in favor of `#[qproperty(get=x)]`.

The `Self::` within the `struct MyObject` technically refers to `MyObject`, not `qobject::MyObject`, but the getter `x` must be implemented on `qobject::MyObject`, so `Self::` doesn't really refer to the right struct, technically.
Furthermore, a getter (or setter for that matter) must always be implemented on the `qobject::` of the given struct, so there's no real point of listing the `Self::`, it's just boilerplate I think and parsing would be simplified by removing it.

Or is there any way one could specify a getter/setter that is not on the `qobject::MyObject`? :thinking:

CarlSchwan:
> @CarlSchwan good to see this change progressing nicely +1
> 
> Quick chime-in on the `#[qproperty(get=Self::x)]` API. I've just noticed that the `Self::` part might be redundant, or even misleading in the worst-case, so it would be simpler to just leave it out in favor of `#[qproperty(get=x)]`. 
> 
> The `Self::` within the `struct MyObject` technically refers to `MyObject`, not `qobject::MyObject`, but the getter `x` must be implemented on `qobject::MyObject`, so `Self::` doesn't really refer to the right struct, technically. Furthermore, a getter (or setter for that matter) must always be implemented on the `qobject::` of the given struct, so there's no real point of listing the `Self::`, it's just boilerplate I think and parsing would be simplified by removing it.
> 
> Or is there any way one could specify a getter/setter that is not on the `qobject::MyObject`? thinking

I don't think there is a way to specify a getter/setter that is not on the `qobject::MyObject` and I don't think this would make sense. Regarding the syntax, I often prefer to make stuff more explicit and adding `Self::` make it explicit and I originally tried to have the custom getter/settter in the `MyObject` instead of the `qobject::MyObject` but I didn't manage to do it. Syntax wise my objective is to have something as convenient as gtk-rs https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/derive.Properties.html which are quite nice.

But there are still quite some work needed to get there.

LeonMatthesKDAB:
> I don't think there is a way to specify a getter/setter that is not on the `qobject::MyObject` and I don't think this would make sense. Regarding the syntax, I often prefer to make stuff more explicit and adding `Self::` make it explicit and I originally tried to have the custom getter/settter in the `MyObject` instead of the `qobject::MyObject` but I didn't manage to do it. Syntax wise my objective is to have something as convenient as gtk-rs https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/derive.Properties.html which are quite nice.
> 
> But there are still quite some work needed to get there.

IMHO, we should still just leave out the `Self::` part then, it's more "explicit", but also points you to the wrong object!

GTK-rs also seems to just use the function name.

LeonMatthesKDAB:
Superseded by #994

Issue #38, opened 2021-12-09
Improve and investigate error messages


ahayzen-kdab:
- [x] Investigate why errors from cxx-qt-gen don't point to the span - #455 
- [x] Investigate if CXX errors can be exposed better (eg from cxx-qt-build) - #461
- [x] ~Investigate why macros don't error in IDEs from cxx-qt-gen errors~ (this was because i had `"rust-analyzer.checkOnSave": false,` set)
- [x] Can we get CXX errors to error in IDEs ? - #461 
- [ ] Check if we need to use the dummy env var to improve IDE experience

# Original Description
- Add error messages for invalid invokable input and return types
- Add error messages for invalid property and invokable names
- Consider how other messages we emitted when the macros code is wrong etc

ahayzen-kdab:
https://blog.jetbrains.com/rust/2022/12/05/what-every-rust-developer-should-know-about-macro-support-in-ides/  is relevant here too with the `RUST_IDE_PROC_MACRO_COMPLETION_DUMMY_IDENTIFIER` env var.

LeonMatthesKDAB:
Also see my comment on #455

ahayzen-kdab:
Interestingly internally `cxx_gen` has two options `gen_header` and `gen_implementation` https://github.com/dtolnay/cxx/blob/5ebaee1a1dd668ab48a231890a26bae5fa81888a/gen/src/mod.rs#L58 these then avoid the generation phase https://github.com/dtolnay/cxx/blob/5ebaee1a1dd668ab48a231890a26bae5fa81888a/gen/src/mod.rs#L178

~Say that these were public or we could modify them somehow, this would allow us to pass the tokens through CXX and find out if it was the parsing/checks phase that has failed or the write phase. Which then means if this first phase failed we could stop the build script and emit a warning, then we know that the macro expansion will fail ?~

But as the later write phases don't fail there is actually no advantage here? So if the `generate_header_and_cc` fails this means the same and we should try to warn.

ahayzen-kdab:
Think i've figured it out :sweat_smile:

ahayzen-kdab:
#547 has massively helped with this. I think this issue is done now ?

ahayzen-kdab:
We have separate issues for any remaining tasks.

PR #455, opened 2023-02-28
cxx-qt: don't use unwrap and convert to a compile error instead

ahayzen-kdab:
This improve error messages when our cxx-qt-gen code breaks.

Eg I have added a duplicate attribute key-value in the code, before we used to get this error when compiling.

```
error: custom attribute panicked
  --> examples/cargo_without_cmake/src/cxxqt_object.rs:11:1
   |
11 | #[cxx_qt::bridge]
   | ^^^^^^^^^^^^^^^^^
   |
   = help: message: called `Result::unwrap()` on an `Err` value: Error("Duplicate keys in the attributes")
```

Now we have a much better error that points to the code block that is at fault.

```
error: Duplicate keys in the attributes
  --> examples/cargo_without_cmake/src/cxxqt_object.rs:23:5
   |
23 |     #[cxx_qt::qobject(qml_uri = "com.kdab.cxx_qt.demo", qml_version = "1.0", qml_version = "1.0")]
   |     ^
```

Related to #38

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for generating correct C++ code for `Pin<T>` Rust types
 - Support namespace attribute on shared types, QObject struct, and extern blocks
 - Asserts for 32bit platforms such as Wasm
+- Errors from the generation not pointing to the span where they occurred
 
 ## [0.4.1](https://github.com/KDAB/cxx-qt/compare/v0.4.0...v0.4.1) - 2022-11-18
 
--- a/crates/cxx-qt/src/lib.rs
+++ b/crates/cxx-qt/src/lib.rs
@@ -141,7 +141,9 @@ pub fn inherit(_args: TokenStream, _input: TokenStream) -> TokenStream {
 
 // Take the module and C++ namespace and generate the rust code
 fn extract_and_generate(module: ItemMod) -> TokenStream {
-    let parser = Parser::from(module).unwrap();
-    let generated_rust = GeneratedRustBlocks::from(&parser).unwrap();
-    write_rust(&generated_rust).into()
+    Parser::from(module)
+        .and_then(|parser| GeneratedRustBlocks::from(&parser))
+        .map(|generated_rust| write_rust(&generated_rust))
+        .unwrap_or_else(|err| err.to_compile_error())
+        .into()
 }


ahayzen-kdab:
This doesn't seem to make the IDE pick up the error yet, but is already an improvement, I'll read more about the IDE side of things.

Be-ing (approved):

LeonMatthesKDAB:
Regarding IDE support:
I'm pretty sure this is actually not caused by issues in the macro expansion, but by how our build script works.
Currently, if the build script fails to parse the macro code, it panics.
This seems right at first glance, but this is actually what's causing the issue (at least in my setup with rust_analyzer + nvim).
If the build script panics, rust_analyzer won't even try to expand our proc_macro, as the appropriate build setup isn't working.

So the right way to fix this would be to silently ignore any parsing errors in the build script, and maybe emit a warning.
Then the same parsing would still fail in the expansion of the procedural macro, which rust_analyzer can deal with and nicely highlight in your IDE of choice :partying_face: 
We do need to be careful though to invoke the same code paths from the build script then to make sure it's not possible to have a parsing error in the build script, but not in the proc_macro itself.

You can already see this working when there's an error in the Rust generation.
In this case the right span will be highlighted in the editor.
In such cases the build script won't panic as it only generates the C++ code.

ahayzen-kdab:
> Regarding IDE support: I'm pretty sure this is actually not caused by issues in the macro expansion, but by how our build script works. Currently, if the build script fails to parse the macro code, it panics. This seems right at first glance, but this is actually what's causing the issue (at least in my setup with rust_analyzer + nvim). If the build script panics, rust_analyzer won't even try to expand our proc_macro, as the appropriate build setup isn't working.
> 
> So the right way to fix this would be to silently ignore any parsing errors in the build script, and maybe emit a warning. Then the same parsing would still fail in the expansion of the procedural macro, which rust_analyzer can deal with and nicely highlight in your IDE of choice partying_face We do need to be careful though to invoke the same code paths from the build script then to make sure it's not possible to have a parsing error in the build script, but not in the proc_macro itself.
> 
> You can already see this working when there's an error in the Rust generation. In this case the right span will be highlighted in the editor. In such cases the build script won't panic as it only generates the C++ code.

Yup that's the next phase of #38 is to try changing the build script not to explode when `cxx_gen::generate_header_and_cc` fails, however i tried switching this to simply produce an empty `GeneratedCode` block and then it instead explodes trying to build the C++ side. So it needs some careful investigation.

But this is i think definitely the issue with the CXX macros, as the build script fails first it never reaches the macro expansion.

Issue #21, opened 2021-12-09
Try to enable deny(missing_docs)


ahayzen-kdab:
Try to enable #![deny(missing_docs)] in all crates

https://doc.rust-lang.org/rustdoc/lints.html#missing_docs

ahayzen-kdab:
`cxx-qt-lib` has many to fix :-)

```
error: could not compile `cxx-qt-lib` (lib) due to 324 previous errors
```

LeonMatthesKDAB:
Suggestion for cxx-qt-lib:

On every type, we add a link to the corresponding Qt documentation.
Then we add `#[allow(missing_docs)]` on all methods that are a 1:1 wrap of a Qt API that doesn't need further comment.

Anything that is different from Qt API, we add a doc comment on.

PR #456, opened 2023-02-28
cxx-qt: activate doctests for the macros and deny missing docs in the crate

ahayzen-kdab:
Related to #21

### DIFF
--- a/crates/cxx-qt/src/lib.rs
+++ b/crates/cxx-qt/src/lib.rs
@@ -3,6 +3,11 @@
 // SPDX-FileContributor: Gerhard de Clercq <gerhard.declercq@kdab.com>
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
+
+#![deny(missing_docs)]
+
+//! The cxx-qt crate provides the procedural attribute macros which are used with cxx-qt.
+
 use proc_macro::TokenStream;
 use syn::{parse_macro_input, ItemMod};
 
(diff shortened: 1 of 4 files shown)


ahayzen-kdab:
Figured out how to get the doctests going on the macro page, we just needed a fake main and then the dependencies.

crates/cxx-qt/src/lib.rs
```diff
@@ -12,23 +17,23 @@ use cxx_qt_gen::{write_rust, GeneratedRustBlocks, Parser};
 ///
 /// # Example
 ///
-/// ```ignore
-/// #[cxx_qt::bridge(namespace = "cxx_qt::my_object")]
```
ahayzen-kdab, line 16: oddly Windows gets confused with the namespace

```
  = note: rust_out.rust_out.b333b0d2-cgu.0.rcgu.o : error LNK2019: unresolved external symbol cxx_qt$my_object$cxxbridge1$MyObject$property_changed referenced in function _ZN8rust_out3ffi10MyObjectQt16property_changed17hd775c3ccb607517eE
          rust_out.rust_out.b333b0d2-cgu.0.rcgu.o : error LNK2019: unresolved external symbol cxx_qt$my_object$cxxbridge1$MyObject$rust referenced in function _ZN8rust_out3ffi10MyObjectQt4rust17hfc2e0d9d9b13c063E
          rust_out.rust_out.b333b0d2-cgu.0.rcgu.o : error LNK2019: unresolved external symbol cxx_qt$my_object$cxxbridge1$MyObject$rust_mut referenced in function _ZN8rust_out3ffi10MyObjectQt8rust_mut17h5a53d6213a79e672E
          C:\Users\RUNNER~1\AppData\Local\Temp\rustdoctestvRP7oB\rust_out.exe : fatal error LNK1120: 3 unresolved externals
```

As the other tests don't fail and specifically this one fails

ahayzen-kdab: Or it might be that the properties internally use `rust`, `rust_mut`, `prop_changed`, but oddly signals do not fail which use `emit_signal` and the `queue` methods which are also `extern "C++"` do not fail :thinking:

ahayzen-kdab: So having no namespace and no properties has worked, now time to reintroduce one of them to see what happens ...

ahayzen-kdab: The namespace is fine, it's something odd in the properties, for now lets not put properties in those macro examples that aren't related to properties anyway...

LeonMatthesKDAB (approved):
Overall, great change!

We shouldn't close #21 yet though, as `#![deny(missing_docs)]` isn't enabled on cxx-qt-lib, where it is arguably most important.
There's currently 259 items in cxx-qt-lib that aren't documented, so we should fix that before closing #21.

PR #458, opened 2023-03-01
cxx-qt-gen: have an explicit error for non public qobjects

### LINKED ISSUE #457: Non public QObject struct has non obvious error message
As part of looking into #38 I found that if you don't mark the QObject struct in Rust as public you get a non obvious warning.

```rust
[build] error[E0432]: unresolved import `super`
[build]   --> examples/qml_minimal/rust/src/cxxqt_object.rs:24:12
[build]    |
[build] 24 |     struct MyObject {
[build]    |            ^^^^^^^^ no `MyObject` in `cxxqt_object`
[build]    |
[build] help: consider importing this type alias instead
[build]    |
[build] 24 |     struct crate::cxxqt_object::qobject::MyObject;
[build]    |            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This happens because we put the QObject struct into a child module, then in the CXX module it has a `type MyObject` where it looks in the parent module for the QObject, but the import `pub use self::cxx_qt_ffi::*;` has failed to reexport the QObject as it was not public.

Our options seem to be
- Fail in the parser if the QObject struct is not `pub` with a nicer error message (this is also potentially a candidate then to go into the checks phase (#213) if we ever create it with the potential refactor of the parser phase to be a true syn Parse)
- Implicitly mark all our QObject structs as `pub` even if the developer doesn't (might be odd?)

ahayzen-kdab:
Before this patch if you didn't mark a qobject struct as public the following error would occur.

```
[build] error[E0432]: unresolved import `super`
[build]   --> examples/qml_minimal/rust/src/cxxqt_object.rs:24:12
[build]    |
[build] 24 |     struct MyObject {
[build]    |            ^^^^^^^^ no `MyObject` in `cxxqt_object`
[build]    |
[build] help: consider importing this type alias instead
[build]    |
[build] 24 |     struct crate::cxxqt_object::qobject::MyObject;
[build]    |            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

After this patch the error is explicit.

```
[build] error: qobject marked structs must be public
[build]   --> examples/qml_minimal/rust/src/cxxqt_object.rs:24:5
[build]    |
[build] 24 |     struct MyObject {
[build]    |     ^^^^^^
```

Closes #457

### DIFF
--- a/crates/cxx-qt-gen/src/generator/cpp/mod.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/mod.rs
@@ -59,7 +59,7 @@ mod tests {
             #[cxx_qt::bridge]
             mod ffi {
                 #[cxx_qt::qobject]
-                struct MyObject;
+                pub struct MyObject;
             }
         });
         let parser = Parser::from(module).unwrap();
@@ -76,7 +76,7 @@ mod tests {
             #[cxx_qt::bridge(cxx_file_stem = "my_object")]
             mod ffi {
                 #[cxx_qt::qobject]
-                struct MyObject;
+                pub struct MyObject;
             }
         });
         let parser = Parser::from(module).unwrap();
@@ -93,7 +93,7 @@ mod tests {
             #[cxx_qt::bridge(namespace = "cxx_qt")]
             mod ffi {
                 #[cxx_qt::qobject]
-                struct MyObject;
+                pub struct MyObject;
             }
         });
         let parser = Parser::from(module).unwrap();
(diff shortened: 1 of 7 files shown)


vimpostor (approved):

ahayzen-kdab:
@vimpostor Oops noticed small typo in the commit message "didn't mark a qobject struct as private" instead it should have said "didn't mark a qobject struct as public", I've fixed that now and rebased. Can you reapprove?

vimpostor (approved):

PR #461, opened 2023-03-03
cxx-qt-build: do not panic for cxx_gen failures instead use Result

ahayzen-kdab:
This allows for CXX errors to appear with a span in the macro expansion

Before this change if you had a unsupported type in an invokable only the following error would occur.

```
[build]   thread 'main' panicked at 'Could not generate C++ from Rust file:
Syn(Error("unsupported type: QUrl"))', crates/cxx-qt-build/src/lib.rs:120:14
```

Now with this change we have an additional error from the macro expansion.

```
[build] error: unsupported type: QUrl
[build]   --> examples/qml_minimal/rust/src/cxxqt_object.rs:57:35
[build]    |
[build] 57 |         pub fn test(&self, _url: &QUrl) {
[build]    |                                   ^^^^
```

Note however this does cause loads of extra errors below but this is a side effect of the rest of the CXX macro not being expanded, this first error is the correct one though.

Related to #38

### DIFF
--- a/crates/cxx-qt-build/src/diagnostics.rs
+++ b/crates/cxx-qt-build/src/diagnostics.rs
@@ -6,28 +6,75 @@
 //! This module takes care of displaying errors emitted by CXX-Qt as nicely-printed diagnostics
 //! using codespan-reporting.
 
-use std::{ops::Range, path::PathBuf};
+use std::{fmt::Display, ops::Range, path::PathBuf};
 
-use proc_macro2::LineColumn;
+use proc_macro2::{LineColumn, Span};
+
+/// We need to wrap the CXX and CXX-Qt errors in a single error type so that they
+/// can be returned from the generation phases. This then allows for us to not unwrap
+/// the CXX failures meaning that we can display the macro expansion correctly.
+#[derive(Debug)]
+pub(crate) enum GeneratedError {
+    Cxx(cxx_gen::Error),
+    CxxQt(cxx_qt_gen::Error),
+}
+
+impl Display for GeneratedError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            GeneratedError::Cxx(err) => write!(f, "{}", err),
+            GeneratedError::CxxQt(err) => write!(f, "{}", err),
+        }
+    }
+}
+
+impl From<cxx_gen::Error> for GeneratedError {
+    fn from(err: cxx_gen::Error) -> Self {
+        Self::Cxx(err)
+    }
+}
+
+impl From<cxx_qt_gen::Error> for GeneratedError {
+    fn from(err: cxx_qt_gen::Error) -> Self {
+        Self::CxxQt(err)
+    }
+}
+
+impl GeneratedError {
+    fn context(&self) -> &str {
+        match self {
+            GeneratedError::Cxx(_) => "cxx",
+            GeneratedError::CxxQt(_) => "cxxqt",
+        }
+    }
+
+    fn span(&self) -> Option<Span> {
+        match self {
+            // TODO: CXX doesn't have span available yet
+            // https://github.com/KDAB/cxx-qt/issues/536
+            GeneratedError::Cxx(_) => None,
+            GeneratedError::CxxQt(syn_err) => Some(syn_err.span()),
+        }
+    }
+}
 
 pub(crate) struct Diagnostic {
     file_path: PathBuf,
-    error: cxx_qt_gen::Error,
+    error: GeneratedError,
 }
 
 impl Diagnostic {
-    pub(crate) fn new(file_path: PathBuf, error: cxx_qt_gen::Error) -> Self {
+    pub(crate) fn new(file_path: PathBuf, error: GeneratedError) -> Self {
         Self { file_path, error }
     }
 
     fn byte_span_in(&self, source: &str) -> Option<Range<usize>> {
-        let syn_err = &self.error;
-        let span = syn_err.span();
+        self.error.span().and_then(|span| {
+            let start_offset = line_column_to_byte_in(span.start(), source)?;
+            let end_offset = line_column_to_byte_in(span.end(), source)?;
 
-        let start_offset = line_column_to_byte_in(span.start(), source)?;
-        let end_offset = line_column_to_byte_in(span.end(), source)?;
-
-        Some(start_offset..end_offset)
+            Some(start_offset..end_offset)
+        })
     }
 
     fn create_codespan_diagnostic(
@@ -39,7 +86,7 @@ impl Diagnostic {
         let syn_err = &self.error;
         let mut diagnostic = codespan_reporting::diagnostic::Diagnostic::error()
             .with_message(format!("{syn_err}"))
-            .with_code("cxxqt".to_owned());
+            .with_code(syn_err.context());
 
         if let Some(span) = self.byte_span_in(source) {
             diagnostic = diagnostic.with_labels(vec![Label::primary((), span)]);
--- a/crates/cxx-qt-build/src/lib.rs
+++ b/crates/cxx-qt-build/src/lib.rs
@@ -12,7 +12,7 @@
 //! the C++ code into a binary with any cxx-qt-lib code and Qt linked.
 
 mod diagnostics;
-use diagnostics::Diagnostic;
+use diagnostics::{Diagnostic, GeneratedError};
 
 use convert_case::{Case, Casing};
 use quote::ToTokens;
@@ -62,7 +62,9 @@ impl GeneratedCpp {
 
         let rust_file_path = rust_file_path.as_ref();
 
-        let file = parse_qt_file(rust_file_path).map_err(to_diagnostic)?;
+        let file = parse_qt_file(rust_file_path)
+            .map_err(GeneratedError::from)
+            .map_err(to_diagnostic)?;
 
         let mut cxx_qt = None;
         let mut qml_metadata = Vec::new();
@@ -103,14 +105,19 @@ impl GeneratedCpp {
                             rust_file_path.display());
                     }
 
-                    let parser = Parser::from(m.clone()).map_err(to_diagnostic)?;
-                    let generated_cpp = GeneratedCppBlocks::from(&parser).map_err(to_diagnostic)?;
+                    let parser = Parser::from(m.clone())
+                        .map_err(GeneratedError::from)
+                        .map_err(to_diagnostic)?;
+                    let generated_cpp = GeneratedCppBlocks::from(&parser)
+                        .map_err(GeneratedError::from)
+                        .map_err(to_diagnostic)?;
                     // TODO: we'll have to extend the C++ data here rather than overwriting
                     // assuming we share the same file
                     cxx_qt = Some(write_cpp(&generated_cpp));
 
-                    let generated_rust =
-                        GeneratedRustBlocks::from(&parser).map_err(to_diagnostic)?;
+                    let generated_rust = GeneratedRustBlocks::from(&parser)
+                        .map_err(GeneratedError::from)
+                        .map_err(to_diagnostic)?;
                     let rust_tokens = write_rust(&generated_rust);
                     file_ident = parser.cxx_file_stem.clone();
                     for (_, qobject) in parser.cxx_qt_data.qobjects {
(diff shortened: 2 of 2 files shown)


ahayzen-kdab:
FWIW to test this i added the following method to the minimal example without the extern block for the QUrl so it is missing.

```rust
        #[qinvokable]
        pub fn test(&self, _url: &QUrl) {

        }
```

Then this is the full error, you can see how the first error is the correct one, the others are a side effect of the CXX expansion not occurring as it failed - I don't think we can silence these as these are from the Rust compiler itself trying to build the code with the failed expansion ? But this is already way better than before :-)

<details><summary>Full error</summary>
<p>

```
[build] warning: cxx-qt-build failed to parse cxx_qt::bridge: Cxx(Syn(Error("unsupported type: QUrl")))
[build] error: unsupported type: QUrl
[build]   --> examples/qml_minimal/rust/src/cxxqt_object.rs:57:35
[build]    |
[build] 57 |         pub fn test(&self, _url: &QUrl) {
[build]    |                                   ^^^^
[build] 
[build] error[E0432]: unresolved import `super::my_object`
[build]   --> examples/qml_minimal/rust/src/cxxqt_object.rs:12:5
[build]    |
[build] 12 | mod my_object {
[build]    |     ^^^^^^^^^ could not find `my_object` in `super`
[build] 
[build] error[E0433]: failed to resolve: use of undeclared type `QString`
[build]   --> examples/qml_minimal/rust/src/cxxqt_object.rs:37:25
[build]    |
[build] 37 |                 string: QString::from(""),
[build]    |                         ^^^^^^^
[build]    |                         |
[build]    |                         use of undeclared type `QString`
[build]    |                         help: a struct with a similar name exists: `String`
[build]    |
[build]    = note: consider importing this struct:
[build]            cxx_qt_lib::QString
[build] 
[build] error[E0412]: cannot find type `QString` in this scope
[build]    --> examples/qml_minimal/rust/src/cxxqt_object.rs:28:17
[build]     |
[build] 28  |         string: QString,
[build]     |                 ^^^^^^^ help: a struct with a similar name exists: `String`
[build]     |
[build]    ::: /var/home/andrew/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/string.rs:367:1
[build]     |
[build] 367 | pub struct String {
[build]     | ----------------- similarly named struct `String` defined here
[build]     |
[build]     = note: consider importing this struct:
[build]             cxx_qt_lib::QString
[build] 
[build] error[E0412]: cannot find type `MyObjectQt` in this scope
[build]   --> examples/qml_minimal/rust/src/cxxqt_object.rs:11:1
[build]    |
[build] 11 | #[cxx_qt::bridge]
[build]    | ^^^^^^^^^^^^^^^^^ not found in this scope
[build]    |
[build]    = note: this error originates in the attribute macro `cxx_qt::bridge` (in Nightly builds, run with -Z macro-backtrace for more info)
[build] 
[build] error[E0412]: cannot find type `QString` in this scope
[build]    --> examples/qml_minimal/rust/src/cxxqt_object.rs:52:39
[build]     |
[build] 52  |         pub fn say_hi(&self, string: &QString, number: i32) {
[build]     |                                       ^^^^^^^ help: a struct with a similar name exists: `String`
[build]     |
[build]    ::: /var/home/andrew/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/string.rs:367:1
[build]     |
[build] 367 | pub struct String {
[build]     | ----------------- similarly named struct `String` defined here
[build]     |
[build]     = note: consider importing this struct:
[build]             cxx_qt_lib::QString
[build] 
[build] error[E0412]: cannot find type `QUrl` in this scope
[build]   --> examples/qml_minimal/rust/src/cxxqt_object.rs:57:35
[build]    |
[build] 57 |         pub fn test(&self, _url: &QUrl) {
[build]    |                                   ^^^^ not found in this scope
[build]    |
[build]    = note: consider importing this struct:
[build]            cxx_qt_lib::QUrl
[build] 
[build] error[E0412]: cannot find type `MyObjectCxxQtThread` in this scope
[build]   --> examples/qml_minimal/rust/src/cxxqt_object.rs:11:1
[build]    |
[build] 11 | #[cxx_qt::bridge]
[build]    | ^^^^^^^^^^^^^^^^^ not found in this scope
[build]    |
[build]    = note: this error originates in the attribute macro `cxx_qt::bridge` (in Nightly builds, run with -Z macro-backtrace for more info)
[build] 
[build] error[E0412]: cannot find type `MyObjectQt` in module `super`
[build]   --> examples/qml_minimal/rust/src/cxxqt_object.rs:11:1
[build]    |
[build] 11 | #[cxx_qt::bridge]
[build]    | ^^^^^^^^^^^^^^^^^ help: a struct with a similar name exists: `MyObject`
[build] ...
[build] 24 |     pub struct MyObject {
[build]    |     ------------------- similarly named struct `MyObject` defined here
[build]    |
[build]    = note: this error originates in the attribute macro `cxx_qt::bridge` (in Nightly builds, run with -Z macro-backtrace for more info)
[build] 
[build] Some errors have detailed explanations: E0412, E0432, E0433.
[build] For more information about an error, try `rustc --explain E0412`.
[build] The following warnings were emitted during compilation:
[build] 
[build] warning: cxx-qt-build failed to parse cxx_qt::bridge: Cxx(Syn(Error("unsupported type: QUrl")))
```

</p>
</details>

LeonMatthesKDAB:
Hm, I'm currently looking into how CXX generates it's own errors.
The weird thing is that they actually do exit the build script unsuccessfully from what I can tell...

However, I then still get perfect error messages in rust_analyzer :thinking: 

I wonder if there's some output from cxx-build that cargo or rust-analyzer can actually understand to provide these nicer diagnostics :thinking: 

Output if I add a code similar to your example somewhere in cxx-qt-lib:
<details>
<summary>Full build script output</summary>

```
    Compiling cxx-qt-lib v0.4.1 (/home/kdab/Documents/projects/3371-Rust-RnD/cxx-qt/crates/cxx-qt-lib)
The following warnings were emitted during compilation:

warning: Could not open /usr/lib64/libQt6Qml.prl file to read libraries to link: No such file or directory (os error 2)

error: failed to run custom build command for `cxx-qt-lib v0.4.1 (/home/kdab/Documents/projects/3371-Rust-RnD/cxx-qt/crates/cxx-qt-lib)`

Caused by:
  process didn't exit successfully: `/home/kdab/Documents/projects/3371-Rust-RnD/cxx-qt/target/debug/build/cxx-qt-lib-c0bcb22de1a2c254/build-script-build` (exit status: 1)
  --- stdout
  cargo:rerun-if-env-changed=QMAKE
  cargo:rerun-if-env-changed=QT_VERSION_MAJOR
  cargo:rustc-link-search=/usr/lib64
  cargo:rustc-link-lib=Qt6Core
  cargo:rustc-link-lib=Qt6Gui
  cargo:rustc-link-lib=Qt6Qml
  cargo:warning=Could not open /usr/lib64/libQt6Qml.prl file to read libraries to link: No such file or directory (os error 2)
  cargo:rustc-cfg=qt_version_major="6"
  cargo:rerun-if-changed=src/core/qbytearray.rs
  cargo:rerun-if-changed=src/core/qcoreapplication.rs
// ....
  cargo:rerun-if-changed=src/qml/qqmlapplicationengine.rs
  cargo:rerun-if-changed=src/qml/qqmlengine.rs

  --- stderr

  error[cxxbridge]: unsupported type: QTest
     ┌─ src/core/qdatetime.rs:31:24
     │
  31 │         fn test(qurl: &QTest);
     │                        ^^^^^ unsupported type
```
</details>

ahayzen-kdab:
@LeonMatthesKDAB does CXX error or warn ? As note out build script only warns at the moment ?

LeonMatthesKDAB:
CXX errors, the build script actually fails.

I think this might be the code responsible for doing this:
``` rust
pub fn bridges(rust_source_files: impl IntoIterator<Item = impl AsRef<Path>>) -> Build {
    let ref mut rust_source_files = rust_source_files.into_iter();
    build(rust_source_files).unwrap_or_else(|err| {
        let _ = writeln!(io::stderr(), "\n\ncxxbridge error: {}\n\n", report(err));
        process::exit(1);
    })
}
```

LeonMatthesKDAB:
It does not panic though, but rather just return a non-zero exit code...
I don't really see what that would change though...

ahayzen-kdab:
Fun, so what happens if you do the same with `cxx-qt-build` ? As it might instruct the rust compiler to try not to build the Rust code but only to perform macro expansion or something :shrug:

LeonMatthesKDAB:
The strange thing is that I don't see this `cxxbridge::error` text exactly in the stderr output of CXX.
So it's possible something else is going on :thinking: 

There's also some machinery in this `report` function that I don't really understand :/

ahayzen-kdab:
The build script output isn't always shown in the console as it's an input to cargo.

ahayzen-kdab:
I wonder if in the bridge code 

```rust
    build(rust_source_files).unwrap_or_else(|err| {
        let _ = writeln!(io::stderr(), "\n\ncxxbridge error: {}\n\n", report(err));
        process::exit(1);
    })
```

Where the `report(err)` occurs, is this writing out the text in a way that appears like it's coming from the macro expansion phase?

```rust
pub(crate) fn report(error: impl StdError) -> impl Display {
    struct Report<E>(E);

    impl<E: StdError> Display for Report<E> {
        fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
            write!(formatter, "{}", self.0)?;
            let mut error: &dyn StdError = &self.0;

            while let Some(cause) = error.source() {
                write!(formatter, "\n\nCaused by:\n    {}", cause)?;
                error = cause;
            }

            Ok(())
        }
    }

    Report(error)
}
```

ahayzen-kdab:
Hmm but then the cause goes missing and we only get `cxx_qt error: unsupported type: QUrl` under the `stderr` part of output. Maybe we aren't enabling a feature or something :shrug:

ahayzen-kdab:
And it's definitely stopping at `cxx_build::bridges` in your `cxx-qt-lib` example

ahayzen-kdab:
Yup so the source() is None when i try the same code for us, `error: Cxx(Syn(Error("unsupported type: QUrl"))), source: None`. Maybe something in `cxx_gen` is causing this vs using `cxx_build` ?

Be-ing:
It may be worth reaching out to the rust-analyzer maintainers for advice how to handle this.

LeonMatthesKDAB:
WTF...

I'm just testing diagnostics on our current main.
They just work with rust_analyzer!

The output of `cmake --build` is still pretty cryptic:
```
thread 'main' panicked at 'Could not generate C++ from Rust file: Syn(Error("unsupported type: QUrl"))', crates/cxx-qt-build/src/lib.rs:127:14
```
But that is something that CXX fixes using the [codespan-reporting](https://docs.rs/codespan-reporting/latest/codespan_reporting/) anyway, which we can use as well...

LeonMatthesKDAB:
It definitely also works with our own errors as well:
image

So maybe we don't need this change and only need to pretty-print our own diagnostics in the build script???

LeonMatthesKDAB:
It even works with diagnostics inside `qinvokable` methods...
image

However, they do take a long time to show up!
Also, it generates a diagnostic on the macro itself:
image

Probably because one of the diagnostic Spans is pointing to output of our macro.

LeonMatthesKDAB:
So, I've got the basic diagnostic display going from our build script:

image

crates/cxx-qt-build/src/diagnostics.rs
```diff
@@ -0,0 +1,142 @@
...
+
+impl Diagnostic {
+    pub(crate) fn new(file_path: PathBuf, error: GeneratedError) -> Self {
+        Self { file_path, error }
+    }
+
+    fn byte_span_in(&self, source: &str) -> Option<Range<usize>> {
+        match &self.error {
+            // TODO: CXX doesn't have span available yet
+            GeneratedError::Cxx(_) => None,
```
ahayzen-kdab, line 56: We need to try and add `span()` to `cxx_gen::Error` upstream, as this will then allow us to point to the code

ahayzen-kdab: This is #536

LeonMatthesKDAB (approved):
LGTM

PR #464, opened 2023-03-06
cxx-qt-lib: add serde support for some types

ahayzen-kdab:
Related to #292

Look at #425 for previous comments, this has been split so that some types can be reviewed in parallel.


ahayzen-kdab:
Abandoning in favour of #425

PR #475, opened 2023-03-13
cxx-qt-build: if files in the qrc file change rerun cargo

### LINKED ISSUE #469: Files inside qrc should be included in `cargo:rerun-if-changed`
I have a Cargo only Rust project in which i use `.qrc("src/qml/qml.qrc")` this works great. However if i'm iterating on the QML files within the qrc file when i run `cargo run -p project` it doesn't recompile the Qt resources.

It appears that if ` .qrc` file is specified in the build script we need to parse the XML (or as rcc to list the absolute paths, is that possible?) and add them to the `cargo:rerun-if-changed`

Note that there is `rcc --list` which may help us here ?

ahayzen-kdab:
Closes #469

### DIFF
--- a/crates/qt-build-utils/src/lib.rs
+++ b/crates/qt-build-utils/src/lib.rs
@@ -630,6 +630,27 @@ Q_IMPORT_PLUGIN({plugin_class_name});
             );
         }
 
+        // Add the qrc file contents to the cargo rerun list
+        let cmd_list = Command::new(self.rcc_executable.as_ref().unwrap())
+            .args(["--list", input_path.to_str().unwrap()])
+            .output()
+            .unwrap_or_else(|_| panic!("rcc --list failed for {}", input_path.display()));
+
+        if !cmd_list.status.success() {
+            panic!(
+                "rcc --list failed for {}:\n{}",
+                input_path.display(),
+                String::from_utf8_lossy(&cmd.stderr)
+            );
+        }
+
+        for path in String::from_utf8_lossy(&cmd_list.stdout)
+            .split('\n')
+            .map(PathBuf::from)
+        {
+            println!("cargo:rerun-if-changed={}", path.display());
+        }
+
         output_path
     }
 }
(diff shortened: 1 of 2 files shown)


crates/qt-build-utils/src/lib.rs
```diff
@@ -632,4 +632,33 @@ Q_IMPORT_PLUGIN({plugin_class_name});
 
         output_path
     }
+
+    /// Run [rcc](https://doc.qt.io/qt-6/resources.html) with --list on a .qrc file
+    /// and return the files found within the resource.
+    ///
+    /// This can be used to indicate to cargo when to rerun as the contents have changed
+    pub fn qrc_files(&mut self, input_file: &impl AsRef<Path>) -> Vec<PathBuf> {
```
Be-ing, line 640: Does this need to be a separate method from `qrc`? How about combining them?

ahayzen-kdab: Potentially, the question becomes, should `qt-build-utils` emit `cargo:rerun-if-changed` for files or is that the job of `cxx-qt-build` ? If it's fine here then it can be moved :-)

ahayzen-kdab: Ah i see there are some for env vars in there already, so done this change :-)

crates/qt-build-utils/src/lib.rs
```diff
@@ -630,6 +630,27 @@ Q_IMPORT_PLUGIN({plugin_class_name});
...
+        if !cmd_list.status.success() {
+            panic!(
+                "rcc --list failed for {}:\n{}",
+                input_path.display(),
+                String::from_utf8_lossy(&cmd.stderr)
+            );
+        }
+
+        for path in String::from_utf8_lossy(&cmd_list.stdout)
+            .split('\n')
```
ahayzen-kdab, line 648: do we think this works on Windows?

Be-ing: Not sure? If it doesn't, it would be easier to fix later if someone finds it to be an issue.

Be-ing (approved):

PR #478, opened 2023-03-15
Allow associated constants in `impl qobject::T`

### LINKED ISSUE #476: Allow associated constants in qobject::T
Especially when using roles in QAIMs, it's useful to associate a constant with each role, for more expressive code:
e.g.: 
``` rust
impl qobject::MarkdownCombiner {
    pub const NAME_ROLE : i32 = 0;
    pub const SELECTED_ROLE : i32 = 1;
}
```

However, cxx-qt currently does not allow this.

It does work on the native Rust struct (e.g. MarkdownCombiner). But that then requires more boilerplate to import the constants.

LeonMatthesKDAB:
Closes #476

### DIFF
--- a/crates/cxx-qt-gen/src/generator/rust/qobject.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/qobject.rs
@@ -15,7 +15,7 @@ use crate::{
     parser::qobject::ParsedQObject,
 };
 use quote::quote;
-use syn::{Ident, ImplItemFn, Item, Result};
+use syn::{Ident, ImplItem, Item, Result};
 
 #[derive(Default)]
 pub struct GeneratedRustQObjectBlocks {
@@ -87,9 +87,10 @@ impl GeneratedRustQObject {
             &qobject.invokables,
             &qobject_idents,
         )?);
-        generated
-            .blocks
-            .append(&mut generate_methods(&qobject.methods, &qobject_idents)?);
+        generated.blocks.append(&mut generate_passthrough_impl(
+            &qobject.passthrough_impl_items,
+            &qobject_idents,
+        )?);
         generated.blocks.append(&mut inherit::generate(
             &qobject_idents,
             &qobject.inherited_methods,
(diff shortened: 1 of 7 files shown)


crates/cxx-qt-gen/src/generator/rust/qobject.rs
```diff
@@ -15,7 +15,7 @@ use crate::{
     parser::qobject::ParsedQObject,
```
ahayzen-kdab, line 15: could add a readme entry

LeonMatthesKDAB: Good point.

LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -205,10 +206,14 @@ impl ParsedQObject {
                 ImplItem::Method(method) => {
                     self.parse_impl_method(method)?;
                 }
+                ImplItem::Const(constant) => {
+                    self.passthrough_impl_items
+                        .push(ImplItem::Const(constant.clone()));
+                }
                 _ => {
```
ahayzen-kdab, line 213: what about `ImplItem::Type`, `ImplType::Macro`, and `ImplType::Verbatim`, some or all of these seems relevant to pass through as well ? Eg there isn't much difference between Const and Type ?

crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -205,10 +206,14 @@ impl ParsedQObject {
                 ImplItem::Method(method) => {
                     self.parse_impl_method(method)?;
                 }
+                ImplItem::Const(constant) => {
+                    self.passthrough_impl_items
+                        .push(ImplItem::Const(constant.clone()));
```
ahayzen-kdab, line 211: assume you can't do `item.clone()` here ?

crates/cxx-qt-gen/src/parser/qobject.rs
LeonMatthesKDAB: Good point :+1: 
Now passed through as well

crates/cxx-qt-gen/src/parser/qobject.rs
LeonMatthesKDAB: I can, actually :D

ahayzen-kdab (approved):
LGTM

PR #480, opened 2023-03-16
Display diagnostics when failing the build script

LeonMatthesKDAB:
As discussed in #461 , the correct route to getting reasonable diagnostics seems to be to:

Not panic in the build script, but rather display nice diagnostics, using e.g. [codespan_reporting](https://docs.rs/codespan-reporting/latest/codespan_reporting/), then std::process::exit(1) to signal failure.
Note: Technically when parsing, we could still use the old approach of not panicking in the build script, but just emitting a warning and having the macro expansion handle the rest. We do gain the ability of providing nicer hints and suggestions though with codespan_reporting.

Current open issue:
cxx-qt-build calls: `cxx_gen::generate_header_and_cc`, which only returns a `cxx_gen::Error`.
Unfortunately this error no longer contains source information in it public API, which means we can't convert it into a nice diagnostic :disappointed: 
We could upstream this though, as it's a matter of getting access to the underlying error, which is stored, but not accessible from the outside.


LeonMatthesKDAB:
Current result:
image

CXX errors are currently still panicking though.
Also, the code is in no shape to be merged yet!
Needs to be rewritten cleanly. Won't have time for this before vacation though unforunately.

crates/cxx-qt-build/src/lib.rs
```diff
@@ -494,4 +507,68 @@ impl CxxQtBuilder {
         }
         self.cc_builder.compile("cxx-qt-gen");
     }
+
+    fn range_for(error: &Error, source: &str) -> Option<std::ops::Range<usize>> {
```
Be-ing, line 511: This function is unrelated to the CxxQtBuilder struct. I think it would be better to make it a free function (private to the module).

LeonMatthesKDAB: Yes, absolutely, as I mentioned, this is in no way production-ready code.
I just dumped it here to try diagnostics out.

LeonMatthesKDAB:
This branch is now ready for review & merge.

CXX errors still cause a panic, as we don't currently have access to the data needed to report CXX errors with nice diagnostics. We will need upstream fixes to change this.

.gitignore
```diff
@@ -16,3 +16,7 @@ rustup/*
 
 # clangd
 compile_commands.json
+
+# cmps (https://github.com/leonmatthes/cmps)
+.cmps/
```
ahayzen-kdab, line 21: is there something in this commit that relates to this change?

LeonMatthesKDAB: No, but I've added license-templates to cmps now :partying_face: 
Could split this into a separate PR if necessary, thought it a bit of a small change for that though.

ahayzen-kdab:
So adding the following to an object where QUrl isn't imported still errors without the correct location being printed?

Do you have an example I can try before and after to see how this changes the output ? As the other pull request ( #461 ) seems to show more info currently ?

```rust
        #[qinvokable]
        pub fn test(&self, _url: &QUrl) {

        }
```

LeonMatthesKDAB:
> So adding the following to an object where QUrl isn't imported still errors without the correct location being printed?
> 
> Do you have an example I can try before and after to see how this changes the output ? As the other pull request ( #461 ) seems to show more info currently ?
> 
> ```rust
>         #[qinvokable]
>         pub fn test(&self, _url: &QUrl) {
> 
>         }
> ```

Right, I think that's an error caused by CXX. Those don't get displayed correctly (yet).
you need to find something where CXX-Qt is generating the error.
For example, try using "&mut self" in the invokable, that's a CXX-Qt error (I think).

The difference compared to the other PR is that we're failing early, like CXX does.
That has the advantage that we don't spam many different failures, but can emit the issue directly, without causing ripple effects.
Additionally, we could even improve our diagnostics, to show hints and suggestions, just like the Rust compiler does.

ahayzen-kdab:
> > So adding the following to an object where QUrl isn't imported still errors without the correct location being printed?
> > Do you have an example I can try before and after to see how this changes the output ? As the other pull request ( #461 ) seems to show more info currently ?
> > ```rust
> >         #[qinvokable]
> >         pub fn test(&self, _url: &QUrl) {
> > 
> >         }
> > ```
> 
> Right, I think that's an error caused by CXX. Those don't get displayed correctly (yet). you need to find something where CXX-Qt is generating the error. For example, try using "&mut self" in the invokable, that's a CXX-Qt error (I think).
> 
> The difference compared to the other PR is that we're failing early, like CXX does. That has the advantage that we don't spam many different failures, but can emit the issue directly, without causing ripple effects. Additionally, we could even improve our diagnostics, to show hints and suggestions, just like the Rust compiler does.

OK, so the other PR fixes CXX errors appearing correctly, this is trying to fix CXX-Qt errors. So we might need a combination of the two.

ahayzen-kdab (approved):
LGTM, cleans up when there are CXX-Qt errors so that the error/warning is less confusing :-)

LeonMatthesKDAB:
Okay, so current analysis and way forward:

# TL;DR
Merge this PR - continue displaying diagnostics from build script and make sure rust-analyzer is up-to-date.

## How Rust-analyzer displays diagnostics:
Rust-analyzer displays diagnostics for proc-macros in two main ways:
1. `cargo check` - the most accurate, rust-analyzer simply runs cargo check and displays any diagnostics returned.
    This no longer works if a build script fails, as we don't get true diagnostics out of this.
    Which is where version 2 comes in.
2. Rust-analyzer expands the macro by itself and shows any diagnostics on the output it generates
    This is really quick and works even without saving the file first.
    The reason I didn't get diagnostics for this was quite convoluted.
    Let's dive into how proc-macros should report errors.

### Proc-macro error reporting
See: https://docs.rs/syn/latest/syn/struct.Error.html#error-reporting-in-proc-macros

Basically, the proc-macro should expand to an invocation of `compile_error!`, instead of producing any real output.
If this is done with the correct span on it, the error will show up as if it originated from something within the macro, which is exactly what we want.

The reason this didn't work for me was because I used a combination of an old rust-analyzer + syn2, where CXX still used syn1.

syn2 emits `::core::compile_error!`, whereas syn1 emits `compile_error!`.
The former is probably correct long-term, but my version of rust-analyzer just couldn't deal with it.
After updating rust-analyzer - everything works now! :partying_face: 

The reason fiddling with the build script changed things was that `cargo check` correctly recognizes `::core::compile_error!`, as well as `compile_error!`.
So if the build script doesn't fail, rust-analyzer can just run `cargo check` to display the diagnostic.
This then does however lead to duplicate diagnostics in many cases.

## Build script error reporting
Ideally we don't actually want to run into macro-expansion if our build script fails.
We want to fail early in the build script.
But if we just panic, the output isn't very nice.

Therefore we can use the `codespan-reporting` crate to pretty-print our error message, then just fail the build script (i.e. `exit(1)`).
That way we get a nice human-readable error message as quickly as possible.

This will mean though that `cargo check` now fails without producing usable diagnostics for `rust-analyzer`.
But that is fine as long as we produce a nice `compile_error!` in our procedural macro as well.

ahayzen-kdab (approved):
LGTM

PR #482, opened 2023-03-17
cxxqtthread: add Sync marker so that it can be used by reference

### LINKED ISSUE #481: `CxxQtThread` doesn't implement `Copy` or `Sync` which then means you have to be careful using async code with it.
`CxxQtThread` doesn't implement `Copy` or `Sync` which then means you have to be careful using async code with it.

Eg

```rust
let qt_thread = self.qt_thread();

tokio::spawn(async move {
   ...
   let read_future = read.for_each(|message| async move {  // or async it complains about no Sync
      qt_thread.queue(...);  // with async move this complains about no Copy
   }
});
```

This fails because `CxxQtThread` doesn't implement `Copy` as each inner loop of the for each needs it's own `qt_thread` or if it wasn't `async move` then `CxxQtThread` needs to be `Sync` as each gets a reference.

Instead this needs to be changed to the following structure

```rust
let qt_thread = self.qt_thread();

tokio::spawn(async move {
   ...
   let read_future = async move {
      while let Some(message) = read.next().await {
          qt_thread.queue(...);
      }
   }
});
```

Which is fine and works with our current code. Just something to consider if we could / would want to add Copy / implement Sync to CxxQtThread or not. Maybe we don't and this is OK as trying to prove that that is safe might be tricky ?

_Originally posted by @ahayzen-kdab in https://github.com/KDAB/cxx-qt/discussions/468#discussioncomment-5266111_

ahayzen-kdab:
This then means it can be used in async blocks too.

Closes #481

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for further types: `QLine`, `QLineF`, `QImage`, `QPainter`, `QFont`, `QPen`
 - `internal_pointer_mut()` function on `QModelIndex`
 - `c_void` in CXX-Qt-lib for easy access to `void *`
+- `CxxQtThread` is now marked as `Sync` so that it can be used by reference
 
 ### Changed
 
--- a/crates/cxx-qt/src/threading.rs
+++ b/crates/cxx-qt/src/threading.rs
@@ -57,7 +57,13 @@ where
     }
 }
 
+// CxxQtThread is safe to be sent across threads and handles
+// locking and checks with the original QObject to prevent issues
 unsafe impl<T> Send for CxxQtThread<T> where T: Threading {}
+
+// CxxQtThread is safe to use as a reference in parallel from multiple
+// places as it protects the queue call and the closure with mutexes
+unsafe impl<T> Sync for CxxQtThread<T> where T: Threading {}
 
 impl<T> CxxQtThread<T>
 where


crates/cxx-qt-gen/src/writer/rust/mod.rs
```diff
@@ -90,8 +90,15 @@ fn cxx_qt_common_blocks(qobject: &GeneratedRustQObject) -> Vec<TokenStream> {
...
     vec![
         quote! {
+            // CxxQtThread is safe to be sent across threads and handles
+            // locking and checks with the original QObject to prevent issues
             unsafe impl Send for #cxx_qt_thread_ident {}
         },
+        quote! {
+            // CxxQtThread is safe to use as a reference in parallel from multiple
+            // places as it protects the queue call and the closure with mutexes
+            unsafe impl Sync for #cxx_qt_thread_ident {}
```
ahayzen-kdab, line 100: Need to check that this does work with `async` blocks

ahayzen-kdab: So this allows it to be used in async blocks, but still doesn't quite work with my usecase due to issues with higher order lifetimes etc. Something to do with the way `tokio::join!(my_future)`  works

ahayzen-kdab: You can see this fun here https://gitlab.com/ahayzen/home-dash-alpha-g1/-/merge_requests/3

The error i get is

```rust
error: higher-ranked lifetime error
  --> src/dashboard/src/backend/weathercontroller.rs:27:13
   |
27 | /             tokio::spawn(async move {
28 | |                 // Connect to the server
29 | |                 let (ws_stream, _) = connect_async(client_ws_url()).await.unwrap();
30 | |                 let (mut write, read) = ws_stream.split();
...  |
66 | |                 tokio::join!(read_future, write_future);
67 | |             });
   | |______________^
   |
   = note: could not prove `[async block@src/dashboard/src/backend/weathercontroller.rs:27:26: 67:14]: std::marker::Send`
```

Which suggests something in the async block it can't prove is `Send`, i might need to add a where clause somewhere.

But what is interesting is that changing `tokio::spawn!(read_future, write_future)` into `read_future.await; write_future.await;` does work. And adding `pin_mut!(read_future, write_future)` doesn't help. And if the `qt_thread.queue(...)` part is removed then it works. So some combination of `tokio::spawn` and using `qt_thread` breaks things with these `async` rather than `async move` blocks.

ahayzen-kdab: @LeonMatthesKDAB I think we just merge this as-is as it still helps for reference and async blocks it just didn't totally solve the usecase i had before.

LeonMatthesKDAB (approved):
There's really no downside to this, so let's finally merge it!

crates/cxx-qt-gen/src/writer/rust/mod.rs
LeonMatthesKDAB: Agreed

Issue #61, opened 2022-01-14
Consider adding the new property value to C++ change events


ahayzen-kdab:
Currently we generate code the is `void valueChanged();` in the C++, this works fine for QML. But when using C++ / widgets sometimes it's easier if the signal passes the new property value as well, eg `void valueChanged(int value)`.

ahayzen-kdab:
Once property changes are emitted immediately do we still need to do this? #281

ahayzen-kdab:
Properties are emitted immediately and with the new API if you had a custom getter + setter + notify you could then add the value if you wanted it.

PR #485, opened 2023-03-20
cxx-qt-gen: (dis)connect from signals

### LINKED ISSUE #473: connect signals and slots in Rust
Currently we can declare and emit signals from Rust, but not connect signals and slots.

ahayzen-kdab:
Closes #473


crates/cxx-qt-gen/src/generator/cpp/signal.rs
```diff
@@ -7,7 +7,7 @@ use crate::{
     generator::{
```
ahayzen-kdab, line 7: need to add the same support for our generated properties so that `self.on_property_changed(|value| ..);` works.

examples/qml_features/rust/src/signals.rs
```diff
@@ -49,6 +51,28 @@ pub mod ffi {
...
+
+        #[qinvokable]
+        pub fn toggle_logging(mut self: Pin<&mut Self>) {
+            if let Some(connections) = self.as_mut().connections_mut().take() {
+                for conn in connections {
+                    conn.disconnect();
+                }
+            } else {
+                let connections = [
+                    self.as_mut().on_connected(|url| {
```
ahayzen-kdab, line 63: should we have the `Qt::ConnectionType` as an argument here ? Eg so you can do `Qt::DirectConnection` or `Qt::QueuedConnection` rather than always assuming `Auto` ?

examples/qml_features/rust/src/signals.rs
```diff
@@ -49,6 +51,28 @@ pub mod ffi {
...
+                        println!("Connected: {}", url);
+                    }),
+                    self.as_mut().on_error(|message| {
+                        println!("Error: {}", message);
```
ahayzen-kdab, line 67: These are just function pointers so don't have the ability to move values into them.

So i guess the first argument should always be the C++ context object like in the `queue(|qobject| ...)` method for threading. Otherwise you can't read or manipulate state...

examples/qml_features/rust/src/signals.rs
Be-ing: yes

examples/qml_features/rust/src/signals.rs
ahayzen-kdab: Done

examples/qml_features/rust/src/signals.rs
ahayzen-kdab: OK wonder if it should be the first or last argument, in Qt it's the last, but then it's after the closure, maybe that's fine though

ahayzen-kdab: Done, added them as the last parameter.

crates/cxx-qt-gen/src/generator/cpp/signal.rs
ahayzen-kdab: We don't need this instead you can use the `#[inherit]` on the signals to reach this which means we don't need to duplicate a load of code.

Eg

```rust
#[cxx_qt::qobject]
pub struct MyObject {
        #[qproperty]
        my_property: bool,
}
#[cxx_qt::qsignals(MyObject)]
pub enum Signals {
    #[inherit]
    MyPropertyChanged,
}
...
self.as_mut().on_my_property_changed(|qobject| println!("{}", qobject.my_property()), ConnectionType::AutoConnection);
```

Then depending what happens with #61 there might be an argument for the new value, otherwise you can just read it back (same as in C++)

ahayzen-kdab: So then this just needs documenting in the book

ahayzen-kdab: Added some notes to the book.

ahayzen-kdab: Although maybe there should be an example of this somewhere....

book/src/qobject/signals_enum.md
```diff
@@ -27,6 +27,26 @@ If a signal is defined on the base class of the QObject then `#[inherit]` can be
 
 Note that `#[cxx_name = "..."]` can also be used on a signal to declare a different name in C++ to Rust.
 
+## Connecting to a signal
+
+For every signal defined in the enum a method is generated `on_<signal_name>`,
```
Be-ing, line 32:
```suggestion
For every signal defined in the enum, a method is generated called `on_<signal_name>`.
```

book/src/qobject/signals_enum.md
```diff
@@ -27,6 +27,26 @@ If a signal is defined on the base class of the QObject then `#[inherit]` can be
...
+this method takes a function pointer as the fist parameter with the first argument being the qobject
+and the remaining arguments being the signal parameters, the second parameter is the Qt connection type.
```
Be-ing, line 34:
```suggestion
This method takes a function as the first parameter. That function's first argument is the qobject
and the remaining arguments are the signal parameters. The second parameter of the `on_<signal_name>` method is the [Qt connection type](https://doc.qt.io/qt-6/qt.html#ConnectionType-enum).
```

book/src/qobject/signals_enum.md
```diff
@@ -27,6 +27,26 @@ If a signal is defined on the base class of the QObject then `#[inherit]` can be
...
+
+Note that by using the `#[inherit]` macro on a signal, connections can be made to property changes
+using the signal name `<property>Changed` with no parameters.
+
+```rust,ignore,noplayground
+{{#include ../../../examples/qml_features/rust/src/signals.rs:book_signals_connect}}
+```
+
+Each connection returns a `QMetaObject::Connection` which then has an extra `disconnect` method
+in Rust which allows you to disconnect the Rust function pointer from the signal.
```
Be-ing, line 44:
```suggestion
Each connection returns a `QMetaObject::Connection` which can be disconnected later by calling its `disconnect` method.
```

crates/cxx-qt-gen/src/generator/cpp/signal.rs
```diff
@@ -82,12 +91,42 @@ pub fn generate_cpp_signals(
...
+                parameters = parameter_types_rust.join(", ")
+            ),
+            source: formatdoc! {
+                r#"
+                ::std::unique_ptr<QMetaObject::Connection>
+                {qobject_ident}::{connect_ident}(::rust::Fn<void({parameters_rust})> func, Qt::ConnectionType type)
+                {{
+                    return ::std::make_unique<QMetaObject::Connection>(QObject::connect(this,
+                            &{qobject_ident}::{signal_ident},
+                            this,
```
Be-ing, line 117: This syntax doesn't provide a way to specify anything other than `this` for the target. Is that a problem?

ahayzen-kdab: Right this is just for adding a function pointer to a signal, if you want to call a slot/method on another object then you should
- store a pointer to the object as a property/field
- connect to the signal you are interested in
- in the function pointer retrieve the qobject
- call the slot/invokable on the qobject via emit(...) or the method name

I think trying to add syntax and safety around making a direct connection between two objects is way more tricky to do and should be a separate task (if we even wanted it), and this is a good enough workaround for now.

book/src/qobject/signals_enum.md
ahayzen-kdab: Done

book/src/qobject/signals_enum.md
ahayzen-kdab: Done

book/src/qobject/signals_enum.md
ahayzen-kdab: Done

crates/cxx-qt-gen/src/generator/cpp/signal.rs
Be-ing: Can you make an example of this workaround to ensure that it does work before we commit to this syntax?

ahayzen-kdab: Added to the nested objects an example causing `OuterObject::Called` to trigger `InnerObject::Called`

examples/qml_features/rust/src/nested_qobjects.rs
```diff
@@ -40,6 +45,25 @@ mod ffi {
...
+        #[qinvokable]
+        pub fn initialise(self: Pin<&mut Self>) {
+            // Example of connecting a signal from one QObject to another QObject
+            // this causes OuterObject::Called to trigger InnerObject::Called
+            self.on_called(
+                |qobject, obj| {
+                    // We need to convert the *mut T to a Pin<&mut T> so that we can reach the methods
+                    if let Some(inner) = unsafe { qobject.inner().as_mut() } {
+                        // TODO: Use `pin!` one it's stable so that this unsafe block can be removed
+                        // https://doc.rust-lang.org/std/pin/macro.pin.html
```
Be-ing, line 57: This is stable now.

ahayzen-kdab: Yup but unfortunately doesn't quite help us enough, it can go from T -> Pin<&mut T>, but we need to get from *mut T to T somehow. This is just copied from the other code below and should be solved in another pull request (i have another branch locally with experiments).

ahayzen-kdab: (As ultimately we need to go from *mut T -> Pin<&mut T>

ahayzen-kdab: Removed this comment for now as it doesn't help us.

examples/qml_features/rust/src/signals.rs
```diff
@@ -49,6 +63,57 @@ pub mod ffi {
...
+                            for conn in connections {
+                                conn.disconnect();
+                            }
+                        }
+                        // ANCHOR_END: book_signals_disconnect
+                    }
+                },
+                ConnectionType::AutoConnection,
+            )
+            .pin_mut()
```
ahayzen-kdab, line 114: Need to get from `UniquePtr<QMetaObjectConnectionGuard>` to `Pin<&mut QMetaObjectConnectionGuard>` which doesn't happen via deref to be able to reach the non-const `release` method.

The other option would be to make inner `mutable` in the C++ object and then make `release` const. But maybe forcing the `pin_mut` is fine?

examples/qml_features/rust/src/signals.rs
```diff
@@ -49,6 +63,57 @@ pub mod ffi {
...
+                                ),
+                            ];
+                            qobject.as_mut().set_connections(Some(connections));
+                            // ANCHOR_END: book_signals_connect
+                        }
+                    } else {
+                        // Disabling logging so disconnect
+                        // ANCHOR: book_signals_disconnect
+                        // Could also empty the connections which would cause drop to trigger a disconnect
+                        // qobject.as_mut().set_connections(None);
```
ahayzen-kdab, line 103: need to add a usecase showing and testing this

examples/qml_features/rust/src/signals.rs
```diff
@@ -49,6 +63,57 @@ pub mod ffi {
...
+                        if let Some(connections) = qobject.as_mut().connections_mut().take() {
+                            for conn in connections {
+                                conn.disconnect();
```
ahayzen-kdab, line 106: need to add a usecase testing this method

book/src/qobject/signals_enum.md
```diff
@@ -27,6 +27,26 @@ If a signal is defined on the base class of the QObject then `#[inherit]` can be
...
+The second parameter of the `on_<signal_name>` method is the [Qt connection type](https://doc.qt.io/qt-6/qt.html#ConnectionType-enum).
+
+Note that by using the `#[inherit]` macro on a signal, connections can be made to property changes
+using the signal name `<property>Changed` with no parameters.
+
+```rust,ignore,noplayground
+{{#include ../../../examples/qml_features/rust/src/signals.rs:book_signals_connect}}
+```
+
+Each connection returns a [`QMetaObject::Connection`](https://doc.qt.io/qt-6/qmetaobject-connection.html) which can be disconnected later by calling its `disconnect` method.
```
ahayzen-kdab, line 44: need to update this doc

examples/qml_features/rust/src/signals.rs
ahayzen-kdab: Changed to trivial.

examples/qml_features/rust/src/signals.rs
ahayzen-kdab: This is the only way of doing it now.

examples/qml_features/rust/src/signals.rs
ahayzen-kdab: There is testing in the C++ tests

book/src/qobject/signals_enum.md
ahayzen-kdab: Updated.

crates/cxx-qt-gen/src/generator/rust/signals.rs
```diff
@@ -203,6 +265,19 @@ mod tests {
...
         );
+        assert_tokens_eq(
+            &generated.cxx_mod_contents[7],
+            quote! {
+                unsafe extern "C++" {
+                    #[doc = "Connect the given function pointer to the signal "]
+                    #[doc = "baseName"]
+                    #[doc = ", so that when the signal is emitted the function pointer is executed."]
+                    #[must_use]
+                    #[rust_name = "on_base_name"]
```
ahayzen-kdab, line 276: the name ends up being odd here as it's the `cxx_name`.

ahayzen-kdab: But this also happened before the signal name internally anyway, so probably a separate change.

examples/qml_features/rust/src/signals.rs
```diff
@@ -67,6 +79,53 @@ pub mod ffi {
...
+                                ),
+                            ];
+                            qobject.as_mut().set_connections(Some(connections));
+                            // ANCHOR_END: book_signals_connect
+                        }
+                    } else {
+                        // Disabling logging so disconnect
+                        // ANCHOR: book_signals_disconnect
+                        // By making connections None, we trigger a drop on the connections
+                        // this then causes disconnections
```
LeonMatthesKDAB, line 120: Nice, this is pretty elegant :+1:

crates/cxx-qt-gen/src/generator/naming/signals.rs
```diff
@@ -51,6 +53,13 @@ impl CombinedIdent {
             rust: ident,
         }
     }
+
+    fn connect_from_signal(ident: &Ident) -> Self {
+        Self {
+            cpp: format_ident!("{}Connect", ident.to_string().to_case(Case::Camel)),
```
LeonMatthesKDAB, line 59: Why not `on{}` with `Case::Pascal` here?
Would that conflict with QML?

If so, please add a comment about it here.

crates/cxx-qt-gen/test_inputs/signals.rs
```diff
@@ -36,6 +36,12 @@ mod ffi {
     impl qobject::MyObject {
         #[qinvokable]
         pub fn invokable(self: Pin<&mut Self>) {
+            self.as_mut().on_data_changed(
+                |_, _, _, _| {
```
LeonMatthesKDAB, line 40: This should be five parameters here, as you always get the `self` qobject as the first parameter, right?
Maybe also add names here, e.g.: 
```suggestion
                |_sender,_first, _second, _third, _fourth| {
```

crates/cxx-qt-lib/src/core/qt.rs
```diff
@@ -22,6 +22,22 @@ mod ffi {
         CaseSensitive,
     }
 
+    #[repr(i32)]
+    enum ConnectionType {
```
LeonMatthesKDAB, line 26: Please add a note here that UniqueConnection is **not** supported, because we're currently always using lambdas as functors.

crates/cxx-qt-lib/src/core/qmetaobjectconnection.rs
```diff
@@ -0,0 +1,60 @@
...
+    unsafe extern "C++" {
+        include!("cxx-qt-lib/common.h");
+
+        #[doc(hidden)]
+        #[rust_name = "qmetaobjectconnection_drop"]
+        fn drop(conn: &mut QMetaObjectConnection);
+    }
+}
+
+/// Represents a handle to a signal-slot (or signal-functor) connection.
```
LeonMatthesKDAB, line 32: Please expand the documentation here a bit.

Especially the fact that a QMetaObjectConnection automatically disconnects when dropped!
As that's something we've changed compared to Qt.

Also, pointing out that you can connect to a signal using `on_SIGNAL_NAME` could quickly be explained here.

crates/cxx-qt-lib/src/core/qt.rs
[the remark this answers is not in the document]
LeonMatthesKDAB: Some other ideas:
1. Use `on_MY_SIGNAL` and `on_MY_SIGNAL_with_type` - because we don't have control over the signal name, this might be easy to overlook though.
2. Use `on_MY_SIGNAL` for Auto-type connections, and `connect_MY_SIGNAL` for the "advanced" API with options.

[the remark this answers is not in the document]
ahayzen-kdab: Number 2) here i think is best, i'll try that..

crates/cxx-qt-gen/src/generator/naming/signals.rs
ahayzen-kdab: I think that's what i was concerned about, that the C++ method then could start conflicting with the QML API and seem "odd". I'll add a comment if that was my concern...

crates/cxx-qt-gen/test_inputs/signals.rs
ahayzen-kdab: Oops, can you tell we don't actually build these examples and they are just for syntax checking :-)

crates/cxx-qt-gen/src/generator/naming/signals.rs
ahayzen-kdab: Done

crates/cxx-qt-gen/test_inputs/signals.rs
ahayzen-kdab: Done

crates/cxx-qt-lib/src/core/qmetaobjectconnection.rs
ahayzen-kdab: Done, but maybe could explain better :thinking:

crates/cxx-qt-lib/src/core/qt.rs
```diff
@@ -22,6 +22,25 @@ mod ffi {
         CaseSensitive,
     }
 
+    /// This enum describes the types of connection that can be used with signals.
+    ///
+    /// Note that UniqueConnection is not supported as
```
LeonMatthesKDAB, line 27:
```suggestion
    /// Note that UniqueConnection is not supported as we only support Lambda functors at the moment.
```

crates/cxx-qt-lib/src/core/qt.rs
LeonMatthesKDAB: I think your comment may have been cut off there :thinking: 
See my suggestion.

crates/cxx-qt-lib/src/core/qt.rs
ahayzen-kdab: haha wonder what happens there must have not flushed the file :-)

crates/cxx-qt-lib/src/core/qt.rs
ahayzen-kdab: Done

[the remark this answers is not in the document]
ahayzen-kdab: Done

crates/cxx-qt-gen/src/generator/rust/signals.rs
ahayzen-kdab: Fixed :-)

book/src/qobject/signals_enum.md
```diff
@@ -27,6 +27,27 @@ If a signal is defined on the base class of the QObject then `#[inherit]` can be
...
+
+Note that by using the `#[inherit]` macro on a signal, connections can be made to property changes
+using the signal name `<property>Changed` with no parameters.
+
+```rust,ignore,noplayground
+{{#include ../../../examples/qml_features/rust/src/signals.rs:book_signals_connect}}
+```
+
+Each connection returns a [`QMetaObject::Connection`](https://doc.qt.io/qt-6/qmetaobject-connection.html) which can be disconnected later by `drop` occurring on the type,
+if you don't want a disconnect to occur call `release`.
```
LeonMatthesKDAB, line 45:
```suggestion
Each connection returns a [`QMetaObject::Connection`](https://doc.qt.io/qt-6/qmetaobject-connection.html) which is used to manage the connection.
Note that the `QMetaObjectConnection` returned by Cxx-Qt behaves a bit different from the Qt C++ implementation.
When the `QMetaObjectConnection` is dropped, it automatically disconnects the connection, similar to how a C++ `std::unique_ptr` or Rusts `Box` behaves.
If you don't want to store the QMetaObjectConnection, call `release`, which will drop the object without disconnectiong.
In this case, it is no longer possibly to disconnect later.
```

book/src/qobject/signals_enum.md
```diff
@@ -27,6 +27,27 @@ If a signal is defined on the base class of the QObject then `#[inherit]` can be
 
 Note that `#[cxx_name = "..."]` can also be used on a signal to declare a different name in C++ to Rust.
 
+## Connecting to a signal
+
+For every signal defined in the enum, a method is generated called `on_<signal_name>`.
+This method takes a function as the first parameter.
+That function's first argument is the qobject and the remaining arguments are the signal parameters.
+The second parameter of the `on_<signal_name>` method is the [Qt connection type](https://doc.qt.io/qt-6/qt.html#ConnectionType-enum).
```
LeonMatthesKDAB, line 35:
```suggestion
For every signal defined in the enum, two methods are generated.

1. `on_<signal_name>`
2. `connect_<signal_name>`

The `on_<signal_name>` method takes a handler function as the parameter, which will be called when the signal is emitted.
That handler function's first argument is the qobject and the remaining arguments are the signal parameters.

The `connect_<signal_name>` function additionally takes the [Qt connection type](https://doc.qt.io/qt-6/qt.html#ConnectionType-enum) as a parameter.
```

crates/cxx-qt-lib/src/core/qt.rs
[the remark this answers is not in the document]
LeonMatthesKDAB: That cleaned up nicely

book/src/qobject/signals_enum.md
ahayzen-kdab: Done

book/src/qobject/signals_enum.md
ahayzen-kdab: done

LeonMatthesKDAB (approved):
LGTM :partying_face:

PR #489, opened 2023-03-24
Cherry pick to 0.5.x

### LINKED ISSUE #488: Cherry picks commits for a 0.5.1
# Commits to cherry pick to a `0.5.x` branch

- [ ] cargo: ensure a version is specified for dependencies of published packages https://github.com/KDAB/cxx-qt/commit/b612f711e56e9ebb805340509c715eb975d2e0c1
- [ ] qt-build-utils: if files in the qrc file change rerun cargohttps://github.com/KDAB/cxx-qt/commit/e9fde3e95fd340a6a977a64abeb2ac335f3d5de7

# Further fixes

- [ ] https://github.com/KDAB/cxx-qt/issues/487

# Possibles

- [ ] https://github.com/KDAB/cxx-qt/issues/481

ahayzen-kdab:
Closes #488

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,17 +27,17 @@ members = [
 edition = "2021"
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/KDAB/cxx-qt/"
-version = "0.5.0"
+version = "0.5.1"
 
 # Note a version needs to be specified on dependencies of packages
 # we publish, otherwise crates.io complains as it doesn't know the version.
 [workspace.dependencies]
 cxx-qt = { path = "crates/cxx-qt" }
 cxx-qt-build = { path = "crates/cxx-qt-build" }
-cxx-qt-gen = { path = "crates/cxx-qt-gen", version = "0.5" }
+cxx-qt-gen = { path = "crates/cxx-qt-gen", version = "0.5.1" }
 cxx-qt-lib = { path = "crates/cxx-qt-lib" }
-cxx-qt-lib-headers = { path = "crates/cxx-qt-lib-headers", version = "0.5" }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -27,15 +27,17 @@ members = [
...
 [workspace.dependencies]
 cxx-qt = { path = "crates/cxx-qt" }
 cxx-qt-build = { path = "crates/cxx-qt-build" }
-cxx-qt-gen = { path = "crates/cxx-qt-gen" }
+cxx-qt-gen = { path = "crates/cxx-qt-gen", version = "0.5" }
 cxx-qt-lib = { path = "crates/cxx-qt-lib" }
-cxx-qt-lib-headers = { path = "crates/cxx-qt-lib-headers" }
-qt-build-utils = { path = "crates/qt-build-utils" }
+cxx-qt-lib-headers = { path = "crates/cxx-qt-lib-headers", version = "0.5" }
+qt-build-utils = { path = "crates/qt-build-utils", version = "0.5" }
```
ahayzen-kdab, line 40: wonder if we should bump these versions as well with microreleases. `cargo update` should sort things out, but then this would ensure if you put `0.5.1` as the version of the main crate and dependencies are `0.5.1` as well

vimpostor: I think it doesn't matter much, but it would certainly be more consistent.
Might be good for reducing risk of confusion, why one crate has version 0.5.1 and the other 0.5.

I also found this old cargo issue with adding support for referencing other versions, but it was closed as won't fix unfortunately: https://github.com/rust-lang/cargo/issues/4221

ahayzen-kdab: Changed to be `0.5.1` for the depends to be explicit.

PR #493, opened 2023-03-29
doc: add `--global` to `git config` when setting up symlink

OlivierLDff:
It is stated that symlink should be enabled before cloning, but copying the current command from readme doesn't work if we are not in a repo.

```
> git config core.symlinks true
fatal: not in a git directory
```

Adding the --global flag write it in the user config.


Be-ing (approved):
thanks

PR #494, opened 2023-03-29
cxx-qt-gen: Fix the format of the error when a file is not found

OlivierLDff:
Before, the missing file was not logged in the panic message, making it hard to debug.

I encountered this error while following the qml-minimal example, I did a typo in my build.rs. Being not familiar with build.rs, it was quite frustrating for me to find the source of the problem.

```
ompiling qml-minimal v0.1.0 (C:\Users\olivi\dev\cxx-qt-tutorial\rust)
error: failed to run custom build command for `qml-minimal v0.1.0 (C:\Users\olivi\dev\cxx-qt-tutorial\rust)`

Caused by:
  process didn't exit successfully: `C:\Users\olivi\dev\cxx-qt-tutorial\rust\target\debug\build\qml-minimal-b0eab65f7e8191c2\build-script-build` (exit code: 101)
  --- stdout
  cargo:rerun-if-changed=src/cxxqt_object_doesntexist.rs
  cargo:rerun-if-env-changed=QMAKE
  cargo:rerun-if-env-changed=QT_VERSION_MAJOR
  cargo:rustc-link-search=C:/Qt/5.15.2/msvc2019_64/lib
  cargo:rustc-link-lib=Qt5Qml
  cargo:rustc-link-lib=Qt5Gui
  cargo:rustc-link-lib=Qt5Core
  cargo:rerun-if-changed=C:\Users\olivi\dev\cxx-qt-tutorial\rust/src/cxxqt_object_doesntexist.rs

  --- stderr
  thread 'main' panicked at 'Could not read path {} to string: Os { code: 2, kind: NotFound, message: "The system cannot find the file specified." }', C:\Users\olivi\scoop\persist\rustup\.cargo\registry\src\github.com-1ecc6299db9ec823\cxx-qt-gen-0.5.1\src\syntax\qtfile.rs:44:57
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
 ```

Now with this fix (I hacked examples/cargo_without_cmake/build.rs) to show a failing case.

```
cargo build -p qml-minimal-no-cmake
   Compiling qml-minimal-no-cmake v0.1.0 (C:\Users\olivi\dev\cxx-qt\examples\cargo_without_cmake)
error: failed to run custom build command for `qml-minimal-no-cmake v0.1.0 (C:\Users\olivi\dev\cxx-qt\examples\cargo_without_cmake)`

Caused by:
  process didn't exit successfully: `C:\Users\olivi\dev\cxx-qt\target\debug\build\qml-minimal-no-cmake-48641d8fda518a0e\build-script-build` (exit code: 101)
  --- stdout
  cargo:rerun-if-changed=src/cxxqt_object_doesntexist.rs
  cargo:rerun-if-changed=qml/qml.qrc
  cargo:rerun-if-env-changed=QMAKE
  cargo:rerun-if-env-changed=QT_VERSION_MAJOR
  cargo:rustc-link-search=C:/Qt/5.15.2/msvc2019_64/lib
  cargo:rustc-link-lib=Qt5Network
  cargo:rustc-link-lib=Qt5Core
  cargo:rustc-link-lib=Qt5Qml
  cargo:rustc-link-lib=Qt5Gui
  cargo:rerun-if-changed=C:\Users\olivi\dev\cxx-qt\examples\cargo_without_cmake/src/cxxqt_object_doesntexist.rs

  --- stderr
  thread 'main' panicked at 'Failed to read file "C:\\Users\\olivi\\dev\\cxx-qt\\examples\\cargo_without_cmake/src/cxxqt_object_doesntexist.rs": The system cannot find the file specified. (os error 2)', C:\Users\olivi\dev\cxx-qt\crates\cxx-qt-gen\src\syntax\qtfile.rs:47:13
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
  
```

### DIFF
--- a/crates/cxx-qt-gen/src/syntax/qtfile.rs
+++ b/crates/cxx-qt-gen/src/syntax/qtfile.rs
@@ -41,7 +41,10 @@ impl ToTokens for CxxQtFile {
 }
 
 pub fn parse_qt_file(path: impl AsRef<std::path::Path>) -> Result<CxxQtFile> {
-    let source = std::fs::read_to_string(path.as_ref()).expect("Could not read path {} to string");
+    let source = std::fs::read_to_string(path.as_ref()).unwrap_or_else(|err| {
+        // todo: fixme with a proper error propagation
+        panic!("Failed to read file {:?}: {}", path.as_ref(), err);
+    });
 
     // We drop the shebang from the generated Rust code
     if source.starts_with("#!") && !source.starts_with("#![") {

### CODE REVIEW

File: crates/cxx-qt-gen/src/syntax/qtfile.rs [hunk 1]
```diff
@@ -41,7 +41,12 @@ impl ToTokens for CxxQtFile {
 }
 
 pub fn parse_qt_file(path: impl AsRef<std::path::Path>) -> Result<CxxQtFile> {
-    let source = std::fs::read_to_string(path.as_ref()).expect("Could not read path {} to string");
+    let source = match std::fs::read_to_string(path.as_ref()) {
+        Ok(source) => source,
+        Err(err) => {
+            panic!("Failed to read file {:?}: {}", path.as_ref(), err);
+        }
+    };
```
Be-ing, line 49: I think `match` is more verbose than needed here. I think we can keep the original `.expect` and use `&format!("Could not read path {} to string", path.as_ref().display())`

ahayzen-kdab: I do find it interesting that Rust doesn't have a function pointer route for expect, like how unwrap has `unwrap_or_else` vs `unwrap_or` so you can have lazy evaluation. Eg it would be nice if there was `expect_slow(|| my_slow_except_thing())`.

Unless, can we hack things with `unwrap_or_else(|| panic!(&format!(...)))` or is that crazy? :laughing: 

As for returning an error, we are currently using `syn::Result`, so it would need a span, and this error occurs before we have any tokens to span with so could be tricky :sweat_smile: Maybe later we'll have our own Error with a enum of IOError / Syn etc but we haven't got that far yet, so this can just panic for now.

lilymonad: `unwrap_or_else` can have weird behavior in the callstack, anyway the best choice here would be to implement an enum Error type, as you said, but you can still for now use [anyhow - Rust](https://docs.rs/anyhow/1.0.70/anyhow/) crate's `Result` type.

Implementing a custom error type seems like a good issue.

Be-ing: I'm not concerned about performance here so I think the simplest solution would be best: `.expect(&format!("Could not read path {} to string", path.as_ref().display()))`

Be-ing: > Unless, can we hack things with unwrap_or_else(|| panic!(&format!(...))) or is that crazy? laughing

Searching for any discussion of an `expect_or_else` function in https://github.com/rust-lang/rust, I found https://github.com/rust-lang/rust/commit/fe588d894fa392fdf787c18a959b342d38d0c71c. So I think that would be an okay solution.

ahayzen-kdab: > `unwrap_or_else` can have weird behavior in the callstack, anyway the best choice here would be to implement an enum Error type, as you said, but you can still for now use [anyhow - Rust](https://docs.rs/anyhow/1.0.70/anyhow/) crate's `Result` type.
> 
> Implementing a custom error type seems like a good issue.

Right, this would be more a medium/long term goal as a separate issue :+1:  so far `syn::Result` has been good enough for most errors as we want to refer to a span most of the time anyway.

For this pull request, either of the solutions (expect, match, unwrap_or_else) are likely fine as discussed, things can be refactored again later when/if we have our own error type.

ahayzen-kdab: Of course if someone is able to get a custom error type working nicely, then feel free to open another pull request :-)

OlivierLDff: So I forced pushed unwrap_or_else, and added a todo comment so we know it's not the best fix anyway, but at least it will give the user all the hint to debug why his build system doesn't work, which is the most important for now

PR #496, opened 2023-03-30
cxx-qt-gen: add a note about bug with msvc when compiling in debug

### LINKED ISSUE #495: doc: mention that `-MTd` is not support with MSVC at the moment
Hi,
Currently experimenting on windows, with msvc, it was quite annoying not to have the `Debug` build working out of the box. I investigate a bit and ended up there: https://github.com/rust-lang/cc-rs/pull/717.

So you know the problem, and it seems out of hand for the moment.

It would be really nice to have a propose workaround. What I'm doing is simply adding `-DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreadedDLL` when doing the configure cmake step. I've also seen people only setting `CXXFLAGS="-MTd"` as env variable. But this seems even more hacky that simply assuming we are stuck with `MT`, so there won't be any debugging support/assert stuff in stl. Which is not a big deal when doing qt app anyway.

So 2 ideas here:
- Mention it in the Readme
- Create a `CMakePresets.json` for msvc that just set the variable.

What is your opinion?

OlivierLDff:
This provide a workaround for users that want to build their code in Debug mode

fix #495

### DIFF
--- a/book/src/getting-started/4-cmake-integration.md
+++ b/book/src/getting-started/4-cmake-integration.md
@@ -116,6 +116,12 @@ $ build/examples/qml_minimal/example_qml_minimal
 
 You should now see the two Labels that display the state of our `MyObject`, as well as the two buttons to call our two Rust functions.
 
+### Windows with MSVC
+
+If you're building CXX-Qt on Windows using MSVC generator, you need to add the `-DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreadedDLL` flag to the cmake command when building with the `Debug` configuration. This flag is necessary to ensure that the correct C Runtime Library is used. Then you can build using `cmake --build build --config Debug`.
+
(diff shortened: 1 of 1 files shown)


book/src/getting-started/4-cmake-integration.md
```diff
@@ -116,6 +116,12 @@ $ build/examples/qml_minimal/example_qml_minimal
 
 You should now see the two Labels that display the state of our `MyObject`, as well as the two buttons to call our two Rust functions.
 
+### Windows with MSVC
+
+If you're building qt-cxx on Windows using MSVC generator, you need to add the `-DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreadedDLL` flag to the cmake command when building with the `Debug` configuration. This flag is necessary to ensure that the correct C Runtime Library is used. Then you can build using `cmake --build build --config Debug`.
```
ahayzen-kdab, line 121: cxx-qt not qt-cxx :-) And generally if it's not the crate name and being used as the name of the project I think we are trying to refer to it as CXX-Qt.

ahayzen-kdab:
Also make sure you rebase onto the latest `main` otherwise we'll have to rebase the branch via GitHub and go through another CI cycle :-)

Be-ing (approved):

PR #497, opened 2023-03-30
cxx-qt-build: Add more information about qmake, how it is used and why

### LINKED ISSUE #492: book: it is unclear how qt version is found on first read
Hi,
I'm trying qml-minimal example on my machine. My background is I'm very familiar with Qt & CMake, but not that much with rust.

From first read, it was unclear to me how the Qt version was found on rust side.
It would have been nice to mention it in https://github.com/KDAB/cxx-qt/blob/92feeae6bd29ad0ed852c8953fd247c3f0a274d4/book/src/getting-started/4-cmake-integration.md?plain=1#L35-L37 that `qmake -query` was used for configuration.

Same here: https://github.com/KDAB/cxx-qt/blob/92feeae6bd29ad0ed852c8953fd247c3f0a274d4/book/src/getting-started/4-cmake-integration.md?plain=1#L79

"Will be used later" is not clear, why?

When you go thru the example it make sense, but that wasn't that obvious in the beginning.

What do you think?

OlivierLDff:
fix #492

### DIFF
--- a/book/src/getting-started/4-cmake-integration.md
+++ b/book/src/getting-started/4-cmake-integration.md
@@ -49,6 +49,10 @@ We'll have to do multiple things:
 - Add `cxx`, `cxx-qt`, as well as `cxx-qt-lib` as dependencies
 - Add `cxx-qt-build` as a build dependency
 
+> Note that in order for cxx-qt to work, the `qmake` executable must be located. This is because cxx-qt relies on `qmake` to locate the necessary Qt libraries and header files on your system.
+>
+> This will be done by setting `QMAKE` environment variable from CMake, so that Qt version found by CMake will be correctly passed down to Cargo.
+
 In the end, your `Cargo.toml` should look similar to this.
 
 ```toml,ignore
@@ -76,7 +80,8 @@ Qt5 and Qt6 with CMake](https://doc.qt.io/qt-6/cmake-qt5-and-qt6-compatibility.h
 {{#include ../../../examples/qml_minimal/CMakeLists.txt:book_cmake_setup}}
 ```
 
-From the Qt CMake target, find the location of the qmake executable. cxx-qt-build will need this later.
+To ensure that cxx-qt-build uses the same version of Qt as your CMake targets, use the `Qt` CMake target to locate the qmake executable. Then, pass `qmake` executable path to `build.rs` with the environment variable `QMAKE` using `corrosion_set_env_vars`.
+
 ```cmake,ignore
 {{#include ../../../examples/qml_minimal/CMakeLists.txt:book_cmake_find_qmake}}
 ```
--- a/book/src/getting-started/5-cargo-executable.md
+++ b/book/src/getting-started/5-cargo-executable.md
@@ -17,6 +17,7 @@ The complete example code is available in [`examples/cargo_without_cmake`](https
 in the cxx-qt repository.
 
 ## Cargo setup
+
 The Cargo.toml file still requires dependencies to `cxx`, `cxx-qt`, `cxx-qt-lib` and `cxx-qt-build` as in our [CMake example](./4-cmake-integration.md). However, we are not building a `staticlib` this time:
 
 ```toml,ignore
@@ -58,6 +59,16 @@ application in C++:
 ```
 
 To build and run the application, use `cargo run`.
+
+> Note that in order for cxx-qt to work, the `qmake` executable must be located. This is because cxx-qt relies on `qmake` to locate the necessary Qt libraries and header files on your system.
+>
+> `cxx-qt` will find qmake in the following order:
+> - Look for an environment variable `QMAKE` that should have the path to `qmake`.
+> - Use `qmake` from the `PATH`. If multiple `qmake` exists in `PATH`, environment variable `QT_VERSION_MAJOR` will control the selected one.
+>
+> To check which version of Qt will be used with `qmake`, you can use the `qmake -query` command. This will display information about the Qt installation, including the version number and installation path.
+>
+> Check [CxxQtBuilder](https://docs.rs/cxx-qt-build/latest/cxx_qt_build/struct.CxxQtBuilder.html) for more information
 
 If you have cloned the CXX-Qt repository, you can run this example from within the repository using:
 


book/src/getting-started/4-cmake-integration.md
```diff
@@ -49,6 +49,10 @@ We'll have to do multiple things:
 - Add `cxx`, `cxx-qt`, as well as `cxx-qt-lib` as dependencies
 - Add `cxx-qt-build` as a build dependency
 
+Note that in order for cxx-qt to work, the `qmake` executable must be located. This is because cxx-qt relies on `qmake` to locate the necessary Qt libraries and header files on your system. This can be done by either adding it to your system's `PATH`, or by setting the `QMAKE` environment variable to its path.
+
+To check which version of Qt will be used with `qmake`, you can use the `qmake -query` command. This will display information about the Qt installation, including the version number and installation path.
```
Be-ing, line 54: This is helpful :+1:

book/src/getting-started/4-cmake-integration.md
```diff
@@ -67,6 +71,8 @@ This is what generates and compiles the C++ code for our `MyObject` class at bui
 Every Rust source file that uses the `#[cxx_qt::bridge]` macro need to be included in this script.
 In our case, this is only the `src/cxxqt_object.rs` file.
 
+> At this point, you can test your library build by running `QMAKE=/path/to/qmake cargo build` inside the `rust` folder. This should output a static library in `rust/target/debug/qml_minimal.lib` on Windows or `rust/target/debug/qml_minimal.a` on Unix system.
```
Be-ing, line 74: Do we really need to say this? I'd like to keep the documentation concise.

book/src/getting-started/4-cmake-integration.md
```diff
@@ -76,7 +82,8 @@ Qt5 and Qt6 with CMake](https://doc.qt.io/qt-6/cmake-qt5-and-qt6-compatibility.h
 {{#include ../../../examples/qml_minimal/CMakeLists.txt:book_cmake_setup}}
 ```
 
-From the Qt CMake target, find the location of the qmake executable. cxx-qt-build will need this later.
+To ensure that cxx-qt-build uses the same version of Qt as your CMake targets, you can use the `Qt` CMake target to locate the qmake executable. The `qmake` executable will be pass to `build.rs` as an environment variable using `corrosion_set_env_vars`.
```
Be-ing, line 85:
```suggestion
To ensure that cxx-qt-build uses the same version of Qt as your CMake targets, use the `Qt` CMake target to locate the qmake executable. Then, the pass `qmake` executable path to `build.rs` with an environment variable using `corrosion_set_env_vars`.
```

book/src/getting-started/4-cmake-integration.md
OlivierLDff: Maybe the second sentence is less useful, but I really think the first one really is.
When you think about it the whole setup is quite complex as it require understanding of cpp, rust, cargo, cmake, qml, qrc, qt. Being able to test at this without bringing further complication is very helpful. 
Having this note here is useful to debug the cargo part of the build without bringing cmake in, finding qt from cmake issues etc...
I aware than when has been done once it is trivial, but on first run, I was happy to just test that part. It worked out of the box my system because I had qmake in the path, I think I would have struggle a bit more or I wouldn't understand everything if I had just runner everything from cmake.

Be-ing: I think this suggestion could open a can of worms for users who may think they need to run Cargo separately from CMake. Users might also get confused if running `cargo build` picks up a different version of qmake than CMake finds. If someone wants to play around with Cargo without involving CMake at all, they can run `cargo build -p qml-minimal-no-cmake` as explained in the following chapter.

book/src/getting-started/4-cmake-integration.md
Be-ing: > To check which version of Qt will be used with `qmake`, you can use the `qmake -query` command. This will display information about the Qt installation, including the version number and installation path.

This isn't always correct. CMake can pick up a different version of Qt than the first `qmake` executable found in `PATH`, particularly if the user sets `CMAKE_PREFIX_PATH`.

OlivierLDff: But we are in the cargo/build.rs section, so everything is based on the qmake exe right?

Of course from the CMake, the `find_package` as so many option and could behave in some many way. But this is not what this section is about.

book/src/getting-started/4-cmake-integration.md
OlivierLDff: Ok, I haven't read this chapter yet, as I'm interesting in integrating cxx-qt in already existing CMake based application.
I will remove this line because I see your point.

I don't know if people will behave like, but I think they will go either to the building with cmake or cargo page. In the cargo page there is no mention at all of qmake.

Maybe I could duplicate the information about qmake to that other page? But I guess that would be another PR

book/src/getting-started/4-cmake-integration.md
ahayzen-kdab: With CMake, the location that CMake chooses for `QMAKE` is passed down into Rust, so if it picked one outside of your `PATH` this should be passed into the Rust env to IIRC

I think this is done via these places
https://github.com/KDAB/cxx-qt/blob/c435d65aba8eef451596def5f29c3b52dae8ea23/examples/qml_minimal/CMakeLists.txt#L29

 https://github.com/KDAB/cxx-qt/blob/c435d65aba8eef451596def5f29c3b52dae8ea23/examples/qml_minimal/CMakeLists.txt#L51-L60

OlivierLDff: So I think this documentation should move to the the cargo build system page.
On the cargo section in the cmake page, I shoudn't mention the PATH at all. Just QMAKE, and say that it will later be passed from whatever qt version CMake found as an env variable. Would it be ok?

OlivierLDff: I haven't look in cargo only build yet. When you say you don't have to ahve `QMAKE` set, you still need to have it in the `PATH` right?

ahayzen-kdab: I think that's how it's working there yes, but i'd have to look at the code again.

The `QMAKE` env var is for ensuring that CMake and Cargo when combined agree on the same `qmake` executable.

OlivierLDff: Ok so on rust side it is check for `QMAKE` env variable, then fallback to `qmake` in `PATH`, then error.
In the CMake documentation, only the `QMAKE` is interesting to the user.

Be-ing: Correct. The reason we must pass the QMAKE environment variable in CMake builds is to ensure that whatever Qt installation CMake picks up due to CMAKE_PREFIX_PATH or whatever other mechanism is the same Qt installation that build.rs uses.

OlivierLDff:
I've followed what you ask, in the CMake chapter, I only talk about how `qmake` is passed from CMake to cargo, and why.
In the Cargo chapter, I explain more in details how it is found. I even linked to the documentation, that goes even more in depth. I wish I would have seen this documentation sooner but I'm not yet very familiar with doc.rs and navigation in there.

Be-ing:
> I wish I would have seen this documentation sooner but I'm not yet very familiar with doc.rs and navigation in there.

We could use some links to docs.rs in the book.

OlivierLDff:
> We could use some links to docs.rs in the book.

It is what I added in the Cargo section. I used the tag latest, I don't know if there is a way to get the exact version?

OlivierLDff:
oups wrong branched pushed

book/src/getting-started/5-cargo-executable.md
```diff
@@ -59,6 +60,16 @@ application in C++:
 
 To build and run the application, use `cargo run`.
 
+> Note that in order for cxx-qt to work, the `qmake` executable must be located. This is because cxx-qt relies on `qmake` to locate the necessary Qt libraries and header files on your system.
+>
+> `cxx-qt` will find qmake in the following order:
+> - Look for an environment variable `QMAKE` that should a path to `qmake`.
```
Be-ing, line 66:
```suggestion
> - Look for an environment variable `QMAKE` that should have the path to `qmake`.
```

Be-ing (approved):

PR #498, opened 2023-03-30
cmake: Add a CMakePresets.json for msvc that handle Debug build with

OlivierLDff:
success

Related to #495 .

It's nice if user want to hack around this repo and the example to be able to build with Debug mode in windows.
Maybe the CMakePresets should be mentioned in Readme, but then a variant for most common generator should be added (like Ninja, Xcode I guess)

What is important here for windows is `"CMAKE_MSVC_RUNTIME_LIBRARY": "MultiThreadedDLL"`.

Then by settings for Debug build with environment:

```
QML_IMPORT_PATH=$(SolutionDir)/vcpkg_installed/x64-windows/debug/Qt6/qml
QT_PLUGIN_PATH=$(SolutionDir)/vcpkg_installed/x64-windows/debug/Qt6/plugins
PATH=$(SolutionDir)/vcpkg_installed/x64-windows/debug/bin
```
or for Release/RelWithDebInfo:
```
QML_IMPORT_PATH=$(SolutionDir)/vcpkg_installed/x64-windows/Qt6/qml
QT_PLUGIN_PATH=$(SolutionDir)/vcpkg_installed/x64-windows/Qt6/plugins
PATH=$(SolutionDir)/vcpkg_installed/x64-windows/bin
```

I'm able to correctly start the app from visual studio. (Not familiar enough with `launch.json` of vscode, but I guess the same can be done).

Screenshot for reference:
image

image

And for reference mix debugging work out of the box:

debug-msvc


CMakePresets.json
```diff
@@ -0,0 +1,60 @@
...
+        {
+            "name": "base-dev",
+            "binaryDir": "${sourceDir}/build/${presetName}",
+            "hidden": true
+        },
+        {
+            "name": "msvc-17",
+            "inherits": "base-dev",
+            "displayName": "Visual Studio 2022",
+            "generator": "Visual Studio 17 2022",
```
Be-ing, line 18: Is anything here specific to this version of Visual Studio besides the generator name? Any reason not to use the Ninja generator?

CMakePresets.json.license
```diff
@@ -0,0 +1,4 @@
+SPDX-FileCopyrightText: 2021, 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
```
ahayzen-kdab, line 1: I find the year here curious, should it not just be 2023?

ahayzen-kdab: Also squash the second commit into the first so each commit passes reuse.

CMakePresets.json
```diff
@@ -0,0 +1,60 @@
...
+            "description": "Configure using Visual Studio 17 2022 generator",
+            "condition": {
+                "type": "equals",
+                "lhs": "${hostSystemName}",
+                "rhs": "Windows"
+            },
+            "cacheVariables": {
+                "CMAKE_MSVC_RUNTIME_LIBRARY": "MultiThreadedDLL"
```
ahayzen-kdab, line 26: And again, assume we can't submit this upstream to Corrosion until rustc / cc have fixes ? Eg is there a reason why Corrosion can't set `CMAKE_MSVC_RUNTIME_LIBRARY = MultiThreadedDLL` for all projects on Windows ? Or does this break other projects ? :thinking: 

As for us that would then help projects that are using CXX-Qt as a library as they don't need to remember to set this in their CMake as well.

CMakePresets.json
OlivierLDff: For ninja I can add this:
```json
        {
            "name": "ninja-debug",
            "inherits": "base-dev-ninja",
            "displayName": "Debug Ninja",
            "description": "Build in Debug with Ninja generator",
            "cacheVariables": {
                "CMAKE_BUILD_TYPE": "Debug"
            }
        },
        {
            "name": "ninja-relwithdebinfo",
            "inherits": "base-dev-ninja",
            "displayName": "RelWithDebInfo Ninja",
            "description": "Build in RelWithDebInfo with Ninja generator",
            "cacheVariables": {
                "CMAKE_BUILD_TYPE": "RelWithDebInfo"
            }
        },
        {
            "name": "ninja-release",
            "inherits": "base-dev-ninja",
            "displayName": "Release Ninja",
            "description": "Build in Release with Ninja generator",
            "generator": "Ninja",
            "cacheVariables": {
                "CMAKE_BUILD_TYPE": "Release"
            }
        },
        {
            "name": "ninja-multi",
            "inherits": "base-dev",
            "displayName": "Ninja Multi-Config",
            "description": "Default build using Ninja Multi-Config generator",
            "generator": "Ninja Multi-Config"
        },
```
but the goal of this pr was just to add a preset for a visual studio solution

CMakePresets.json
OlivierLDff: From the [documentation](https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html#:~:text=the%20MSVC_RUNTIME_LIBRARY%20property%20on%20all%20targets%20as%20they%20are%20created) changing CMAKE_MSVC_RUNTIME_LIBRARY  from corrosion after creating any target would not work for target created before changing CMAKE_MSVC_RUNTIME_LIBRARY .
I haven't tried that it's just my understanding from the doc.
Once again, writing to a global variable in not a good idea

CMakePresets.json.license
OlivierLDff: I just copied the file from the repo somewhere else.
Is the year even necessary? I saw some oss project that stopped writing the year.

CMakePresets.json
ahayzen-kdab: Note that before we did try adding all the combinations of CMakePresets but it kinda exploded into many possibilities #325, with varying options of Qt 5 vs 6, Ninja, vcpkg etc

CMakePresets.json.license
ahayzen-kdab: Right, https://www.linuxfoundation.org/blog/blog/copyright-notices-in-open-source-software-projects and plenty of other large tech vendors have started dropping the year.

For now in this project we have had at least the year the file was introduced, lets be consistent for now and keep adding the year the file was made. Maybe later we can decide to purge the year from everywhere in it's own change :-)

ahayzen-kdab: (and more articles linked here https://hynek.me/til/copyright-years/ )

OlivierLDff: Oh yes that's the one I've stump on across

CMakePresets.json
OlivierLDff: I see that now.
I think it's always nice to have a default preset that "just work" for testing purpose and easy onboarding. Then people can hack their own way.
So having a default with Qt6 let's say is ok.

So from your PR I would suggest creating a base config with:
"CMAKE_CXX_COMPILER_LAUNCHER": "sccache"
"RUSTC_WRAPPER": "sccache"
"USE_QT5": false
"VCPKG": true

I feel it is always better to have documentation like that, than outdated Readme information.
What you be even better is for the presets to be exactly what is being used in CI. So one preset for MSVC, one for linux with ninja, one for mac (with ninja too?), so the same as linux maybe.

OlivierLDff: That's why going the CI way is a good option where there is no debate. If people want something else, either they create custom presets, or use cmake manually. CMakePresets should just provide a starting point, and show all the variable people usually tweak?

OlivierLDff:
I've pushed a second proposal based on @ahayzen-kdab CMakeUserPresets.json.
I think that changing the default value of `VCPKG` based on user OS is bad and highlight dev preference.
I think VCPKG should be defaulted to ON by default, so that default configuration  "just build". Same it should be set to ON in CMakePresets.json, since presets are expected to "just work" no matter how the host system is configured.

Be-ing:
Ehh, this is turning into lots of different presets. which brings up the [same question as before](https://github.com/KDAB/cxx-qt/pull/325#issuecomment-1285996410). Could we hack around the Windows debug build issue another way by adding this to CMakeLists.txt?

```
if(WIN32 AND (CMAKE_BUILD_TYPE STREQUAL "Debug" OR CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo"))
  set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreadedDLL")
endif()
```

OlivierLDff:
It's only 5 presets. It could be trim to 2 ninja multi config and msvc. The fact I included all configuration to inherit is so it is simple to use from cmakeuserpresets.json

OlivierLDff:
Ok so I took a drastic modification in the Preset, there is only 2 left.
- MSVC on windows.
- Ninja Multi otherwise.
I think it is the 2 sane default that should work anywhere.

Leaving Ninja on windows, if clang++ is available it will use that instead of msvc, leading to conflict with Qt version used from vcpkg.

Be-ing:
I am reluctant to recommend use of the MSBuild generators because MSBuild doesn't support ccache/sccache with [CMAKE_LANG_COMPILER_LAUNCHER](https://cmake.org/cmake/help/latest/envvar/CMAKE_LANG_COMPILER_LAUNCHER.html).

OlivierLDff:
Does it even matter for this project to use ccache/sccache ? Most of the code compiled is rust, so it doesn't really matter what build system is used or c++ compiler?

OlivierLDff:
In the end the choice is yours, if you are against CMakePresets, then writing `CMAKE_MSVC_RUNTIME_LIBRARY` is the best workaround. Be aware that `CMAKE_BUILD_TYPE` doesn't really matter for multi config generator and you should avoid it. Better use generator expression.

```cmake
if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
  # rust cc::Build always use MultiThreadedDLL, so for c++ code handled from CMake to correctly link with generated code by Cargo, this variable is globally set on all CMake target.
  # This is a workaround, and is not optimal.
  set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreadedDLL")
endif()
```

Be-ing:
I think adding that little bit to make debug builds work on Windows regardless of presets would be helpful. I personally don't really understand the value of adding this CMakePresets.json, but if it helps some people then sure. So I'll neither merge nor block this and leave the decision to others.

ahayzen-kdab:
> I think adding that little bit to make debug builds work on Windows regardless of presets would be helpful. I personally don't really understand the value of adding this CMakePresets.json, but if it helps some people then sure. So I'll neither merge nor block this and leave the decision to others.

I think for now lets put this in the CMakeLists so that it solves the issue for both users of CMakePresets and users who use CMake manually. Then presets can be a separate investigation / task whether we want to include one.

Then i guess the CMakeLists workaround should be documented / visible in the code snippets somewhere so that people who are using CXX-Qt as a library know to put this in their CMakeLists too.

OlivierLDff:
Another alternative that I saw in corrosion: https://github.com/corrosion-rs/corrosion/blob/e516737bf4987d000353045133c2f82b23553490/test/cxxbridge/cxxbridge_rust2cpp/CMakeLists.txt#L13-L16

PR #243, opened 2022-09-12
make use of cargo workspace for tests

Be-ing:
There is no need to run cargo test/clippy/fmt for each crate individually in the workspace. Reusing the workspace target directory that Corrosion uses drastically increases the speed of the tests by reusing compilation artifacts from the main build.

Depends on #229 to avoid merge conflicts

PR #500, opened 2023-03-30
tests: make tests run in CI on windows

### LINKED ISSUE #111: Investigate why lots of tests are failing on Windows
Currently these tests are failing on Windows `cxx_qt_gen_cargo_tests|example_qml.*|qml.*tests|test.*|cpp_clang_format` some of these are related for formatted and some are not building correctly. We need to setup a build env on Windows to check what is happening as no one has tried it yet :-)

OlivierLDff:
They were failing due to missing dll. This commit
- inject the location of Qt dll in the PATH
- Qt plugin directory in QT_PLUGIN_PATH
- Qml import path in QML_IMPORT_PATH & QML2_IMPORT_PATH

They are not necessary to every tests, but at least they should be robust for the future when more tests will be added.

This commit use `ENVIRONMENT_MODIFICATION` instead of `ENVIRONMENT`, since adding a value to the PATH in windows is a mess otherwise with CMake, because `;` is used as a list separator.

It also move `add_subdirectory` of book and example at the end of main CMakeLists.txt file, in order not to duplicate code to set `RUNTIME_ENV`.

Next step is to refactor all the `add_test` function in some helper function `add_cxxqt_test` or something like that.

fix #111

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -138,9 +138,6 @@ function(add_valgrind_test NAME_WITH_PREFIX BINARY WORKING_DIRECTORY)
     )
 endfunction()
 
-add_subdirectory(book)
-add_subdirectory(examples)
-
 # QMAKE environment variable is needed by qt-build-utils to ensure that Cargo
 # uses the same installation of Qt as CMake does.
 if(NOT USE_QT5)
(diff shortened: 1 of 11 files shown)


Be-ing:
If you can get some of the tests to run but not all, feel free to add some back to the exclusion list. Any step forward would be helpful!

OlivierLDff:
Yes the goal of this pr is just to make test passed that are related to missing all. I removed all the filter for now to see what is working and what is not.

examples/qml_minimal/CMakeLists.txt
```diff
@@ -122,6 +122,11 @@ function(add_qml_test TEST_NAME)
 
     add_test(NAME ${APP_NAME}_${TEST_NAME}_test COMMAND ${CMAKE_CURRENT_BINARY_DIR}/${TEST_NAME}/${APP_NAME}_${TEST_NAME}_test${EXE_SUFFIX} WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${TEST_NAME})
 
+    set_tests_properties(
```
Be-ing, line 125:
```suggestion
    # RUNTIME_ENV comes from the CMakeLists.txt at the root of this repository.
    set_tests_properties(
```

examples/qml_extension_plugin/tests/CMakeLists.txt
```diff
@@ -19,12 +19,26 @@ target_link_libraries(${APP_NAME} Qt::QuickTest)
...
+    "$<TARGET_FILE_DIR:core_qmlplugin>"
+)
+
+add_test(
+    NAME ${APP_NAME}
     COMMAND ${test_cmd}
     WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}"
 )
 
+set_tests_properties(
```
Be-ing, line 38:
```suggestion
# RUNTIME_ENV comes from the CMakeLists.txt at the root of this repository.
set_tests_properties(
```

We've been writing the CMakeLists.txt within the examples so they could stand on their own outside of this repo, but it would be overkill to setup `RUNTIME_ENV` repeatedly just for the tests. So let's leave a comment for people who may be reading this as an example of how to setup their own repository using CXX-Qt.

Be-ing (commented):
This is awesome, thanks so much. Just a few minor comments

Be-ing:
Maybe we can revive #243 after merging this :thinking:

examples/qml_extension_plugin/tests/CMakeLists.txt
OlivierLDff: My idea was to later propose to have the cmakelists as non standalone. And create a standalone folder in examples folder with really standalone, not including test and stuff.

CMakeLists.txt
```diff
@@ -150,25 +147,86 @@ if(NOT Qt6_FOUND)
...
     # Add cargo as a test
-    add_test(NAME ${TEST_NAME_PREFIX}_cargo_tests COMMAND cargo test --all-features --all-targets --manifest-path ${MANIFEST_PATH})
-    set_property(TEST ${TEST_NAME_PREFIX}_cargo_tests PROPERTY ENVIRONMENT ${CARGO_ENV})
+    set(CARGO_TEST_NAME ${TEST_NAME_PREFIX}_cargo_tests)
+    add_test(NAME ${CARGO_TEST_NAME}
+             COMMAND cargo test --all-features --all-targets --manifest-path
+                     ${MANIFEST_PATH}
+    )
+    set_tests_properties(
+        ${CARGO_TEST_NAME} PROPERTIES ENVIRONMENT_MODIFICATION "${CARGO_ENV}"
```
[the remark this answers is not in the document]
OlivierLDff: Because on windows PATH is separated by `;` not `:`.
Cmake strings are separated by `;`. So having a list of environment variable containing `;` is not possible

OlivierLDff:
I've force push a slight modification that also make valgrind test work with ninja multi config. So that every test can pass when using the CMakePresets.json from my other PR.

examples/qml_features/CMakeLists.txt
```diff
@@ -64,46 +64,27 @@ qt_import_qml_plugins(${APP_NAME})
...
+)
 
-add_valgrind_test(${APP_NAME}_test ${CMAKE_CURRENT_BINARY_DIR}/tests/${APP_NAME}_test${EXE_SUFFIX} ${CMAKE_CURRENT_BINARY_DIR}/tests)
+add_valgrind_test(
+    ${APP_TEST_NAME} "${TEST_CMD}" ${CMAKE_CURRENT_BINARY_DIR}
+)
 
-set_target_properties(${APP_NAME}_test PROPERTIES
-    RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/tests
+set_tests_properties(
```
Be-ing, line 88:
```suggestion
# RUNTIME_ENV comes from the CMakeLists.txt at the root of this repository.
set_tests_properties(
```

OlivierLDff: This should be ok

Be-ing (approved):

PR #502, opened 2023-03-31
cxx-qt-gen: move to syn 2

ahayzen-kdab:

PR #503, opened 2023-03-31
cxx-qt-gen: remove tokens_to_syn and use parse_quote instead

ahayzen-kdab:
Requires #502


LeonMatthesKDAB (approved):
Well I guess that's been the easiest +500 line change I've reviewed yet xD

No need to argue with less boilerplate :+1:

PR #511, opened 2023-04-03
ctest: ensure that cargo fmt is run across the workspace

ahayzen-kdab:
This appears to have been missed in
https://github.com/KDAB/cxx-qt/commit/048c7e9ab3bde427ed321044b4abed9e37f404a5


Be-ing (approved):

OlivierLDff:
Oh yes you are right,I removed it from the set_tests_properties  since cargo felt doesn't need the runtime env. I don't why I didn't copy the add_test

ahayzen-kdab:
> Oh yes you are right,I removed it from the set_tests_properties since cargo felt doesn't need the runtime env. I don't why I didn't copy the add_test

Ah yes it doesn't actually need the CARGO_ENV, there was some fun where the `--manifest-path` doesn't work with the workspace `Cargo.toml` for some reason, but using the working directory does :shrug:

PR #512, opened 2023-04-05
ci: only save compiler cache when the branch is main

ahayzen-kdab:
This should prevent the cache increasing in size due to a cache per branch, which can cause the main cache to be removed.


Be-ing (approved):
Let's give this a try.

PR #513, opened 2023-04-05
add QQuickPaintedItem example using Constructor trait

### LINKED ISSUE #474: Need to be able to choose the parent when deriving from different bases
Currently we assume that it'll be `QObject* parent = nullptr`, but for some Qt types (eg QQuickPaintedItem or QQuickItem) the parent is different.

We need to have a way in the `cxx_qt::qobject` macro to be able to specify a different parent other than `QObject`.

There seems to be these options

- If we can assume all parent types will be a raw pointer then just have `#[cxx_qt::qobject(base = "QQuickPaintedItem", parent = "QQuickItem"]`
- If we can't assume this then we might need something like `#[cxx_qt::qobject(base = "QQuickPaintedItem", parent_ptr = "QQuickItem"]` or `#[cxx_qt::qobject(base = "QQuickPaintedItem", parent = "QQuickItem*", parent_default = "nullptr"]`
- Have a separate macro for items that aren't directly derived from `QObject`, eg have `#[cxx_qt::qquickitem(base = "QQuickPaintedItem")]` this then means that we can generate the different asserts and parent etc and it's clear to the developer which base the type is deriving from

ahayzen-kdab:
This then allows you to create a QQuickItem such as a QQuickPaintedItem.

Closes #474

### DIFF
--- a/examples/qml_features/qml/pages/CustomParentClassPage.qml
+++ b/examples/qml_features/qml/pages/CustomParentClassPage.qml
@@ -1,0 +1,61 @@
+// SPDX-FileCopyrightText: 2023 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+import QtQuick 2.12
+import QtQuick.Controls 2.12
+import QtQuick.Layouts 1.12
+
+import com.kdab.cxx_qt.demo 1.0
+
+Page {
+    header: ToolBar {
+        RowLayout {
+            anchors.fill: parent
+
+            ToolButton {
+                text: qsTr("Red")
+
+                onClicked: customPainter.color = "red"
+            }
+
+            ToolButton {
+                text: qsTr("Green")
+
+                onClicked: customPainter.color = "green"
+            }
+
+            ToolButton {
+                text: qsTr("Blue")
+
+                onClicked: customPainter.color = "blue"
+            }
+
+            Item {
+                Layout.fillWidth: true
+            }
+        }
+    }
+
+
+    ColumnLayout {
+        anchors.left: parent.left
+        anchors.right: parent.right
+        anchors.verticalCenter: parent.verticalCenter
+
+        CustomParentClass {
+            id: customPainter
+            color: "red"
+            Layout.alignment: Qt.AlignHCenter
+            height: 200
+            width: 200
+        }
+
+        Label {
+            Layout.fillWidth: true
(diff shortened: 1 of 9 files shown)


examples/qml_features/qml/pages/CustomParentClassPage.qml
```diff
@@ -0,0 +1,64 @@
...
+        anchors.verticalCenter: parent.verticalCenter
+
+        CustomParentClass {
+            id: customPainter
+            color: "red"
+            Layout.alignment: Qt.AlignHCenter
+            height: 200
+            width: 200
+
+            // TODO: once we can connect to signals on the Rust side, this could be done there
```
ahayzen-kdab, line 53: This is #485 we will still need a initialise call to do that connection but that's another task to have initialise in Rust somewhere...

examples/qml_features/rust/src/custom_parent_class.rs
```diff
@@ -0,0 +1,57 @@
...
+        type QPainter;
+        include!(<QtGui/QPainter>);
+
+        #[rust_name = "fill_rect"]
+        fn fillRect(self: Pin<&mut QPainter>, rectangle: &QRectF, color: &QColor);
+    }
+
+    #[cxx_qt::qobject(
+        base = "QQuickPaintedItem",
+        parent = "QQuickItem",
```
ahayzen-kdab, line 29: need to decide on API here

ahayzen-kdab, line 29: need to document this in the book if we did go this route etc

ahayzen-kdab:
#546 may help (bumping vcpkg version) with the macOS errors here as @Be-ing found a similar looking error in the Qt bugs https://bugreports.qt.io/browse/QTBUG-95665?focusedCommentId=578483

Be-ing:
Linking Qt6Quick statically requires linking `.cpp.o` files. Unfortunately, rustc and cargo do not have a stable interface to pass those to the linker. There is a `-l link-arg` option but that is [unstable](https://github.com/rust-lang/rust/issues/99427). I tested building Qt6 statically on Linux and adding

```rust
println!("cargo:rustc-link-lib=link-arg=/home/be/qt6-installed/lib/objects-Release/Quick_resources_2/.rcc/qrc_scenegraph_shaders.cpp.o");
```

to build.rs for the qml-features crate got `cargo test` to link.

ahayzen-kdab:
This is now blocked on the fixes in #600 to get static macOS Qt with QtQuick working...

Be-ing (approved):

Be-ing:
Woohoo! @CarlSchwan we finally have the pieces in place to implement custom QQuickItems in Rust :)

Issue #523, opened 2023-04-28
Incremental builds are failing causing slow builds


ahayzen-kdab:
This is due to https://github.com/corrosion-rs/corrosion/issues/386

When building the project with CMake it causes all your crates to be rebuilt as well, moving back to 0.3.0 is a workaround for now until the next release.

ahayzen-kdab:
Moving to 0.4.1 of corrosion has solved this https://github.com/KDAB/cxx-qt/commit/7a7d5429f300e1e5701760abc368140f9e9cab72 .

PR #524, opened 2023-05-01
cmake: use older corrosion version to fix incremental builds

ahayzen-kdab:
This is a workaround until a bug is fixed upstream which breaks incremental builds in projects with multiple targets, causing very slow builds.
https://github.com/corrosion-rs/corrosion/issues/386

Related to #523


LeonMatthesKDAB (approved):
LGTM

PR #525, opened 2023-05-01
github: push the cache from all branches

ahayzen-kdab:
GitHub expires caches if they aren't accessed after 7 days, this then means once the cache has expired a commit needs to reach main to generate a new cache entry.

Instead by pushing back the cache after each build we only have one slow commit, the downside is that the cache contains data not relevant to the current branch. But sccache should cycle less relevant data as the cache size is hit.

### DIFF
--- a/.github/workflows/github-cxx-qt-tests.yml
+++ b/.github/workflows/github-cxx-qt-tests.yml
@@ -146,11 +146,12 @@ jobs:
       run: cargo install sccache mdbook mdbook-linkcheck
 
     - name: "Compiler cache"
-      # Only restore here, later we save for only the main branch
-      # and always use the same key as we start from the main branch cache
-      uses: actions/cache/restore@v3
+      uses: actions/cache@v3
       with:
+        # Cache the sccache data
         path: ${{ matrix.compiler_cache_path }}
+        # Share the same cache key for branches
+        # this ensures that we don't create separate cache entries per branch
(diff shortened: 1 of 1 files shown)


LeonMatthesKDAB (commented):
Looks good to me, but I think @Be-ing should also approve.

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -146,11 +146,12 @@ jobs:
       run: cargo install sccache mdbook mdbook-linkcheck
 
     - name: "Compiler cache"
-      # Only restore here, later we save for only the main branch
-      # and always use the same key as we start from the main branch cache
-      uses: actions/cache/restore@v3
+      uses: actions/cache@v3
```
ahayzen-kdab, line 149: i don't think this worked fully, as it didn't update the cache "not saving cache" ?

```
Post job cleanup.
Cache hit occurred on the primary key Ubuntu [2](https://github.com/KDAB/cxx-qt/actions/runs/4850575112/jobs/8643604078#step:32:2)2.04 (gcc) Qt5_compiler_cache, not saving cache.
```

ahayzen-kdab:
As this doesn't work, instead going to revert the previous changes and go back to having multiple caches but of a smaller size.

Then the corrosion fixes should help us build faster again #526

PR #526, opened 2023-05-01
Revert cache change

ahayzen-kdab:


LeonMatthesKDAB (approved):
This makes sense, because when only updating the cache on main, it "times out" after 2 weeks of nothing being committed to main, which is especially annoying as then all PRs will be really slow, making it harder to merge something.

So instead of only updating the cache on main, use a reduced cache size, so we can have more, smaller caches that should still be fast enough.

PR #534, opened 2023-05-17
cxx-qt-lib: ensure that QVariant{Hash,List,Map} are registered for QML

### LINKED ISSUE #533: QVariantMap
The cxx-qt-lib appears to not have a QVariantMap. However, it does have a QMap-QString-QVariant. Which I can use, but since it is not recognized as a QVariantMap in QML, it must first pass through C++ before QML can pick it up. I've worked around this with a C++ bridge that passes the params to rust functions and the returned QMap-QString-QVariant back to QML as a QVariantMap and QT seems to do the conversion automatically since it's just an alias for the same thing.

Any idea why Rust can't send those back to QML without first going through C++?

ahayzen-kdab:
Closes #533


crates/cxx-qt-lib/src/std_types.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+#include <cstdint>
+
+namespace rust {
+namespace cxxqtlib1 {
+
+// Ensure that std int types are registered
+// so that they can be used with QML in Qt 5
+
+static const int register_i8 =
```
ahayzen-kdab, line 22: this causes the extension plugin tests to fail.

Is this the same problem as the `+whole-archive` option where things are compiled out as they aren't used ?

Eg this `std_types.cpp` doesn't get executed ...

ahayzen-kdab: Moving the extension plugin example to use WHOLE_ARCHIVE in the CMake like other examples fixes this...

LeonMatthesKDAB (commented):
We will also want to update the book section [Using a Custom Type with Containers or QVariant](https://kdab.github.io/cxx-qt/book/concepts/types.html#using-a-custom-type-with-containers-or-qvariant) to explain that you will have to call qmlRegisterType for your own types as well.

ahayzen-kdab:
> We will also want to update the book section [Using a Custom Type with Containers or QVariant](https://kdab.github.io/cxx-qt/book/concepts/types.html#using-a-custom-type-with-containers-or-qvariant) to explain that you will have to call qmlRegisterType for your own types as well.

Done

LeonMatthesKDAB (approved):
LGTM

PR #535, opened 2023-05-18
cxx-qt-lib-headers: ensure that rerun-if-changed is set for headers

ahayzen-kdab:


ahayzen-kdab:
Was this intentionally missed or is this a case when we do need it ?

LeonMatthesKDAB (approved):
Looks like we just missed this.

PR #541, opened 2023-05-18
Add ability to auto-update test_outputs/ directory

LeonMatthesKDAB:
Use CXXQT_UPDATE_EXPECTED=/path/to/crates/cxx-qt-gen/ to update the test_outputs/ directory.
This is very useful for bulk updates when making changes to small cross-cutting concerns.

### DIFF
--- a/crates/cxx-qt-gen/src/lib.rs
+++ b/crates/cxx-qt-gen/src/lib.rs
@@ -22,13 +22,18 @@ pub use syn::{Error, Result};
 mod tests {
     use super::*;
 
-    use clang_format::{clang_format, ClangFormatStyle, CLANG_FORMAT_STYLE};
+    use clang_format::{ClangFormatStyle, CLANG_FORMAT_STYLE};
     use generator::{cpp::GeneratedCppBlocks, rust::GeneratedRustBlocks};
     use parser::Parser;
     use pretty_assertions::assert_str_eq;
     use proc_macro2::TokenStream;
     use quote::ToTokens;
-    use std::io::Write;
+    use std::{
+        env,
+        fs::OpenOptions,
+        io::Write,
+        path::{Path, PathBuf},
+    };
     use writer::{cpp::write_cpp, rust::write_rust};
 
     #[ctor::ctor]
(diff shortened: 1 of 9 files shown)


crates/cxx-qt-gen/src/lib.rs
```diff
@@ -73,45 +78,84 @@ mod tests {
...
+    }
+
+    fn update_expected(test_name: &str, rust: &str, header: &str, source: &str) -> bool {
+        // Ideally we'd be able to get the path from `file!()`, but that unfortunately only
+        // gives us a relative path, which isn't really all that useful.
+        // So require the path of the crate to be set via an environment variable.
+        //
+        // In the simplest case this can be achieved by running:
+        //
+        //      CXXQT_UPDATE_EXPECTED=$(pwd) cargo test
```
ahayzen-kdab, line 99: Probably is the simplest way of doing it as this is a developer tool.

Assume that making a bin target or having a separate crate that we don't publish which updates the files would be way over the top and could leak into crates.io as a installable thing via cargo. And would add more complexity.

LeonMatthesKDAB: Yeah, I'm fine with the Environment variable approch per se.
Easy to implement & maintain and will continue to work as we add new tests.
I was hoping to not have to provide the path to it though.

I.e. instead of having to go:
```
CXXQT_UPDATE_EXPECTED=/path/to/crates/cxx-qt-gen
```
I would have liked:
```
CXXQT_UPDATE_EXPECTED=1
```

We could just add a small shell script that does the updating though.
``` bash
#!/bin/bash

set -ex

SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
cd ${SCRIPT_DIR}
CXXQT_UPDATE_EXPECTED="${SCRIPT_DIR}" cargo test tests::generates
```

crates/cxx-qt-gen/update_expected.sh
```diff
@@ -0,0 +1,14 @@
...
+# SPDX-FileCopyrightText: 2023 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+# SPDX-FileContributor: Leon Matthes <leon.matthes@kdab.com>
+#
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+
+set -ex
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+cd ${SCRIPT_DIR}
```
ahayzen-kdab, line 12: shellcheck warns for the following (basically add quotes)

```
In crates/cxx-qt-gen/update_expected.sh line 12:
cd ${SCRIPT_DIR}
   ^-----------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
cd "${SCRIPT_DIR}"

For more information:
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
```

crates/cxx-qt-gen/update_expected.sh
```diff
@@ -0,0 +1,14 @@
...
+
+# SPDX-FileCopyrightText: 2023 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+# SPDX-FileContributor: Leon Matthes <leon.matthes@kdab.com>
+#
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+
+set -ex
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```
ahayzen-kdab, line 11: interesting way of doing it, i usually do this

```bash
SCRIPT=$(realpath "$0")
```

crates/cxx-qt-gen/src/lib.rs
```diff
@@ -73,45 +78,84 @@ mod tests {
         code
     }
 
+    fn update_expected_file(path: PathBuf, source: &str) {
+        println!("Updating expected file: {:?}", path);
```
ahayzen-kdab, line 82: do we ever see this trace line ?

crates/cxx-qt-gen/update_expected.sh
LeonMatthesKDAB: Found this on Stackoverflow :man_shrugging: 
Doesn't really matter I guess...

crates/cxx-qt-gen/src/lib.rs
LeonMatthesKDAB: Yes, if the "test" fails, i.e. file access fails because the provided path points to the wrong directory, etc.
Then some of the unwrapping later will make the test fail and output will be printed.

crates/cxx-qt-gen/update_expected.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/lib.rs
LeonMatthesKDAB: Done

crates/cxx-qt-gen/update_expected.sh
ahayzen-kdab: Just seems super complicated in comparison and we've already used the realpath + dirname route elsewhere :thinking:

ahayzen-kdab: Or `SCRIPT_DIR=$(dirname "$(realpath "$0")")` also works for a one liner, but you get the double quote fun

LeonMatthesKDAB: Done

ahayzen-kdab (approved):
LGTM :-)

PR #544, opened 2023-05-22
update_expected: ensure cwd is the cxx-qt-gen folder

ahayzen-kdab:
Otherwise all cargo projects in the cwd are built when this isn't required to update the expected.


LeonMatthesKDAB (approved):
Good point.

Issue #550, opened 2023-05-26
Support implementing a custom constructor from Rust


LeonMatthesKDAB:
See the answer in [#339 ](https://github.com/KDAB/cxx-qt/discussions/339#discussioncomment-6002853) for intended API.

This would solve:
* Providing non-default constructor
* Relaxing requirement for `Default` on a qobject type.
* Allow calling non-default base class constructors in C++ (e.g. when deriving from QQuickItem)

Depends on #562

LeonMatthesKDAB:
Accidentally closed this in #566, even though that didn't fully test & document the feature yet

PR #551, opened 2023-05-26
cxx-qt: Add CxxQtType trait

LeonMatthesKDAB:
This trait is automatically implemented on any `qobject::T`. It currently exposes the inner Rust type for a given QObject.

First step of implementing #550

This also requires us to split our macros into a cxx-qt-macro crate, as otherwise only exporting proc-macros is allowed.
We could now potentially move CxxQtThread into the cxx-qt crate now to remove the dependency on cxx-qt-lib.

### DIFF
--- a/crates/cxx-qt-gen/src/writer/rust/mod.rs
+++ b/crates/cxx-qt-gen/src/writer/rust/mod.rs
@@ -31,9 +31,9 @@ fn cxx_bridge_common_blocks(qobject: &GeneratedRustQObject) -> Vec<TokenStream> 
     vec![
         quote! {
             unsafe extern "C++" {
-                /// Retrieve an immutable reference to the Rust struct backing this C++ object
                 #[cxx_name = "unsafeRust"]
-                fn rust(self: &#cpp_struct_ident) -> &#rust_struct_ident;
+                #[doc(hidden)]
+                fn cxx_qt_ffi_rust(self: &#cpp_struct_ident) -> &#rust_struct_ident;
 
                 #[doc = "Generated CXX-Qt method which creates a new"]
                 #[doc = #cpp_struct_ident_str]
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -33,6 +33,7 @@ version = "0.5.3"
 # we publish, otherwise crates.io complains as it doesn't know the version.
 [workspace.dependencies]
 cxx-qt = { path = "crates/cxx-qt" }
+cxx-qt-macro = { path = "crates/cxx-qt-macro" }
```
ahayzen-kdab, line 36: can you split this split into a separate pull request. As the trait stuff we might want to see things develop a little bit further to check things work before merging. And it eases review if things are small

LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/writer/rust/mod.rs
```diff
@@ -41,9 +41,9 @@ fn cxx_bridge_common_blocks(qobject: &GeneratedRustQObject) -> Vec<TokenStream>
                 /// <https://github.com/dtolnay/cxx/issues/683>
                 type #cxx_qt_thread_ident;
 
-                /// Retrieve an immutable reference to the Rust struct backing this C++ object
                 #[cxx_name = "unsafeRust"]
-                fn rust(self: &#cpp_struct_ident) -> &#rust_struct_ident;
+                #[doc(hidden)]
+                fn ffi_rust(self: &#cpp_struct_ident) -> &#rust_struct_ident;
```
ahayzen-kdab, line 46: do we use `ffi_` as our common "hidden" method naming ? As we have the same problem with `qt_thread()` ? or should we use like `cxx_qt_` or `cxx_qt_ffi_` ?

Namespacing to our things to at least `cxx_qt` would avoid any collisions accidentally

LeonMatthesKDAB: We could do `ccx_qt_ffi` :thinking: 
Let me check if that also fixes our "similarity" issues.

LeonMatthesKDAB: image
At least for `rust()` and `rust_mut()`, you no longer get suggested the `ffi_rust` stuff.

@ahayzen-kdab Can you check that `cxx_qt_ffi` also works for `qt_thread`?
As that may have more overlap with the longer method name.

ahayzen-kdab: Yup it makes it different enough too

```
build] error[E0599]: no method named `qt_thread` found for struct `Pin<&mut threading::ffi::ThreadingWebsiteQt>` in the current scope
[build]   --> examples/qml_features/rust/src/threading.rs:90:34
[build]    |
[build] 90 |             let qt_thread = self.qt_thread();
[build]    |                                  ^^^^^^^^^ method not found in `Pin<&mut ThreadingWebsiteQt>`
[build]    |
[build]    = help: items from traits can only be used if the trait is in scope
[build] help: the following trait is implemented but not in scope; perhaps add a `use` for it:
[build]    |
[build] 11 | use cxx_qt::Threading;
[build]    |
```

ahayzen-kdab:
Just needs rebasing, then looks good :-)

PR #553, opened 2023-05-29
Support result return type

ahayzen-kdab:
Requires #589


crates/cxx-qt-gen/src/generator/utils/rust.rs
```diff
@@ -0,0 +1,79 @@
...
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use syn::{GenericArgument, PathArguments, Result, ReturnType, Type, TypeReference};
+
+/// When a [syn::ReturnType] is used in a CXX bridge we need to ensure that
+/// there is only one generic
+pub(crate) fn syn_return_type_for_cxx_bridge(ty: &ReturnType) -> Result<ReturnType> {
+    // TODO: if Result<T, E> change to Result<T>
+
+    // FIXME: problem is that we can't detect anyhow::Result<T>
```
ahayzen-kdab, line 13: problem is this can't work as we can't determine the type.

The new API described in #531 is the better approach.

ahayzen-kdab:
We will still need some work here with the new API, the Rust side won't change as now the developer writes the signature and the implementation.

However we need to ensure that our type parser for creating the C++ type can support `Result<T>` and knows what to do with it when it's a return type.

So some parts of this pull request will still be valid.

Be-ing (approved):

PR #554, opened 2023-05-29
cxx-qt-gen: remove support for cxx_type and cxx_return_type

### LINKED ISSUE #289: Consider if cxx_type and return_cxx_type are still required
Do we still need the cxx_type and return_cxx_type ? If they can be removed then signal emits don't need a wrapper. And macro attribute parsing may be simpler. And the templates for converts can be removed.

ahayzen-kdab:
These aren't needed anymore as most Qt types are trivial.
This prevents us directly binding to things like signal later.
And these can all be worked around in C++ with proxies.
    
Closes #289


LeonMatthesKDAB (approved):
LGTM

PR #556, opened 2023-05-29
cxx-qt-gen: remove generation of wrapper for emitting signals

ahayzen-kdab:
This is now unused as the type conversion has been removed.

Requires #554 
Requires #571

### DIFF
--- a/crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
@@ -97,7 +97,7 @@ mod tests {
         assert_str_eq!(generated.metaobjects[1], "Q_PROPERTY(::std::unique_ptr<QColor> opaqueProperty READ getOpaqueProperty WRITE setOpaqueProperty NOTIFY opaquePropertyChanged)");
 
         // methods
-        assert_eq!(generated.methods.len(), 10);
+        assert_eq!(generated.methods.len(), 8);
         let (header, source) = if let CppFragment::Pair { header, source } = &generated.methods[0] {
             (header, source)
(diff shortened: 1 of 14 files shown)


ahayzen-kdab:
This doesn't work yet as we have two different cases for inheriting signals

  * inherit a signal from the base class
  * inherit a signal as a property has generated it

For the first we need to generate a connect + emit method, the second we need to generate a connect method but no emit as the qproperty has done this already.

```rust
    #[cxx_qt::qsignals(CustomBaseClass)]
    pub enum Signals<'a> {
        /// Inherit the DataChanged signal from the QAbstractListModel base class
        #[inherit]
        DataChanged {
            /// Top left affected index
            top_left: &'a QModelIndex,
            /// Bottom right affected index
            bottom_right: &'a QModelIndex,
            /// Roles that have been modified
            roles: &'a QVector_i32,
        },
    }

    #[cxx_qt::qsignals(RustSignals)]
    pub enum Connection<'a> {
        // Example of using #[inherit] so that connections to the logging_enabled property can be made
        #[inherit]
        // We don't ever emit this enum, so silence clippy warnings
        #[allow(dead_code)]
        /// The Q_SIGNAL emitted when the Q_PROPERTY logging_enabled changes
        LoggingEnabledChanged,
    }
```

ahayzen-kdab:
This will be solved by the new API as in that when there is a qproperty a myPropertyChanged signal will be generated and the connect method. So we don't have the enum craziness.

ahayzen-kdab:
This now works and ready for review :-)

LeonMatthesKDAB (approved):
LGTM, always good to see things simplified.

PR #564, opened 2023-05-29
cxx-qt-gen: use impl cxx_qt::Threading for T {} to make threading opt-in

### LINKED ISSUE #560: Add `impl cxx_qt::Threading for qobject::MyObject {}`
To make the code generation optional.

Imagined trait interface:
``` rust
trait Threading {
    fn qt_thread() -> CxxQtThread<Self>;
}
```
We may have to make the return type an associated `type CxxQtThread;`.

ahayzen-kdab:
Instead of always generating threading opt in via a trait.

Closes #560

### DIFF
--- a/crates/cxx-qt-gen/src/generator/cpp/mod.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/mod.rs
@@ -9,6 +9,7 @@ pub mod invokable;
 pub mod property;
 pub mod qobject;
 pub mod signal;
+pub mod threading;
 pub mod types;
 
 use crate::parser::Parser;
--- a/crates/cxx-qt-gen/src/generator/cpp/qobject.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/qobject.rs
@@ -6,7 +6,7 @@
 use crate::generator::{
     cpp::{
         fragment::CppFragment, inherit, invokable::generate_cpp_invokables,
-        property::generate_cpp_properties, signal::generate_cpp_signals,
+        property::generate_cpp_properties, signal::generate_cpp_signals, threading,
     },
     naming::{namespace::NamespaceName, qobject::QObjectName},
 };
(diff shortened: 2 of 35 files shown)


crates/cxx-qt-gen/src/generator/cpp/qobject.rs
```diff
@@ -142,7 +152,6 @@ mod tests {
         .unwrap();
         assert_eq!(cpp.ident, "MyObject");
         assert_eq!(cpp.rust_ident, "MyObjectRust");
-        assert_eq!(cpp.cxx_qt_thread_ident, "MyObjectCxxQtThread");
```
ahayzen-kdab, line 145: need to fix tests

crates/cxx-qt-gen/src/generator/cpp/mod.rs
```diff
@@ -9,6 +9,7 @@ pub mod invokable;
 pub mod property;
 pub mod qobject;
 pub mod signal;
+pub mod threading;
```
ahayzen-kdab, line 12: need to fix book

crates/cxx-qt-gen/src/generator/rust/threading.rs
```diff
@@ -0,0 +1,111 @@
...
+                }
+            },
+            quote! {
+                extern "Rust" {
+                    #[namespace = #namespace_internals]
+                    type #cxx_qt_thread_queued_fn_ident;
+                }
+            },
+        ],
+        implementation: vec![
```
ahayzen-kdab, line 64: need a trait implemented somewhere

crates/cxx-qt-gen/src/generator/cpp/qobject.rs
ahayzen-kdab: We have tests now :-)

crates/cxx-qt-gen/src/parser/cxxqtdata.rs
```diff
@@ -291,6 +292,15 @@ impl ParsedCxxQtData {
                 return Ok(None);
             // Find if we are an impl block for a qobject
             } else if let Some(qobject) = self.qobjects.get_mut(&path_to_single_ident(path)?) {
+                // Find if we are implemeting a trait for the QObject
+                if let Some((not, path, _)) = &imp.trait_ {
+                    // Find if we are enabling Threading
+                    if not.is_none() && path_compare_str(path, &["cxx_qt", "Threading"]) {
```
LeonMatthesKDAB, line 298: I think this should be restructured slightly.
Check the path_compare_str() first.
Then add assertions that not is_none() and return an error otherwise (i.e. "Negative impls for cxx_qt::Threading not allowed").
Also, please assert that the actual implementation is empty.

crates/cxx-qt-gen/src/writer/cpp/header.rs
```diff
@@ -31,7 +31,6 @@ fn create_block(block: &str, items: &[&str]) -> String {
                 // Remove any empty entries from the blocks
                 .filter(|item| !item.is_empty())
                 .cloned()
```
LeonMatthesKDAB, line 33: This `.cloned` call can also go now

crates/cxx-qt-gen/src/generator/rust/threading.rs
ahayzen-kdab: This will go in a separate pull request.

crates/cxx-qt-gen/src/parser/cxxqtdata.rs
ahayzen-kdab: done

LeonMatthesKDAB (approved):
LGTM

PR #565, opened 2023-05-30
cxx-qt-gen: Fix diff order

LeonMatthesKDAB:
Now new generated code is marked green and removed code is marked red.

### DIFF
--- a/crates/cxx-qt-gen/src/lib.rs
+++ b/crates/cxx-qt-gen/src/lib.rs
@@ -139,9 +139,9 @@ mod tests {
         let rust = sanitize_code(format_rs_source(&write_rust(&generated_rust).to_string()));
 
         if !update_expected(test_name, &rust, &header, &source) {
-            assert_str_eq!(header, sanitize_code(expected_cpp_header.to_owned()));
-            assert_str_eq!(source, sanitize_code(expected_cpp_source.to_owned()));
-            assert_str_eq!(rust, sanitize_code(expected_rust_output.to_owned()));
+            assert_str_eq!(sanitize_code(expected_cpp_header.to_owned()), header);
+            assert_str_eq!(sanitize_code(expected_cpp_source.to_owned()), source);
+            assert_str_eq!(sanitize_code(expected_rust_output.to_owned()), rust);
         }
     }
 


crates/cxx-qt-gen/src/lib.rs
```diff
@@ -139,9 +139,9 @@ mod tests {
         let rust = sanitize_code(format_rs_source(&write_rust(&generated_rust).to_string()));
 
         if !update_expected(test_name, &rust, &header, &source) {
-            assert_str_eq!(header, sanitize_code(expected_cpp_header.to_owned()));
-            assert_str_eq!(source, sanitize_code(expected_cpp_source.to_owned()));
-            assert_str_eq!(rust, sanitize_code(expected_rust_output.to_owned()));
+            assert_str_eq!(sanitize_code(expected_cpp_header.to_owned()), header);
```
ahayzen-kdab, line 142: Note that everywhere else we have output on the left and expected on the right ?

Have we got that wrong everywhere?

LeonMatthesKDAB: Hm, unsure.
assert_str_equal doesn't actually have an "expected" and "actual", just "left" and "right"
https://docs.rs/pretty_assertions/latest/pretty_assertions/struct.StrComparison.html

I believe for cases where we use `assert_eq!` it probably doesn't matter, as that uses different formatting anyhow.
But something to investigate :thinking:

ahayzen-kdab: So do we need to swap all cases of `assert_str_eq` ?

LeonMatthesKDAB: Hm, I guess we could...

Though these are certainly the most annoying.
So maybe merge these now and fix the others piece by piece?
As doing it all at once would certainly be terribly conflicting with basically everything.

ahayzen-kdab: I just fear that when we have two test failures. If one comes from this part of the code red diff lines will mean one thing whereas test failures from elsewhere red lines mean another.

Ideally we should ensure all the cases are consistent together. Maybe we just need to have this land during a quiet period when we don't have many active large branches :sweat_smile:

ahayzen-kdab: As discussed lets go with this for now and look into others later.

ahayzen-kdab (approved):
LGTM

PR #566, opened 2023-05-30
Constructor trait

### LINKED ISSUE #550: Support implementing a custom constructor from Rust

LeonMatthesKDAB:
Closes #550

### DIFF
--- a/crates/cxx-qt-gen/src/generator/cpp/constructor.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/constructor.rs
@@ -1,0 +1,135 @@
+// SPDX-FileCopyrightText: 2023 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Leon Matthes <leon.matthes@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use super::qobject::GeneratedCppQObjectBlocks;
+use crate::{
+    generator::cpp::{types::CppType, GeneratedCppQObject},
+    parser::{constructor::Constructor, cxxqtdata::ParsedCxxMappings},
+    CppFragment,
+};
+
+use indoc::formatdoc;
+use syn::{Result, Type};
+
+fn default_constructor(
+    qobject: &GeneratedCppQObject,
+    initializers: String,
+) -> GeneratedCppQObjectBlocks {
+    GeneratedCppQObjectBlocks {
+        methods: vec![CppFragment::Pair {
+            header: format!(
+                "explicit {class_name}(QObject* parent = nullptr);",
+                class_name = qobject.ident
+            ),
+            source: formatdoc!(
+                r#"
+            {class_name}::{class_name}(QObject* parent)
+              : {base_class}(parent)
+              , m_rustObj({namespace_internals}::createRs())
+              {initializers}
+            {{
+            }}
+            "#,
+                class_name = qobject.ident,
+                base_class = qobject.base_class,
+                namespace_internals = qobject.namespace_internals,
+            ),
+        }],
+        ..Default::default()
+    }
+}
+
+fn argument_names(arguments: &[Type]) -> Vec<String> {
+    arguments
+        .iter()
+        .enumerate()
+        .map(|(index, _)| format!("arg{index}"))
+        .collect()
+}
+
+fn expand_arguments(arguments: &[Type], cxx_mappings: &ParsedCxxMappings) -> Result<String> {
+    Ok(arguments
+        .iter()
+        .zip(argument_names(arguments).into_iter())
+        .map(|(ty, name)| {
+            CppType::from(ty, cxx_mappings).map(|ty| format!("{ty} {name}", ty = ty.as_cxx_ty()))
+        })
+        .collect::<Result<Vec<_>>>()?
+        .join(", "))
+}
+
+pub fn generate(
+    qobject: &GeneratedCppQObject,
+    constructors: &Vec<Constructor>,
+    member_initializers: &[String],
+    cxx_mappings: &ParsedCxxMappings,
+) -> Result<GeneratedCppQObjectBlocks> {
+    let initializers = member_initializers
+        .iter()
+        .map(|initializer| format!(", {initializer}"))
+        .collect::<Vec<_>>()
+        .join("\n");
+
+    if constructors.is_empty() {
+        return Ok(default_constructor(qobject, initializers));
+    }
+
+    let mut generated = GeneratedCppQObjectBlocks::default();
+
+    let class_name = qobject.ident.as_str();
+    let namespace_internals = &qobject.namespace_internals;
+    let base_class = &qobject.base_class;
+    for (index, constructor) in constructors.iter().enumerate() {
+        let argument_list = expand_arguments(&constructor.arguments, cxx_mappings)?;
+        let constructor_argument_names = argument_names(&constructor.arguments);
+
+        generated.methods.push(CppFragment::Pair {
+            header: format!("explicit {class_name}({argument_list});"),
+            source: formatdoc! {
+                r#"
+                {class_name}::{class_name}({argument_list})
+                    : {class_name}({namespace_internals}::routeArguments{index}({move_arguments}))
+                    {{ }}
+                "#,
+                move_arguments = constructor_argument_names.iter().map(|arg| format!("std::move({arg})")).collect::<Vec<_>>().join(", "),
+            },
+        });
+
+        let base_args = if let Some(base_args) = &constructor.base_arguments {
+            argument_names(base_args)
+                .into_iter()
+                .map(|arg| format!("std::move(args.baseArguments.{arg})"))
+                .collect::<Vec<_>>()
+                .join(", ")
+        } else {
+            "".to_string()
+        };
+        // For each constructor defined in CXX-Qt we need a pair of one public and one private
+        // constructor.
+        // The reason for this is that CXX-Qt needs to be able to route the list of raw arguments
+        // provided in C++ to a Plain-Old-Data type that contains the arguments already routed
+        // through Rust.
+        // This second constructor which takes the routed arguments is private, so that only CXX-Qt
+        // can use it.
+        generated.private_methods.push(CppFragment::Pair {
+            header: format!(
+                "explicit {class_name}({namespace_internals}::CxxQtConstructorArguments{index}&& args);"
+            ),
+            source: formatdoc! {
+                r#"
+                {class_name}::{class_name}({namespace_internals}::CxxQtConstructorArguments{index}&& args)
+                    : {base_class}({base_args})
+                    , m_rustObj({namespace_internals}::newRs{index}(std::move(args.newArguments)))
+                    {initializers}
+                    {{
+                        {namespace_internals}::initialize{index}(*this, std::move(args.initializeArguments));
+                    }}
+                "#,
+            },
+        })
+    }
+
+    Ok(generated)
+}
--- a/crates/cxx-qt-gen/src/generator/cpp/threading.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/threading.rs
@@ -10,7 +10,7 @@ use crate::generator::{
 use indoc::formatdoc;
 use syn::Result;
 
-pub fn generate(qobject_idents: &QObjectName) -> Result<GeneratedCppQObjectBlocks> {
+pub fn generate(qobject_idents: &QObjectName) -> Result<(String, GeneratedCppQObjectBlocks)> {
     let mut result = GeneratedCppQObjectBlocks::default();
 
     let ident = &qobject_idents.cpp_class.cpp;
@@ -27,10 +27,9 @@ pub fn generate(qobject_idents: &QObjectName) -> Result<GeneratedCppQObjectBlock
         static_assert(sizeof({cxx_qt_thread_ident}) == sizeof(::std::size_t[4]), "unexpected size");
         "#
     }));
-    result.members.push(CppFragment::Pair {
-        header: format!("::std::shared_ptr<::rust::cxxqtlib1::CxxQtGuardedPointer<{ident}>> m_cxxQtThreadObj;"),
-        source: format!(", m_cxxQtThreadObj(::std::make_shared<::rust::cxxqtlib1::CxxQtGuardedPointer<{ident}>>(this))"),
-    });
+    result.members.push(format!(
+        "::std::shared_ptr<::rust::cxxqtlib1::CxxQtGuardedPointer<{ident}>> m_cxxQtThreadObj;"
+    ));
     result.methods.push(CppFragment::Pair {
         header: format!("{cxx_qt_thread_ident} qtThread() const;"),
         source: formatdoc! {
@@ -50,7 +49,9 @@ pub fn generate(qobject_idents: &QObjectName) -> Result<GeneratedCppQObjectBlock
         "#
     });
 
-    Ok(result)
+    let member_initializer = format!("m_cxxQtThreadObj(::std::make_shared<::rust::cxxqtlib1::CxxQtGuardedPointer<{ident}>>(this))");
+
+    Ok((member_initializer, result))
 }
 
 #[cfg(test)]
@@ -65,7 +66,7 @@ mod tests {
     fn test_generate_cpp_threading() {
         let qobject_idents = create_qobjectname();
 
-        let generated = generate(&qobject_idents).unwrap();
+        let (initializer, generated) = generate(&qobject_idents).unwrap();
 
         // forward declares
         assert_eq!(generated.forward_declares.len(), 1);
@@ -77,19 +78,13 @@ mod tests {
 
         // members
         assert_eq!(generated.members.len(), 1);
-
-        let (header, source) = if let CppFragment::Pair { header, source } = &generated.members[0] {
-            (header, source)
-        } else {
-            panic!("Expected pair")
-        };
         assert_str_eq!(
-            header,
+            &generated.members[0],
             "::std::shared_ptr<::rust::cxxqtlib1::CxxQtGuardedPointer<MyObject>> m_cxxQtThreadObj;"
         );
         assert_str_eq!(
-            source,
-            ", m_cxxQtThreadObj(::std::make_shared<::rust::cxxqtlib1::CxxQtGuardedPointer<MyObject>>(this))"
+            initializer,
+            "m_cxxQtThreadObj(::std::make_shared<::rust::cxxqtlib1::CxxQtGuardedPointer<MyObject>>(this))"
         );
 
         // methods
--- a/crates/cxx-qt-gen/src/generator/rust/constructor.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/constructor.rs
@@ -1,0 +1,316 @@
+// SPDX-FileCopyrightText: 2023 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Leon Matthes <leon.matthes@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use crate::{
+    generator::{
+        naming::{namespace::NamespaceName, qobject::QObjectName, CombinedIdent},
+        rust::{qobject::GeneratedRustQObjectBlocks, types},
+    },
+    parser::constructor::Constructor,
+};
+
+use convert_case::{Case, Casing};
+use proc_macro2::{Span, TokenStream};
+use quote::{format_ident, quote};
+use syn::{parse_quote, parse_quote_spanned, spanned::Spanned, Expr, Ident, Item, Result, Type};
+
+const CONSTRUCTOR_ARGUMENTS: &str = "CxxQtConstructorArguments";
+const BASE_ARGUMENTS: &str = "CxxQtConstructorBaseArguments";
+const NEW_ARGUMENTS: &str = "CxxQtConstructorNewArguments";
+const INITIALIZE_ARGUMENTS: &str = "CxxQtConstructorInitializeArguments";
+
+fn map_types<F: FnMut((usize, &Type)) -> TokenStream>(
+    args: &Option<Vec<Type>>,
+    f: F,
+) -> Vec<TokenStream> {
+    args.as_ref()
+        .map(|args| args.iter().enumerate().map(f).collect())
+        .unwrap_or_default()
+}
+
+fn extract_arguments_from_tuple(args: &Option<Vec<Type>>, tuple_name: Ident) -> Vec<TokenStream> {
+    map_types(args, |(index, _ty)| {
+        let arg_name = format_ident!("arg{index}");
+        let index = syn::LitInt::new(index.to_string().as_str(), Span::call_site());
+        quote! {
+            #arg_name: #tuple_name.#index
+        }
+    })
+}
+
+fn extract_arguments_from_struct(args: &Option<Vec<Type>>, struct_name: Ident) -> Vec<TokenStream> {
+    map_types(args, |(index, _ty)| {
+        let arg_name = format_ident!("arg{index}");
+        quote! {
+            #struct_name.#arg_name
+        }
+    })
+}
+
+fn argument_members(args: &Option<Vec<Type>>) -> Vec<TokenStream> {
+    map_types(args, |(index, ty)| {
+        let arg_name = format_ident!("arg{index}");
+        quote! {
+            #arg_name: #ty
+        }
+    })
+}
+
+fn generate_default_constructor(
+    qobject_idents: &QObjectName,
+    namespace: &NamespaceName,
+) -> GeneratedRustQObjectBlocks {
+    let rust_struct_ident = &qobject_idents.rust_struct.rust;
+    let create_rs_ident = format_ident!(
+        "create_rs_{object_name}",
+        object_name = rust_struct_ident.to_string().to_case(Case::Snake)
+    );
+    let namespace_internals = &namespace.internal;
+
+    GeneratedRustQObjectBlocks {
+        cxx_mod_contents: vec![parse_quote! {
+        extern "Rust" {
+            #[cxx_name = "createRs"]
+            #[namespace = #namespace_internals]
+            fn #create_rs_ident() -> Box<#rust_struct_ident>;
+        }
+        }],
+        cxx_qt_mod_contents: vec![parse_quote! {
+            /// Generated CXX-Qt method which creates a boxed rust struct of a QObject
+            pub fn #create_rs_ident() -> std::boxed::Box<#rust_struct_ident> {
+                core::default::Default::default()
+            }
+        }],
+    }
+}
+
+fn generate_arguments_struct(
+    namespace_internals: &str,
+    struct_name: &CombinedIdent,
+    argument_list: &Option<Vec<Type>>,
+) -> Item {
+    let argument_members = argument_members(argument_list);
+    let not_empty = if argument_list.as_ref().is_some_and(|list| !list.is_empty()) {
+        None
+    } else {
+        Some(quote! { not_empty: i8 })
+    };
+    let rust_name = &struct_name.rust;
+    // use to_string here, as the cxx_name needs to be in quotes for the attribute macro.
+    let cxx_name = &struct_name.cpp.to_string();
+    parse_quote! {
+        #[namespace = #namespace_internals]
+        #[cxx_name = #cxx_name]
+        #[doc(hidden)]
+        struct #rust_name {
+            #(#argument_members,)*
+            #not_empty // Make sure there's always at least one struct member, as CXX
+                       // doesn't support empty shared structs.
+        }
+    }
+}
+
+fn generate_arguments_initialization(
+    struct_name: &Ident,
+    instance_name: Ident,
+    argument_list: &Option<Vec<Type>>,
+) -> Expr {
+    let init_arguments = extract_arguments_from_tuple(argument_list, instance_name);
+    println!("init_arguments: {:?}", init_arguments);
+    let not_empty = if argument_list.as_ref().is_some_and(|list| !list.is_empty()) {
+        None
+    } else {
+        Some(quote! { not_empty: 0 })
+    };
+
+    parse_quote! {
+        #struct_name {
+            #(#init_arguments,)*
+            #not_empty
+        }
+    }
+}
+
+pub fn generate(
+    constructors: &[Constructor],
+    qobject_idents: &QObjectName,
+    namespace: &NamespaceName,
+) -> Result<GeneratedRustQObjectBlocks> {
+    if constructors.is_empty() {
+        return Ok(generate_default_constructor(qobject_idents, namespace));
+    }
+
+    let mut result = GeneratedRustQObjectBlocks::default();
+    let namespace_internals = &namespace.internal;
+    let rust_struct_name_rust = &qobject_idents.rust_struct.rust;
+    let rust_struct_name_snake = &qobject_idents
+        .rust_struct
+        .rust
+        .to_string()
+        .to_case(Case::Snake);
+    let qobject_name_rust = &qobject_idents.cpp_class.rust;
+
+    for (index, constructor) in constructors.iter().enumerate() {
+        let arguments_rust = format_ident!("{CONSTRUCTOR_ARGUMENTS}{rust_struct_name_rust}{index}");
+        let base_arguments_rust = format_ident!("{BASE_ARGUMENTS}{rust_struct_name_rust}{index}");
+        let new_arguments_rust = format_ident!("{NEW_ARGUMENTS}{rust_struct_name_rust}{index}");
+        let initialize_arguments_rust =
+            format_ident!("{INITIALIZE_ARGUMENTS}{rust_struct_name_rust}{index}");
+
+        let arguments_cxx = format!("{CONSTRUCTOR_ARGUMENTS}{index}");
+        let base_arguments_cxx = format_ident!("{BASE_ARGUMENTS}{index}");
+        let new_arguments_cxx = format_ident!("{NEW_ARGUMENTS}{index}");
+        let initialize_arguments_cxx = format_ident!("{INITIALIZE_ARGUMENTS}{index}");
+
+        let new_rust = format_ident!("new_rs_{rust_struct_name_snake}_{index}");
+        let new_cxx = format!("newRs{index}");
+
+        let initialize_rust = format_ident!("initialize_{rust_struct_name_snake}_{index}");
+        let initialize_cxx = format!("initialize{index}");
+
+        let route_arguments_rust =
+            format_ident!("route_arguments_{rust_struct_name_snake}_{index}");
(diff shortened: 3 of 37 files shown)


ahayzen-kdab:
Random thought, how do this work in combination with `#[derive(Default)]` or `impl Default for T` on the `MyObject` struct ?

Is it allowed to have a `Constructor` as well ? I guess yes ? it doesn't really affect things just means that the rust constructor could call `T::default()` or the developer could do what they want instead.

crates/cxx-qt-gen/src/generator/cpp/constructor.rs
```diff
@@ -0,0 +1,128 @@
...
+        let base_args = if let Some(base_args) = &constructor.items.base_arguments {
+            argument_names(base_args)
+                .into_iter()
+                .map(|arg| format!("std::move(args.baseArguments.{arg})"))
+                .collect::<Vec<_>>()
+                .join(", ")
+        } else {
+            "".to_string()
+        };
+        generated.private_methods.push(CppFragment::Pair {
```
ahayzen-kdab, line 109: do we need some commentary in this block to say why we need both a public and private constructor for each constructor set ?

crates/cxx-qt-gen/src/generator/cpp/threading.rs
```diff
@@ -19,10 +19,10 @@ pub fn generate(qobject_idents: &QObjectName) -> Result<GeneratedCppQObjectBlock
...
         "using {cxx_qt_thread_ident} = ::rust::cxxqtlib1::CxxQtThread<{ident}>;"
     ));
-    result.members.push(CppFragment::Pair {
-        header: format!("::std::shared_ptr<::rust::cxxqtlib1::CxxQtGuardedPointer<{ident}>> m_cxxQtThreadObj;"),
-        source: format!(", m_cxxQtThreadObj(::std::make_shared<::rust::cxxqtlib1::CxxQtGuardedPointer<{ident}>>(this))"),
-    });
+    result.members.push(format!(
+        "::std::shared_ptr<::rust::cxxqtlib1::CxxQtGuardedPointer<{ident}>> m_cxxQtThreadObj;"
+    ));
+    let member_initializer = format!("m_cxxQtThreadObj(::std::make_shared<::rust::cxxqtlib1::CxxQtGuardedPointer<{ident}>>(this))");
```
ahayzen-kdab, line 25: could put this after all the `result.***.push(..)` blocks so it's not hidden  in the middle.

crates/cxx-qt-gen/src/generator/rust/constructor.rs
```diff
@@ -0,0 +1,310 @@
...
+                    #(#initialize_argument_members,)*
+                    notEmpty: i8 // Make sure there's always at least one struct member, as CXX
+                                 // doesn't support empty shared structs.
+                }
+            },
+            parse_quote! {
+                extern "Rust" {
+                    #[namespace = #namespace_internals]
+                    #[cxx_name = #route_arguemnts_cxx]
+                    // This function needs to marked unsafe, as some arguments may be pointers.
```
ahayzen-kdab, line 242: Can you not use the `is_unsafe_cxx_type` to determine this, if it only happens sometimes ?

If it always happens then rephrase to state there'll always be pointers

crates/cxx-qt-gen/src/generator/rust/constructor.rs
```diff
@@ -0,0 +1,310 @@
...
+                    #[namespace = #namespace_internals]
+                    #[cxx_name = #initialize_cxx]
+                    fn #initialize_rust(qobject: Pin<&mut #qobject_name_rust>, args: #initialize_arguments_rust);
+                }
+            },
+        ]);
+        result.cxx_qt_mod_contents.append(&mut vec![parse_quote! {
+            #[doc(hidden)]
+            pub fn #route_arguments_rust(#(#route_arguments_parameters),*) -> #arguments_rust {
+                // these won't be used if any of them don't have any members
```
ahayzen-kdab, line 258: two negatives in this sentence are really hard to read, is it simpler to write it in the opposite way?

crates/cxx-qt-gen/src/generator/rust/constructor.rs
```diff
@@ -0,0 +1,310 @@
...
+            #[doc(hidden)]
+            pub fn #new_rust(new_arguments: #new_arguments_rust) -> Box<#rust_struct_name_rust> {
+                let new_arguments = (#(#extract_new_arguments,)*);
+                Box::new(<#qobject_name_rust as cxx_qt::Constructor<(#(#argument_types,)*)>>::new(new_arguments))
+            }
+        },
+        parse_quote! {
+            #[doc(hidden)]
+            pub fn #initialize_rust(
+                qobject: Pin<&mut #qobject_name_rust>,
```
ahayzen-kdab, line 292: ensure these are fully qualified `core::pin::Pin`

crates/cxx-qt-gen/src/generator/rust/constructor.rs
```diff
@@ -0,0 +1,310 @@
...
+        cxx_mod_contents: vec![parse_quote! {
+        extern "Rust" {
+            #[cxx_name = "createRs"]
+            #[namespace = #namespace_internals]
+            fn #create_rs_ident() -> Box<#rust_struct_ident>;
+        }
+        }],
+        cxx_qt_mod_contents: vec![parse_quote! {
+            /// Generated CXX-Qt method which creates a boxed rust struct of a QObject
+            pub fn #create_rs_ident() -> std::boxed::Box<#rust_struct_ident> {
```
ahayzen-kdab, line 82: should probably use `alloc::boxed::Box` instead

crates/cxx-qt-gen/src/generator/rust/constructor.rs
```diff
@@ -0,0 +1,310 @@
...
+                std::default::Default::default()
```
ahayzen-kdab, line 83: use `core::default::Default`

crates/cxx-qt-gen/src/generator/rust/constructor.rs
```diff
@@ -0,0 +1,310 @@
...
+                        notEmpty: 0
+                    },
+                }
+            }
+        },
+        parse_quote! {
+            #[doc(hidden)]
+            pub fn #new_rust(new_arguments: #new_arguments_rust) -> Box<#rust_struct_name_rust> {
+                let new_arguments = (#(#extract_new_arguments,)*);
+                Box::new(<#qobject_name_rust as cxx_qt::Constructor<(#(#argument_types,)*)>>::new(new_arguments))
```
ahayzen-kdab, line 286: do you need `alloc::boxed::Box` here for both the Boxes to fully qualify them like elsewhere?

crates/cxx-qt-gen/src/parser/constructor.rs
```diff
@@ -0,0 +1,126 @@
...
+// SPDX-License-Identifier: MIT OR Apache-2.0
+use syn::{
+    Path,
+    AngleBracketedGenericArguments, Error, ItemImpl,
+    PathArguments, PathSegment, Result,Type, GenericArgument,
+    spanned::Spanned, ImplItem
+};
+
+#[derive(Default)]
+pub struct ImplItems {
```
ahayzen-kdab, line 13: some doc lines about the fields could be useful here

crates/cxx-qt-gen/src/parser/constructor.rs
```diff
@@ -0,0 +1,126 @@
...
+    pub new_arguments: Option<Vec<Type>>,
+    pub base_arguments: Option<Vec<Type>>,
+    pub initialize_arguments: Option<Vec<Type>>,
+
+    pub passthrough: Vec<ImplItem>,
+}
+
+pub struct Constructor {
+    pub arguments: Vec<Type>,
```
ahayzen-kdab, line 22: some doc lines here could be useful too

crates/cxx-qt-gen/src/parser/constructor.rs
```diff
@@ -0,0 +1,126 @@
...
+        if let [
+            // cxx_qt::
+            _cxx_qt, 
+            // Constructor
+            PathSegment {
+            arguments:
+                PathArguments::AngleBracketed(AngleBracketedGenericArguments { args: generics_punct, .. }),
+            ..
+        }] = *constructor_path
+        {
```
ahayzen-kdab, line 66: should you check somewhere here that you have actually been given `Constructor` and not some other random trait?

crates/cxx-qt-gen/src/parser/constructor.rs
```diff
@@ -0,0 +1,126 @@
...
+            let generics : Vec<_> = generics_punct.iter().collect();
```
ahayzen-kdab, line 67: remove the extra space after `generics` ?

crates/cxx-qt-gen/src/parser/constructor.rs
```diff
@@ -0,0 +1,126 @@
...
+                // if the generics are empty, using the span of `generics_punct` causes an invalid
+                // span to be returned!
+                let span = if generics.is_empty() { trait_path.span() } else { generics_punct.span() };
+                return Err(Error::new(span, "cxx_qt::Constructor expects a single generic argument, use a tuple to provide multiple arguments!"))
+            }
+
+            if let [GenericArgument::Type(ty)] = *generics {
+                Self::parse_argument_list(ty.clone())
+            }
+            else {
```
ahayzen-kdab, line 78: normally it's `} else {` on a single line

crates/cxx-qt-gen/src/parser/constructor.rs
```diff
@@ -0,0 +1,126 @@
...
+        {
+            let generics : Vec<_> = generics_punct.iter().collect();
+            if generics.len() != 1 {
+                // if the generics are empty, using the span of `generics_punct` causes an invalid
+                // span to be returned!
+                let span = if generics.is_empty() { trait_path.span() } else { generics_punct.span() };
+                return Err(Error::new(span, "cxx_qt::Constructor expects a single generic argument, use a tuple to provide multiple arguments!"))
+            }
+
+            if let [GenericArgument::Type(ty)] = *generics {
```
ahayzen-kdab, line 75: the `[...]` is curious here, is that pattern matching on the first element in a slice / array ?

crates/cxx-qt-gen/src/writer/cpp/source.rs
```diff
@@ -49,36 +43,12 @@ fn qobjects_source(generated: &GeneratedCppBlocks) -> Vec<String> {
...
-                members.extend(vec![
-                    ", m_rustObjMutex(::std::make_shared<::std::recursive_mutex>())".to_string(),
-                ]);
-            }
-
-            members.extend(qobject.blocks.members.iter().filter_map(pair_as_source).collect::<Vec<String>>());
-            members.join("\n  ")
-        },
+        methods = qobject.blocks.methods.iter().chain(qobject.blocks.private_methods.iter()).filter_map(pair_as_source).collect::<Vec<String>>().join("\n"),
         deconstructors = qobject.blocks.deconstructors.join("\n  "),
```
ahayzen-kdab, line 52: i wonder if in a further commit deconstructors should be handled in a similar way and built up to a normal public method in the `generator` phase rather than happening in the `writer` phase.

Especially if we are going to have a trait in the future to allow devs to run code in the C++ context in the deconstructor. But maybe this isn't required.

Anyway a separate future change i guess?

crates/cxx-qt/Cargo.toml
```diff
@@ -19,3 +19,5 @@ categories = ["api-bindings", "gui"]
 
 [dependencies]
 cxx-qt-macro.workspace = true
+cxx.workspace = true
+cxx-qt-lib.workspace = true
```
ahayzen-kdab, line 23: `cxx-qt` should not depend on `cxx-qt-lib`, i also can't see why this was added ?

crates/cxx-qt/Cargo.toml
```diff
@@ -19,3 +19,5 @@ categories = ["api-bindings", "gui"]
 
 [dependencies]
 cxx-qt-macro.workspace = true
+cxx.workspace = true
```
ahayzen-kdab, line 22: why was this added, it seems unused ?

ahayzen-kdab: Ah maybe for the docs tests ?

Then it goes under `[dev-dependencies]` not `[dependencies]`

crates/cxx-qt/Cargo.toml
ahayzen-kdab: move under `dev-dependencies` if it's used by the doc tests, but curiously how does it fail without?

crates/cxx-qt/src/lib.rs
```diff
@@ -50,3 +50,118 @@ pub trait CxxQtType {
...
+/// This trait can be implemented inside the [`#[cxx_qt::bridge]`](bridge) on any [CxxQtType] to define a
+/// custom constructor in C++ for the QObject.
+///
+/// The `Arguments` must be a tuple of CXX types that will be the arguments to the constructor in C++.
+///
+/// Note that implementing this trait will stop Cxx-Qt from generating a default constructor.
+///
+/// # Minimal Example
+/// ```no_run
+/// # #[macro_use] extern crate cxx;
```
ahayzen-kdab, line 63: curious, why is that line required when we didn't need it in the doc tests in `cxx-qt-macro`?

crates/cxx-qt/src/lib.rs
```diff
@@ -50,3 +50,118 @@ pub trait CxxQtType {
...
+/// #[cxx_qt::bridge]
+/// mod ffi {
+///     // Struct without `Default` implementation
+///     #[cxx_qt::qobject]
+///     pub struct MyStruct {
+///         integer: i32,
+///         string: String
```
ahayzen-kdab, line 70: could always bring in `cxx_qt_lib::QString`

crates/cxx-qt/src/lib.rs
```diff
@@ -50,3 +50,118 @@ pub trait CxxQtType {
...
+///         {
+///             Constructor::initialize(*this, route_arguments(args).InitializeArguments);
+///         }
+/// }
+/// ```
+/// Note that in reality, `route_arguments` will only be called once and all arguments
+/// will be moved, never copied.
+pub trait Constructor<Arguments>: CxxQtType {
+    /// The arguments that are passed to the [`new()`](Self::new) function to construct the inner Rust struct.
+    /// This must be a tuple of Cxx compatible types.
```
ahayzen-kdab, line 115: CXX instead of Cxx

crates/cxx-qt/src/lib.rs
```diff
@@ -50,3 +50,118 @@ pub trait CxxQtType {
...
+    ///
+    /// This way QObjects can be constructed that need additional arguments for constructing the
+    /// inner Rust type.
+    // TODO: Enable Defaults
+    // # Default
+    // If nothing is specified, CxxQt will default this to `Arguments`.
```
ahayzen-kdab, line 121: CXX-Qt

crates/cxx-qt/src/lib.rs
```diff
@@ -50,3 +50,118 @@ pub trait CxxQtType {
...
+    type NewArguments;
+    /// The arguments that should be passed to the constructor of the base class.
+    /// This must be a tuple of Cxx compatible types.
+    // TODO: Enable Defaults
+    // # Default
+    // If nothing is specified CxxQt will default this to an empty tuple.
```
ahayzen-kdab, line 127: CXX-Qt

crates/cxx-qt/src/lib.rs
```diff
@@ -50,3 +50,118 @@ pub trait CxxQtType {
...
+    type BaseArguments;
+    /// The arguments that should be used to initialize the QObject in the [`initialize()`](Self::initialize) function.
+    /// This must be a tuple of Cxx compatible types.
+    // TODO: Enable Defaults
+    // # Default
+    // If nothing is specified CxxQt will default this to an empty tuple.
```
ahayzen-kdab, line 133: CXX-Qt

crates/cxx-qt/src/lib.rs
```diff
@@ -50,3 +50,118 @@ pub trait CxxQtType {
...
+    type InitializeArguments;
+
+    /// This function is called by CxxQt to route the arguments to the correct places.
```
ahayzen-kdab, line 136: CXX-Qt

crates/cxx-qt/src/lib.rs
```diff
@@ -50,3 +50,118 @@ pub trait CxxQtType {
...
+    ///
+    /// This way QObjects can be constructed that need additional arguments for constructing the
+    /// inner Rust type.
+    // TODO: Enable Defaults
+    // # Default
+    // If nothing is specified, CxxQt will default this to `Arguments`.
+    type NewArguments;
+    /// The arguments that should be passed to the constructor of the base class.
+    /// This must be a tuple of Cxx compatible types.
+    // TODO: Enable Defaults
```
ahayzen-kdab, line 125: Is this going to happen in this pull request or future ones?

crates/cxx-qt/src/lib.rs
```diff
@@ -50,3 +50,118 @@ pub trait CxxQtType {
...
+    #[allow(unused_variables)]
+    fn route_arguments(
+        arguments: Arguments,
+    ) -> (
+        Self::NewArguments,
+        Self::BaseArguments,
+        Self::InitializeArguments,
+    );
+
+    /// This function is called to construct the inner Rust struct of the CxxQt QObject.
```
ahayzen-kdab, line 153: CXX-Qt

examples/qml_features/cpp/main.cpp
```diff
@@ -9,12 +9,15 @@
...
 
 #include "custom_object.h"
+#include "cxx-qt-gen/rust_invokables.cxxqt.h"
 
 int
 main(int argc, char* argv[])
 {
   QGuiApplication app(argc, argv);
 
+  RustInvokables invokables(0.f, 0.f, 1.f, nullptr);
```
ahayzen-kdab, line 19: Would be better to create a separate constructor.rs example in qml_features, then keep RustInvokables as it was.

Otherwise people searching through the qml_features examples won't be able to find custom constructors easily.

examples/qml_features/rust/src/invokables.rs
```diff
@@ -13,23 +13,38 @@ pub mod ffi {
...
     }
 
-    impl Default for RustInvokables {
-        fn default() -> Self {
-            Self {
-                red: 0.0,
-                green: 0.4667,
-                blue: 0.7843,
-            }
+    impl cxx_qt::Constructor<(f32, f32, f32, UniquePtr<QObject>)> for qobject::RustInvokables {
```
ahayzen-kdab, line 31: why is this `UniquePtr<QObject>` and then `*mut QObject` why not `*mut QObject` in both ?

examples/qml_features/rust/src/invokables.rs
```diff
@@ -13,23 +13,38 @@ pub mod ffi {
...
+        type BaseArguments = (*mut QObject,);
```
ahayzen-kdab, line 32: This looks good enough for me to port https://github.com/KDAB/cxx-qt/blob/d50012461d76d69884be82941f3b36ae915dea14/examples/qml_features/rust/src/custom_parent_class.rs over to show  how to do a custom base argument so that `QQuickPaintedItem` is possible.

I'll do this once things have been rebased here.

crates/cxx-qt-gen/src/generator/cpp/constructor.rs
```diff
@@ -0,0 +1,128 @@
+// SPDX-FileCopyrightText: 2023 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
```
ahayzen-kdab, line 1: don't forget to add info about the constructor trait to the book :-)

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 
 - Removed support for `cxx_type` and `cxx_return_type` and related conversion methods.
+- Removed `newCppObject` function that allowed creation of default-constructed QObject from Rust.
```
ahayzen-kdab, line 35: add an entry in here for adding the contructor trait which allows support for
- custom arguments
- base arguments (which unlocks implementing types like QQuickPaintedItem)
- initialise method for starting things in Rust at the constructor (like threads)

LeonMatthesKDAB:
> Random thought, how do this work in combination with `#[derive(Default)]` or `impl Default for T` on the `MyObject` struct ?
> 
> Is it allowed to have a `Constructor` as well ? I guess yes ? it doesn't really affect things just means that the rust constructor could call `T::default()` or the developer could do what they want instead.

That's not an issue, as the Default only applies to Rust in that case and the Constructor implementation is totally independent and may even call default if one wants.

CHANGELOG.md
LeonMatthesKDAB: Let's post-pone this until the Constructor trait is more thoroughly tested.
Then we'll add documentation for the feature as well.

crates/cxx-qt-gen/src/generator/cpp/constructor.rs
LeonMatthesKDAB: Like with the Changelog I think we should post-pone this until the feature is more stable.
Let's keep it as a "hidden" feature for now.

crates/cxx-qt-gen/src/generator/cpp/constructor.rs
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/generator/cpp/threading.rs
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/generator/rust/constructor.rs
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/generator/rust/constructor.rs
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/generator/rust/constructor.rs
LeonMatthesKDAB: Turns out I can.
Done.

crates/cxx-qt-gen/src/generator/rust/constructor.rs
LeonMatthesKDAB: Agreed, reprhased it.

crates/cxx-qt-gen/src/parser/constructor.rs
LeonMatthesKDAB: Done.

crates/cxx-qt-gen/src/parser/constructor.rs
LeonMatthesKDAB: Done.

crates/cxx-qt-gen/src/parser/constructor.rs
LeonMatthesKDAB: This is already being checked by the previous parser stage in parser/qobject.rs.
So it's kind of redundant to do that here.

crates/cxx-qt-gen/src/parser/constructor.rs
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/parser/constructor.rs
LeonMatthesKDAB: yes it is and it's super useful.
Got inspired by this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment?retiredLocale=de
Turns out the same works in Rust.

crates/cxx-qt-gen/src/parser/constructor.rs
LeonMatthesKDAB: Hm, yes and I'm very confused why rustfmt doesn't seem to want to fix this...

crates/cxx-qt-gen/src/writer/cpp/source.rs
LeonMatthesKDAB: yes, I think this will be a change we'll make at some point. But it can wait for now.

examples/qml_features/rust/src/invokables.rs
LeonMatthesKDAB: I think I just did that to test whether types like UniquePtr and raw pointers would work.

crates/cxx-qt/Cargo.toml
LeonMatthesKDAB: Ah, I think that was failing because I tried to use QString, might not be necessary now that I'm using String :thinking:

crates/cxx-qt/src/lib.rs
LeonMatthesKDAB: I was trying to avoid that to keep the code in the example minimal and not clutter it with distracting imports.
And hiding the import might have users confused why the example doesn't compile if they copy it into their own codebase.

crates/cxx-qt/src/lib.rs
LeonMatthesKDAB: Hm, seems to work without this as well.
Might have been fixed by adding cxx as a dev-dependency, not sure.

crates/cxx-qt/src/lib.rs
LeonMatthesKDAB: Done.

crates/cxx-qt/src/lib.rs
LeonMatthesKDAB: Done

crates/cxx-qt/src/lib.rs
LeonMatthesKDAB: In future ones, I just wanted to document here what the plan was.
Note that it's using double `/`, so it's a normal comment, not part of the documentation (yet).

crates/cxx-qt/src/lib.rs
LeonMatthesKDAB: Done

crates/cxx-qt/src/lib.rs
LeonMatthesKDAB: Done

crates/cxx-qt/src/lib.rs
LeonMatthesKDAB: Done

crates/cxx-qt/src/lib.rs
LeonMatthesKDAB: Done

crates/cxx-qt/Cargo.toml
LeonMatthesKDAB: Still necessary, but moved to dev-dependencies now.

crates/cxx-qt/Cargo.toml
LeonMatthesKDAB: Done

examples/qml_features/cpp/main.cpp
LeonMatthesKDAB: I've migrated to implementing this on the two types that currently need a `Component.onComplete` callback, so that it's actually useful.
Hopefully that'll be easy enough to find.
We can always add a separate `Constructor` example later.

crates/cxx-qt-gen/src/generator/rust/constructor.rs
```diff
@@ -0,0 +1,297 @@
...
+                    initializeArguments: #initialize_arguments_rust,
+                }
+            },
+            parse_quote! {
+                #[namespace = #namespace_internals]
+                #[cxx_name = #base_arguments_cxx]
+                #[doc(hidden)]
+                struct #base_arguments_rust {
+                    #(#base_argument_members,)*
+                    notEmpty: i8 // Make sure there's always at least one struct member, as CXX
```
ahayzen-kdab, line 202: unfortunate, assume you can't use phantom data ? And is a bool smaller than a i8 ?

ahayzen-kdab: Also have you seen upstream if there is an issue for this, could be something to fix if it's simple ?

If there is an issue or you make one, link in a comment

crates/cxx-qt-gen/src/writer/cpp/mod.rs
```diff
@@ -233,6 +251,15 @@ mod tests {
                         },
                         CppFragment::Header("Q_SIGNAL void countChanged();".to_owned()),
                         ],
+                        private_methods: vec![
+                            CppFragment::Pair{
+                                header: "void privateMethod() const;".to_owned(),
+                                source: indoc! {r#"
+                                    void SecondObject::privateMethod() const {
```
ahayzen-kdab, line 258: move `{` to a new line like above

crates/cxx-qt-gen/src/writer/cpp/mod.rs
```diff
@@ -507,6 +535,15 @@ mod tests {
           // setter
         }
 
+        void MyObject::privateMethod() const {
```
ahayzen-kdab, line 538: move `{` to new line

crates/cxx-qt-gen/src/writer/cpp/mod.rs
```diff
@@ -594,25 +618,22 @@ mod tests {
           // setter
         }
 
+        void SecondObject::privateMethod() const {
```
ahayzen-kdab, line 621: move `{` to new line

crates/cxx-qt-gen/src/writer/cpp/mod.rs
```diff
@@ -666,6 +687,15 @@ mod tests {
           // setter
         }
 
+        void MyObject::privateMethod() const {
```
ahayzen-kdab, line 690: move `{` to new line

crates/cxx-qt/src/lib.rs
```diff
@@ -76,3 +76,120 @@ pub trait Threading: Locking + Sized {
...
+
+/// This trait can be implemented inside the [`#[cxx_qt::bridge]`](bridge) on any [CxxQtType] to define a
+/// custom constructor in C++ for the QObject.
+///
+/// The `Arguments` must be a tuple of CXX types that will be the arguments to the constructor in C++.
+///
+/// Note that implementing this trait will stop Cxx-Qt from generating a default constructor.
+///
+/// # Minimal Example
+/// ```
```
ahayzen-kdab, line 88: put `rust` after the ``` ?

crates/cxx-qt/src/lib.rs
```diff
@@ -76,3 +76,120 @@ pub trait Threading: Locking + Sized {
...
+///     }
+/// }
+///
+/// # // Note that we need a fake main function for doc tests to build.
+/// # fn main() {}
+/// ```
+///
+/// # Pseudo Code for generated C++ Constructor
+/// You can imagine this trait as creating a constructor roughly like this:
+/// ``` cpp
```
ahayzen-kdab, line 128: remove the space before cpp ?

ahayzen-kdab:
In general i think this looks good now, just a few comments left :-)

crates/cxx-qt/src/lib.rs
Be-ing: not needed; rustdoc presumes code blocks are Rust unless specified otherwise

crates/cxx-qt-gen/src/generator/rust/constructor.rs
LeonMatthesKDAB: I've created an issue in CXX for this.
https://github.com/dtolnay/cxx/issues/1245

This can't be PhantomData I believe, as that wouldn't fix the underlying issue.
In C++ an "empty" struct is actually still 1 byte in size for *reasons*.
In Rust it's actually 0 bytes. PhantomData is 0 bytes as well, so wouldn't change that.
So the memory layout would be different between them without this notEmpty member

An `i8` should be the same size as a `bool` as the minimum addressable size in Rust is one byte.

crates/cxx-qt-gen/src/writer/cpp/mod.rs
```diff
@@ -143,6 +143,23 @@ mod tests {
                                 "Q_SIGNAL void toggleChanged();".to_owned(),
                             ),
                         ],
+                        private_methods: vec![CppFragment::Pair{
+                                header: "void privateMethod() const;".to_owned(),
+                                source: indoc! {r#"
+                                    void MyObject::privateMethod() const {
```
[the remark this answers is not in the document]
LeonMatthesKDAB: I believe I added this on purpose.
The reason is that the tests should actively assert what happens with anything that's differently formatted.
I.e. that we're NOT using clang-format in the writer.
The test is doing the same thing with Return values, there are versions where the `void` is on the same, as well as the previous line.
The writer shouldn't change what's generated.

crates/cxx-qt-gen/src/writer/cpp/mod.rs
LeonMatthesKDAB: Answered, see previous comment.

crates/cxx-qt-gen/src/parser/constructor.rs
```diff
@@ -0,0 +1,158 @@
...
+    /// Arguments to the new function.
+    /// The `new` function needs to return the inner Rust struct for the QObject.
+    pub new_arguments: Option<Vec<Type>>,
+    /// Arguments to be passed to the base class constructor.
+    pub base_arguments: Option<Vec<Type>>,
+    /// Arguments to the initialize function.
+    /// The `initialize` function is run after the QObject is created.
+    pub initialize_arguments: Option<Vec<Type>>,
+
+    pub imp: ItemImpl,
```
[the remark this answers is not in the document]
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/writer/cpp/mod.rs
LeonMatthesKDAB: Already answered in the first comment :)

crates/cxx-qt-gen/src/writer/cpp/mod.rs
LeonMatthesKDAB: Already answered in the first comment :)

crates/cxx-qt/src/lib.rs
LeonMatthesKDAB: I agree with @Be-ing , `rust` is the default, so no need to specify

crates/cxx-qt/src/lib.rs
LeonMatthesKDAB: Done.

Also updated the doc comment to reflect the change to a declarative API.

PR #574, opened 2023-06-02
use --locked argument with `cargo install`

Be-ing:
https://github.com/mozilla/sccache/issues/1786


ahayzen-kdab:
Will be interesting what happens to Windows here, Qt 5 might fail to build and Qt 6 might want to rebuild vcpkg (but needs to happen in main to push to packages)...

Be-ing:
Abandoning in favor of #576

PR #575, opened 2023-06-02
Remove newCppObject method for now.

LeonMatthesKDAB:
Will probably be readded as part of the Constructor trait later.

This wasn't documented anywhere and isn't used by anything at the moment.

### DIFF
--- a/crates/cxx-qt-gen/src/writer/cpp/header.rs
+++ b/crates/cxx-qt-gen/src/writer/cpp/header.rs
@@ -86,17 +86,11 @@ fn qobjects_header(generated: &GeneratedCppBlocks) -> Vec<String> {
             static_assert(::std::is_base_of<QObject, {ident}>::value, "{ident} must inherit from QObject");
             {namespace_end}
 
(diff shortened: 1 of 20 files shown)


crates/cxx-qt-gen/src/writer/cpp/header.rs
```diff
@@ -86,17 +86,11 @@ fn qobjects_header(generated: &GeneratedCppBlocks) -> Vec<String> {
             static_assert(::std::is_base_of<QObject, {ident}>::value, "{ident} must inherit from QObject");
             {namespace_end}
 
-            namespace {namespace_internals} {{
```
ahayzen-kdab, line 89: probably add an entry in the `CHANGELOG.md` saying it was removed, incase anyone was using it.

LeonMatthesKDAB: Done

ahayzen-kdab (approved):
LGTM :-)

PR #579, opened 2023-06-07
cxx-qt-gen: syn 2 supports unsafety in ItemForeignMod

ahayzen-kdab:
We do not need to consider verbatim code paths for parsing unsafe extern "ABI" anymore, this is supportted in syn 2.

eg compare ItemForeignMod and notice Option<Unsafe>
syn 1 https://docs.rs/syn/1.0.109/syn/struct.ItemForeignMod.html
syn 2 https://docs.rs/syn/2.0.0/syn/struct.ItemForeignMod.html

### DIFF
--- a/crates/cxx-qt-gen/src/parser/cxxqtdata.rs
+++ b/crates/cxx-qt-gen/src/parser/cxxqtdata.rs
@@ -3,13 +3,13 @@
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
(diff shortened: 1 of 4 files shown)


Be-ing (approved):

LeonMatthesKDAB:
Hehe, something tells me dtolnay reaaaally wanted that feature as well :D 

Great to see that simplified :+1:

PR #580, opened 2023-06-07
cxx-qt-gen: use extern "RustQt" block for inherit and signals

### LINKED ISSUE #557: Move `#[signal]` and `#[inherit]` to `extern "RustQt"` block
See target API in #555 

This would unify use of `inherit`, and make it the same syntax, no matter whether you're inheriting a normal function or a signal.
Also, signals would no longer be an enum, but rather just functions, which also makes our code generation a lot simpler.

ahayzen-kdab:
Requires #579 

Closes #557


LeonMatthesKDAB (approved):
Actually, this change is cleaner that I thought it'd be.

Issue #559, opened 2023-05-29
Move QObject type & property into `extern "RustQt"`


LeonMatthesKDAB:
As discussed in #555 

- [x] Move the struct definition outside the bridge
- [x] Use a declaration inside the `extern "RustQt"` block.

ahayzen-kdab:
Note that once there is a `type MyObjectQt = super::MyObject;` or whatever syntax that ends up with.

The ForeignItemFn in extern "RustQt" can be just passed through in the generator rust phase to the inner extern "Rust" block, as the self type is then correct. We don't need to store the self type and rebuild the function.

ahayzen-kdab:
Although note we still have the wrapper fun coming from C++ -> Rust, so maybe this can't be removed for invokables.

For inherit and signals going the other way though this should be possible.

ahayzen-kdab:
First step is to move the `qproperty` attribute from the fields onto the struct.

Then once that is done, we can move the struct outside the bridge and type alias with the qproperty attribute inside.

ahayzen-kdab:
We might be able to turn the parser phase here into something more like a continuous ParseStream, but would be a larger refactor.

ahayzen-kdab:
This is done now.

PR #583, opened 2023-06-12
cxx-qt-gen: do not generate getter/setter for fields anymore

ahayzen-kdab:
As we move qproperty to an attribute we need to disable generating helpers for fields as we won't be able to find them when the struct is outside the bridge.

There are three options going forwards
- let the developer create their own getters/setters (this commit)
- add a #[field(T, NAME)] as an attribute to the qobject later
- remove the (un)safe for qproperty with notify so that rust_mut is safe

Related to #559

### DIFF
--- a/crates/cxx-qt-gen/src/parser/qobject.rs
+++ b/crates/cxx-qt-gen/src/parser/qobject.rs
@@ -99,7 +99,7 @@ impl ParsedQObject {
 
         // Parse any properties in the struct
         // and remove the #[qproperty] attribute
-        let (properties, fields) = Self::parse_struct_fields(&mut qobject_struct.fields)?;
+        let (properties, _) = Self::parse_struct_fields(&mut qobject_struct.fields)?;
 
         // Ensure that the QObject is marked as pub otherwise the error is non obvious
         // https://github.com/KDAB/cxx-qt/issues/457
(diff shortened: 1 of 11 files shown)


examples/demo_threading/rust/src/lib.rs
```diff
@@ -112,16 +112,16 @@ mod ffi {
             let sensors_changed = Arc::new(AtomicBool::new(false));
 
             // Make relevent clones so that we can pass them to the threads
-            let accumulator_sensors = Arc::clone(self.as_mut().sensors_map_mut());
+            let accumulator_sensors = Arc::clone(&unsafe { self.as_mut().rust_mut() }.sensors_map);
```
ahayzen-kdab, line 115: note how this all becomes "unsafe" when it shouldn't be. And if `rust_mut()` was safe the API would still be quite nice, as it'd be saying just, I need to mutate the rust side then work on the field as normal.

crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -119,7 +119,11 @@ impl ParsedQObject {
             inherited_methods: vec![],
             passthrough_impl_items: vec![],
             properties,
-            fields,
+            // Do not generate helpers for fields as they are going to move out of the bridge
```
ahayzen-kdab, line 122: needs changelog update

ahayzen-kdab: done

examples/qml_features/rust/src/custom_base_class.rs
```diff
@@ -150,6 +150,11 @@ pub mod ffi {
                 self.as_mut().end_remove_rows();
             }
         }
+
+        /// Safe helper to access mutate the vector field
+        pub fn vector_mut<'a>(self: Pin<&'a mut Self>) -> &'a mut Vec<(u32, f64)> {
```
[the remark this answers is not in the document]
ahayzen-kdab: #585 tries this out

ahayzen-kdab:
Think we are going with this + removing the unsafeness. So closing in favour of #585

PR #585, opened 2023-06-13
cxx-qt-gen: change rust_mut to be safe and remove field generation

ahayzen-kdab:
Related to #559 

Related to #583


examples/qml_features/rust/src/containers.rs
```diff
@@ -139,7 +144,7 @@ impl ffi::RustContainersQt {
     fn insert_map(mut self: Pin<&mut Self>, key: QString, value: QVariant) {
         // SAFETY: map is a Q_PROPERTY so ensure we manually trigger changed
         unsafe {
```
ahayzen-kdab, line 146: should the property mutable getter be unsafe still ? as you can bypass it with `rust_mut().property = ...`

ahayzen-kdab: Or do we even need the property mutable getter ? Instead just have

- `property(&self) -> T`
- `set_property(self: Pin<&mut Self>, value: T)`
- `rust().property` (`&T`)
- `rust_mut().property` (`&mut T`)
- `property_changed(self: Pin<&mut Self>)`

Then if you go through `rust_mut()` it's up to the developer to call `property_changed` themselves.

crates/cxx-qt-gen/src/generator/rust/property/mod.rs
```diff
@@ -174,9 +157,7 @@ mod tests {
                         if self.rust().trivial_property == value {
```
ahayzen-kdab, line 157: we could implement Deref quite easily which allows for `.rust()` to go here. DerefMut gets confused with the Pin stuff....

ahayzen-kdab: `Deref` fun is here #586

PR #586, opened 2023-06-21
cxx-qt-gen: impl Deref for qobject::T to reach T easily

ahayzen-kdab:
Related to #559

Requires #585

### DIFF
--- a/crates/cxx-qt-gen/src/generator/rust/property/getter.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/property/getter.rs
@@ -43,7 +43,7 @@ pub fn generate(
                     #[doc = "Getter for the Q_PROPERTY "]
                     #[doc = #ident_str]
                     pub fn #getter_rust(&self) -> &#ty {
-                        &self.rust().#ident
+                        &self.#ident
                     }
                 }
             },
--- a/crates/cxx-qt-gen/src/writer/rust/mod.rs
+++ b/crates/cxx-qt-gen/src/writer/rust/mod.rs
@@ -35,19 +35,30 @@ fn cxx_qt_common_blocks(qobject: &GeneratedRustQObject) -> Vec<TokenStream> {
     let cpp_struct_ident = &qobject.cpp_struct_ident;
     let rust_struct_ident = &qobject.rust_struct_ident;
 
-    vec![quote! {
-        impl cxx_qt::CxxQtType for #cpp_struct_ident {
-            type Rust = #rust_struct_ident;
+    vec![
+        quote! {
+            impl core::ops::Deref for #cpp_struct_ident {
+                type Target = #rust_struct_ident;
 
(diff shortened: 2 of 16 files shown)


crates/cxx-qt-gen/src/writer/rust/mod.rs
```diff
@@ -58,6 +58,15 @@ fn cxx_qt_common_blocks(qobject: &GeneratedRustQObject) -> Vec<TokenStream> {
     let create_rs_ident = mangle("create_rs", rust_struct_ident);
 
     vec![
+        quote! {
+            impl core::ops::Deref for #cpp_struct_ident {
```
ahayzen-kdab, line 62: note we can't quite do `DerefMut` as the `Pin<&mut T>` makes that not work as the C++ type is not `Unpin`.

crates/cxx-qt-gen/src/generator/rust/property/getter.rs
```diff
@@ -43,7 +43,7 @@ pub fn generate(
                     #[doc = "Getter for the Q_PROPERTY "]
                     #[doc = #ident_str]
                     pub fn #getter_rust(&self) -> &#ty {
-                        &self.rust().#ident
```
LeonMatthesKDAB, line 46: Hm, it seems like these getters have become kind of pointless in Rust at least.
We do still use them in C++ though, so maybe something to think about removing later.

LeonMatthesKDAB (approved):
Overall a very welcome change, there are probably a few more `as_ref` calls that we may have missed, but that's not a big deal.

tests/basic_cxx_qt/rust/src/locking.rs
```diff
@@ -42,33 +42,32 @@ pub mod ffi {
...
 // TODO: this will change to qobject::RustLockingDisabled once
 // https://github.com/KDAB/cxx-qt/issues/559 is done
 impl ffi::RustLockingDisabledQt {
     fn get_counter(&self) -> u32 {
-        self.rust().counter.load(Ordering::Acquire)
+        self.counter.load(Ordering::Acquire)
     }
 
     fn increment(self: Pin<&mut Self>) {
         let counter = self.as_ref().get_counter();
```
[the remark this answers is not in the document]
ahayzen-kdab: Done

crates/cxx-qt-gen/src/generator/rust/property/getter.rs
ahayzen-kdab: Yup, but we still need something to bind the C++ to for it's getter, as we can't use member in C++ etc.

Interestingly we can remove the "wrapper" getter we have via the same solution as #589 if it's accepted upstream :-)

LeonMatthesKDAB (approved):
Looks like a great ergonomic change

PR #587, opened 2023-06-21
cxx-qt-gen: don't import Pin in hidden module resolving IDE integration

### LINKED ISSUE #417: Auto-imported Pin breaks IDE integration
It looks like https://github.com/KDAB/cxx-qt/commit/5797eb75fcd88e453cff976e2dc74b66e1d556e7 added Pin as an implicit import. However, this means that code that manually imports `std::pin::Pin` will fail:

```
error[E0252]: the name `Pin` is defined multiple times
 --> src/bridge.rs:4:9
  |
1 | #[cxx_qt::bridge]
  | ----------------- previous import of the type `Pin` here
...
4 |     use std::pin::Pin;
  |         ^^^^^^^^^^^^^ `Pin` reimported here
  |
  = note: `Pin` must be defined only once in the type namespace of this module
```

Meanwhile, if I *don't* import Pin, then IntelliJ Rust wouldn't be able to resolve it:

image

This is a nuisance because it keeps red lines in my code when it's not relevant, and it also breaks code completion to some extent.

Perhaps, it would be better to keep generated imports under an alias that doesn't collide (e.g. `use std::pin::Pin as _Pin`)?

ahayzen-kdab:
Now that impls are outside of the bridge the developer needs to import Pin anyway.

Closes #417

Requires #553

### DIFF
--- a/crates/cxx-qt-gen/src/generator/rust/signals.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/signals.rs
@@ -43,10 +43,13 @@ pub fn generate_rust_signals(
             })
             .collect::<Vec<TokenStream>>();
 
-        let self_type = if signal.mutable {
-            quote! { Pin<&mut #qobject_name> }
+        let (self_type, self_type_cxx) = if signal.mutable {
+            (
+                quote! { core::pin::Pin<&mut #qobject_name> },
+                quote! { Pin<&mut #qobject_name> },
+            )
         } else {
-            quote! { &#qobject_name }
+            (quote! { &#qobject_name }, quote! { &#qobject_name })
         };
 
         let mut unsafe_block = None;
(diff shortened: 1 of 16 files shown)


crates/cxx-qt-gen/src/generator/rust/signals.rs
```diff
@@ -43,10 +43,13 @@ pub fn generate_rust_signals(
             })
             .collect::<Vec<TokenStream>>();
 
-        let self_type = if signal.mutable {
-            quote! { Pin<&mut #qobject_name> }
+        let (self_type, self_type_cxx) = if signal.mutable {
+            (
+                quote! { core::pin::Pin<&mut #qobject_name> },
```
ahayzen-kdab, line 48: once #553 is ready we'll have a method we can just pass types through to auto add fully qualified syntax.

Be-ing (approved):

Issue #404, opened 2023-01-09
Support Result<T, E> as a return type from Rust -> C++


ahayzen-kdab:
https://cxx.rs/binding/result.html#returning-result-from-rust-to-c

We need to add support for returning a Result in invokables, this is tricky as the CXX and method types are different. Then Result is only supported in a return type and not a parameter (?).

Consider implementing the API ideas in https://github.com/dtolnay/cxx/issues/1144 as part of this.

So

- [ ] Have a separate cxx_utils module within cxx-qt-gen (which could then be made a crate)
- [ ] Change the generator/cpp/types code into modular pieces
- [ ] Move generator/rust/types into the same module
- [ ] Add a separate method for when the type is a return type to support `Result<T, E>`

Issue #20, opened 2021-12-08
Rename C++ members and methods to avoid collisions


ahayzen-kdab:
Rename C++ members so that more Q_PROPERTY names are available eg use m_internalRustObj, m_propertyObj, m_mutexInternalRustObj, m_mutexPropertyObj etc

So there could be the following prefixes
- internal
- property
- mutex

Also consider if there are any names we cannot avoid collisions and we should error if they are used in Rust (and if this should move to another issue).

ahayzen-kdab:
We also need to consider how to avoid collisions on the Rust side eg for wrappers, a separate struct is possible but tricky as C++ would need to call through that to the actual object.

Instead consider using a prefix of `cxx_qt_` on our helper methods, with the possibility to override the prefix name in the `build.rs` like the C++ namespace.

cc @LeonMatthesKDAB

ahayzen-kdab:
Also note that the task of ensuring all the C++ methods are namespaced is linked to avoiding collisions #14

LeonMatthesKDAB:
> We also need to consider how to avoid collisions on the Rust side eg for wrappers, a separate struct is possible but tricky as C++ would need to call through that to the actual object.
> 
> Instead consider using a prefix of `cxx_qt_` on our helper methods, with the possibility to override the prefix name in the `build.rs` like the C++ namespace.
> 
> cc @LeonMatthesKDAB

+1 for using a prefix like `cxx_qt_` on our generated code.
That would also make it easier to see in a stack-trace, where cxx-qt ends and user code starts.

ahayzen-kdab:
This is becoming increasingly less relevant, as methods now map to the rust ones. We might have a checks phase which could ensure that properties and invokables or reserved names don't collide - but we are also trying to lean on the compiler. For now lets leave this open and review in the next cycles if anything is needed.

ahayzen-kdab:
Moving our `invokableWrapper` in C++ to private + named better could also help.

LeonMatthesKDAB:
That doesn't seem to be a problem anymore, as we have barely anything inside the C++ side of things.

PR #589, opened 2023-06-22
cxx-qt-gen: remove wrapper method for C++ -> Rust invokables

ahayzen-kdab:
This then avoids us needing to generate Rust methods with fully qualified types on the Rust side and removes a load of generation.

Related to #404

### DIFF
--- a/crates/cxx-qt-gen/src/generator/cpp/invokable.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/invokable.rs
@@ -67,11 +67,11 @@ pub fn generate_cpp_invokables(
             .collect::<Result<Vec<CppNamedType>>>()?;
 
         let body = format!(
-            "m_rustObj->{ident}({parameter_names})",
+            "{ident}({parameter_names})",
             ident = idents.wrapper.cpp,
-            parameter_names = vec!["*this"]
-                .into_iter()
-                .chain(parameters.iter().map(|parameter| parameter.ident.as_str()))
+            parameter_names = parameters
+                .iter()
+                .map(|parameter| parameter.ident.as_str())
                 .collect::<Vec<&str>>()
                 .join(", "),
         );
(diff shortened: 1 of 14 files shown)


crates/cxx-qt-gen/src/generator/cpp/invokable.rs
```diff
@@ -140,6 +140,25 @@ pub fn generate_cpp_invokables(
                 },
             },
         });
+
+        // Note that we are generating a header to match the extern "Rust" method
+        // in Rust for our invokable.
+        //
+        // CXX generates the source and we just need the matching header.
+        //
+        // TODO: this could be private ?
```
ahayzen-kdab, line 149: This could be sorted out as part of #20 and in the changes with #566 might help.

crates/cxx-qt-gen/src/generator/cpp/invokable.rs
```diff
@@ -140,6 +140,25 @@ pub fn generate_cpp_invokables(
...
+        //
+        // TODO: will this always be noexcept ? If the return type is Result is this then removed?
```
ahayzen-kdab, line 151: This could be solved in #553 separately as we don't support Result yet anyway.

crates/cxx-qt-gen/src/generator/rust/invokable.rs
```diff
@@ -50,46 +45,29 @@ pub fn generate_rust_invokables(
...
-            .parameters
-            .iter()
-            .map(|parameter| parameter.ident.clone())
-            .collect::<Vec<Ident>>();
-
         let fragment = RustFragmentPair {
             cxx_bridge: vec![quote! {
-                // TODO: is an unsafe block valid?
+                // Note: extern "Rust" block does not need to be unsafe
                 extern "Rust" {
```
ahayzen-kdab, line 61: we should confirm that `extern "Rust"` with an opaque C++ type and it's behaviour we are using here is valid upstream and ensure there is a test.

ahayzen-kdab:
Note if we use this trick, the same can be used for getters / setters etc for properties.

ahayzen-kdab:
After this change we could potentially pass through the invokable extern function blocks.

crates/cxx-qt-gen/test_outputs/invokables.h
```diff
@@ -27,15 +27,25 @@ class MyObject : public QObject
 
 public:
   Q_INVOKABLE void invokable() const;
+  void invokableWrapper() const noexcept;
```
ahayzen-kdab, line 30: should / can these be private ?

Be-ing (approved):

PR #593, opened 2023-06-26
cxx-qt-lib: use +whole_archive for std_types so that cargo-only works

### LINKED ISSUE #592: Cargo only example fails in `main` with Qt 5 due to std types confusion
```bash
$ cargo run -p qml-minimal-no-cmake
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/qml-minimal-no-cmake`
QQmlApplicationEngine failed to load component
qrc:/main.qml:26:17: Invalid property assignment: unsupported type ""
```

Using `main` the cargo-only build fails to run with Qt 5 due to `std_types.cpp` not occurring, it however works with CMake.

This is likely due to `+whole-archive` missing ?

ahayzen-kdab:
When building with cargo we need to ensure that the statics aren't optimised out for registering types with Qt 5.
    
Otherwise we cannot use numbers in QML.
    
Closes #592


crates/cxx-qt-build/src/lib.rs
```diff
@@ -506,6 +507,26 @@ impl CxxQtBuilder {
...
+        // as they are stored in statics in the source.
+        if qtbuild.version().major == 5 {
+            let std_types_contents = include_str!("std_types_qt5.cpp");
+            let std_types_path = format!("{out_dir}/std_types_qt5.cpp");
+            let mut source =
+                File::create(&std_types_path).expect("Could not create std_types source");
+            write!(source, "{std_types_contents}").expect("Could not write std_types source");
+
+            cc_builder_whole_archive.file(&std_types_path);
+            // TODO: this line shouldn't be required,
```
ahayzen-kdab, line 524: this shouldn't be required, needs investigating

ahayzen-kdab: Actually in CMake builds double up whole_archive parts too.

Eg if you remove `qml/qml.qrc` from `CMakeLists.txt` and add it to the `build.rs` file, then it only works with cargo-only builds and not CMake builds.

So having `std_types.cpp` twice to get cargo-only and cmake-only builds both working is not that different and maybe OK for now.

ahayzen-kdab: We maybe need to investigate separately why the `qt-static-initializers` is not included in the final binary with CMake. It kinda feels like either the other cc builder needs to link to it or CMake itself needs to link to it.

ahayzen-kdab: Although looks like the `compile` method of `cc::Build` should be causing the `cargo:rustc-link-lib` to be set, so maybe something else is going wrong when CMake links this.

ahayzen-kdab: Digging further into this it seems to be a general issue with main as stated before the same happens with qrc files.

Things which are built into `libqt-static-initializers.a` do not end up in the staticlib from cargo `libqml_minimal.a`, but things from `libcxx-qt-gen.a` do (even though the two libraries are built in very similar ways). Note that when cargo builds to an executable they do however remain.

It seems that cargo/rustc itself is doing something odd when compiling to a staticlib rather than an executable which means that the first library is lost / optimised out.

Note that swapping around the order of the `compile(...)` of the `cc::Build` causes the build to fail as it can't find things that are likely in the first library and missing.

So either we need to figure out how to cause cargo/rustc not to lose the non-last library or we need to move everything to being in one lib that is WHOLE_ARCHIVE

ahayzen-kdab: So now this branch also fixes that qrc and QML modules wouldn't have worked with CMake builds if you put them only in the rust build script and not the CMakeLists.

The question now becomes is, should we just always build with whole_archive rather than having this split in the rust side ? As CMake builds link with whole_archive anyway?

Or can we improve the Rust side when building to a staticlib to not remove things?

ahayzen-kdab: This doesn't work with cargo-only builds as you get duplicate implementations :see_no_evil: I'll try creating another branch which removes the double builder and tries that route...

ahayzen-kdab: This has separate issues, i think this should be discussed in another issue / pull request as `main` suffers from this too.

ahayzen-kdab: Split this into #597

crates/cxx-qt-build/src/lib.rs
```diff
@@ -424,11 +425,29 @@ impl CxxQtBuilder {
...
+                //
+                // However, we need to not have -bundle for qt-static-initializers to work
+                // with CMake builds, otherwise the statement below occurs where it's missing
+                // from the final binary.
+                //
+                // When building a staticlib -bundle means that the native static library
+                // is simply not included into the archive and some higher level build
+                // system will need to add it later during linking of the final binary.
+                // https://doc.rust-lang.org/rustc/command-line-arguments.html#option-l-link-lib
+                panic!("rustc 1.69.x is not supported with CXX-Qt");
```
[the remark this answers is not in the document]
ahayzen-kdab: done

LeonMatthesKDAB (dismissed):
LGTM

ahayzen-kdab:
Note that i tried to create a multi crate project, but this doesn't work as you run into 
```
error: link modifiers combination `+bundle,+whole-archive` is unstable when generating rlibs
```

I tried to have a project where i had the following crate structure each with their own QObject and cxx-qt-build, and then main using public methods from sub1 and sub2.

```
main (staticlib)
  - sub1 (rlib)
  - sub2 (rlib)
```

crates/cxx-qt-build/src/lib.rs
```diff
@@ -11,6 +11,8 @@
 //! for CXX-Qt or CXX macros and generate any resulting C++ code. It also builds
 //! the C++ code into a binary with any cxx-qt-lib code and Qt linked.
 
+extern crate version_check as rustc;
```
Be-ing, line 14: I find this aliasing of the crate confusing. It makes it seem like the `version_check` crate is something shipped with rustc, but it's a normal dependency from crates.io.

crates/cxx-qt-build/src/lib.rs
```diff
@@ -506,6 +525,25 @@ impl CxxQtBuilder {
...
+        // Note that we need this to be compiled into the whole_archive builder
+        // as they are stored in statics in the source.
+        if qtbuild.version().major == 5 {
+            let std_types_contents = include_str!("std_types_qt5.cpp");
+            let std_types_path = format!(
+                "{out_dir}/std_types_qt5.cpp",
+                out_dir = env::var("OUT_DIR").unwrap()
+            );
+            let mut source =
+                File::create(&std_types_path).expect("Could not create std_types source");
```
Be-ing, line 541: Do we need to do this elaborate copying of the C++ file? Can we just compile the C++ file directly without copying it first?

crates/cxx-qt-build/src/lib.rs
ahayzen-kdab: Oops yup, i just copied it from their readme. Changed it to `version_check`.

crates/cxx-qt-build/src/lib.rs
ahayzen-kdab: Not sure we can, as the `CARGO_MANIFEST_DIR` turns out to be the example rather than the cxx-qt-build crate ?

Once we have support for +bundle and +whole-archive then we should be able to put this in `cxx-qt` then we should be able to remove the copying. Unless there is another way...

ahayzen-kdab: Even using `file!()` doesn't give you an absolute path so it'd be quite tricky to do i think ?

ahayzen-kdab: Left a comment saying that this can be improved once we have +bundle and +whole-archive.

Be-ing (approved):

Issue #590, opened 2023-06-25
qml_minimal example fails to bulid due to "byte index 1 is out of bounds of ``"


spacekookie:
Hey there,

sorry if this is the wrong channel to ask questions. I was looking around for a Matrix/IRC/Discord channel in the documentation but couldn't find anything.

I'm new to both cxx-qt and qt development, so apologies also if this is an obvious problem :D 

I'm trying to build the `qml_minimal` example from the cxx-qt repository. I'm using a NixOS shell environment to install the build requirements into my environment (`nix-shell -p rustc cargo qt5.full mold clang cmake` - it also happens with the `qt6.full` package). I don't know if there's something missing from it though.

I try building the `qml-minimal` example via this command:

`cargo run -p qml-minimal-no-cmake`

And here's the full stacktrace of my build (it looks like the cxx macro fails?): https://paste.rs/GBOwa.txt

Anyway, I hope you can give me some pointers as to what is wrong on my system/ what I'm missing to make this example build. Thanks in advance!

ahayzen-kdab:
Thanks for taking to the time to report a bug, here is fine for questions we don't have any chat channels yet, or a discussion in github is fine for non-bugs :-) 

This error does look curious, it seems to suggest that one of our `.rs` files with CXX bridges is failing to be parsed by CXX / syn / somewhere else.

I assume that it is stopping on these block and not getting any futher? https://github.com/KDAB/cxx-qt/blob/386a7aaa2d24b68fd29b333b1357aed778e43089/crates/cxx-qt-lib/build.rs#L183-L184

Could you try commenting out some of the `rust_bridges` (in the code block below), to see if it is one specific `.rs` file that is causing an issue ?

Otherwise I can try to get a Nix environment setup to see if the issue is reproducible :-) 

I also assume you are trying this from `main` ? Please also note that we are going through a slight API reshuffle so `main` is halfway between what is documented in the book and what is in #555. Also note that `main` has a regression for Qt 5 builds meaning that cargo-only builds will have issues with numbers in QML #592.

https://github.com/KDAB/cxx-qt/blob/386a7aaa2d24b68fd29b333b1357aed778e43089/crates/cxx-qt-lib/build.rs#L30-L175

ahayzen-kdab:
Trying this out with Qt 6 i get a different error with Nix, due to the `.prl` file missing / in a different place.

```bash
[...]

PR #594, opened 2023-06-26
qt-build-utils: default to qmake, qmake6, qmake-qt5

ahayzen-kdab:
When using Nix with Qt 6 the output of qmake6 -query is incorrect and ends up pointing to the incorrect locations. But qmake -query points to the correct ones.

This solves the issue for Nix but looks like Debian only has qmake for Qt 5 and qmake6 for Qt 6.

Related to #590

### DIFF
--- a/crates/qt-build-utils/src/lib.rs
+++ b/crates/qt-build-utils/src/lib.rs
@@ -254,7 +254,7 @@ impl QtBuild {
         }
 
         // Fedora 36 renames Qt5's qmake to qmake-qt5
-        let candidate_executable_names = ["qmake6", "qmake-qt5", "qmake"];
+        let candidate_executable_names = ["qmake", "qmake6", "qmake-qt5"];
         for (index, executable_name) in candidate_executable_names.iter().enumerate() {
             match verify_candidate(executable_name) {
                 Ok((executable_name, version)) => {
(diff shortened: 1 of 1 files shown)


ahayzen-kdab:
> I guess this would still work with Debian though, it would just choose Qt 5 by default. So IMHO we can go ahead with this, as this kind of expresses the intent pretty well:
> 
>     1. Choose whatever "default" Qt installation has been chosen
> 
>     2. Try Qt6
> 
>     3. Try Qt5
> 
> 
> However, we may need to ensure this defaulting makes sense when a specific QT_VERSION_MAJOR has been provided.

So lets pretend that

```
qmake -> Qt 5
qmake6 -> Qt 6
```

If you specify QT_VERSION_MAJOR=6, what we want to happen is for it to try `qmake`, go hey the version is wrong, and continue looking. (which it kinda looks like it does already?).

If you specify QT_VERSION_MAJOR=5, what we want to happen is for it to try `qmake`, then go hey that's the right version. (which again kinda looks like it might be doing that, need to check).

ahayzen-kdab:
Lets report this upstream to Nix first and see what they say...

Be-ing:
I agree. I'm not inclined to hack around a packaging bug that only affects one distro and risk regressions on other distros when it's clearly a [bug](https://github.com/NixOS/nixpkgs/issues/239971).

Be-ing:
The bug was fixed upstream in Nix. Shall we close this?

ahayzen-kdab:
> The bug was fixed upstream in Nix. Shall we close this?

Yup, it may take a while for it to flow from nixpkgs unstable through to a stable release, but it's fixed.

And for now if anyone else gets stuck on this just set your `QMAKE` env var to point to `qmake` until `qmake6` is fixed.

PR #596, opened 2023-06-27
cxx-qt-build: enable parallel feature of cc crate

ahayzen-kdab:
This allows for parallel compilation of C/C++ code.

### DIFF
--- a/crates/cxx-qt-build/Cargo.toml
+++ b/crates/cxx-qt-build/Cargo.toml
@@ -13,7 +13,7 @@ description = "Helpers for integrating `cxx-qt` crate into a Cargo build"
 repository.workspace = true
 
 [dependencies]
-cc = "1.0.74"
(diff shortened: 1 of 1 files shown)


crates/cxx-qt-build/Cargo.toml
```diff
@@ -13,7 +13,7 @@ description = "Helpers for integrating `cxx-qt` crate into a Cargo build"
 repository.workspace = true
 
 [dependencies]
-cc = "1.0.74"
+cc = { version = "1.0.74", features = ["parallel"] }
```
ahayzen-kdab, line 16: We had this before, then we reverted it or it was lost ? #220 was there a reason ? eg Windows file locks or something ?

ahayzen-kdab: Oh no that's just `cxx-build` :-) So maybe we just missed enabling it for our own `cc` crate

PR #598, opened 2023-06-28
WIP: examples: create a multi project to test multiple crates

ahayzen-kdab:


examples/meta_project/rust/sub1/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+version = "0.1.0"
+authors = [
+  "Andrew Hayzen <andrew.hayzen@kdab.com>",
+]
+edition = "2021"
+license = "MIT OR Apache-2.0"
+
+# This will instruct Cargo to create an rlib our main crate is the staticlib
+[lib]
+crate-type = ["rlib"]
```
ahayzen-kdab, line 16: this fails with
```
error: link modifiers combination `+bundle,+whole-archive` is unstable when generating rlibs
```

Be-ing: I tested with nightly and this error still occurs.

Be-ing:
Blocked on https://github.com/rust-lang/rust/issues/108081

examples/meta_project/rust/sub1/Cargo.toml
petrochenkov: You need `#![feature(packed_bundled_libs)]` to ungate it if you can use nightly.

ahayzen-kdab: Thanks! With the nightly compiler and that feature enabled this works :tada: 

Now the question is whether this will ever become stable, whether there is a timeline for this, and if we can help in any way.

Be-ing:
Upstream PR to make this work: https://github.com/rust-lang/rust/pull/113301

Be-ing:
The upstream PR has been merged and will be released in stable Rust 1.74, which is scheduled for 16 November 2023. I suggest to leave this as a draft until then.

ahayzen-kdab:
@Be-ing Awesome news, thanks! :-)

Be-ing:
Rust 1.74 has been released so this now builds with stable Rust! :partying_face:

ahayzen-kdab:
@Be-ing sweet! Great work on getting that stabilised into Rust :muscle: We look at getting this landed and the reshuffle internally in the next cycle I think as 0.6 is imminent with a deadline we need to reach.

ahayzen-kdab:
Seems to be getting a linking issue with double resources, need to investigate. Wonder if it's the QML module changes.

Be-ing:
> Seems to be getting a linking issue with double resources, need to investigate. Wonder if it's the QML module changes.

Fixed in https://github.com/KDAB/cxx-qt/pull/755

ahayzen-kdab:
Locally this builds now, but it fails to run as it can't find the sub QML QObjects, needs some further investigation or maybe a clean build.

Be-ing:
I opened a PR for your fork to get this working: https://github.com/ahayzen-kdab/cxx-qt/pull/4 The errors were simply from missing `#[qml_element]` attributes :)

ahayzen-kdab:
Seems macOS works, Windows has had a wobble with file locking / mapped stuff, Linux explodes using ld.gold.

Is this one of those, only lld works situations ?

Be-ing:
Seems we're continuing the CXX-Qt tradition of stumbling over the weirdest edge cases of toolchains :sweat_smile: 

Somehow Cargo does specify linking the same qt-static-initializers library twice with `--whole-archive`, so it makes sense why gold fails... but why does Cargo do that? And only does that for `cargo test`?? Perhaps a Cargo bug?

As for the Windows Qt5 build error:
```
 error: failed to write file '\\?\C:\cxx-qt\build\cargo\build\x86_64-pc-windows-msvc\release\deps\libsub1-ef2617c41816f8ec.rlib': The requested operation cannot be performed on a file with a user-mapped section open. (os error 1224)
 ```
 I don't know. Another cargo/rustc bug? A Windows bug?
 
 A quick hack around both of these problems would be replacing the `--all-targets` CLI option to `cargo test` in the root CMakeLists.txt with `--workspace --exclude qml-meta-project`. The qml-meta-project crate doesn't even have Cargo tests, so there's no much point building it in that context.

ahayzen-kdab:
> Seems we're continuing the CXX-Qt tradition of stumbling over the weirdest edge cases of toolchains 😅

:sweat_smile: 

Yeah, a switch around to ensure it builds (and maybe runs) would be useful, I'll probably look into this next month as I'm tight on time.

Be-ing:
I don't think it's worth deep diving into the root causes of these errors right now considering that qml-meta-project doesn't even have Cargo tests. So I think switching to `--workspace --exclude qml-meta-project` would be sufficient for finally merging this PR.

Be-ing:
Hopefully this will be enough to merge finally: https://github.com/ahayzen-kdab/cxx-qt/pull/5

ahayzen-kdab:
Split only running cachegen when there are QML files in #811

ahayzen-kdab:
Wonder if it's the Rust side that needs to not be parallel rather than the C++/CMake side, but still shouldn't happen.

ahayzen-kdab:
This will need further work and changes for the way QML modules are now passed around and as part of the 0.8 cycle improvements where a single crate could become a QML module.

PR #599, opened 2023-07-01
Pass the qobject uri's to the moc compiler with the -Muri option.

jimmyvanhest:
For some targets the moc compiler requires that the -Muri option should be passed for all qobjects. If this is not done the resulting qml application would not be able to find the qobject and will fail to load.

### DIFF
--- a/crates/cxx-qt-build/src/lib.rs
+++ b/crates/cxx-qt-build/src/lib.rs
@@ -496,7 +496,14 @@ impl CxxQtBuilder {
         let mut cc_builder_whole_archive_files_added = false;
         // Run moc on C++ headers with Q_OBJECT macro
         for qobject_header in self.qobject_headers {
-            let moc_products = qtbuild.moc(&qobject_header.path);
+            let moc_uri_opt = qobject_header
+                .qml_metadata
+                .iter()
+                .flat_map(|qml_metadata| ["-Muri=", &qml_metadata.uri, " "])
+                .flat_map(|s| s.chars())
+                .collect::<String>()
+                .trim_end();
+            let moc_products = qtbuild.moc(&qobject_header.path, &moc_uri_opt);
             self.cc_builder.file(moc_products.cpp);
             for qml_metadata in qobject_header.qml_metadata {
                 self.cc_builder.define("QT_STATICPLUGIN", None);
--- a/crates/cxx-qt-build/src/lib.rs
+++ b/crates/cxx-qt-build/src/lib.rs
@@ -496,14 +496,11 @@ impl CxxQtBuilder {
         let mut cc_builder_whole_archive_files_added = false;
         // Run moc on C++ headers with Q_OBJECT macro
         for qobject_header in self.qobject_headers {
-            let moc_uri_opt = qobject_header
+            let uris = qobject_header
                 .qml_metadata
                 .iter()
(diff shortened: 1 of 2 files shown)


crates/cxx-qt-build/src/lib.rs
```diff
@@ -478,7 +478,14 @@ impl CxxQtBuilder {
         let mut cc_builder_whole_archive_files_added = false;
         // Run moc on C++ headers with Q_OBJECT macro
         for qobject_header in self.qobject_headers {
-            let moc_products = qtbuild.moc(&qobject_header.path);
+            let moc_uri_opt = qobject_header
+                .qml_metadata
+                .iter()
+                .flat_map(|qml_metadata| ["-Muri=", &qml_metadata.uri, " "])
```
ahayzen-kdab, line 484: is the `" "` here trying to make a space between multiple uris or something ? So that you result in `-Muri=abc -Muri=def` ?

If so I think it would be easier to read if you built them to a list then joined with a " " ?

eg something like this seems more readable, unless i'm misunderstanding what is happening ?

```rust
            let moc_uri_opt = qobject_header
                .qml_metadata
                .iter()
                .flat_map(|qml_metadata| ["-Muri=", &qml_metadata.uri])
                .collect::<Vec<&str>>()
                .join(" ");
```

Something like

crates/qt-build-utils/src/lib.rs
```diff
@@ -450,7 +450,7 @@ impl QtBuild {
     /// Run moc on a C++ header file and save the output into [cargo's OUT_DIR](https://doc.rust-lang.org/cargo/reference/environment-variables.html).
     /// The return value contains the path to the generated C++ file, which can then be passed to [cc::Build::files](https://docs.rs/cc/latest/cc/struct.Build.html#method.file),
     /// as well as the path to the generated metatypes.json file, which can be passed to [register_qml_types](Self::register_qml_types).
-    pub fn moc(&mut self, input_file: impl AsRef<Path>) -> MocProducts {
+    pub fn moc(&mut self, input_file: impl AsRef<Path>, uri_opt: &str) -> MocProducts {
```
ahayzen-kdab, line 453: Could you add some documentation about what `uri_opt` does to the doc lines above?

Also assume `uri_opt` should always be mandatory and not a `Option<&str>` ? :thinking:

crates/cxx-qt-build/src/lib.rs
jimmyvanhest: You are correct that this was the intent. Your suggestion looks nice but would produce a slightly different output `-Muri= abc -Muri= def`(note the spaces).

I chose to do something I should have done from the start and just pass all uris to the moc function so it can figure out how to create the arguments required for the moc compiler. this would make it much easier to use this in the future.

jimmyvanhest: There is still a small issue when uris is an empty iterator. I will make a fix for that at a later time.

crates/qt-build-utils/src/lib.rs
jimmyvanhest: I added some documentation. I did not know how it should be documented so I chose something that looks good.

crates/cxx-qt-build/src/lib.rs
jimmyvanhest: should be fixed ow

Be-ing (approved):

PR #600, opened 2023-07-01
Prl file parsing fixes

jimmyvanhest:
The following three issues with prl file parsing have been fixed:
- previously QT_INSTALL_PREFIX would be treated the same as QT_INSTALL_LIBS which is not correct.
- some prl files names(mainly from android) would not follow the assumed patern.
- building for wasm requires that the libqwasm.prl must be included.

### DIFF
--- a/crates/qt-build-utils/src/lib.rs
+++ b/crates/qt-build-utils/src/lib.rs
@@ -421,6 +421,23 @@ impl QtBuild {
                 &prl_path,
             );
         }
+
+        let emscripten_targeted = match env::var("CARGO_CFG_TARGET_OS") {
+            Ok(val) => val == "emscripten",
+            Err(_) => false,
+        };
+        if emscripten_targeted {
+            let platforms_path = format!("{}/platforms", self.qmake_query("QT_INSTALL_PLUGINS"));
+            println!("cargo:rustc-link-search={platforms_path}");
+            self.cargo_link_qt_library(
+                builder,
+                "qwasm",
+                &prefix_path,
+                &lib_path,
+                "qwasm",
+                &format!("{platforms_path}/libqwasm.prl"),
+            );
+        }
     }
 
     /// Get the include paths for Qt, including Qt module subdirectories. This is intended
--- a/crates/qt-build-utils/src/lib.rs
+++ b/crates/qt-build-utils/src/lib.rs
@@ -317,15 +317,15 @@ impl QtBuild {
     ) {
         println!("cargo:rustc-link-lib={link_lib}");
 
-        match std::fs::read_to_string(&prl_path) {
+        match std::fs::read_to_string(prl_path) {
             Ok(prl) => {
                 for line in prl.lines() {
                     if let Some(line) = line.strip_prefix("QMAKE_PRL_LIBS = ") {
                         parse_cflags::parse_libs_cflags(
                             builder,
                             name,
-                            line.replace(r"$$[QT_INSTALL_LIBS]", &lib_path)
-                                .replace(r"$$[QT_INSTALL_PREFIX]", &prefix_path)
+                            line.replace(r"$$[QT_INSTALL_LIBS]", lib_path)
+                                .replace(r"$$[QT_INSTALL_PREFIX]", prefix_path)
                                 .as_bytes(),
                         );
                     }
@@ -341,7 +341,13 @@ impl QtBuild {
     }
 
     /// Some prl files don't follow a consistent naming scheme. Try to find it by looking at all files in lib_path.
-    fn find_qt_module_prl(&self, lib_path: &str, prefix: &str, version_major: u32, qt_module: &str) -> String {
+    fn find_qt_module_prl(
+        &self,
+        lib_path: &str,
+        prefix: &str,
+        version_major: u32,
+        qt_module: &str,
+    ) -> String {
         match Path::new(lib_path).read_dir() {
             Ok(lib_dir) => {
                 for entry in lib_dir {
@@ -367,7 +373,10 @@ impl QtBuild {
             }
         }
 
-        format!("{}/{}Qt{}{}.prl",lib_path, prefix, version_major, qt_module)
+        format!(
+            "{}/{}Qt{}{}.prl",
+            lib_path, prefix, version_major, qt_module
+        )
     }
 
     /// Tell Cargo to link each Qt module.
--- a/crates/qt-build-utils/src/lib.rs
+++ b/crates/qt-build-utils/src/lib.rs
@@ -340,6 +340,36 @@ impl QtBuild {
         }
     }
 
+    /// Some prl files don't follow a consistent naming scheme. Try to find it by looking at all files in lib_path.
+    fn find_qt_module_prl(&self, lib_path: &str, prefix: &str, version_major: u32, qt_module: &str) -> String {
+        match Path::new(lib_path).read_dir() {
+            Ok(lib_dir) => {
+                for entry in lib_dir {
+                    match entry {
+                        Ok(entry) => {
+                            let file_name = entry.file_name();
+                            let file_name = file_name.to_string_lossy();
+                            let prl_file = file_name.ends_with(".prl");
+                            let matches_module = file_name
+                                .contains(&format!("Qt{}{}", self.version.major, qt_module));
+                            if prl_file && matches_module {
+                                return entry.path().display().to_string();
+                            }
+                        }
+                        Err(e) => {
+                            println!("cargo:warning=Could not read {} entry: {}", lib_path, e);
+                        }
+                    }
+                }
+            }
+            Err(e) => {
+                println!("cargo:warning=Could not read dir {}: {}", lib_path, e);
+            }
+        }
+
+        format!("{}/{}Qt{}{}.prl",lib_path, prefix, version_major, qt_module)
+    }
+
     /// Tell Cargo to link each Qt module.
     pub fn cargo_link_libraries(&self, builder: &mut cc::Build) {
         let prefix_path = self.qmake_query("QT_INSTALL_PREFIX");
@@ -378,10 +408,7 @@ impl QtBuild {
             } else {
                 (
                     format!("Qt{}{qt_module}", self.version.major),
-                    format!(
-                        "{}/{}Qt{}{}.prl",
-                        lib_path, prefix, self.version.major, qt_module
-                    ),
+                    self.find_qt_module_prl(&lib_path, prefix, self.version.major, qt_module),
                 )
             };
 
--- a/crates/qt-build-utils/src/lib.rs
+++ b/crates/qt-build-utils/src/lib.rs
@@ -306,8 +306,43 @@ impl QtBuild {
         .to_string()
     }
 
+    fn cargo_link_qt_library(
+        &self,
+        builder: &mut cc::Build,
+        name: &str,
+        prefix_path: &str,
+        lib_path: &str,
+        link_lib: &str,
+        prl_path: &str,
+    ) {
+        println!("cargo:rustc-link-lib={link_lib}");
+
+        match std::fs::read_to_string(&prl_path) {
+            Ok(prl) => {
+                for line in prl.lines() {
+                    if let Some(line) = line.strip_prefix("QMAKE_PRL_LIBS = ") {
+                        parse_cflags::parse_libs_cflags(
+                            builder,
+                            name,
+                            line.replace(r"$$[QT_INSTALL_LIBS]", &lib_path)
+                                .replace(r"$$[QT_INSTALL_PREFIX]", &prefix_path)
+                                .as_bytes(),
+                        );
+                    }
+                }
+            }
+            Err(e) => {
+                println!(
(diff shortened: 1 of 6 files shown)


crates/qt-build-utils/src/lib.rs
```diff
@@ -306,8 +306,73 @@ impl QtBuild {
...
+        builder: &mut cc::Build,
+        name: &str,
+        prefix_path: &str,
+        lib_path: &str,
+        link_lib: &str,
+        prl_path: &str,
+    ) {
+        println!("cargo:rustc-link-lib={link_lib}");
+
+        match std::fs::read_to_string(&prl_path) {
```
ahayzen-kdab, line 320: run `cargo clippy` to see suggestions here that CI is failing on

crates/qt-build-utils/src/lib.rs
```diff
@@ -306,8 +306,73 @@ impl QtBuild {
...
+                println!(
+                    "cargo:warning=Could not open {} file to read libraries to link: {}",
+                    &prl_path, e
+                );
+            }
+        }
+    }
+
+    /// Some prl files don't follow a consistent naming scheme. Try to find it by looking at all files in lib_path.
+    fn find_qt_module_prl(&self, lib_path: &str, prefix: &str, version_major: u32, qt_module: &str) -> String {
```
ahayzen-kdab, line 344: run `cargo fmt` to fix formatting issues CI is failing on

crates/qt-build-utils/src/lib.rs
jimmyvanhest: Oops was sure I did that. Should be fixed.

crates/qt-build-utils/src/lib.rs
jimmyvanhest: Oops was sure I did that. Should be fixed.

crates/qt-build-utils/src/parse_cflags.rs
```diff
@@ -167,13 +167,18 @@ pub(crate) fn parse_libs_cflags(name: &str, link_args: &[u8]) {
...
-                        match extract_lib_from_filename(target, &file_name.to_string_lossy()) {
-                            Some(lib_basename) => {
-                                println!("cargo:rustc-link-search={}", dir.display());
-                                println!("cargo:rustc-link-lib={lib_basename}");
-                            }
-                            None => {
-                                println!("cargo:warning=File path {} found in .prl file for {name}, but could not extract library base name to pass to linker command line", path.display());
+                        if file_name.to_string_lossy().ends_with(".o") {
+                            #[cfg(feature = "include_qt_objects")]
+                            _builder.object(path);
```
Be-ing, line 172: Oh this is an interesting hack; I didn't realize the cc crate could do this. This should fix linking to Qt6Quick statically (#513). Why is this behind a feature flag though? IIUC this should always be done.

Please add a TODO comment linking to https://github.com/rust-lang/rust/issues/99427#issuecomment-1562092085 and explaining that this should be replaced when rustc has a CLI for linking object files.

crates/qt-build-utils/src/parse_cflags.rs
```diff
@@ -103,7 +103,7 @@ fn split_flags(link_args: &[u8]) -> Vec<String> {
     words
 }
 
-pub(crate) fn parse_libs_cflags(name: &str, link_args: &[u8]) {
+pub(crate) fn parse_libs_cflags(_builder: &mut cc::Build, name: &str, link_args: &[u8]) {
```
Be-ing, line 106: Move `builder` to be the last argument and add a TODO to remove it (see other review comment).

crates/qt-build-utils/src/lib.rs
```diff
@@ -343,35 +417,35 @@ impl QtBuild {
...
+                &link_lib,
+                &prl_path,
+            );
+        }
+
+        let emscripten_targeted = match env::var("CARGO_CFG_TARGET_OS") {
+            Ok(val) => val == "emscripten",
+            Err(_) => false,
+        };
+        if emscripten_targeted {
```
Be-ing, line 438: Can't this be condensed to a single `if let` statement?

crates/qt-build-utils/src/lib.rs
```diff
@@ -306,8 +306,82 @@ impl QtBuild {
...
+            Err(e) => {
+                println!(
+                    "cargo:warning=Could not open {} file to read libraries to link: {}",
+                    &prl_path, e
+                );
+            }
+        }
+    }
+
+    /// Some prl files don't follow a consistent naming scheme. Try to find it by looking at all files in lib_path.
```
Be-ing, line 343: Can you be more specific in this comment? What is an example of a filename that doesn't match the usual pattern?

Be-ing (changes requested):
Thanks for working on build fixes for Android and emscripten! We haven't been trying these platforms ourselves, so this is a very welcome contribution!

crates/qt-build-utils/src/parse_cflags.rs
Be-ing: When rustc has support for linking object files, we should contribute that upstream to the pkg-config crate which is where this code came from. Add a dependency on the `cc` crate wouldn't work for that, but it's fine for this temporary hack.

jimmyvanhest: The reason I put this behind a feature flag is for integration with other build systems like cmake and qmake.

qmake when building for wasm in particular, pulls in a lot of dependencies automatically resulting in lots of duplicate symbol errors at the linking stage when this feature is turned on.

crates/qt-build-utils/src/parse_cflags.rs
jimmyvanhest: I added the TODO comment at the same place where builder is used.

Also _builder can't drop its leading _ because clippy would complain without the feature flag turned on.

crates/qt-build-utils/src/lib.rs
jimmyvanhest: I could condense it down to a single if let statement but it doesn't look really nice to me
`if let Ok(true) = env::var("CARGO_CFG_TARGET_OS").map(|val| val == "emscripten") {`

I tired doing it with 2 if statements but this changes the indenting and adds some linebreaks after `cargo fmt`
```
        if let Ok(target_os) = env::var("CARGO_CFG_TARGET_OS").map(|val| val == "emscripten") {
            if target_os == "emscripten" {
                let platforms_path =
                    format!("{}/platforms", self.qmake_query("QT_INSTALL_PLUGINS"));
                println!("cargo:rustc-link-search={platforms_path}");
                self.cargo_link_qt_library(
                    "qwasm",
                    &prefix_path,
                    &lib_path,
                    "qwasm",
                    &format!("{platforms_path}/libqwasm.prl"),
                    builder,
                );
            }
        }
```

crates/qt-build-utils/src/lib.rs
jimmyvanhest: Should be done please review

ahayzen-kdab:
(Note that you might want to rebase your branches onto `main` as i noticed when merging one of the others it was based on `v0.5.3` which is a bit old now, so you could get conflicts)

crates/qt-build-utils/src/lib.rs
Be-ing: Okay, I'm not going to nitpick on this.

crates/qt-build-utils/src/lib.rs
```diff
@@ -306,8 +306,84 @@ impl QtBuild {
...
+        match Path::new(lib_path).read_dir() {
+            Ok(lib_dir) => {
+                for entry in lib_dir {
+                    match entry {
+                        Ok(entry) => {
+                            let file_name = entry.file_name();
+                            let file_name = file_name.to_string_lossy();
+                            let prl_file = file_name.ends_with(".prl");
+                            let matches_module = file_name
+                                .contains(&format!("Qt{}{}", self.version.major, qt_module));
```
Be-ing, line 362: This isn't specific enough. It finds libQt6QuickDialogs2Utils.prl when linking Qt6Quick which doesn't list all the `.o` files that Qt6Quick needs to link. Can you post a list of those filenames for Android? Maybe we can hardcode a list of Android architectures to append or something?

crates/qt-build-utils/src/parse_cflags.rs
Be-ing: I have been playing with merging this branch together with #513 and got it to link by deduplicating the `.o` files (plus a quick hack around https://github.com/KDAB/cxx-qt/pull/600/files/cebe9711c434f336d15cc98512d79cccfeeb47f4#diff-e4bb99471e34daf6f9630c7df936bd268f023426f24f3f39ba7658079ff2ffe9). I made a PR for your fork with this fix: https://github.com/jimmyvanhest/cxx-qt/pull/1 Please test that and see if it avoids the duplicate symbol link errors you were encountering before.

jimmyvanhest:
hmm latest changes failed on macOS Qt6 with duplicate symbol errors on the qml_extension_plugin example. Strange that only this platform would pull in additional object files that are now handled also handled by cxx-qt.

The following change might fix this, but I have no mac to test this on and would not like to spam the pipeline with things that might not actually do anything. For Linux Qt6 this would not make a difference work though

examples/qml_extension_plugin/plugin/rust/Cargo.toml:
```
[build-dependencies]
-cxx-qt-build.workspace = true
+cxx-qt-build = { workspace = true, features = [ "link_qt_external" ] }
```

jimmyvanhest:
What we are doing here now would technically be more correct but would break compatibility for some system and Qt installations, thus requiring major, or since this is still in major 0 a minor, version change.

Be-ing:
> hmm latest changes failed on macOS Qt6 with duplicate symbol errors on the qml_extension_plugin example. Strange that only this platform would pull in additional object files that are now also handled by cxx-qt.

macOS isn't the issue. I can reproduce this on Linux with a static build of Qt6. The issue is that both the Cargo side and the CMake side are linking those `.o` files... :thinking: Maybe we have to only link Qt with CMake when linking dynamically? I'll experiment with that tomorrow.

Be-ing:
Oof, we're in quite a maze of conflicting requirements and I can't think of a great solution that Just Works automatically in every case. The duplicate symbol errors come from Cargo linking the `.o` files into the staticlib and linking Qt's CMake targets also telling CMake to link the `.o` files. We can't have both Cargo and CMake linking the `.o` files.

There might be a way to tell CMake to not link the `.o` files when the Qt targets are static libraries. However, I think that would get quite messy with reaching into unstable implementation details of Qt's CMake modules. Perhaps we could try to get in touch with someone in the Qt Company that works on Qt's build system and ask them to add an option for not linking those `.o` files. If they agree, our downstream users would need to add a line to their CMakeLists.txt to enable that very niche option. Explaining why that option would need to be turned on would be awkward. Likely some downstream users would skip that and their end users would run into puzzling link errors when trying to link Qt statically.

We want Cargo to link the `.o` files when Cargo is building an executable, whether that's a bin crate or `cargo test` executables. Unfortunately Cargo doesn't provide any means for build.rs to determine what type of crate its building nor whether Cargo is building a test executable.

We could use a Cargo feature to toggle whether to link the `.o` files. That would allow Cargo to successfully link bin crates. We could make `cargo test` work by passing `--no-default-features`/`--feature skip-linking-objects` (depending on whether it's default on or off). Because we run `cargo test` indirectly from CMake in our repository, I think this may be the least worst option, though downstream users would likely get confused if they fall into this same rabbit hole using `cargo test` on a staticlib crate and statically linking Qt.

Be-ing:
I made another PR for your fork adding back the Cargo feature for linking `.o` files and enabling it for our bin crate and `cargo test`: https://github.com/jimmyvanhest/cxx-qt/pull/2. I tested merging that with #513 and linking to a static build of Qt 6 on Linux and cmake & ctest both work.

Be-ing:
I made a new PR https://github.com/KDAB/cxx-qt/pull/605 with your commits plus a few of my own on top of those. We've merged that to unblock #513. Thank you for contributing this!

jimmyvanhest:
Ah great that it's merged.

I did not have much free time this week too look at it hence the silence from me.

I will tests all of it out with my setup for wasm and android.

And a quick question, is there any indication when a new release will be made? I would need that since I'd like to move away from using this crate as a submodule in my project and just use cargo.

ahayzen-kdab:
> And a quick question, is there any indication when a new release will be made? I would need that since I'd like to move away from using this crate as a submodule in my project and just use cargo.

We are currently in the middle of an API transition to 0.6, currently projecting for maybe end of August for the 0.6 series.

However I can try to do a backport to the 0.5.x series and make a 0.5.4 release if this doesn't regress / change the behaviour for 0.5.x :thinking:

jimmyvanhest:
> We are currently in the middle of an API transition to 0.6, currently projecting for maybe end of August for the 0.6 series.

neat.

> However I can try to do a backport to the 0.5.x series and make a 0.5.4 release if this doesn't regress / change the behaviour for 0.5.x thinking

While handy I won't require it very much. Using a submodule till then would not be bothersome for me, so don't bother with it(unless you really want of course). It was just a curiosity because the last release was a while ago.

PR #601, opened 2023-07-01
Disabled QDateTime and QTimeZone for unsupported target emscripten.

jimmyvanhest:


ahayzen-kdab (approved):
Awesome! I like the using of the emscripten target to remove QDateTime etc.

A further improvement could be to also add this in the tests/ and maybe examples/. But I could see this getting more complicated, so making cxx-qt-lib work should be the first step.

ahayzen-kdab:
@jimmyvanhest Could you create a further change which adds the `#[cfg(not(target_os = "emscripten"))]` to the `cxx-qt-lib-headers/src/lib.rs`, just so that we don't pointlessly write out `qdatetime.h` and `qtimezone.h` into the include dir.

PR #609, opened 2023-07-15
update Corrosion

Be-ing:
to work around https://github.com/corrosion-rs/corrosion/issues/418 with Rust 1.71


ahayzen-kdab (approved):
LGTM :-)

PR #613, opened 2023-07-18
cxx-qt-gen: remove qobject:: from impl a trait for a qobject in a bridge

ahayzen-kdab:
Related to #559


Be-ing (approved):
Still seems like some magical scoping to have the type alias declared in the `extern "RustQt"` block but the trait outside of that :thinking: But anyway getting rid of the old `qobject::` is an improvement.

PR #614, opened 2023-07-18
cxx-qt-gen: remove qobject:: from parameters in methods in a bridge

ahayzen-kdab:
Related to #559

Requires #613


examples/qml_minimal/rust/src/cxxqt_object.rs
```diff
@@ -33,10 +33,10 @@ pub mod qobject {
     // ANCHOR: book_rustobj_invokable_signature
     unsafe extern "RustQt" {
```
Be-ing, line 34: Tangential to this PR, but should this be combined with the above `extern "RustQt"` block?

Be-ing (dismissed):

examples/qml_minimal/rust/src/cxxqt_object.rs
ahayzen-kdab: Yup it can, they are just separate so there are separate blocks for the book to anchors to. But yup can be combined or separate it doesn't matter.

LeonMatthesKDAB (approved):
LGTM

PR #615, opened 2023-07-18
WIP: cxx-qt-gen: pass through signals directly

ahayzen-kdab:
Although not sure it's worth the gain as we still need to generate the other methods from the same info.
    
Related to #559

Requires #614

### DIFF
--- a/crates/cxx-qt-gen/src/generator/rust/signals.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/signals.rs
@@ -24,14 +24,13 @@ pub fn generate_rust_signals(
     // Create the methods for the other signals
     for signal in signals {
         let idents = QSignalName::from(signal);
-        let signal_name_rust = idents.name.rust;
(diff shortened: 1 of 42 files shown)


crates/cxx-qt-gen/src/generator/rust/signals.rs
```diff
@@ -55,15 +54,12 @@ pub fn generate_rust_signals(
...
-        let attrs = &signal.method.attrs;
-
         let fragment = RustFragmentPair {
             cxx_bridge: vec![
                 quote! {
                     #unsafe_block extern "C++" {
-                        #(#attrs)*
-                        #[rust_name = #signal_name_rust_str]
-                        #unsafe_call fn #signal_name_cpp(self: #self_type, #(#parameters),*);
+                        #[cxx_name = #signal_name_cpp_str]
```
ahayzen-kdab, line 61: not sure this is worth the gain, as we still need to generate the other methods.

Something to consider later after the API shifts are complete.

PR #618, opened 2023-07-19
cxx-qt-gen: remove hidden module

ahayzen-kdab:
Requires #617 #587 #620

And needs

- [x] Fully qualified types in generated methods, like parameters from signals `UniquePtr` -> `cxx::UniquePtr` (should be able to make a mechanism in the new `utils/` to qualify a given syn type, then use that for other types like `Pin` too (slightly related to #587) Have a generic mechanism in utils to do Pin, UniquePtr, SharedPtr etc which then removes the double definitions in #587 
- [x] `impl MyObject` needs to be `impl module::MyObject` or have a type/use to bring it in
- [x] Rust setters might need to have `use cxx_qt::CxxQtType` to be able to use `rust_mut()` in their methods, but we also don't want to import and "leak" that into the outside scope, so should just be in the method scope if required
- [x] Removed `uses` from `ParsedCxxQtData`
- [x] Qualify free methods calls to the bridge with the module ident
- [x] Arguments in the constructor need to pass through qualify

### DIFF
--- a/crates/cxx-qt-gen/src/generator/rust/qobject.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/qobject.rs
@@ -151,6 +151,7 @@ impl GeneratedRustQObject {
             &qobject.constructors,
             &qobject_idents,
             &namespace_idents,
+            qualified_mappings,
         )?);
 
         generated.blocks.append(&mut cxxqttype::generate(
--- a/crates/cxx-qt-gen/src/generator/rust/qobject.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/qobject.rs
@@ -18,7 +18,7 @@ use crate::{
     parser::qobject::ParsedQObject,
 };
 use quote::quote;
-use syn::{parse_quote, Ident, ImplItem, Item, Path, Result};
+use syn::{Ident, ImplItem, Item, Path, Result};
 
 #[derive(Default)]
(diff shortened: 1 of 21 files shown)


crates/cxx-qt-gen/src/generator/rust/qobject.rs
```diff
@@ -45,7 +47,10 @@ pub struct GeneratedRustQObject {
 }
 
 impl GeneratedRustQObject {
-    pub fn from(qobject: &ParsedQObject) -> Result<GeneratedRustQObject> {
+    pub fn from(
+        qobject: &ParsedQObject,
+        qualified_mappings: &BTreeMap<Ident, Path>,
+    ) -> Result<GeneratedRustQObject> {
```
Be-ing: I think it is misleading to call this function `from` because it is not an impl of the core::convert::From trait. Can you come up with a new name for this function? Maybe `new`?

ahayzen-kdab: Right, however note that this was copied from the others that are also `from` for both one and multiple parameters.

I can fix this as a separate commit at the end of this chain or a separate pull request...

Be-ing: okay

Be-ing (approved):

PR #620, opened 2023-07-21
cxx-qt-gen: fully qualify bridge parameters in implementations

ahayzen-kdab:
Related to #404

### DIFF
--- a/crates/cxx-qt-gen/src/generator/utils/rust.rs
+++ b/crates/cxx-qt-gen/src/generator/utils/rust.rs
@@ -3,7 +3,99 @@
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
-use syn::{GenericArgument, PathArguments, Type, TypeReference};
+use syn::{GenericArgument, PathArguments, PathSegment, ReturnType, Type, TypeReference};
+
+/// Return a qualified version of the type that can by used outside of a CXX bridge
+///
+/// Eg Pin -> core::pin::Pin or UniquePtr -> cxx::UniquePtr
+pub(crate) fn syn_type_cxx_bridge_to_qualified(ty: &syn::Type) -> syn::Type {
+    match ty {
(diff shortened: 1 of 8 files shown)


crates/cxx-qt-gen/src/generator/utils/rust.rs
```diff
@@ -3,7 +3,97 @@
...
+                let qualified_prefix = match segment.ident.to_string().as_str() {
+                    "CxxString" | "CxxVector" | "SharedPtr" | "UniquePtr" | "WeakPtr" => {
+                        Some(vec!["cxx"])
+                    }
+                    "Box" => Some(vec!["std", "boxed"]),
+                    "Pin" => Some(vec!["core", "pin"]),
+                    "String" => Some(vec!["std", "string"]),
+                    "Vec" => Some(vec!["std", "vec"]),
+                    _ => None,
+                };
```
Be-ing, line 56: Is this an exhaustive list of all the types we will need to handle? Or are we going to keep needing to adding special cases when users ask for them? What determines that a type needs to be handled here?

ahayzen-kdab: Added a comment.

It's basically any type that CXX supports that isn't in primitive types https://doc.rust-lang.org/stable/std/primitive/index.html or prelude types https://doc.rust-lang.org/stable/std/prelude/index.html - which comes down to CxxString, CxxVector, SharedPtr, UniquePtr, WeakPtr, Pin.

We need this because inside a `cxx::bridge` it is supported to just write `UniquePtr<T>` but as we generate shims outside of the bridge from this type information we need to qualify it.

For custom types we don't need this as the developer needs to add the type to the bridge as `type T;` then a change in the next branch it reads all the mappings so if there is a `type T;` inside the `ffi` bridge module it then qualifies to `ffi::T` in the shims :-)

Be-ing (approved):

PR #621, opened 2023-07-21
Allow lifetime parameters in cxx_qt::Constructors

LeonMatthesKDAB:
This makes it possible to have constructors take arguments by-reference, which is the default in C++.

### DIFF
--- a/crates/cxx-qt-gen/test_inputs/invokables.rs
+++ b/crates/cxx-qt-gen/test_inputs/invokables.rs
@@ -52,12 +52,14 @@ mod ffi {
 
     impl cxx_qt::Threading for MyObject {}
 
-    impl
+    impl<'a>
         cxx_qt::Constructor<
-            (i32, *mut QObject),
+            (i32, &'a QString),
             BaseArguments = (*mut QObject,),
-            NewArguments = (i32,),
+            NewArguments = (&'a QString,),
         > for MyObject
     {
     }
+
+    impl cxx_qt::Constructor<()> for MyObject {}
 }
--- a/crates/cxx-qt-gen/src/generator/cpp/constructor.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/constructor.rs
@@ -153,6 +153,8 @@ mod tests {
             base_arguments: vec![],
             new_arguments: vec![],
             initialize_arguments: vec![],
+            lifetime: None,
+            // dummy impl
             imp: parse_quote! { impl X {} },
         }
     }
@@ -275,6 +277,7 @@ mod tests {
                 new_arguments: vec![parse_quote! { i16}, parse_quote! { i32 }],
                 initialize_arguments: vec![parse_quote! { i32 }, parse_quote! { i64 }],
                 base_arguments: vec![parse_quote! { i64 }, parse_quote! { *mut QObject }],
+                lifetime: Some(parse_quote! { 'a_lifetime }),
                 ..mock_constructor()
             }],
             &["initializer".to_string()],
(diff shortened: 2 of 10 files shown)


LeonMatthesKDAB:
Wondering now whether it's worth it to add this, it's such a huge pain to add, for relatively little benefit, as it's at least possible to just take most arguments by-value or use pointers as a workaround if necessary.

ahayzen-kdab:
> Wondering now whether it's worth it to add this, it's such a huge pain to add, for relatively little benefit, as it's at least possible to just take most arguments by-value or use pointers as a workaround if necessary.

I think we need to consider the use cases where you would want to use references and there isn't a workaround. We are already quite niche though being a C++ constructor with a reference :thinking:

LeonMatthesKDAB:
After rebasing this onto main again, the diff is actually not that bad...
The generation is still relatively complicated, so unsure if it's worth it to maintain.

I'll leave that decision up to you @Be-ing @ahayzen-kdab

crates/cxx-qt-gen/src/syntax/types.rs
```diff
@@ -119,6 +119,85 @@ pub fn extract_qobject_ident(ty: &Type) -> Result<(Ident, Option<Mut>)> {
...
+
+fn extract_generic_argument_lifetimes(argument: &GenericArgument) -> Result<Vec<Lifetime>> {
+    match argument {
+        GenericArgument::Lifetime(lifetime) => Ok(vec![lifetime.clone()]),
+        GenericArgument::Type(ty) => extract_lifetimes(ty),
+        _ => Err(err_unsupported_type(argument)),
+    }
+}
+
+fn extract_pathsegment_lifetimes(segment: &PathSegment) -> Result<Vec<Lifetime>> {
```
ahayzen-kdab, line 134: wonder if this stuff could be in it's own lifetimes.rs under syntax as they are sortof types helpers but very specifically lifetime helpers ?

crates/cxx-qt-gen/test_inputs/invokables.rs
```diff
@@ -52,12 +52,14 @@ mod ffi {
...
     impl cxx_qt::Threading for MyObject {}
 
-    impl
+    impl<'a>
         cxx_qt::Constructor<
-            (i32, *mut QObject),
+            (i32, &'a mut QString),
             BaseArguments = (*mut QObject,),
-            NewArguments = (i32,),
+            NewArguments = (&'a QString,),
```
ahayzen-kdab, line 59: any reason this isn't `mut` ? or the other isn't const ?

LeonMatthesKDAB: I just wanted to show off the `default` for these two types.
E.g. the `*mut QObject` should map to a `QObject*` (usually the parent pointer). The `&'a QString` should map to a `const QString &`, which is usually what you would use in C++.

ahayzen-kdab: But in C++ you get `QString&` for the constructor as the template type as the top is `&mut QString` ?

ahayzen-kdab: (at least looking at invokables.h in the test outputs)

crates/cxx-qt-gen/src/syntax/types.rs
LeonMatthesKDAB: Hm, good point, especially for those helpers :thinking:

LeonMatthesKDAB: Done

crates/cxx-qt-gen/test_inputs/invokables.rs
LeonMatthesKDAB: Aaah, that `mut` you mean.
Good point.
Fixed.

ahayzen-kdab (approved):
LGTM, lets go with this for now, in the end not too crazy :-) We can always strip it back out before the release if we find something that doesn't work.

PR #622, opened 2023-07-25
examples: add basic widgets example

### LINKED ISSUE #143: QWidgets example
It would be helpful to have an example how to use CXX-Qt in a QWidgets application.

ahayzen-kdab:
Lots of hacks, but we can launch a `QPushButton` !

I'll likely pick things off from this branch as separate pull requests over time.

Closes #143

### DIFF
--- a/crates/cxx-qt-lib-extras/src/gui/qpushbutton.rs
+++ b/crates/cxx-qt-lib-extras/src/gui/qpushbutton.rs
@@ -42,6 +42,9 @@ mod ffi {
 }
 
 impl ffi::QPushButton {
+    // TODO: we need to think about how ownership in widgets is going to work
+    // as for items like QPushButton you may have a parent and have that
+    // free the object and not when the unique ptr goes out of scope
     pub fn new() -> cxx::UniquePtr<Self> {
         ffi::qpushbutton_init_default()
     }
--- a/crates/cxx-qt-lib-extras/src/gui/qpushbutton.rs
+++ b/crates/cxx-qt-lib-extras/src/gui/qpushbutton.rs
@@ -1,0 +1,50 @@
+// SPDX-FileCopyrightText: 2024 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+#[cxx_qt::bridge]
+mod ffi {
+    unsafe extern "C++" {
+        include!("cxx-qt-lib/qstring.h");
+        type QString = cxx_qt_lib::QString;
+    }
+
+    unsafe extern "C++Qt" {
+        include!("cxx-qt-lib-extras/qpushbutton.h");
+
+        #[qobject]
+        type QPushButton;
+
+        // TODO: we should use upcasting methods here and implement QAbstractButton and QWidget
+        // so that we don't need to duplicate all of the methods
+
+        /// This signal is emitted when the button is activated (i.e., pressed down then released while the mouse cursor is inside the button)
+        #[qsignal]
+        fn clicked(self: Pin<&mut QPushButton>, checked: bool);
+
+        /// Set the text shown on the button
+        #[rust_name = "set_text"]
+        fn setText(self: Pin<&mut QPushButton>, text: &QString);
+
+        /// Shows the widget and its child widgets.
+        fn show(self: Pin<&mut QPushButton>);
+    }
+
+    #[namespace = "rust::cxxqtlib1"]
+    unsafe extern "C++Qt" {
+        include!("cxx-qt-lib/common.h");
+
+        #[doc(hidden)]
+        #[rust_name = "qpushbutton_init_default"]
+        fn make_unique() -> UniquePtr<QPushButton>;
+    }
+}
+
+impl ffi::QPushButton {
+    pub fn new() -> cxx::UniquePtr<Self> {
+        ffi::qpushbutton_init_default()
+    }
+}
+
+pub use ffi::QPushButton;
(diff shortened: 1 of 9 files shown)


examples/widgets_minimal/cpp/qpushbutton.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
+
+#include <QtWidgets/QPushButton>
+
+#include <memory>
+
+#include "rust/cxx.h"
+
+::std::unique_ptr<QPushButton>
+qpushbuttonNew();
```
ahayzen-kdab, line 10: should we have a generic templated constructor in cxx-qt-lib to help with this?

crates/cxx-qt-gen/src/generator/cpp/externcxxqt.rs
```diff
@@ -32,6 +32,10 @@ pub fn generate(
             let mut block = GeneratedCppExternCxxQtBlocks::default();
             let data = generate_cpp_signal(signal, &signal.qobject_ident, cxx_mappings)?;
             block.includes = data.includes;
+            // Ensure that we include MaybeLockGuard<T> that is used in multiple places
```
ahayzen-kdab, line 35: not sure how we missed this elsewhere

examples/widgets_minimal/src/qpushbutton_cxx.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    unsafe extern "C++Qt" {
+        include!(<QtWidgets/QPushButton>);
+        type QPushButton;
+
+        include!("cxx-qt-lib/qstring.h");
+        type QString = cxx_qt_lib::QString;
+
+        #[qsignal]
+        #[allow(dead_code)]
```
ahayzen-kdab, line 17: as we only connect to the signal and don't emit it causes dead code, wonder if we should always mark as dead ?

ahayzen-kdab: (as this is an existing signal rather than one being created from Rust, so for inherit / C++Qt types you might not see an emit)

crates/cxx-qt-gen/src/generator/cpp/externcxxqt.rs
ahayzen-kdab: Split into #804

examples/widgets_minimal/cpp/qpushbutton.h
LeonMatthesKDAB: I think we have this now thanks to one of our other PRs that landed recently.
The QPainter one I think :thinking:

examples/widgets_minimal/src/main.rs
```diff
@@ -0,0 +1,35 @@
...
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! This example provides demostrations of building a Qt Widgets CXX-Qt application
+
+use cxx_qt_lib::QString;
+use cxx_qt_lib_extras::{QApplication, QPushButton};
+
+fn main() {
```
ahayzen-kdab, line 11: we should have some kind of test as well, even if C++ to see if this works, unsure how so far....

examples/widgets_minimal/src/qpushbutton_cxx.rs
ahayzen-kdab: Isn't required anymore

examples/widgets_minimal/src/main.rs
```diff
@@ -0,0 +1,35 @@
...
+    let mut app = QApplication::new();
+
+    // TODO: we should really pass a parent in here
+    let mut push_button = QPushButton::new();
```
ahayzen-kdab, line 15: Should we use UniquePtr or should we use raw pointer and set a parent like Qt, or should we support both?

jaques-sam:
Thanks for doing this!
Can this be continued? If cxx-qt can manage to get these bindings in, cxx-qt will get much more traction I think. The need for Rust-QWidgets interop is a missing link!

crates/cxx-qt-lib-extras/src/gui/qpushbutton.rs
```diff
@@ -0,0 +1,54 @@
...
+        type QString = cxx_qt_lib::QString;
+    }
+
+    unsafe extern "C++Qt" {
+        include!("cxx-qt-lib-extras/qpushbutton.h");
+
+        #[qobject]
+        type QPushButton;
+
+        // TODO: we should use upcasting methods here and implement QAbstractButton and QWidget
```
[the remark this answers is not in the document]
jaques-sam: Possibly with something like this:

```rust
    unsafe extern "C++" {
        include!("qwidget.h");
        type QWidget = super::super::widget::QWidget;
    }
```

And like so:
`let widget_ptr = Pin::into_inner_unchecked(label) as *mut QPushButton;`

and:
`Pin::new_unchecked(&mut *self.ptr), widget_ptr.cast());`

examples/widgets_minimal/src/main.rs
jaques-sam: Raw pointers -> definitely not, as it will make it unsafe.

ahayzen-kdab:
> Thanks for doing this! Can this be continued? If cxx-qt can manage to get these bindings in, cxx-qt will get much more traction I think. The need for Rust-QWidgets interop is a missing link!

The problem with Qt Widgets API is it is mostly raw pointers, so when you end up wrapping it you end up with an unsafe API that just maps the C++ code but in Rust.

With the previous versions of CXX-Qt we extended the support for `extern "C++Qt"` blocks and for inheritance/casting, so instead of these types being baked into cxx-qt-lib and having a huge maintenance burden of the whole Qt API, it is at least possible for developers to write their own bridges.

Eg see the examples in https://kdab.github.io/cxx-qt/book/bridge/extern_cppqt.html and https://kdab.github.io/cxx-qt/book/concepts/casting.html

I would however be cautious attempting to write a whole GUI with Qt Widgets in Rust, a lot of the glue code is quite painful to write, instead consider writing the frontend side still in C++ and move the backend to Rust by using the QAbstractListModel abstractions that work well in Rust.

Issue #242, opened 2022-09-08
ahead-of-time compilation of QML


Be-ing:
It would be nice if qt-build could call qmlcachegen (Qt5) or qmlsc (Qt6).

https://www.qt.io/blog/the-new-qtquick-compiler-technology

However qmlsc is not open source yet:

> qmlsc is available for commercial customers and some of its features are merged into qmlcachegen, which continues to be available with all versions of Qt.

https://doc.qt.io/qt-6/qtqml-tool-qmlsc.html

ahayzen-kdab:
There is also the `qmltc` https://doc.qt.io/qt-6/qtqml-tool-qmltc.html https://www.qt.io/blog/qml-type-compilation

And we should support commercial tools as well.

Be-ing:
> qmlcachegen is helpful in any case when qmltc isn't used.

I asked about this on a [recent Qt blog post](https://www.qt.io/blog/compiling-qml-to-c-a-4x-speedup) and this isn't quite right. qmlcachegen is always helpful. qmltc can be used together with qmlcachegen; they work on different aspects of speeding up QML.

Be-ing:
> Qt 6.5 will add a new [QQmlApplicationEngine::loadFromModule function](https://doc-snapshots.qt.io/qt6-6.5/qqmlapplicationengine.html#loadFromModule) which will be able to load qmltc compiled QML types via their QML module URI and type name as QStrings, so we won't need to generate Rust bindings to the generated C++ classes.

This is detailed in a new blog post now that Qt 6.5 has been released: https://www.qt.io/blog/whats-new-for-qml-modules-in-6.5

PR #623, opened 2023-07-25
move QML element registration to build.rs instead of bridge

Be-ing:
This avoids the need to repeatedly specify the URI and version of the QML module in the bridges. QObjects that are registered as QML elements are now marked by
[cxx_qt::qobject(qml_element = "OptionalName")]

This is a precursor to generating qmldir files which are required for qmlcachegen and qmltc
(https://github.com/KDAB/cxx-qt/issues/242)

### DIFF
--- a/crates/cxx-qt-build/src/lib.rs
+++ b/crates/cxx-qt-build/src/lib.rs
@@ -17,7 +17,7 @@ use diagnostics::{Diagnostic, GeneratedError};
 use convert_case::{Case, Casing};
 use quote::ToTokens;
 use std::{
-    collections::{HashMap, HashSet},
+    collections::HashSet,
     env,
     fs::File,
     io::Write,
@@ -26,7 +26,7 @@ use std::{
 
 use cxx_qt_gen::{
     parse_qt_file, write_cpp, write_rust, CppFragment, CxxQtItem, GeneratedCppBlocks,
-    GeneratedRustBlocks, Parser, QmlElementMetadata,
+    GeneratedRustBlocks, Parser,
 };
 
 // TODO: we need to eventually support having multiple modules defined in a single file. This
@@ -40,19 +40,25 @@ use cxx_qt_gen::{
 struct GeneratedCppFilePaths {
     plain_cpp: PathBuf,
     qobject: Option<PathBuf>,
-    qobject_header: Option<QObjectHeader>,
-}
-
-struct QObjectHeader {
-    path: PathBuf,
-    qml_metadata: Vec<QmlElementMetadata>,
+    qobject_header: Option<PathBuf>,
 }
 
 struct GeneratedCpp {
     cxx_qt: Option<CppFragment>,
     cxx: cxx_gen::GeneratedCode,
     file_ident: String,
-    qml_metadata: Vec<QmlElementMetadata>,
+}
+
+/// Metadata for registering a QML module
+struct QmlModule {
+    /// The URI of the QML module
+    pub uri: String,
+    /// The minor version of the QML module
+    pub version_minor: usize,
+    /// The major version of the QML module
+    pub version_major: usize,
+    /// The .rs files with #[qml_element] attribute(s)
+    pub rust_files: Vec<PathBuf>,
 }
 
 impl GeneratedCpp {
@@ -67,7 +73,6 @@ impl GeneratedCpp {
             .map_err(to_diagnostic)?;
 
         let mut cxx_qt = None;
-        let mut qml_metadata = Vec::new();
         // TODO: later change how the resultant filename is chosen, can we match the input file like
         // CXX does?
         //
@@ -120,11 +125,6 @@ impl GeneratedCpp {
                         .map_err(to_diagnostic)?;
                     let rust_tokens = write_rust(&generated_rust);
                     file_ident = parser.cxx_file_stem.clone();
-                    for (_, qobject) in parser.cxx_qt_data.qobjects {
-                        if let Some(q) = qobject.qml_metadata {
-                            qml_metadata.push(q);
-                        }
-                    }
 
                     // We need to do this and can't rely on the macro, as we need to generate the
                     // CXX bridge Rust code that is then fed into the cxx_gen generation.
@@ -145,7 +145,6 @@ impl GeneratedCpp {
             cxx_qt,
             cxx,
             file_ident,
-            qml_metadata,
         })
     }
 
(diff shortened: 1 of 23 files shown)


crates/cxx-qt-build/src/lib.rs
```diff
@@ -241,11 +237,17 @@ fn generate_cxxqt_cpp_files(
...
         println!("cargo:rerun-if-changed={path}");
 
-        let generated_code = GeneratedCpp::new(&path)?;
+        let generated_code = match GeneratedCpp::new(&path) {
+            Ok(v) => v,
+            Err(diagnostic) => {
+                diagnostic.report();
+                std::process::exit(1);
+            }
+        };
```
Be-ing, line 246: Moving this error handling here avoids some duplicate code because this function is now called twice in CxxQtBuilder::build. There's nothing else useful to do with the error and the process exits here, so there's no point propagating the error upward.

crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -394,17 +350,14 @@ pub mod tests {
     #[test]
     fn test_qml_metadata() {
         let item: ForeignTypeIdentAlias = parse_quote! {
-            #[cxx_qt::qobject(qml_uri = "foo.bar", qml_version = "1.0")]
+            #[cxx_qt::qobject(qml_element)]
             type MyObject = super::MyObjectRust;
         };
```
Be-ing, line 355: I was thinking of having the syntax be like:
```rust
#[cxx_qt::qobject]
#[qml_element]
#[qml_singleton]
```

but this was easier to implement and also makes sense, so :shrug:

examples/qml_features/rust/src/containers.rs
```diff
@@ -33,7 +33,7 @@ pub mod qobject {
     }
 
     unsafe extern "RustQt" {
-        #[cxx_qt::qobject(qml_uri = "com.kdab.cxx_qt.demo", qml_version = "1.0")]
+        #[cxx_qt::qobject(qml_element)]
```
Be-ing, line 36: This is nicer than repeating that copypasta over and over :smiley:

crates/cxx-qt-gen/src/parser/qobject.rs
ahayzen-kdab: I think eventually the syntax will be either
```rust
#[qml_element]
#[qml_singleton]
```
or
```rust
#[qobject]
#[qml_element]
#[qml_singleton]
```

But what you have at the moment is fine as a stepping stone, we can sort out exact the syntax in follow ups :-)

examples/qml_features/rust/src/containers.rs
ahayzen-kdab: :tada:

crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -124,63 +121,22 @@ impl ParsedQObject {
             &qobject_ty.attrs[attr_index],
             AttributeDefault::Some(|span| LitStr::new("", span)),
         )?;
-        let qml_uri = attrs_map.get(&quote::format_ident!("qml_uri"));
-        let qml_version = attrs_map.get(&quote::format_ident!("qml_version"));
-        let qml_name = attrs_map.get(&quote::format_ident!("qml_name"));
```
ahayzen-kdab, line 129: i would note that we have lost `qml_name` support in this, i think this was for `QML_NAMED_ELEMENT(T)`. But this can be readded in another pull request i guess ?
```cpp
class SqlEventDatabase : public QObject
{
    Q_OBJECT
    QML_NAMED_ELEMENT(EventDatabase)

    // ...
};
```

https://doc.qt.io/qt-6/qqmlengine.html#QML_NAMED_ELEMENT

crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -414,17 +367,14 @@ pub mod tests {
     #[test]
     fn test_qml_metadata_named() {
         let item: ForeignTypeIdentAlias = parse_quote! {
-            #[cxx_qt::qobject(qml_uri = "foo.bar", qml_version = "1", qml_name = "MyQmlElement")]
-            type MyNamedObject = super::MyObjectRust;
+            #[cxx_qt::qobject(qml_element = "OtherName")]
```
ahayzen-kdab, line 370: ah i see you've put it within the `qml_element = "T"`

crates/cxx-qt-gen/src/parser/qobject.rs
ahayzen-kdab: Ah i see further down this is combined into `qml_element = "T"`

crates/cxx-qt-gen/src/parser/qobject.rs
ahayzen-kdab: Slight bit of magic but probably makes sense and avoids doing `qml_element, qml_named_element = "OtherName"` which is repetitive.

crates/qt-build-utils/src/lib.rs
```diff
@@ -545,16 +541,13 @@ impl QtBuild {
             include_args += &format!("-I {} ", include_path.display());
         }
 
-        let mut uri_args = String::new();
-        for uri in uris {
-            uri_args += &format!("-Muri={} ", uri);
+        let mut cmd = Command::new(self.moc_executable.as_ref().unwrap());
+
+        if let Some(uri) = uri {
+            cmd.arg(&format!("-Muri={} ", uri));
```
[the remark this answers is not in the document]
Be-ing: good catch, fixed

examples/qml_features/rust/build.rs
```diff
@@ -9,19 +9,26 @@ use cxx_qt_build::CxxQtBuilder;
...
-        .file("src/multiple_qobjects.rs")
-        .file("src/nested_qobjects.rs")
-        .file("src/serialisation.rs")
-        .file("src/signals.rs")
-        .file("src/singleton.rs")
-        .file("src/properties.rs")
-        .file("src/threading.rs")
-        .file("src/types.rs")
-        .file("src/uncreatable.rs")
+        .qml_module(
```
ahayzen-kdab, line 12: was the plan still to move the arguments into a struct like

```rust
    CxxQtBuilder::new()
        .qml_module(QmlModule {
            uri: "com.example.qobject"
            version_major: 1
            version_minor: 0
            rust_files: ["src/cxxqt_object.rs"]
            qml_files: ["ui/foo.qml"]
        })
        .build();
```

As when we have mutliple args having named args is much nicer. And having a struct allows in the future the potential of builder pattern too.

Be-ing: Maybe, but then I think the struct would have to deal with lifetimes. This implementation is easy and flexible in what the user can provide with CxxQtBuilder::qml_module taking `&str` and `&[impl AsRef<Path>]` then converting those to owned values to create the struct.

crates/cxx-qt-build/src/lib.rs
ahayzen-kdab: Sure, doesn't look like this would break anything in @LeonMatthesKDAB 's diagnostics battle :-)

ahayzen-kdab (approved):
LGTM :-)

PR #627, opened 2023-07-26
cxx-qt-gen: add support for c++ only methods without qinvokable

### LINKED ISSUE #298: Defining C++ methods and Q_INVOKABLES from Rust
See #279

ahayzen-kdab:
Closes #298

### DIFF
--- a/crates/cxx-qt-gen/src/parser/invokable.rs
+++ b/crates/cxx-qt-gen/src/parser/invokable.rs
@@ -42,10 +42,12 @@ pub struct ParsedQInvokable {
     pub parameters: Vec<ParsedFunctionParameter>,
     /// Any specifiers that declared on the invokable
     pub specifiers: HashSet<ParsedQInvokableSpecifiers>,
+    /// Whether the method is qinvokable
+    pub is_qinvokable: bool,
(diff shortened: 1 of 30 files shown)


crates/cxx-qt-gen/src/parser/invokable.rs
```diff
@@ -42,18 +42,27 @@ pub struct ParsedQInvokable {
     pub parameters: Vec<ParsedFunctionParameter>,
     /// Any specifiers that declared on the invokable
     pub specifiers: HashSet<ParsedQInvokableSpecifiers>,
+    /// Whether the method is qinvokable
+    pub is_qinvokable: bool,
 }
```
Be-ing, line 47: This naming is now confusing. A struct called `ParsedQInvokable` shouldn't be able to maybe not be a QINVOKABLE. Can you come up with a new name for this? Maybe `ParsedMethod` or something?

ahayzen-kdab: Done renamed most of the invokable naming to method :-)

Be-ing (approved):

Issue #555, opened 2023-05-29
Stabilizing 1.0 API


LeonMatthesKDAB:
After discussions following #531, we believe we've found an improved API that we want to target for Cxx-Qt 1.0.

Anyone using Cxx-Qt, feel free to leave feedback on this issue.

## Solved API problems
- #527 - `extern "Qt"` block & `extern "RustQt"` block
- #531 -  you explicitly define the QObject type now!
- #529 - Documentation should now show up correctly
- #294 - Multiple bridges should no longer collide, as we no longer require the qobject:: prefix
- #404 - Results can now be represented the same way as in CXX

## Drawbacks
- Around 6 lines of boilerplate, compared to the current API for the example below

## 1.0 API Overview
The point of this API is to be even closer to CXX than before, by only putting declarations within the bridge itself.
We've found this preferable for `#[inherit]`, as well as for `#[qsignal]` for a while now, but haven't found a good API for QProperty.
We believe by simply using a `#[qproperty()]` attribute this can be solved nicely!
``` rust
// the module name now defines the module that your qobjects show up in.
#[cxxqt::bridge]
mod qobject {
    // Like an `extern "C++"` block, but every type is assumed to be a QObject/QGadget
    extern "Qt" {
        // Just passed through to the `extern "C++"` block without additional generation
        type QAbstractItemModel = cxx_qt_lib::QAbstractItemModel;

        type QButton;

        // Passed through straight to CXX + generates on_clicked/connect_clicked methods
        #[qsignal]
        fn clicked(self: &QButton);
    }

    // Any type in here is a QObject that comes from a Rust struct, like with `extern "Rust"`
    extern "RustQt" {
        // This mimicks C++ pretty closely, as everything is now split into its own attribute
        #[base="QAbstractItemModel"]
        #[qml_element]
        #[qproperty(i32, prop)] // needs to access field .prop
        #[qproperty(i32, prop, read=get_my_prop, write)]
        #[qproperty(QString, prop, read=get_my_prop, write=set_my_prop)]
        type MyObject = super::MyObject;
        // ^^^^^^^^^^^^^^^^^^^^^^^^^^ This solves the duality of types, you define a new type
        // that's within a `mod qobject` yourself, you can choose any name you like, etc.

        // invokables are now declared in the bridge, but defined outside.
        // That solves the issues we had with `Result<..>`
        #[qinvokable]
[...]

PR #634, opened 2023-07-27
cxx-qt-gen: use #[qobject] #[A] instead of #[cxx_qt::qobject(A)]

ahayzen-kdab:
Related to #555


ahayzen-kdab:
Changed to just `#[qobject]` for now, later we can potentially assume that all types in an extern "RustQt" block are QObjects, but this was a simpler first step without the debate and tidies up some code :-)

examples/qml_features/rust/src/custom_base_class.rs
```diff
@@ -34,7 +34,9 @@ pub mod qobject {
     // ANCHOR: book_inherit_qalm
     // ANCHOR: book_qobject_base
     extern "RustQt" {
-        #[cxx_qt::qobject(base = "QAbstractListModel", qml_element)]
```
ahayzen-kdab, line 37: assume we think that below is a better syntax ?

Or do we think that `#[qobject(base = "QAbstractListModel", qml_element)]` is better, but i think this is harder to do with syn only methods rather than our custom parsing that was removed in this change :-) 

The new syntax below isn't far off what you'd do in C++ anyway though as for that each of them is a separate macro
```cpp
class CustomBaseClass : public QAbstractListModel {
    Q_OBJECT
    QML_ELEMENT
    ...
}
```

ahayzen-kdab: It also matches CXX well with
```rust
#[namespace = "A"]
#[cxx_name = "B"]
#[rust_name = "C"]
type D = ...;
```

(and we already support `namespace` here, cxx_name / rust_name would be a further extension i'll add as another issue)

ahayzen-kdab: Done as #635

crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -66,31 +65,29 @@ pub struct ParsedQObject {
...
 
         // Load the namespace, if it is empty then the ParsedCxxQtData will inject any global namespace
-        let namespace = attrs_map
-            .get(&quote::format_ident!("namespace"))
-            .map_or_else(|| "".to_owned(), |base| base.value());
-
-        // Remove the macro from the type
-        let mut qobject_ty = qobject_ty.clone();
-        qobject_ty.attrs.remove(attr_index);
+        let namespace = if let Some(index) = attribute_find_path(&qobject_ty.attrs, &["namespace"])
```
LeonMatthesKDAB, line 84: It probably makes sense to introduce a `attribute_take_path` function.
e.g.:
``` rust
pub fn attribute_find_path(attrs: &mut Vec<Attribute>, path: &[&str]) -> Option<Attribute> { ... } 
```
Which would directly take the attribute out if it's found.
That would reduce duplicate code 4 times just in this file.

LeonMatthesKDAB (commented):
I think we can use Rust Options methods a bit more extensively to make this more readable.
Otherwise looks awesome though :)

crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -66,31 +65,29 @@ pub struct ParsedQObject {
...
-        attr_index: usize,
-    ) -> Result<Self> {
-        let qml_metadata = Self::parse_qml_metadata(qobject_ty, attr_index)?;
-
-        let attrs_map = attribute_tokens_to_map::<Ident, LitStr>(
-            &qobject_ty.attrs[attr_index],
-            AttributeDefault::Some(|span| LitStr::new("", span)),
-        )?;
+    /// Parse a ForeignTypeIdentAlias into a [ParsedQObject] with the index of the #[qobject] specified
+    pub fn from_foreign_item_type(qobject_ty: &ForeignTypeIdentAlias) -> Result<Self> {
```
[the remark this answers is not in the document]
ahayzen-kdab: done

crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -114,30 +111,42 @@ impl ParsedQObject {
...
+                false
+            };
+
+            // Determine if this element is a singleton
+            let singleton = if let Some(index) = attribute_find_path(attrs, &["qml_singleton"]) {
+                attrs.remove(index);
+                true
+            } else {
+                false
+            };
```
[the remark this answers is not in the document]
ahayzen-kdab: done

crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -114,30 +111,42 @@ impl ParsedQObject {
...
+
+            // Determine if this element is uncreatable
+            let uncreatable = if let Some(index) = attribute_find_path(attrs, &["qml_uncreatable"])
+            {
+                attrs.remove(index);
+                true
+            } else {
+                false
+            };
```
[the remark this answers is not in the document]
ahayzen-kdab: done

crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -66,31 +65,29 @@ pub struct ParsedQObject {
...
         // Find if there is any base class
-        let base_class = attrs_map
-            .get(&quote::format_ident!("base"))
-            .map(|base| base.value());
+        let base_class = if let Some(index) = attribute_find_path(&qobject_ty.attrs, &["base"]) {
+            let attr = qobject_ty.attrs.remove(index);
+            Some(expr_to_string(&attr.meta.require_name_value()?.value)?)
+        } else {
+            None
+        };
```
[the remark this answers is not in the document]
ahayzen-kdab: done

crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -66,31 +65,29 @@ pub struct ParsedQObject {
...
-        // Remove the macro from the type
-        let mut qobject_ty = qobject_ty.clone();
-        qobject_ty.attrs.remove(attr_index);
+        let namespace = if let Some(index) = attribute_find_path(&qobject_ty.attrs, &["namespace"])
+        {
+            let attr = qobject_ty.attrs.remove(index);
+            expr_to_string(&attr.meta.require_name_value()?.value)?
+        } else {
+            "".to_owned()
+        };
```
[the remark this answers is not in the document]
ahayzen-kdab: done, was missing bracket before transpose and used `unwrap_or_else` instead for lazy evaluation

crates/cxx-qt-gen/src/parser/qobject.rs
ahayzen-kdab: i'll add another commit with this

ahayzen-kdab: done

examples/qml_features/rust/src/custom_base_class.rs
LeonMatthesKDAB: I think for continuity with Qt it's nice to have these as separate attributes.
That would also allow us to automatically apply `#[qobject]` to all types in a `RustQt` bridge, as that would then just mean removing the `#[qobject]` attribute.

LeonMatthesKDAB (approved):
LGTM

PR #642, opened 2023-07-31
cxx-qt-gen: improve the generator / writer split

ahayzen-kdab:
This potentially helpers later with generic signal connections.

### DIFF
--- a/crates/cxx-qt-gen/src/generator/cpp/locking.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/locking.rs
@@ -1,0 +1,97 @@
+// SPDX-FileCopyrightText: 2023 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use crate::generator::{
+    cpp::{fragment::CppFragment, qobject::GeneratedCppQObjectBlocks},
+    naming::qobject::QObjectName,
+};
+use indoc::formatdoc;
+use syn::Result;
+
+pub fn generate(qobject_idents: &QObjectName) -> Result<(String, GeneratedCppQObjectBlocks)> {
+    let mut result = GeneratedCppQObjectBlocks::default();
+
+    let lock_guard_mutex = "::std::lock_guard<::std::recursive_mutex>";
+    let qobject_ident = qobject_idents.cpp_class.cpp.to_string();
+
+    result.includes.insert("#include <mutex>".to_owned());
+
+    result.private_methods.push(CppFragment::Pair {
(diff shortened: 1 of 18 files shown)


crates/cxx-qt-gen/src/generator/cpp/locking.rs
```diff
@@ -0,0 +1,42 @@
...
+pub fn generate(qobject_idents: &QObjectName) -> Result<(String, GeneratedCppQObjectBlocks)> {
+    let mut result = GeneratedCppQObjectBlocks::default();
+
+    let lock_guard_mutex = "::std::lock_guard<::std::recursive_mutex>";
+    let qobject_ident = qobject_idents.cpp_class.cpp.to_string();
+
+    result.includes.insert("#include <mutex>".to_owned());
+
+    result.methods.push(CppFragment::Pair {
+        header: format!("[[nodiscard]] {lock_guard_mutex} unsafeRustLock() const;"),
```
ahayzen-kdab, line 22: only differing part of this from a user point of view is the new `unsafeRustLock()` method, the idea is that templated methods could use this later (but I haven't got them working yet).

But maybe there is a usecase for wanting to lock a Rust obj via C++ anyway and it's not too bad to make it a public method.

ahayzen-kdab: Although it's looking like the templated connect for signals might not be possible, so maybe having this as a private method is better for now. I'll switch it to that, then we can make it public when we need it...

ahayzen-kdab: Done, made it private for now.

crates/cxx-qt-gen/test_outputs/properties.h
```diff
@@ -26,10 +26,11 @@ class MyObject : public QObject
 
 public:
   ~MyObject();
+
+public:
   MyObjectRust const& unsafeRust() const;
   MyObjectRust& unsafeRustMut();
 
-public:
```
Be-ing, line 32: Why do these need to become public?

ahayzen-kdab: They were public before ...

```cpp
public:
  ~MyObject();

  MyObjectRust const& unsafeRust() const;
  MyObjectRust& unsafeRustMut();

public:
  ::std::int32_t const& getPrimitive() const;
  Q_SLOT void setPrimitive(::std::int32_t const& value);
...
```

... and they are public after ?

```cpp
public:
  ~MyObject();

public:
  MyObjectRust const& unsafeRust() const;
  MyObjectRust& unsafeRustMut();

  ::std::int32_t const& getPrimitive() const;
  Q_SLOT void setPrimitive(::std::int32_t const& value);
```

Just the `unsafeRust` and `unsafeRustMut` are now generated in the same `public:` block as the `getPrimitive` whereas before they were separate.

Be-ing: ah, I didn't see that they were public before

Be-ing (approved):

Issue #625, opened 2023-07-26
qml_minimal not compiling on Windows


jacquetc:
Hello,

I'm trying out this nice crate. Already encountered issue #450 on MacOS (still investigating) and now trying it on  Windows 10. I followed step by step your book, then I tried qml_minimal directly. Nothing compile.

Any idea ?

Regards,
Cyril

**Steps to reproduce:** 
- Qt Official Installer (open source)
- Qt 5.6.1 MSVC2019 64bit
- Rust 1.71
- modify Cargo.toml to change versions and comment out the feature:
```
`# SPDX-FileCopyrightText: 2021 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
# SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
# SPDX-FileContributor: Gerhard de Clercq <gerhard.declercq@kdab.com>
#
# SPDX-License-Identifier: MIT OR Apache-2.0
# ANCHOR: book_all
[package]
name = "qml-minimal"
version = "0.1.0"
authors = [
  "Andrew Hayzen <andrew.hayzen@kdab.com>",
  "Gerhard de Clercq <gerhard.declercq@kdab.com>",
  "Leon Matthes <leon.matthes@kdab.com>"
]
edition = "2021"
license = "MIT OR Apache-2.0"

# This will instruct Cargo to create a static
# library which CMake can link against
# ANCHOR: book_static_lib
[lib]
crate-type = ["staticlib"]
# ANCHOR_END: book_static_lib

# ANCHOR: book_dependencies
[dependencies]
# Use `cxx = "1.0.95"` here instead!
cxx = "1.0.95"
# Use `cxx-qt = "0.5"` here instead!
cxx-qt = "0.5.3"
# Use `cxx-qt-lib = "0.5"` here instead!
cxx-qt-lib = "0.5.3"
# ANCHOR_END: book_dependencies

# cxx-qt-build generates C++ code from the `#[cxx_qt::bridge]` module
# and compiles it together with the Rust static library
# ANCHOR: book_build_dsependencies
[build-dependencies]
# Use `cxx-qt-build = "0.5"` here instead!
cxx-qt-build = "0.5.3"

[features]
# This feature must be enabled for `cargo test` when linking Qt 6 statically.
# link_qt_object_files = [ "cxx-qt-build/link_qt_object_files" ]
# ANCHOR_END: book_build_dependencies

# ANCHOR_END: book_all

[...]

PR #644, opened 2023-08-01
fix "file too big" error in MinGW Debug

jacquetc:
#625

### DIFF
--- a/crates/cxx-qt-build/src/lib.rs
+++ b/crates/cxx-qt-build/src/lib.rs
@@ -497,8 +497,10 @@ impl CxxQtBuilder {
             builder.flag_if_supported("/std:c++17");
             builder.flag_if_supported("/Zc:__cplusplus");
             builder.flag_if_supported("/permissive-");
+            builder.flag_if_supported("/bigobj");
             // GCC + Clang
             builder.flag_if_supported("-std=c++17");
+            // MinGW requires big-obj otherwise debug builds fail
             builder.flag_if_supported("-Wa,-mbig-obj");
             // Enable Qt Gui in C++ if the feature is enabled
             #[cfg(feature = "qt_gui")]
--- a/crates/cxx-qt-build/src/lib.rs
+++ b/crates/cxx-qt-build/src/lib.rs
@@ -499,6 +499,7 @@ impl CxxQtBuilder {
             builder.flag_if_supported("/permissive-");
             // GCC + Clang
             builder.flag_if_supported("-std=c++17");
+            builder.flag_if_supported("-Wa,-mbig-obj");
             // Enable Qt Gui in C++ if the feature is enabled
             #[cfg(feature = "qt_gui")]
             builder.define("CXX_QT_GUI_FEATURE", None);
--- a/crates/cxx-qt-lib/build.rs
+++ b/crates/cxx-qt-lib/build.rs
@@ -270,6 +270,7 @@ fn main() {
     builder.flag_if_supported("/permissive-");
     // GCC + Clang
     builder.flag_if_supported("-std=c++17");
+    builder.flag_if_supported("-Wa,-mbig-obj");
 
     builder.compile("cxx-qt-lib");
 }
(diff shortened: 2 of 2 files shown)


crates/cxx-qt-build/src/lib.rs
```diff
@@ -499,6 +499,7 @@ impl CxxQtBuilder {
             builder.flag_if_supported("/permissive-");
```
ahayzen-kdab, line 499: can you check with MSVC as well, do we need to enable `/bigobj` there for debug builds?

https://learn.microsoft.com/en-us/cpp/build/reference/bigobj-increase-number-of-sections-in-dot-obj-file?view=msvc-170

ahayzen-kdab: Or maybe this isn't a problem yet there, just curious how we are hitting it in MinGW debug builds.

crates/cxx-qt-lib/build.rs
```diff
@@ -270,6 +270,7 @@ fn main() {
     builder.flag_if_supported("/permissive-");
     // GCC + Clang
     builder.flag_if_supported("-std=c++17");
+    builder.flag_if_supported("-Wa,-mbig-obj");
```
ahayzen-kdab, line 273: could you add a comment above this line in both cases stating that `MinGW requires big-obj otherwise debug builds fail` or similar

crates/cxx-qt-build/src/lib.rs
jacquetc: I don't find any drawback to add it by default for MSVC. It only increases slightly the compilation time and can break some 3rd party debug tools if they don't support the biobj format

jacquetc: Better be safe than sorry. I pushed the change

crates/cxx-qt-lib/build.rs
jacquetc: Done

ahayzen-kdab:
Cool thanks! lets see what CI thinks.

And does this now mean that MinGW builds and works on Windows ? As i've then created another task #645 to create a CI runner for it, so we can ensure we don't regress :-)

jacquetc:
> Cool thanks! lets see what CI thinks.
> 
> And does this now mean that MinGW builds and works on Windows ? As i've then created another task #645 to create a CI runner for it, so we can ensure we don't regress :-)

That is a good assumption. I have still errors, but about my own code, so I can't be sure before fixing my own mess. :)

crates/cxx-qt-build/src/lib.rs
```diff
@@ -497,8 +497,11 @@ impl CxxQtBuilder {
             builder.flag_if_supported("/std:c++17");
             builder.flag_if_supported("/Zc:__cplusplus");
             builder.flag_if_supported("/permissive-");
+            builder.flag_if_supported("/bigobj");
```
ahayzen-kdab, line 500: `cargo fmt` is failing here because there is a space after the `;`

Be-ing:
> That is a good assumption. I have still errors, but about my own code, so I can't be sure before fixing my own mess. :)

Please try [building the examples in this repo](https://github.com/KDAB/cxx-qt#building) with MinGW so you don't have the uncertainty about your downstream code.

crates/cxx-qt-build/src/lib.rs
jacquetc: amended

Be-ing:
Can you explicitly confirm this does build now with MinGW? I'm unclear if that's what you meant by the :+1: on my last comment.

ahayzen-kdab:
Looks like you have local changes or something broken upstream in cc, but they haven't had a release in a while.

Could try `cargo update` and ensure you don't have any patches.

jacquetc:
I had first to delete the .cargo/registry and then it compiled. 
Qt 6 MingGW Debug : Ok
Qt 5 MinGW Debug : Ok

Be-ing (approved):

Be-ing:
Thanks for your contributions! I'm glad we have MinGW support now!!

LeonMatthesKDAB:
Right, don't feel obliged to continue on this issue, we can probably look into this ourselves, as the object files are likely just as big on linux. It's just something to keep in mind when we finally get around to optimizing for build times.

PR #646, opened 2023-08-02
cxx-qt-gen: move to clang-format 0.2 and drop ctor dependency

ahayzen-kdab:


crates/cxx-qt-gen/src/writer/cpp/mod.rs
```diff
@@ -35,8 +35,8 @@ pub fn write_cpp(generated: &GeneratedCppBlocks) -> CppFragment {
     let source = write_cpp_source(generated);
 
     CppFragment::Pair {
-        header: clang_format(&header).unwrap_or(header),
-        source: clang_format(&source).unwrap_or(source),
+        header: clang_format_with_style(&header, &ClangFormatStyle::File).unwrap_or(header),
```
LeonMatthesKDAB, line 38: A lot of code-duplication in this file.
Please introduce a function like this:
``` rust
fn format(source: &str) -> Result<_> {
    clang_format_with_style(source, &ClangFormatStyle::File)
}
```

ahayzen-kdab: done

LeonMatthesKDAB (approved):
LGTM

Issue #577, opened 2023-06-07
Add support for `extern "C++Qt"` with type, fn, and signals


ahayzen-kdab:
Our current generation is mostly about the `extern "RustQt"` side, once #557 is done we'll have `extern "RustQt"` with support for inherit and signals. Then #558 adds support for invokables and #559 for type and properties.

But we also want to support foreign types (eg `type QButton`), so we need to support `extern "C++Qt"`.

- [x] Add support in parser phase for C++Qt
- [x] Add support in generation phases for a type that is foreign (it only needs to generate CXX bridges and the signal connect parts?)
- [x] Ensure that `type QButton;`, `fn show(self: Pin<&mut QButton>);`, `#[qsignal] fn clicked(self: Pin<&mut QButton>)` work, but note that qproperty, inherit, qinvokable should not work ?
- [x] Change connections to be DirectConnection by default and make choosing a connection type as an unsafe closure (as QueuedConnection could have a mutable reference to something and be executed at any time in the future)

Consider any notes from #527

ahayzen-kdab:
<details>
<summary>C++ prototypes of the mixins / templates route</summary>

#include <chrono>
#include <functional>
#include <iostream>
#include <memory>
#include <optional>
#include <vector>

#include <mutex>
using namespace std;
using namespace std::chrono;

class Locking {
public:
  std::lock_guard<std::recursive_mutex> lock() {

    std::cout << "Locking" << std::endl;
    return std::lock_guard<std::recursive_mutex>(m_mutex);
  };

private:
  std::recursive_mutex m_mutex;
};

class Object {};

class MyObject : public Locking {};

template <typename T, typename Derived = void> struct MaybeLock {
  MaybeLock(T &) {}
};

template <typename T>
struct MaybeLock<T, std::enable_if_t<std::is_base_of_v<Locking, T>>> {
  MaybeLock(Locking &locking) : m_lock(locking.lock()) {}

  std::lock_guard<std::recursive_mutex> m_lock;
};

template <typename Object> void lock(Object &obj) {
  MaybeLock<Object> guard(obj);
}

template <typename Object, typename... Args>
std::function<void(Args...)>
templateConnect(Object &obj, void (Object::*signal)(Args...),
                std::function<void(Object &, Args...)> function) {
  return [&, func = std::move(function)](Args... args) {
    MaybeLock<Object> maybeLock(obj);

    func(obj, ::std::move(args)...);
  };
}

int main(int arg, char **argv) {
[...]

PR #648, opened 2023-08-03
cxx-qt-gen: use mixins for CxxQtType, CxxQtLocking, CxxQtThreading

ahayzen-kdab:
These mixins in the C++ side then match the traits we have on the Rust side.

This then means as a next phase we can have `MaybeLockGuard<T>` which can lock a given type if it derives from `CxxQtLocking` or do nothing if it doesn't. Which then means we can always put that in the code generation simplifying that further. Which then means for connection generation for signals of external types (like `QPushButton`) we don't need to have different generation for if it's a `CxxQtLocking` object or not :tada:  (expect this in further pull requests).

Related to #577

### DIFF
--- a/crates/cxx-qt-gen/src/generator/cpp/qobject.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/qobject.rs
@@ -139,9 +139,7 @@ impl GeneratedCppQObject {
 
         // If this type has locking enabled then add generation
         if qobject.locking {
-            let (initializer, mut blocks) = locking::generate(&qobject_idents)?;
-            generated.blocks.append(&mut blocks);
-            member_initializers.push(initializer);
+            generated.blocks.append(&mut locking::generate()?);
         }
 
         // If this type has threading enabled then add generation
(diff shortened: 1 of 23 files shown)


crates/cxx-qt-gen/src/generator/cpp/qobject.rs
```diff
@@ -27,10 +27,10 @@ pub struct GeneratedCppQObjectBlocks {
     pub private_methods: Vec<CppFragment>,
     /// List of members for the QObject
     pub members: Vec<String>,
-    /// List of deconstructor source
-    pub deconstructors: Vec<String>,
     /// List of includes
     pub includes: BTreeSet<String>,
+    /// Base class of the QObject
+    pub base_class: Vec<String>,
```
LeonMatthesKDAB, line 33:
```suggestion
    pub base_classes: Vec<String>,
```

ahayzen-kdab: done

crates/cxx-qt-gen/src/generator/cpp/qobject.rs
```diff
@@ -138,28 +132,42 @@ impl GeneratedCppQObject {
...
+            class_initializers.push(initializer);
+        // If this type has locking enabled then add generation
+        } else if qobject.locking {
+            let (initializer, mut blocks) = locking::generate()?;
+            generated.blocks.append(&mut blocks);
+            class_initializers.push(initializer);
         }
+
+        // We need the QObject base class to be first which is added last
+        generated.blocks.base_class.reverse();
```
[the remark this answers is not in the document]
ahayzen-kdab: done

LeonMatthesKDAB (approved):
LGTM

PR #651, opened 2023-08-04
cxx-qt-gen: add MaybeLockGuard which can optionally lock

ahayzen-kdab:
This means that all places where we need to lock can use this
and not have to worry if the type actually implements locking
or not. As MaybeLockGuard detects if the type has the
CxxQtLocking mixin.
    
This then simplifies connections later with extern C++ types
like QPushButton.
    
Related to #577

Requires #648


LeonMatthesKDAB (approved):
LGTM

PR #658, opened 2023-08-10
cxx-qt-gen: do not pass lock_guard around instead just generate it

ahayzen-kdab:
Later we have locks that require different parameters than *this so instead just generate them when needed which is easily done from the already known qobject_ident.

Related to #577


LeonMatthesKDAB (approved):
LGTM

PR #660, opened 2023-08-11
cxx-qt-gen: add C++ generator and writer of extern "C++Qt"

ahayzen-kdab:
Related to #577


crates/cxx-qt-gen/src/generator/cpp/signal.rs
```diff
@@ -60,6 +60,60 @@ fn parameter_types_and_values(
     })
 }
 
+/// Generate C++ blocks for a free signal on an existing object not a generated QObject, eg QPushButton::clicked
```
Be-ing, line 63:
```suggestion
/// Generate C++ blocks for a free signal on an existing QObject (not generated by CXX-Qt), eg QPushButton::clicked
```

crates/cxx-qt-gen/src/generator/cpp/signal.rs
```diff
@@ -60,6 +60,60 @@ fn parameter_types_and_values(
...
+    signal: &ParsedSignal,
+    cxx_mappings: &ParsedCxxMappings,
+) -> Result<CppFragment> {
+    // Prepare the idents we need
+    let qobject_ident = signal.qobject_ident.to_string();
+    let qobject_ident_namespaced = cxx_mappings.cxx(&qobject_ident);
+    let idents = QSignalName::from(signal);
+    let signal_ident = idents.name.cpp.to_string();
+    // TODO: in the future we might improve the naming of the methods
+    // to avoid collisions ideas include using a similar $ separator to CXX
```
Be-ing, line 74:
```suggestion
    // to avoid collisions (maybe use a $ separator similar to CXX?)
```

Be-ing (changes requested):
A few small suggestions to clarify the comments, otherwise looks good

crates/cxx-qt-gen/src/generator/cpp/signal.rs
ahayzen-kdab: done

Be-ing (approved):

Issue #616, opened 2023-07-18
Determine if passing through foreign funcs is worth it


ahayzen-kdab:
We can potentially pass through the input of certain items to the output with minimal modification

- [ ] Invokables
- [ ] Signals
- [ ] Inherit

Determine if it is worth it to simplify the parser or not, for signals #615 shows that this may not be worth the effort as there are other methods that need to be generated anyway and the method still needs extra attributes added to it.

LeonMatthesKDAB:
We've now gone the following route:

We never modify the input, but we still always fully reconstruct the output ourselves and never pass anything through.
This gives us certainty about our possible outputs so should be more stable and maintainable down the line.

Issue #595, opened 2023-06-26
Explore using Fn/FnOnce trampolines for queuing onto CxxQtThread and for signal connections


LeonMatthesKDAB:
See: https://github.com/dtolnay/cxx/issues/668

In the above discussion this approach is linked: https://adventures.michaelfbryan.com/posts/rust-closures-in-ffi/
It may save us from having to do so much heap-allocation and wrapping for executing closures in the cxx-qt-thread.
So explore if this is feasible and which (dis-)advantages it brings.

Ideally we could also use this approach to allow closures when connecting to signals.

ahayzen-kdab:
We also need to allow for closures (FnMut) in signal connections. Otherwise you cannot move a binding into the closure to use in the connection. Eg the following doesn't work.

```rust
let qt_thread = self.qt_thread();
anotherObject.on_signal(|ctx| {
  // cannot move the qt_thread here
  // as the on_signal is just a function ptr ( fn )
  // instead of a closure ( FnMut )
  qt_thread.queue(|ctx| { ... });
});
```

Issue #319, opened 2022-10-14
Make cxx-qt-lib truely optional or depend on it with cxx-qt or does it work currently?


ahayzen-kdab:
Currently cxx-qt-gen assumes that headers such as `convert.h` and `cxxqt_thread.h` will exist, these currently come via cxx-qt-build.

- [x] Ensure that we can build without cxx-qt-lib
  - [x] cxx-qt-gen depends on `cxx_qt_lib::ConnectionType` and `cxx_qt_lib::QMetaObjectConnection` for signals :thinking: 
- [ ] Decide if these header only includes that we always assume exist are in the correct place
- [ ] Check if we need any further linking assumptions (just QtCore?)
- [ ] Can this then be used to register the std numeric types for Qt 5 ? #317 
- [x] Ensure that cxx-qt-build can take a key-value collection of headers and put them in a folder that is in the include target

Be-ing:
What would be the benefit of this, even if it is possible?

ahayzen-kdab:
You could then use our generation with other Qt / CXX libraries and also then use cxx-qt-lib with other C++ binding tools. At the moment cxx-qt-gen assumes that the convert and cxxqtthread headers are coming from somewhere. And if you are using cxx-qt-lib by itself you don't need the convert and cxxqtthread.

So we should ensure that the positioning of these headers is correct, any dependencies are correct, and that each can be built on it's own. Lots of it looks correct already, but having a full and clear separation between layers could prove useful later.

ahayzen-kdab:
Related to #355 but lets investigate in a future cycle.

ahayzen-kdab:
If we ever did this we would likely want to consider moving `cxx_qt_lib_headers::write_headers(format!("{header_root}/cxx-qt-lib"));` out of `cxx-qt-build` and have this as something in the build script or something.

```rust
CxxQtBuilder::new()
   .with_headers("cxx-qt-lib", cxx_qt_lib::headers()
```

Eg something like the above to pass a key-value collection of headers with a folder prefix that should be written somewhere that can then be included.

This has the advantage that developers can write their own independent crates in the same way and they can still be used without cxx-qt-build and cxx-qt-lib being tightly coupled.

Something we should consider if / how we wanted to do this before 1.0 so tagging to remember.

ahayzen-kdab:
[...]

PR #662, opened 2023-08-11
WIP: cxx-qt-gen: add Rust generator and writer of extern "C++Qt"

ahayzen-kdab:
Related to #577


crates/cxx-qt-gen/src/generator/rust/externcxxqt.rs
```diff
@@ -0,0 +1,89 @@
...
+                cpp: format_ident!("{}_{}Connect", signal.qobject_ident, signal.ident.cpp),
+                rust: format_ident!("{}_{}Connect", signal.qobject_ident, signal.ident.rust),
+            };
+            generated.append(
+                &mut generate_rust_signal(
+                    signal,
+                    &signal.qobject_ident,
+                    qualified_mappings,
+                    Some(free_connect_ident),
+                    // TODO: inject correct value here
```
ahayzen-kdab, line 80: need to build correct namespace here

crates/cxx-qt-gen/src/generator/rust/externcxxqt.rs
```diff
@@ -0,0 +1,89 @@
...
+            implementation: vec![],
+        };
+        generated
+            .cxx_mod_contents
+            .append(&mut fragment.cxx_bridge_as_items()?);
+
+        // Build the signals
+        for signal in &extern_cxxqt_block.signals {
+            let free_connect_ident = CombinedIdent {
+                cpp: format_ident!("{}_{}Connect", signal.qobject_ident, signal.ident.cpp),
```
ahayzen-kdab, line 71: could probably use naming module here

crates/cxx-qt-gen/src/generator/rust/signals.rs
```diff
@@ -16,107 +16,151 @@ use crate::{
...
+    let self_type_cxx = if signal.mutable {
+        parse_quote! { Pin<&mut #qobject_name> }
+    } else {
+        parse_quote! { &#qobject_name }
+    };
+    let self_type_qualified = syn_type_cxx_bridge_to_qualified(&self_type_cxx, qualified_mappings);
+    let qualified_impl = syn_ident_cxx_bridge_to_qualified_impl(qobject_name, qualified_mappings);
+    let self_call = if free_connect_ident.is_some() {
+        // TODO: we need the module ident here
+        quote! { qobject:: }
```
ahayzen-kdab, line 71: need to use the module ident here

crates/cxx-qt-gen/src/generator/rust/signals.rs
```diff
@@ -16,107 +16,151 @@ use crate::{
...
 use syn::{parse_quote, FnArg, Ident, Path, Result};
 
-pub fn generate_rust_signals(
-    signals: &Vec<ParsedSignal>,
-    qobject_idents: &QObjectName,
+pub fn generate_rust_signal(
+    signal: &ParsedSignal,
+    qobject_name: &Ident,
     qualified_mappings: &BTreeMap<Ident, Path>,
+    free_connect_ident: Option<CombinedIdent>,
```
ahayzen-kdab, line 23: maybe there is a better way this can be done, similar to the cpp generator route?

crates/cxx-qt-gen/src/generator/rust/externcxxqt.rs
ahayzen-kdab: might need to pass in full mappings rather than just qualified?

ahayzen-kdab: works, need to share in naming module with C++ gen

ahayzen-kdab:
No example yet as we need to add support for closures in signal connection as well as function pointers to make them useful. This will likely be similar to how threading is done and #595 can be reused

crates/cxx-qt-gen/src/generator/rust/signals.rs
```diff
@@ -11,11 +11,137 @@ use crate::{
...
+
+    let attrs = &signal.method.attrs;
+
+    let fragment = RustFragmentPair {
+        cxx_bridge: vec![
+            quote! {
+                #unsafe_block extern "C++" {
+                    #(#attrs)*
+                    #[rust_name = #signal_name_rust_str]
+                    #unsafe_call fn #signal_name_cpp(self: #self_type_cxx, #(#parameters_cxx),*);
```
LeonMatthesKDAB, line 91: Doesn't this basically re-generate the declaration the user wrote initially? As emitting the signal is always just a normal function call that we need to bind to.

Could you try just removing the `#[qsignal]` specifier and then adding the ForeignItemFn the the passthrough_items?
As then we could get rid of this generation here.

crates/cxx-qt-gen/src/generator/rust/signals.rs
```diff
@@ -11,11 +11,137 @@ use crate::{
...
+                    pub fn #connect_ident_rust(self: #self_type_qualified, func: fn(#self_type_qualified, #(#parameters_qualified),*)) -> cxx_qt_lib::QMetaObjectConnection
+                    {
+                        #module_ident::#free_connect_ident_rust(self, func, cxx_qt_lib::ConnectionType::AutoConnection)
+                    }
+                }
+            },
+        ],
+    };
+
+    let mut generated = GeneratedRustQObject::default();
```
LeonMatthesKDAB, line 134: Why doesn't this just return a `GeneratedExternCxxQt` directly?

crates/cxx-qt-gen/test_outputs/passthrough_and_naming.rs
```diff
@@ -208,6 +208,59 @@ pub mod ffi {
...
+    unsafe extern "C++" {
+        #[rust_name = "error_occurred"]
+        #[rust_name = "errorOccurred"]
+        fn errorOccurred(self: Pin<&mut ExternObject>);
+    }
+    unsafe extern "C++" {
+        #[doc(hidden)]
+        #[namespace = "rust::cxxqtgen1::externcxxqt::mynamespace"]
+        #[must_use]
+        # [rust_name = ExternObject_connect_error_occurred]
```
LeonMatthesKDAB, line 257: This will probably cause a clippy lint or even compiler warning about "function name should be snake_case"

crates/cxx-qt-gen/src/generator/rust/signals.rs
```diff
@@ -11,11 +11,137 @@ use crate::{
...
+            },
+            quote! {
+                impl #qualified_impl {
+                    #[doc = "Connect the given function pointer to the signal "]
+                    #[doc = #signal_name_cpp_str]
+                    #[doc = ", so that when the signal is emitted the function pointer is executed."]
+                    #[must_use]
+                    pub fn #connect_ident_rust(self: #self_type_qualified, func: fn(#self_type_qualified, #(#parameters_qualified),*)) -> cxx_qt_lib::QMetaObjectConnection
+                    {
+                        #module_ident::#free_connect_ident_rust(self, func, cxx_qt_lib::ConnectionType::AutoConnection)
```
LeonMatthesKDAB, line 127: This is the same function definition for `connect_X` as it is for `on_X`!
Make sure `connect` actually allows passing in a connection type!

crates/cxx-qt-gen/src/generator/rust/signals.rs
```diff
@@ -11,11 +11,137 @@ use crate::{
...
+                impl #qualified_impl {
+                    #[doc = "Connect the given function pointer to the signal "]
+                    #[doc = #signal_name_cpp_str]
+                    #[doc = ", so that when the signal is emitted the function pointer is executed."]
+                    #[doc = "\n"]
+                    #[doc = "Note that this method uses a AutoConnection connection type."]
+                    #[must_use]
+                    pub fn #on_ident_rust(self: #self_type_qualified, func: fn(#self_type_qualified, #(#parameters_qualified),*)) -> cxx_qt_lib::QMetaObjectConnection
+                    {
+                        #module_ident::#free_connect_ident_rust(self, func, cxx_qt_lib::ConnectionType::AutoConnection)
```
LeonMatthesKDAB, line 115: Note for future:
ConnectionType needs to be in cxx_qt, not cxx_qt_lib.

crates/cxx-qt-gen/src/generator/naming/signals.rs
```diff
@@ -25,6 +25,17 @@ impl From<&ParsedSignal> for QSignalName {
     }
 }
 
+impl QSignalName {
+    pub fn free_namespace_from_qobject_namespace(qobject_namespace: Option<&String>) -> String {
```
LeonMatthesKDAB, line 29: Could you move this up into the `naming/mod.rs` file?
As it doesn't really have anything to do with QSignal, but is more generally usable as the "private" namespace for our generation.

crates/cxx-qt-gen/src/generator/rust/signals.rs
ahayzen-kdab: Yup this would be as part of #319 and once we have packed_bundled_libs support in rustc i think this becomes easier as well.

crates/cxx-qt-gen/src/generator/rust/signals.rs
ahayzen-kdab: Oops, done!

crates/cxx-qt-gen/src/generator/rust/signals.rs
ahayzen-kdab: done

crates/cxx-qt-gen/src/generator/rust/signals.rs
ahayzen-kdab: I think we can in this case potentially easily ... Note that #616 is where i'll look further into it for other places as well.

I'll see if i can make the parser happy for the generation here.

ahayzen-kdab: done

crates/cxx-qt-gen/test_outputs/passthrough_and_naming.rs
ahayzen-kdab: Even using #670 and running clippy in the example doesn't complain :thinking:

LeonMatthesKDAB: Alright, then it's fine.
We can always fix if we do find anything.

LeonMatthesKDAB (approved):
LGTM

Issue #34, opened 2021-12-09
Consider how to support enums through to QML


ahayzen-kdab:
We need to support enums to allow for complex objects like QAbstractListModel etc.

- [x] Enums on QObjects from Rust
- [x] Enums in a namespace from Rust
- [x] Existing enums (via CXX)

ahayzen-kdab:
I need to try this again but i almost got it working before using `Q_ENUM_NS` from C++ as a workaround until we do add support with an attribute to shared enums. (eg #402 will hopefully have helped with that)

LeonMatthesKDAB:
We should be able to support both `Q_ENUM`, as well as `Q_ENUM_NS` using [CXX Extern enums](https://cxx.rs/shared.html#extern-enums).

e.g.:
``` cpp
#pragma once

#include <cstdint>

struct MyObject {
  enum class MyEnum : std::int8_t { A, B };
};

namespace rust {
namespace MyObject {
using MyEnum = ::MyObject::MyEnum;
}
} // namespace rust
```

```rust
#[cxx::bridge]
pub mod ffi {

    #[cxx_name = "MyEnum"]
    #[namespace = "rust::MyObject"]
    enum MyEnum {
        A,
        B,
    }

    unsafe extern "C++" {
        include!("src/myobject.h");

        #[cxx_name = "MyEnum"]
        #[namespace = "rust::MyObject"]
        type MyEnum;
    }
}
```

Then we would support these two kinds of declarations:
```rust
#[qenum]
#[namespace="MyNamespace"]
enum MyEnum { ... }
```
```rust
#[qenum(MyObject)]
enum MyEnum { ... }

extern "RustQt" {
    type MyObject = ...;
}
```

LeonMatthesKDAB:
It turns out we can use QML_ELEMENT with Q_NAMESPACE to expose a Q_ENUM_NS it to QML.

See: https://doc.qt.io/qt-6/qqmlengine.html#QML_ELEMENT

LeonMatthesKDAB:
Done

ahayzen-kdab:
This is done.

PR #663, opened 2023-08-16
Add support for #[qenum] associated with a QObject

LeonMatthesKDAB:
This is the initial support for #34.

It only supports `Q_ENUM` within a given `Q_OBJECT` class. `Q_ENUM_NS` is not (yet) supported.
Documentation will be added in a follow-up commit.

### DIFF
--- a/crates/cxx-qt-gen/src/generator/cpp/mod.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/mod.rs
@@ -11,6 +11,7 @@ pub mod inherit;
 pub mod locking;
 pub mod method;
 pub mod property;
+pub mod qenum;
 pub mod qobject;
 pub mod signal;
 pub mod threading;
--- a/crates/cxx-qt-gen/src/generator/cpp/qenum.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/qenum.rs
@@ -1,0 +1,111 @@
+// SPDX-FileCopyrightText: 2023 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Leon Matthes <leon.matthes@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use indoc::formatdoc;
+use syn::Result;
+
+use crate::{
+    generator::utils::cpp::Indent,
+    parser::{mappings::ParsedCxxMappings, qenum::ParsedQEnum},
+};
+
+use super::qobject::GeneratedCppQObjectBlocks;
+
+pub fn generate(
+    qenums: &[ParsedQEnum],
+    cxx_mappings: &ParsedCxxMappings,
+) -> Result<GeneratedCppQObjectBlocks> {
+    let mut generated = GeneratedCppQObjectBlocks::default();
+
+    for qenum in qenums {
+        let enum_name = &qenum.ident.to_string();
+
+        let mut qualified_name = cxx_mappings.cxx(enum_name);
+        // TODO: this is a workaround for cxx_mappings.cxx not always returning a fully-qualified
+        // identifier.
+        // Once https://github.com/KDAB/cxx-qt/issues/619 is fixed, this can be removed.
+        if !qualified_name.starts_with("::") {
+            qualified_name.insert_str(0, "::");
+        }
+
+        let enum_values = qenum
+            .variants
+            .iter()
+            .map(ToString::to_string)
+            .collect::<Vec<_>>()
+            .join(",\n");
+
+        generated.includes.insert("#include <cstdint>".to_string());
+        let enum_definition = formatdoc! { r#"
+            enum class {enum_name} : ::std::int32_t {{
+            {enum_values}
+            }};
+        "#, enum_values = enum_values.indented(2) };
+        generated.forward_declares.push(enum_definition.clone());
+        generated.metaobjects.push(formatdoc! {r#"
+            #ifdef Q_MOC_RUN
+            {enum_definition}
+              Q_ENUM({enum_name})
+            #else
+              using {enum_name} = {qualified_name};
+              Q_ENUM({enum_name})
+            #endif
+        "#, enum_definition = enum_definition.indented(2) });
+    }
+
+    Ok(generated)
+}
+
+#[cfg(test)]
+mod tests {
+    use std::assert_eq;
+
+    use super::*;
+    use indoc::indoc;
+    use pretty_assertions::assert_str_eq;
+    use syn::parse_quote;
+
+    #[test]
+    fn generates() {
+        let qenums = [ParsedQEnum::parse(parse_quote! {
+            enum MyEnum {
+                A, B, C
+            }
+        })
+        .unwrap()];
+
+        let generated = generate(&qenums, &ParsedCxxMappings::default()).unwrap();
+        assert_eq!(generated.includes.len(), 1);
+        assert!(generated.includes.contains("#include <cstdint>"));
+        assert_eq!(generated.metaobjects.len(), 1);
+        assert_str_eq!(
+            indoc! {r#"
+                #ifdef Q_MOC_RUN
+                  enum class MyEnum : ::std::int32_t {
+                    A,
+                    B,
+                    C
+                  };
+                  Q_ENUM(MyEnum)
+                #else
+                  using MyEnum = ::MyEnum;
+                  Q_ENUM(MyEnum)
+                #endif
+            "#},
+            generated.metaobjects[0],
+        );
+        assert_eq!(generated.forward_declares.len(), 1);
+        assert_str_eq!(
+            indoc! { r#"
+                enum class MyEnum : ::std::int32_t {
+                  A,
+                  B,
+                  C
+                };
+            "# },
+            generated.forward_declares[0],
+        );
+    }
+}
--- a/crates/cxx-qt-gen/src/parser/qenum.rs
+++ b/crates/cxx-qt-gen/src/parser/qenum.rs
@@ -1,0 +1,163 @@
+// SPDX-FileCopyrightText: 2023 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Leon Matthes <leon.matthes@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use quote::ToTokens;
+use syn::{Ident, ItemEnum, Result, Variant};
+
+use crate::syntax::path::path_compare_str;
+
+pub struct ParsedQEnum {
+    /// The ident of the QEnum
+    pub ident: Ident,
+    /// the values of the QEnum
+    pub variants: Vec<Ident>,
+    /// The original enum item
+    pub item: ItemEnum,
+}
+
+impl ParsedQEnum {
+    fn parse_variant(variant: &Variant) -> Result<Ident> {
+        fn err(spanned: &impl ToTokens, message: &str) -> Result<Ident> {
+            Err(syn::Error::new_spanned(spanned, message))
+        }
+
+        if !variant.fields.is_empty() {
+            return err(variant, "QEnum variants cannot have fields");
+        }
+        if let Some(attr) = variant
+            .attrs
+            .iter()
+            .find(|attr| !path_compare_str(attr.path(), &["doc"]))
+        {
+            return err(
+                attr,
+                "QEnum variants can only have #[doc=\"...\"] attributes",
+            );
+        }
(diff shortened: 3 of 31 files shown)


crates/cxx-qt-gen/src/writer/cpp/mod.rs
```diff
@@ -300,10 +300,10 @@ mod tests {
         class MyObject : public QStringListModel
         {
           Q_OBJECT
+        public:
           Q_PROPERTY(int count READ count WRITE setCount NOTIFY countChanged)
```
ahayzen-kdab, line 304: why are the metaobjects now public ? Normally these are above aren't they ? 

eg https://doc.qt.io/qt-6/properties.html#a-simple-example shows the `Q_OBJECT`, `Q_PROPERTY` above the `public:` and the enum and `Q_ENUM` below ?

Maybe just adding to `methods` is the right way to do the enum? even though it isn't a "method" but it's something in the public space ? (maybe we need a better name there)

examples/qml_features/rust/src/custom_base_class.rs
```diff
@@ -31,6 +31,15 @@ pub mod qobject {
         type QVector_i32 = cxx_qt_lib::QVector<i32>;
     }
 
+    #[qenum(CustomBaseClass)]
```
ahayzen-kdab, line 34: would be good to show this in use in QML and also to have a unit test in the QML tests (the test being the most important)

You could be able to try doing `model.data(index, CustomBaseClass.Value);` or something somewhere in JavaScript to demo that.

crates/cxx-qt-gen/src/writer/cpp/mod.rs
LeonMatthesKDAB: The reason for this is that we really want the enum class itself (or rather the `using` directive for it) to be public in the class (as you correctly guessed).

I've considered putting the `Q_ENUM` into another part of the generation, but honestly that seemed like a bit of a hack.
As a `Q_ENUM` is as much a meta-object as `Q_PROPERTY` I think it should go into the same category.

So the simplest solution is to move all metaobjects into the public category.
This doesn't change any behavior, as Q_PROPERTY is just defined as nothing in C++.
And semantically one could argue that it's correct anyhow, as Q_PROPERTY values are indeed publicly accessible.
It does go a bit against convention, though there don't seem to be any strong guidelines or reasoning that support this convention.

examples/qml_features/rust/src/custom_base_class.rs
LeonMatthesKDAB: I'll try adding a test to the QML tests.

LeonMatthesKDAB: Done.

crates/cxx-qt-gen/test_outputs/qenum.h
```diff
@@ -0,0 +1,47 @@
+#pragma once
+
+#include <cstdint>
+#include <cxx-qt-common/cxxqt_locking.h>
+#include <cxx-qt-common/cxxqt_maybelockguard.h>
+#include <cxx-qt-common/cxxqt_type.h>
+
+namespace cxx_qt::my_object {
+class MyObject;
+namespace cxx_qt_qenums {
```
ahayzen-kdab, line 10: as discussed there might be some namespace changes?

examples/qml_features/rust/src/custom_base_class.rs
Be-ing: Hmm... it kinda bugs me that `CustomBaseClass` isn't in scope here because the type alias is declared in the `extern "RustQt"` block below, making this feel a bit magical. We could move this into the `extern "RustQt"` block, but that would break from upstream CXX where shared enums are outside the `extern` blocks... thoughts? I don't have a strong opinion either way.

ahayzen-kdab: Isn't this the same as CXX though? you can define a shared enum or struct and use it in another block below or above ? The `type T` is enough to make that `T` in scope for the whole bridge?

And the enum here was originally trying to emulate the shared enums of CXX but for Qt.

The Q_ENUM_NS variant (coming in a later PR?) would just be `#[qenum]` however :-)

crates/cxx-qt-gen/src/writer/cpp/mod.rs
ahayzen-kdab: Note that `Q_OBJECT` and `Q_GADGET` expand to something that has as `private:` at the end, we would need to assume here that `Q_PROPERTY` would not also have any `private:` or `protected:` in it's expansion (which it doesn't currently).

LeonMatthesKDAB: As that would be breaking change, this can't happen until Qt7, so let's not worry about this for now.

crates/cxx-qt-gen/test_outputs/qenum.h
LeonMatthesKDAB: Done.

The enum now doesn't modify the namespace at all.
So the user get's full choice of the namespace the enum ends up in.

It now acts the same as declaring an enum in CXX normally would, but additionally registers the enum as a Q_ENUM inside the given QObject.
This seems to be the best way to keep compatibility with both CXX and Qt behavior.

As that may polute the global namespace with QEnum definitions, we should add support for the `#[namespace="..."]` attribute on qenums in a follow-up commit.
However, that requires fixing a few other bugs, including that `#[namespace="..."]` on a `#[qobject]` results in incorrect code generation in C++.

crates/cxx-qt-gen/src/generator/cpp/qenum.rs
```diff
@@ -0,0 +1,111 @@
...
+) -> Result<GeneratedCppQObjectBlocks> {
+    let mut generated = GeneratedCppQObjectBlocks::default();
+
+    for qenum in qenums {
+        let enum_name = &qenum.ident.to_string();
+
+        let mut qualified_name = cxx_mappings.cxx(enum_name);
+        // TODO: this is a workaround for cxx_mappings.cxx not always returning a fully-qualified
+        // identifier.
+        // Once https://github.com/KDAB/cxx-qt/issues/619 is fixed, this can be removed.
```
ahayzen-kdab, line 28: Right, one idea i had was to make the `cxx` method always add a `::` or to add every type even if it doesn't have a namespace to the namespaces mappings.

But likely something to improve with the tree structure again too.

crates/cxx-qt-gen/src/parser/qenum.rs
```diff
@@ -0,0 +1,162 @@
...
+    pub ident: Ident,
+    /// the values of the QEnum
+    pub variants: Vec<Ident>,
+    /// The original enum item
+    pub item: ItemEnum,
+}
+
+impl ParsedQEnum {
+    fn parse_variant(variant: &Variant) -> Result<Ident> {
+        fn err(spanned: &impl ToTokens, message: &str) -> Result<Ident> {
```
ahayzen-kdab, line 22: what the reason for this being a `fn` rather than a closure ? like `let err = |spanned: T, message: T| -> T { ... }` ?

examples/qml_features/tests/tst_custom_base_class.qml
```diff
@@ -82,4 +82,9 @@ TestCase {
         compare(instantiator.count, 1);
         compare(dataChangedSpy.count, 1);
     }
+
+    function test_roles_qenum() {
+        compare(CustomBaseClass.Id, 0);
+        compare(CustomBaseClass.Value, 1);
```
ahayzen-kdab, line 88: Can you add this code to this test, this ensures that enums work in QML through the abstract list model. And show how you can use the imperative API to read the model from QML :-) 

```
    function test_roles_qenum() {
        compare(CustomBaseClass.Id, 0);
        compare(CustomBaseClass.Value, 1);

        const model = createTemporaryObject(componentCustomBaseClass, null, {});

        model.add();
        compare(model.data(model.index(0, 0), CustomBaseClass.Id), 0);
        compare(model.data(model.index(0, 0), CustomBaseClass.Value), 0);

        model.add();
        compare(model.data(model.index(1, 0), CustomBaseClass.Id), 1);
        compare(model.data(model.index(1, 0), CustomBaseClass.Value), 1.0 / 3.0);
    }
```

crates/cxx-qt-gen/src/generator/cpp/mod.rs
```diff
@@ -11,6 +11,7 @@ pub mod inherit;
 pub mod locking;
 pub mod method;
 pub mod property;
+pub mod qenum;
```
ahayzen-kdab, line 14: Maybe add an entry into the CHANGELOG as well ?

crates/cxx-qt-gen/src/generator/cpp/qenum.rs
LeonMatthesKDAB: Right, I've actually tried doing that, but that generated stuff like: `:: ::std::unique_ptr`, so I think we should do this properly in a follow-up commit.

ahayzen-kdab: Right, separate commit and tracked with that issue :-)

crates/cxx-qt-gen/src/parser/qenum.rs
LeonMatthesKDAB: You can't do `&impl ToTokens` in a closure unfortunately, as that's using generics.

ahayzen-kdab: Aha thought there was a reason, shame :-/

crates/cxx-qt-gen/src/parser/qenum.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    pub fn parse(qenum: ItemEnum) -> Result<Self> {
+        if qenum.variants.is_empty() {
+            return Err(syn::Error::new_spanned(
+                qenum,
+                "QEnum must have at least one variant",
+            ));
+        }
+
+        if let Some(attr) = qenum
```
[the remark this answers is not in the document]
LeonMatthesKDAB: Yeah, probably.
Done

crates/cxx-qt-gen/src/generator/cpp/mod.rs
LeonMatthesKDAB: I don't want to do that yet, as with Constructor, I think it makes sense to merge this change, but not make it "public" in the CHANGELOG until we've added documentation, support for `namespace`, etc.

examples/qml_features/tests/tst_custom_base_class.qml
LeonMatthesKDAB: Done.
Also added another QEnum for the background thread state and tested that here.

examples/qml_features/qml/pages/CustomBaseClassPage.qml
```diff
@@ -76,4 +77,13 @@ Page {
...
         }
     }
+
+    BusyIndicator {
+        anchors {
+            right: scrollView.right
+            bottom: scrollView.bottom
+            margins: 15
+        }
+        running: customBaseClass.state == CustomBaseClass.Running
```
ahayzen-kdab, line 87: this should work with triple equals and would be good to test that it does 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality

examples/qml_features/tests/tst_custom_base_class.qml
```diff
@@ -82,4 +82,26 @@ TestCase {
...
+        model.add();
+        compare(model.data(model.index(1, 0), CustomBaseClass.Id), 1);
+        compare(model.data(model.index(1, 0), CustomBaseClass.Value), 1.0 / 3.0);
+
+        compare(model.state, CustomBaseClass.Idle);
+        model.addOnThread(1);
+        // This shouldn't cause a flaky test, as the `addOnThread` cannot reset the
+        // state to `Idle` until the Event-Loop of this thread is available again.
+        // Which should only be after this thread has finished.
+        compare(model.state, CustomBaseClass.Running);
```
ahayzen-kdab, line 105: can you add a further line to the end to check that we eventually go back to idle?

```
tryCompare(model.state, CustomBaseClass.Idle);
```

LeonMatthesKDAB: Done.
I added a method to configure the delay, so we can use 0 ms in the test to not artificially inflate test runtimes.
Because the qt_thread queues something for the main thread event loop, the check to Running should still work.

ahayzen-kdab (approved):
LGTM :-)

PR #667, opened 2023-08-24
WIP: cxx-qt-gen: use original bridge definition in CXX and remove camel/snake conversion

### LINKED ISSUE #532: Consider allowing `#[doc]` in `cxx_qt::inherit` blocks
In #518 we start generating docs for any methods that are inherited, but instead we should consider if supporting passing through documentation instead and require the developer to document their inherit methods.

Currently if one tries to do this our generator states that the doc attribute is not supported.

### LINKED ISSUE #616: Determine if passing through foreign funcs is worth it

ahayzen-kdab:
Closes #616

Closes #532

### DIFF
--- a/crates/cxx-qt-gen/src/generator/cpp/method.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/method.rs
@@ -169,7 +169,10 @@ mod tests {
     fn test_generate_cpp_invokables() {
         let invokables = vec![
             ParsedMethod {
-                method: parse_quote! { fn void_invokable(self: &MyObject); },
+                method: parse_quote! {
+                    #[cxx_name = "voidInvokable"]
+                    fn void_invokable(self: &MyObject);
+                },
                 qobject_ident: format_ident!("MyObject"),
                 mutable: false,
                 safe: true,
@@ -178,7 +181,10 @@ mod tests {
                 is_qinvokable: true,
             },
             ParsedMethod {
-                method: parse_quote! { fn trivial_invokable(self: &MyObject, param: i32) -> i32; },
+                method: parse_quote! {
+                    #[cxx_name = "trivialInvokable"]
+                    fn trivial_invokable(self: &MyObject, param: i32) -> i32;
+                },
                 qobject_ident: format_ident!("MyObject"),
                 mutable: false,
                 safe: true,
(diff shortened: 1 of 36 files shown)


ahayzen-kdab:
This works quite well from a generation side, only the `qproperty` is potentially fun to sort out syntax wise.

However for `qproperty` we do not pass it straight through and generate methods from it :thinking: 

The syntax could be something like
```rust
#[qproperty(i32, my_property, cxx_name = "myProperty")]
#[qproperty(i32, myProperty, rust_name = "myProperty")]
#[qproperty(i32, my_property, cxx_name = "myProperty", read = Self::custom_read)]
```

But note that we might have custom getters and setters in there later too :thinking: 

Another route could be to require to specify the "name" in C++ as the property name and then specify the field it maps to (`Name = Self::Field`) eg
```rust
#[qproperty(i32, myProperty = Self::my_property)]
#[qproperty(i32, myProperty = Self::my_property, read = Self::custom_read)]
```

Then it could also fallback to if you don't specify the right side it assumes that they are the same
```rust
#[qproperty(i32, myProperty)]  // would look for Self::myProperty
```

Or a further idea could be to make everything key-values, combine the C++ name with the type
```rust
#[qproperty(myProperty = i32, member = Self::myProperty)]
#[qproperty(myProperty = i32, read = Self::custom_read)]
```

Or if we can use the generic syntax

```rust
#[qproperty(myProperty<i32>, member = Self::myProperty)]
#[qproperty(myProperty<i32>, read = Self::custom_read)]
```

Which could also drop the Self

```rust
#[qproperty(myProperty<i32>, member = myProperty)]
#[qproperty(myProperty<i32>, read = custom_read)]
```

LeonMatthesKDAB:
Hm, I do like the automatic camel-case/snake-case conversion, as snake-case code does look very out of place in QML/C++.
Though that's potentially just me not being used to it.

I believe we could still make the passthrough work with case conversion, by simply checking whether there's a `cxx_name` or `rust_name` associated with the `qinvokable` and if not, injecting only that attribute and leaving the rest alone, which would still be basically a pass-through, with just one additional attribute.

It's something that we should discuss at the company meeting, whether people think camelCase is important in QML.

LeonMatthesKDAB:
Another suggestion for naming of properties:
```rust
    extern "RustQt" {
        #[qobject]
        #[qproperty(
            i32, my_property,
            cxx_name="my_prop",
            setter=set_my_prop)]
        type MyThing = super::MyThingRust;

        #[cxx_name="..."]
        fn set_my_prop(self: Pin<&mut MyThing>, value: i32);
    }
```
It basically boils down to requiring you to declare the actual setter in the bridge yourself, if you don't want to use the auto-generated one.
That way you have full control.

ahayzen-kdab:
This has been done via various changes with @BenFordTytherington 's refactorings. Instead of passing through the original we now rebuild in the generation phase this allows for ensuring we have a supported API etc.

PR #670, opened 2023-08-24
examples: add example of using a signal on an external object

ahayzen-kdab:
Related to #577

Requires #669 

Requires #681


examples/qml_features/rust/src/externcxxqt.rs
```diff
@@ -0,0 +1,109 @@
...
+
+impl ffi::ExternalCxxQtHelper {
+    unsafe fn connect_to_external(self: Pin<&mut Self>, external: *mut ffi::ExternalQObject) {
+        // let qt_thread = self.qt_thread();
+        if let Some(external) = external.as_mut() {
+            let mut pinned_external = Pin::new_unchecked(external);
+            pinned_external
+                .as_mut()
+                .on_triggered(|_| {
+                    // FIXME: can't move this inside the function pointer
```
ahayzen-kdab, line 71: cc @LeonMatthesKDAB we need the closure support in signals to complete this example :-)

examples/qml_features/rust/src/externcxxqt.rs
```diff
@@ -0,0 +1,106 @@
...
+                })
+                .release();
+
+            let qt_thread = self.qt_thread();
+            pinned_external
+                .as_mut()
+                .on_triggered_private_signal(move |_| {
+                    qt_thread
+                        .queue(|mut qobject| {
+                            qobject.as_mut().rust_mut().count += 1;
```
ahayzen-kdab, line 87: should increment private_count

examples/qml_features/rust/src/externcxxqt.rs
```diff
@@ -0,0 +1,106 @@
...
+    unsafe fn connect_to_external(self: Pin<&mut Self>, external: *mut ffi::ExternalQObject) {
+        if let Some(external) = external.as_mut() {
+            let qt_thread = self.qt_thread();
+            let mut pinned_external = Pin::new_unchecked(external);
+            pinned_external
+                .as_mut()
+                .on_triggered(move |_| {
+                    qt_thread
+                        .queue(|mut qobject| {
+                            qobject.as_mut().rust_mut().count += 1;
```
ahayzen-kdab, line 74: should just use set_count(count() + 1)

examples/qml_features/tests/tst_externcxxqt.qml
```diff
@@ -0,0 +1,60 @@
...
+        helper.connectToExternal(obj);
+
+        compare(triggeredSpy.count, 0);
+        compare(triggeredPrivateSpy.count, 0);
+
+        helper.triggerOnExternal(obj, 2);
+
+        tryCompare(triggeredSpy, "count", 2);
+        tryCompare(triggeredPrivateSpy, "count", 2);
+    }
```
ahayzen-kdab, line 59: should also check the count and private_count properties

examples/qml_features/qml/pages/ExternCxxQtPage.qml
```diff
@@ -0,0 +1,69 @@
...
+
+    ColumnLayout {
+        anchors.left: parent.left
+        anchors.right: parent.right
+        anchors.verticalCenter: parent.verticalCenter
+
+        Label {
+            Layout.fillWidth: true
+            horizontalAlignment: Text.AlignHCenter
+            text: qsTr("Connecting to signals in external QObject can run closures as normal.")
```
LeonMatthesKDAB, line 48: I think this explanation could be a bit more thorough.
```suggestion
            text: qsTr("CXX-Qt supports closures when connecting to signals.\n" +
                            "In this example the 'Trigger' button can be used to trigger a signal.\n" +
                            "A different object is captured by the invoked closure and this object increments the count.\n" +
                            "This way easy bindings between objects are possible with CXX-Qt.")
```
I think this page is otherwise a bit confusing when viewed without the accompanying count...

In general, I think `qml_features` suffers a lot from this.
The features are all shown on a technical level.
But the accompanying examples often don't really show anything exciting when viewing the QML GUI.

Also, this page doesn't really have much to do with `extern "C++Qt"`, but rather with the closures in signals...

Anyhow, I think a redo of some of the `qml_features` is in order anyhow.
Something for post-release maybe, so we can play more with the (more) stable API.

LeonMatthesKDAB (commented):
the code looks good.

I think the example itself doesn't really tell you much unless you view the code with it though.

We should improve the wording a bit to fix this short-term and maybe redo some of the examples in the long term.

examples/qml_features/qml/pages/ExternCxxQtPage.qml
ahayzen-kdab: done, and yes i think `qml_features` needs a proper overhaul at some point, another good task for newcomers / R&D time :thinking:

LeonMatthesKDAB (approved):

PR #673, opened 2023-08-28
Fixed comparison

b1nhack:
Rust Qt Binding Generator is under constant maintenance


b1nhack:
> 

Yeah, you are right :)

PR #674, opened 2023-08-29
Add Support for Q_ENUM_NS and Q_NAMESPACE

LeonMatthesKDAB:


crates/cxx-qt-gen/src/generator/rust/qenum.rs
```diff
@@ -4,29 +4,43 @@
...
+    qenums
+        .iter()
+        .flat_map(|qenum| {
+            let qenum_item = &qenum.item;
+            let qenum_ident = &qenum.ident;
+            let namespace = &qenum.namespace;
+            let namespace = if namespace.is_empty() {
+                quote! {}
+            } else {
+                quote! { #[namespace= #namespace ] }
```
ahayzen-kdab, line 20: add a space before the `=`

crates/cxx-qt-gen/src/parser/cxxqtdata.rs
```diff
@@ -43,6 +50,8 @@ impl ParsedCxxQtData {
         Self {
             cxx_mappings: ParsedCxxMappings::default(),
             qobjects: BTreeMap::<Ident, ParsedQObject>::default(),
+            qenums: Vec::default(),
```
ahayzen-kdab, line 53: why `Vec::default()` vs just `vec![]` like elsewhere?

ahayzen-kdab: same for two lines below

crates/cxx-qt-gen/src/parser/qenum.rs
```diff
@@ -54,12 +56,20 @@ impl ParsedQEnum {
...
 
+        let namespace = attribute_find_path(&qenum.attrs, &["namespace"])
+            .map(|attr_index| {
+                let attr = &qenum.attrs[attr_index];
+                expr_to_string(&attr.meta.require_name_value()?.value)
+            })
+            .transpose()?
+            .unwrap_or_default();
+
         // TODO: Add support for `namespace`, `cxx_name` and `rust_name` attributes.
```
ahayzen-kdab, line 67: namespace has been added now? so can be removed from the comment

crates/cxx-qt-gen/src/parser/qenum.rs
```diff
@@ -96,18 +118,33 @@ mod tests {
                 Variant2,
             }
         };
+        let qenum: ItemEnum = syn::parse2(original_item.clone()).unwrap();
 
         let parsed = ParsedQEnum::parse(qenum).unwrap();
         assert_eq!(parsed.ident, "MyEnum");
+        assert_eq!(parsed.namespace, "");
```
ahayzen-kdab, line 125: if there is no namespace or QObject we should error ?

crates/cxx-qt-gen/src/parser/qnamespace.rs
```diff
@@ -0,0 +1,107 @@
...
+    /// The name of the namespace
+    pub namespace: String,
+    /// whether qml_element was specified
+    pub qml_element: bool,
+}
+
+impl ParsedQNamespace {
+    pub fn parse(mut mac: ItemMacro) -> Result<Self> {
+        let namespace: LitStr = syn::parse2(mac.mac.tokens)?;
+        let namespace = namespace.value();
```
ahayzen-kdab, line 20: we should check here that the namespace is not empty?

crates/cxx-qt-gen/src/parser/qnamespace.rs
```diff
@@ -0,0 +1,107 @@
...
+        Ok(Self {
+            namespace,
+            qml_element,
+        })
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use std::assert_eq;
```
ahayzen-kdab, line 47: why does this need to be included ? (and in `crates/cxx-qt-gen/src/generator/cpp/constructor.rs` in `main`) as elsewhere it doesn't need to be included ?

crates/cxx-qt-gen/src/generator/rust/qenum.rs
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/parser/cxxqtdata.rs
LeonMatthesKDAB: No reason, really... can change.

crates/cxx-qt-gen/src/parser/qenum.rs
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/parser/qnamespace.rs
LeonMatthesKDAB: That's not done here.
The QEnum parser doesn't care whether the QEnum is assigned to anything, it's still a valid `ParsedQEnum`.

CxxQtData is responsible for matching the QEnum to the corresponding namespace or QObject.
Which will then error out if there's no namespace or QObject to associate the QEnum with.

crates/cxx-qt-gen/src/parser/qenum.rs
LeonMatthesKDAB: That's not done here.
The QEnum parser doesn't care whether the QEnum is assigned to anything, it's still a valid `ParsedQEnum`.

CxxQtData is responsible for matching the QEnum to the corresponding namespace or QObject.
Which will then error out if there's no namespace or QObject to associate the QEnum with.

crates/cxx-qt-gen/src/parser/qnamespace.rs
LeonMatthesKDAB: I've added a check that the namespace contains no whitespace and isn't empty.
We may want to do more checks in future, but this should be good enough to cover most cases.

crates/cxx-qt-gen/src/parser/qnamespace.rs
LeonMatthesKDAB: it does, in fact, not need to be included :sweat_smile:

ahayzen-kdab (commented):
Looks good, was there any reason for this still being a draft? And it needs rebasing...

ahayzen-kdab (approved):
Lets go with this, the book is being rewritten so docs can come later. And any other fixes can arrive later.

PR #675, opened 2023-08-29
qt-build-utils: do not rerun when generated QML module files change

### LINKED ISSUE #672: Build script always reruns due to QML module
When running ` cargo build -p qml-minimal-no-cmake` twice, the second time always still runs the build script.

I suspect this is due to the `rerun-if-changed` of the generated QML module as these include the `OUT_DIR` inside them, as seen in the snippet below.

We likely need to figure out a way around this

- We can't have a stable `OUT_DIR` folder ? As in cargo-only builds we don't have such a directory ?
- We can't skip the generation of the QML module if the sources haven't changed, as we still need the output of the plugin to be located from somewhere ?
- Can we support / hook into rustc's incremental compiler somehow so that this can be a block of input -> output that is then incrementally built ? Meaning that the second time it doesn't need to do anything ?

```
  cargo:rerun-if-changed=/var/home/andrew/Projects/cxx-qt/examples/cargo_without_cmake/src/cxxqt_object.rs
  cargo:rerun-if-changed=/var/home/andrew/Projects/cxx-qt/target/debug/build/qml-minimal-no-cmake-d5c5131364817605/out/qml_modules/com/kdab/cxx_qt/demo/qml_module_resources.qrc
  cargo:rerun-if-changed=/var/home/andrew/Projects/cxx-qt/target/debug/build/qml-minimal-no-cmake-d5c5131364817605/out/qml_modules/com/kdab/cxx_qt/demo/plugin.qmltypes
  cargo:rerun-if-changed=/var/home/andrew/Projects/cxx-qt/target/debug/build/qml-minimal-no-cmake-d5c5131364817605/out/qml_modules/com/kdab/cxx_qt/demo/qmldir
  cargo:rerun-if-changed=/var/home/andrew/Projects/cxx-qt/target/debug/build/qml-minimal-no-cmake-d5c5131364817605/out/qml_modules/com/kdab/cxx_qt/demo/qmldir
  cargo:rerun-if-changed=/var/home/andrew/Projects/cxx-qt/examples/qml_minimal/qml/main.qml
  cargo:rerun-if-changed=
```

ahayzen-kdab:
Closes #672

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -219,6 +219,9 @@ add_test(NAME cpp_clang_format COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/clang_format
 # Add test which checks all files have a valid license
 add_test(NAME reuse_lint COMMAND reuse lint)
 
+# Ensure that reuse doesn't run while files are changing
+set_tests_properties(reuse_lint PROPERTIES RUN_SERIAL TRUE)
+
 # Add test which checks that a build rerun doesn't recompile and uses caches instead
 add_test(NAME cargo_build_rerun COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/check_cargo_build_rerun.sh" "${CMAKE_CURRENT_SOURCE_DIR}")
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -219,6 +219,12 @@ add_test(NAME cpp_clang_format COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/clang_format
 # Add test which checks all files have a valid license
 add_test(NAME reuse_lint COMMAND reuse lint)
 
+# Add test which checks that a build rerun doesn't recompile and uses caches instead
+add_test(NAME cargo_build_rerun COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/check_cargo_build_rerun.sh" "${CMAKE_CURRENT_SOURCE_DIR}")
+
+# Ensure that cargo_build_rerun doesn't run while we are already building
+set_tests_properties(cargo_build_rerun PROPERTIES RUN_SERIAL TRUE)
+
 add_subdirectory(book)
 add_subdirectory(examples)
 add_subdirectory(tests)
--- a/.github/workflows/github-cxx-qt-tests.yml
+++ b/.github/workflows/github-cxx-qt-tests.yml
@@ -65,7 +65,7 @@ jobs:
             # once Ubuntu 22.04 is we can move to clang-format-14 everywhere
             # for now we need at least clang-format-12 otherwise include reordering fails after clang-format off
             # https://github.com/KDAB/cxx-qt/issues/121
-            ctest_args: --exclude-regex '^(example_qml_features_test|example_qml_minimal_myobject_test|reuse_lint|cpp_clang_format|.*valgrind)$'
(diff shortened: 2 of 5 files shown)


ahayzen-kdab:
This allows for multiple builds with no code change not to rebuild.

But need to check that if the code changes it does still work ?

vimpostor:
> But need to check that if the code changes it does still work ?

Could we add a test for this? If testing this is too impractical it's fine, but I think we can just add a third stage inside `check_cargo_build_rerun.sh` where we apply a small diff to some QML module source file and grep for `Compiling` again.

ahayzen-kdab:
> > But need to check that if the code changes it does still work ?
> 
> Could we add a test for this? If testing this is too impractical it's fine, but I think we can just add a third stage inside `check_cargo_build_rerun.sh` where we apply a small diff to some QML module source file and grep for `Compiling` again (but with the condition inverted).

Good idea, i'll take a look :thinking:

ahayzen-kdab:
Which also nicely shows how this doesn't quite work correctly yet :-D

ahayzen-kdab:
OK think i've fixed it, we needed to track rerun-if-changed for the files in a qml_module as well

vimpostor:
Awesome!

vimpostor (approved):

CMakeLists.txt
```diff
@@ -219,6 +219,9 @@ add_test(NAME cpp_clang_format COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/clang_format
 # Add test which checks all files have a valid license
 add_test(NAME reuse_lint COMMAND reuse lint)
 
+# Add test which checks that a build rerun doesn't compile and uses caches
+add_test(NAME cargo_build_rerun COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/check_cargo_build_rerun.sh" "${CMAKE_CURRENT_SOURCE_DIR}")
```
ahayzen-kdab, line 223: I think there is the potential of a "race" condition here.

As if this is in the middle of running, another test might happen to be also compiling the `qml-minimal-no-cmake` example.

Is there a way we can force CTests to run a test after all the others have completed ? Or just run it by itself etc ?

ahayzen-kdab: Glorious found that you supposedly do `set_tests_properties(test_name PROPERTIES RUN_SERIAL TRUE)`

Also set this on our `reuse` test as before we have had it where the book test generates new html files and reuse thinks they are unlicensed :-D

CMakeLists.txt
```diff
@@ -219,6 +219,12 @@ add_test(NAME cpp_clang_format COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/clang_format
 # Add test which checks all files have a valid license
 add_test(NAME reuse_lint COMMAND reuse lint)
 
+# Add test which checks that a build rerun doesn't compile and uses caches
```
vimpostor, line 225: Maybe one last very minor nitpick, if I may: "doesn't compile" sounds to me like literally not compiling, as in the build fails with a compilation error, which is obviously not what is meant here. Maybe something like this is clearer:
```diff
-# Add test which checks that a build rerun doesn't compile and uses caches
+# Add test which checks that a build rerun doesn't recompile and uses caches instead
```

ahayzen-kdab: Done!

vimpostor (approved):

PR #680, opened 2023-09-04
Add qobject_header_include() to add support for moc -I<path>

### LINKED ISSUE #679: How to specify -I (include paths) to moc compiler with CxxQtBuilder.qobject_header()?
For this snippet to pass the moc compiler I need to pass in a `-IOpenRGB` for it to work correctly.

```cpp
  class OpenRGBPluginInterfaceWrapper : public QObject, public OpenRGBPluginInterface
  {
      Q_OBJECT
      Q_PLUGIN_METADATA(IID OpenRGBPluginInterface_IID)
      Q_INTERFACES(OpenRGBPluginInterface)
  public:
      OpenRGBPluginInterfaceWrapper(QObject* parent);
      ~OpenRGBPluginInterfaceWrapper();
```

Otherwise I'll get errors like:
```bash
: moc OpenRGBSamplePlugin.h         
OpenRGBSamplePlugin.h:23: Parse error at "IID"
```
and (commenting out Q_PLUGIN_METADATA).
```bash
OpenRGBSamplePlugin.h:24: Error: Undefined interface
```

This works:
```bash
moc -IOpenRGB OpenRGBSamplePlugin.h
```

I've looked through the code for cxx-qt-build and it's not clear if there is a way to do this.

haata:
POC, probably not the best way to do this, but it was the simplest.

- Needed when the moc compiler does not have all the necessary files present in the cwd.
- Fixes #679

### DIFF
--- a/crates/qt-build-utils/src/lib.rs
+++ b/crates/qt-build-utils/src/lib.rs
@@ -531,7 +531,8 @@ impl QtBuild {
     /// as well as the path to the generated metatypes.json file, which can be passed to [register_qml_module](Self::register_qml_module).
     ///
     /// * uri - Should be passed if the input_file is part of a QML module
-    pub fn moc(&mut self, input_file: impl AsRef<Path>, uri: Option<&str>) -> MocProducts {
+    /// * include_paths - Additional include paths to pass to moc
+    pub fn moc(&mut self, input_file: impl AsRef<Path>, uri: Option<&str>, include_paths: &Vec<PathBuf>) -> MocProducts {
         if self.moc_executable.is_none() {
             self.moc_executable = Some(self.get_qt_tool("moc").expect("Could not find moc"));
         }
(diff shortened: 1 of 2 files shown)


ahayzen-kdab:
Thanks!

This likely needs a bit of thought about the API (maybe `moc` can take a struct if we are going to add more args and then use default() like QmlModule does).

We also need to look into the use case of `Q_INTERFACES` and other macros as we haven't explored using that yet.

crates/qt-build-utils/src/lib.rs
```diff
@@ -531,7 +531,8 @@ impl QtBuild {
     /// as well as the path to the generated metatypes.json file, which can be passed to [register_qml_module](Self::register_qml_module).
     ///
     /// * uri - Should be passed if the input_file is part of a QML module
-    pub fn moc(&mut self, input_file: impl AsRef<Path>, uri: Option<&str>) -> MocProducts {
+    /// * include_paths - Additional include paths to pass to moc
+    pub fn moc(&mut self, input_file: impl AsRef<Path>, uri: Option<&str>, include_paths: &Vec<PathBuf>) -> MocProducts {
```
ahayzen-kdab, line 535: we should go through the other `moc` arguments while we are here and consider which others might be useful.

LeonMatthesKDAB:
I agree with @ahayzen-kdab's that using a `MOCArguments` struct would be good.
This struct could just boil everything down to a `Vec<String>` , which is then passed to the moc.

That would also allow us to provide users an escape hatch to just add any string as an argument, so we don't have to support every moc argument type.

LeonMatthesKDAB:
Superceded by #796

Issue #682, opened 2023-09-05
Windows WSL2 WSLg not rendering inside window and release only build


AddictArts:
Windows 11 x64

Passing along feedback that like other people have reported release only builds are needed on the Windows 11 side. The default or debug does not complete and create an executable. The below worked for me. The cargo only build did not work either, could not find Qt stuff. Error was QtMissing ... See also #498 and #676

```
cmake --build build --config Release
```

WSL2
WSLg
Ubuntu 22

The cmake build works and the cargo only. However, it does not render anything in the window, just a white canvas, no buttons, etc controls. The button is clickable and I see terminal output, "Hi from rust ...". Unfortunately the empty window has no workaround that I could find. I don't know if this is a Qt issue and WSLg.

Last issue Qt organization, if you download from them you are forced to use Qt6.5 and this project is not ready for that. This is unfortunate and not a good developer experience. Don't understand the poor support for prior releases. This is not anyone's here fault, just noting the latest may cause issues.

ahayzen-kdab:
Thanks for taking the time to report the issue.

I would be interested if someone could point to where the debug vs release build is going wrong on Windows. As a lot of this is coming from Corrosion now. Is this the issue with the `"CMAKE_MSVC_RUNTIME_LIBRARY": "MultiThreadedDLL"` that needs to be set ? Or is this another issue ? (and we have some fun of where to put that, eg in the root CMakeList for when people are building the repository or in each of the examples CMakeLists?).

I don't think any of us have tried this on WSLg, not many of us have even tried it on Windows.

From the Qt online installer you should be able to download any release you want via the "archive" checkbox on the right side then click "filter" (eg i can pick any version from 5.9 -> 6.6 beta).

ahayzen-kdab:
Wonder if we are hitting the difference between `CMAKE_BUILD_TYPE=Debug` that we set in the CMakeList and `--config Debug`, appears that different generators might do different things there? We could try removing those lines tomorrow and see what happens...

AddictArts:
Thanks @ahayzen-kdab 

Maybe I have this wrong, I set `$Env:CMAKE_MSVC_RUNTIME_LIBRARY = "MultiThreadedDLL"` Did not help and same error

```
LINK : warning LNK4098: defaultlib 'MSVCRT' conflicts with use of other libs; use /NODEFAULTLIB:library
```

AddictArts:
[...]

Issue #676, opened 2023-08-30
Following the book to make the first build work is difficult.


ajantti:
First thing to note is I'm on Windows.
I tried following the book to build the example. It didn't really work:
1) First it wouldn't find Qt (and if you follow the book, Qt is not automatically downloaded). The book should at least explain how to make CMake understand where Qt is.
2) Of course the whole thing with "# Use `cxx = "1.0.83"` here instead!" is a bit silly. Why does the book give me code that doesn't work and then tells me to fix it?
3) Then I ran into problems with _ITERATOR_DEBUG_LEVEL not matching, signaling that half is being built in Debug config and half in Release, so I had to build it with --config Release, and I still don't know how to make a debug build.
4) At this point I figured I'd try to use the example from the repository, so I copied the files from the link given in the book instead. But then I ran into problems with cxx_qt_build::QmlModule not existing. Turns out the book tells me to use the stable versions of the crates, but then links to the development branch of the repo for "check this link if anything doesn't work". (Maybe would be a good idea to make the book link to the latest release tags that is connected to the book.)
After setting Cargo.toml to get the crates from the repository instead, I got it to build, but running was difficult:
5) First it had problems with not finding the .dlls. So I added them to PATH.
6) Then it wouldn't find the platform plugins.
7) Then it wouldn't find the qml files.
8) Of course all these would be solved by running windeployqt, but the given CMakeLists.txt doesn't seem to be handling that, and following CMake's instructions didn't work for me either, so I gave up and ran it manually.
9) Then I even had problems with the qrc. I don't know exactly why the AUTORCC is not working, but shouldn't it be added as a file in the CMakeLists.txt?

Actually, I might as well write it here, (I can make a new issue too), but following the instructions on the Readme.md also had problems:
1) Running `cmake -S . -B build` was fine, but I needed to do the same Release thing with `cmake --build build --config Release`
2) After that, the exe was in `./build/examples/qml_minimal/Release/example_qml_minimal.exe` not `./build/examples/qml_minimal/example_qml_minimal.exe` like mentioned
3) And then I got the same stuff as before about platform plugins.

Sorry for a bit of a ranty thing. I appreciate the work, and thought you should know what the experience is like.

LeonMatthesKDAB:
Hi, thank you for the feedback, it's super valuable, sorry this was such a painful experience for you.

[...]

PR #683, opened 2023-09-06
cmake: set MSVC_RUNTIME_LIBRARY to MultiThreadedDLL

ahayzen-kdab:
Related to #682
Related to #676

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Do not use -bundle otherwise CMake builds are missing qt-static-initalizers (note this is broken in rustc 1.69)
 - Do not import `Pin` in hidden module as invokables are outside now, resolving IDE integration
+- Rust always links against a non-debug Windows runtime with *-msvc targets, so we need to link to MultiThreadedDLL
 
 ### Removed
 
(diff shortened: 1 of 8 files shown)


OlivierLDff:
Will it work for more complex projects? Because then you will have target linking to the default, and the app & rust lib to MultiThreadedDLL.
I would recommend setting the global `CMAKE_MSVC_RUNTIME_LIBRARY` so every target do so.

ahayzen-kdab:
> Will it work for more complex projects? Because then you will have other targets linking to the default, and the app & rust lib to MultiThreadedDLL. Not sure if this is how it work. I would recommend setting the global `CMAKE_MSVC_RUNTIME_LIBRARY` so every target do so.

Can we set this in all of the example CMakeLists.txt's ? I know it's slightly pointless as it should come from the route. But we see people copy the CMakeList out of the example folder and then miss that there is something set globally. So i think there should be something in each of the projects too.

I'm also trying to fix my Windows setup so i can test some of this stuff myself :sweat_smile:

OlivierLDff:
Yes setting it in each example is a good idea, maybe with a warning that it would be best to set it in the command line.

ahayzen-kdab:
This seems to work :-)

@OlivierLDff does the change look sane to you ?

vimpostor (approved):

PR #684, opened 2023-09-08
cargo: bump versions from cargo outdated and use workspace for cc

ahayzen-kdab:


Cargo.lock
```diff
@@ -0,0 +1,1350 @@
...
+version = "0.8.16"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5a22b2d63d4d1dc0b7f1b6b2747dd0088008a9be28b6ddf0b1e7d335e3037294"
+dependencies = [
+ "cfg-if",
+]
+
+[[package]]
+name = "cxx"
+version = "1.0.107"
```
ahayzen-kdab, line 273: Should we lock all of these with the minimum versions specified in the Cargo.toml ? 

I've tried reverting some to those versions and tested the build still works.

Or should we run `cargo update` and ensure the lock file is up to date ? even if the version in the `Cargo.toml` is lower ? :thinking:

ahayzen-kdab: https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries recommends not to have a lock file for libraries. So maybe we try without for now ?

vimpostor: Yeah for now let's follow the official guidelines and go without a lock. When (if ever) someone reports an actual issue they  are having with this, we can still reevaluate.

vimpostor (approved):

PR #687, opened 2023-09-13
WIP: book: rewrite various sections for 0.6

ahayzen-kdab:


book/src/concepts/nested_objects.md
```diff
@@ -7,6 +7,10 @@ SPDX-License-Identifier: MIT OR Apache-2.0
 
 # Nested Objects
```
ahayzen-kdab, line 8: wonder if inheritance and nested objects should be split into another chapter under "Use cases" / "User stories" / "Advanced" / "Complex examples" or something :thinking: 

Then keep the "concepts" to be very general basic concepts like bridge, build systems, duality of C++ and Rust QObject, built-in types

PR #698, opened 2023-10-13
cxx-qt-lib: fix incorrect cmp order

ahayzen-kdab:
This meant that QString and QDateTime had reversed Ord results.


vimpostor (approved):
Perfect, and with test! :)

FTR, the regression was introduced in 6745f22db already

ahayzen-kdab:
@vimpostor could you re-review, i fixed the rustfmt issue and added a QDateTime test :-)

vimpostor (approved):

PR #700, opened 2023-10-23
QT_STRICT_ITERATORS was removed in qt6

Montel:


CMakeLists.txt
```diff
@@ -110,9 +110,11 @@ add_compile_definitions(
 )
 
 if(UNIX AND NOT APPLE)
-    add_compile_definitions(
-        QT_STRICT_ITERATORS
-    )
+    if (NOT USE_QT5)
```
ahayzen-kdab, line 113: Should this instead use something like `Qt5_FOUND` or similar ?

As if you had `USE_QT5=Off` you can still end up building with Qt 5 if only Qt 5 is available. The CMake option is to force Qt5 when Qt6 is also available.

```cmake
if(NOT USE_QT5)
    find_package(Qt6 COMPONENTS Core Gui Test)
endif()
if(NOT Qt6_FOUND)
    find_package(Qt5 5.15 COMPONENTS Core Gui Test REQUIRED)
endif()
```

ahayzen-kdab: Ah but this block comes before the `find_package` block that is below :sweat_smile:

Montel: Fixed

ahayzen-kdab (approved):
LGTM :-)

PR #701, opened 2023-10-23
Use new QT_NO_CONTEXTLESS_CONNECT variable in qt6

Montel:
Defining the macro before including any Qt header will disable the overload of QObject::connect that does not take a receiver/context argument.


CMakeLists.txt
```diff
@@ -109,6 +109,10 @@ add_compile_definitions(
     QT_USE_QSTRINGBUILDER
 )
 
+if (NOT USE_QT5)
```
ahayzen-kdab, line 112: Similar to #700 wonder if this needs to use Qt5_FOUND or whether something like `QT_VERSION_MAJOR` could be compared?

PR #702, opened 2023-10-23
Add missing qsTr(...)

Montel:


examples/demo_threading/qml/MainWindow.qml
```diff
@@ -386,7 +386,7 @@ Window {
         y: 22
 
         SideText {
-            text: "nº Online Sensors: " + "<i><b>%1</i></b>".arg(energyUsage.sensors)
+            text: qsTr("nº Online Sensors: ") + "<i><b>%1</i></b>".arg(energyUsage.sensors)
```
ahayzen-kdab, line 389: This should probably actually be the following so that translators can flip the order of the `A: B` etc.

Note it also has a weird ordering of i and b in closing and opening that could be fixed so that the open and close tags are in the same order.

`qsTr("nº Online Sensors: <i><b>%1</b></i>").arg(energyUsage.sensors)`

Montel:
done

ahayzen-kdab (approved):
LGTM :-)

PR #704, opened 2023-10-23
Add explicit moc includes to sources for moc-covered headers

Montel:
* speeds up incremental builds as changes to a header will not always need the full mocs_compilation.cpp for all the target's headers rebuild, while having a moc file sourced into a source file only adds minor extra costs, due to small own code and the used headers usually already covered by the source file, being for the same class/struct
* seems to not slow down clean builds, due to empty mocs_compilation.cpp resulting in those quickly processed, while the minor extra cost of the sourced moc files does not outweigh that in summary. Measured times actually improved by some percent points. (ideally CMake would just skip empty mocs_compilation.cpp & its object file one day)
* enables compiler to see all methods of a class in same compilation unit to do some sanity checks
* potentially more inlining in general, due to more in the compilation unit
* allows to keep using more forward declarations in the header, as with the moc code being sourced into the cpp file there definitions can be ensured and often are already for the needs of the normal class methods

### DIFF
--- a/crates/cxx-qt-gen/test_outputs/inheritance.cpp
+++ b/crates/cxx-qt-gen/test_outputs/inheritance.cpp
@@ -20,3 +20,5 @@ MyObject::MyObject(QObject* parent)
   , ::rust::cxxqtlib1::CxxQtLocking()
 {
 }
+
+#include "moc_inheritance.cpp"
--- a/crates/cxx-qt-gen/test_outputs/invokables.cpp
+++ b/crates/cxx-qt-gen/test_outputs/invokables.cpp
@@ -120,3 +120,5 @@ MyObject::MyObject(
 }
 
 } // namespace cxx_qt::my_object
+
+#include "moc_invokables.cpp"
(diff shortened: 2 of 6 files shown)


crates/cxx-qt-gen/test_outputs/inheritance.cpp
```diff
@@ -20,3 +20,5 @@ MyObject::MyObject(QObject* parent)
   , ::rust::cxxqtlib1::CxxQtLocking()
 {
 }
+
+#include "moc_inheritance.cpp"
```
ahayzen-kdab, line 24: Note these files are just test outputs for the test inputs. Eg the expected C++ output for the given C++ input.

Should these have explicit moc includes ? As we would then need to add this to the generation phase which then needs to know the resultant file name.

Or should these be added to the `cxx-qt-lib` sources instead ?

Montel: ok I didn't know that it was not compiled.

LeonMatthesKDAB:
If I understand the previous discussion correctly, this can be closed?

PR #707, opened 2023-10-23
Add cmakepreset 2

Montel:


CMakePresets.json
```diff
@@ -0,0 +1,145 @@
...
+                "base"
+            ]
+        },
+        {
+            "name": "release",
+            "displayName": "Build as release mode.",
+            "cacheVariables": {
+                "CMAKE_BUILD_TYPE": "Release",
+		"BUILD_TESTING": "OFF",
+                "USE_QT5" : "ON"
```
ahayzen-kdab, line 100: if you look at #498 and #498 there are a minefield of possible options and we were worried this file could become huge with so many options.

What do you think is the best way of covering the cases here ? And which ones should be left for `CMakeUserPresets` ? Eg i think probably `sccache` (from my old pull request) should be a user preset.

CMakePresets.json
```diff
@@ -0,0 +1,145 @@
+{
```
ahayzen-kdab, line 1: can you rebase this so that the merge commits are removed?

OlivierLDff:
I think the best would be to do like in corrosion: https://github.com/corrosion-rs/corrosion/pull/443
The CMakePresets correspond to what is used in the CI. Then users have at least a base to create their own presets

ahayzen-kdab:
> I think the best would be to do like in corrosion: [corrosion-rs/corrosion#443](https://github.com/corrosion-rs/corrosion/pull/443) The CMakePresets correspond to what is used in the CI. Then users have at least a base to create their own presets

Right, this is likely a good route, so then we'd have 6 ? options of Linux, macOS, Windows with Qt 5 and Qt 6 that CI can use the presets. Then devs can add user presets for any additional things and derive from the closest one.

PR #714, opened 2023-10-26
Update copyright (2023)

Montel:


README.md
```diff
@@ -130,7 +130,7 @@ ctest --test-dir build
 
 ## Licensing
 
-CXX-Qt is Copyright (C) 2022, Klarälvdalens Datakonsult AB, and is available under
+CXX-Qt is Copyright (C) 2023, Klarälvdalens Datakonsult AB, and is available under
```
ahayzen-kdab, line 133: Wonder if we should remove the year from there instead? So then we don't nee do update it.

https://hynek.me/til/copyright-years/
https://www.linuxfoundation.org/blog/blog/copyright-notices-in-open-source-software-projects

Montel: ok no problem for me :)
I will update patch without year.

LeonMatthesKDAB (approved):
LGTM

My university lessons in German law support this change as well.

KDAB still holds the Copyright, no matter what year it is.

PR #715, opened 2023-10-26
fix clang format

Montel:


crates/cxx-qt-gen/include/cxxqt_thread.h
```diff
@@ -124,7 +124,6 @@ namespace rust {
 
 template<typename T>
 struct IsRelocatable<::rust::cxxqtlib1::CxxQtThread<T>> : ::std::true_type
-{
-};
+{};
```
ahayzen-kdab, line 127: why isn't CI picking this up if it is a clang-format issue ? As running our `clang_format_check.sh` script on the directory doesn't lead to any issues for me.

Is this one of the "I have a different version of clang-format" issues ? Currently we are using whatever is in Ubuntu 22.04.

When we redo the CI we would probably also move to pre-commit so that devs all have the same clang-format version.

ahayzen-kdab: So must be something in a newer version of clang-format that you have
```
 /home/runner/cxx-qt/crates/cxx-qt-gen/include/cxxqt_thread.h:127:2: error: code should be clang-formatted [-Wclang-format-violations]
{};
 ^
```

Montel: Yep perhaps. I use last tumbleweed version.

PR #716, opened 2023-10-27
Check if valgrind is installed on system

Montel:


CMakeLists.txt
```diff
@@ -118,6 +118,9 @@ if(NOT Qt6_FOUND)
     find_package(Qt5 5.15 COMPONENTS Core Gui Test REQUIRED)
 endif()
```
ahayzen-kdab, line 119: the diff here is strange, think you need to rebase onto `upstream/main` ?

ahayzen-kdab (approved):
LGTM :-)

PR #718, opened 2023-10-27
revert "Make modules required"

ahayzen-kdab:
Otherwise when build in an environment with only Qt5 you have to still specify USE_QT5=true.

This reverts commit 834cd850a49e451445fd866eb18772485b63e1da.

### DIFF
--- a/examples/cargo_without_cmake/CMakeLists.txt
+++ b/examples/cargo_without_cmake/CMakeLists.txt
@@ -12,7 +12,7 @@
 # from any directory in this repository to build this crate without using CMake at all.
 
 if(NOT USE_QT5)
-    find_package(Qt6 COMPONENTS Core Gui Qml QuickControls2 QmlImportScanner REQUIRED)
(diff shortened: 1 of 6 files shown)


ahayzen-kdab:
@Montel so before these scenarios used to work

|   | USE_QT5=ON | USE_QT5=OFF (default) |
| ------------- | ------------- | ------- |
| Only Qt 5 installed  | Qt 5 | Qt 5 |
| Only Qt 6 installed  | Fails  | Qt 6 |
| Both Qt 5 and 6 installed | Qt 5 | Qt 6 |

With the change that was made making the Qt 6 modules required, the Qt 5 fallback case fails

|   | USE_QT5=ON | USE_QT5=OFF (default) |
| ------------- | ------------- | ------- |
| Only Qt 5 installed  | Qt 5 | Fails |
| Only Qt 6 installed  | Fails  | Qt 6 |
| Both Qt 5 and 6 installed | Qt 5 | Qt 6 |

So i wonder if we want this change or not ?

Montel (approved):
Ok I see. I didn't have qt5 here.

PR #723, opened 2023-11-01
book: use docs page for traits rather than book

ahayzen-kdab:
This removes duplication of documentation as we are trying to have one source.


book/src/bridge/traits.md
```diff
@@ -19,8 +19,8 @@ impl UniquePtr<A> {} // explicit CXX trait implementation of UniquePtr for A
 impl cxx_qt::Trait for A {} // explicit CXX-Qt trait implementation of Trait for A
 ```
 
-- [CxxQtType](./cxxqttype.md) - trait to reach the Rust implementation of a QObject
-- [Constructor](./constructor.md) - custom constructor
-- [Initialize](./initialize.md) - execute Rust code when the object is constructed
-- [Locking](./locking.md) - marker trait whether locking is enabled
-- [Threading](./threading.md) - marker trait whether CXX-Qt threading should be enabled
+- [CxxQtType](https://docs.rs/cxx-qt/latest/cxx_qt/trait.CxxQtType.html) - trait to reach the Rust implementation of a QObject
```
ahayzen-kdab, line 22: we may need to add more docs to the trait itself ?

book/src/SUMMARY.md
```diff
@@ -25,9 +25,4 @@ SPDX-License-Identifier: MIT OR Apache-2.0
     - [extern "C++Qt"](./bridge/extern_cppqt.md)
     - [Shared types](./bridge/shared_types.md)
     - [Attributes](./bridge/attributes.md)
-- [Reference: traits](./traits/index.md)
-    - [CxxQtType](./traits/cxxqttype.md)
-    - [Constructor](./traits/constructor.md)
-    - [Initialize](./traits/initialize.md)
-    - [Locking](./traits/locking.md)
-    - [Threading](./traits/threading.md)
```
ahayzen-kdab, line 33: wonder if we should have a page in concept explaining threading and locking

LeonMatthesKDAB: I think it's fine if the explanation lives on `docs.rs`.

book/src/bridge/traits.md
LeonMatthesKDAB: Yes, though not necessarily for `CxxQtType`, as I think it's a less important trait.

 I think in general we're removing too much information here.
We should copy some of the text from `traits/` into their code documentation.

LeonMatthesKDAB (changes requested):
I think overall we're losing too much information here.

We should just copy relevant parts from these explanatory pages into the in-code documentation.

book/src/bridge/traits.md
ahayzen-kdab: I think `CxxQtType` is still important to mention, as it's not obvious to new people how do you go from the C++ context to the Rust context.

ahayzen-kdab:
@LeonMatthesKDAB OK increased some of the traits, don't know if you want to pull more into the `Constructor` one at all ?

LeonMatthesKDAB (approved):
LGTM.

I think the Constructor documentation is good enough for now.

Issue #719, opened 2023-10-29
QModelIndex::internalId() needed for trees


kermitfrog:
Possible tasks:

- [ ] Add Qt types to cxx-qt-lib, eg `quintptr` so it can be used easily in bridges https://doc.qt.io/qt-6/qttypes.html
- [ ] Add an alias in a C++ header somewhere for `c_void` as suggested in https://github.com/dtolnay/cxx/issues/1049
- [x] Add support for methods in existing types which return these types and were missed before, eg `QModelIndex::internalId`

Original description:

I am trying to create a tree-model to provide data to QML. 
The normal way is to use QAbstractItemModel, but there is a problem:
The view might request an items parent by calling QModelIndex::parent().

In a list or table, where items never have parents, this would simply return an invalid QModelIndex().
The current item can then be found simply by it's row and column.

In a tree, the parent is needed as well, for which I need to know the current Item, for which I need the parent ...
That problem is usually solved by storing a pointer or id in the QModelIndex, which is set by QAbstractItemModel::createIndex() and retrieved by QModelIndex::internalId() (or internalPointer()) -- but cxx_qt does not provide access to these methods.

I thought about a workaround by following the parents up to the root level, remembering all the rows and then down again to find the right item, but as the documentation of QAbstractItemModel::parent states:

> When reimplementing this function in a subclass, be careful to avoid calling QModelIndex member functions, such as QModelIndex::parent(), since indexes belonging to your model will simply call your implementation, leading to infinite recursion.

so that would not work.

The only other way I can think of, that might possibly work, is subclassing QModelIndex, which I'd like to avoid.

ahayzen-kdab:
Thanks for taking the time to report this issue :-)

We haven't wrapped those methods in CXX-Qt yet, we could add support for the internalPointer quite easily by the looks of things or if added support for `quintptr` that would also be simple to add.

Currently you could extend the `QModelIndex` type with your own type, for example

You could define your own type to extend the `cxx_qt_lib` one.

```rust
#[repr(transparent)]
pub struct QModelIndexExtended(cxx_qt_lib::QModelIndex);

unsafe impl cxx::ExternType for QModelIndexExtended {
    type Id = cxx::type_id!("QModelIndex");
    type Kind = cxx::kind::Trivial;
}
```

Then in your bridge define the type with additional methods

```rust
    unsafe extern "C++" {
        include!("cxx-qt-lib/qmodelindex.h");
[...]

PR #724, opened 2023-11-04
Wrapper for QModelIndex::internalId

mmMike:
Implementation of QModelIndex::internalId for #719 using a wrapper function.
The direct solution 

        #[rust_name = "internalId"]
        fn internal_id(self: &QModelIndex) -> usize;

didn't compile. There was a type mismatch between ::std::size_t (unsigned long) and quintptr (unsigned long long) on my system (Apple Silicon, stable-aarch64-apple-darwin).

This PR enables trees via QAbstractItemModel.

### DIFF
--- a/crates/cxx-qt-lib/src/core/qmodelindex.cpp
+++ b/crates/cxx-qt-lib/src/core/qmodelindex.cpp
@@ -17,3 +17,14 @@ assert_alignment_and_size(QModelIndex,
                             sizeof(::std::size_t));
 
 static_assert(::std::is_trivially_copyable<QModelIndex>::value);
+
+namespace rust {
+namespace cxxqtlib1 {
+
+::std::size_t
+qmodelindexInternalId(const QModelIndex& index) {
+  return index.internalId();
+}
+
+}
+}
--- a/crates/cxx-qt-lib/src/core/qmodelindex.cpp
+++ b/crates/cxx-qt-lib/src/core/qmodelindex.cpp
@@ -22,7 +22,8 @@ namespace rust {
 namespace cxxqtlib1 {
 
 ::std::size_t
-qmodelindexInternalId(const QModelIndex& index) {
+qmodelindexInternalId(const QModelIndex& index)
+{
   return index.internalId();
 }
 
--- a/crates/cxx-qt-lib/src/core/qmodelindex.cpp
+++ b/crates/cxx-qt-lib/src/core/qmodelindex.cpp
@@ -24,7 +24,8 @@ namespace cxxqtlib1 {
 ::std::size_t
 qmodelindexInternalId(const QModelIndex& index)
 {
-  return index.internalId();
+  // TODO: need to add support for quintptr
+  return static_cast<::std::size_t>(index.internalId());
 }
 
 }
--- a/crates/cxx-qt-lib/src/core/qmodelindex.rs
+++ b/crates/cxx-qt-lib/src/core/qmodelindex.rs
@@ -67,6 +67,9 @@ pub struct QModelIndex {
 }
 
 impl QModelIndex {
+    /// Returns a `usize` used by the model to associate the index with the internal data structure.
+    //
+    // TODO: need to add support for quintptr
     pub fn internal_id(&self) -> usize {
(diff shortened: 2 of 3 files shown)


crates/cxx-qt-lib/src/core/qmodelindex.cpp
```diff
@@ -17,3 +17,15 @@ assert_alignment_and_size(QModelIndex,
...
 
 static_assert(::std::is_trivially_copyable<QModelIndex>::value);
+
+namespace rust {
+namespace cxxqtlib1 {
+
+::std::size_t
+qmodelindexInternalId(const QModelIndex& index)
+{
+  return index.internalId();
```
ahayzen-kdab, line 27: could you add a static_cast here and a TODO eg

```cpp
   // TODO: need to add support for quintptr
   return static_cast<::std::size_t>(index.internalId());
```

crates/cxx-qt-lib/src/core/qmodelindex.rs
```diff
@@ -62,6 +66,12 @@ pub struct QModelIndex {
     _m: MaybeUninit<usize>,
 }
 
+impl QModelIndex {
+    pub fn internal_id(&self) -> usize {
```
ahayzen-kdab, line 70: could you add a documentation line and a TODO

```rust
/// Returns a `usize` used by the model to associate the index with the internal data structure.
//
// TODO: need to add support for quintptr
```

ahayzen-kdab:
Looks good thanks! Just some small comments and then i'd like to discuss with @LeonMatthesKDAB if we think it's useful to land this using `usize` or whether we should try to add support for things like `quintptr` etc first.

mmMike:
> Looks good thanks! Just some small comments and then i'd like to discuss with @LeonMatthesKDAB if we think it's useful to land this using `usize` or whether we should try to add support for things like `quintptr` etc first.

I don't see a benefit in waiting for quintptr to be honest. These two functions are also in Qt side by side, so why wait? Another point in this direction: I guess this function might be more convenient in the Rust setting, given that pointers to objects are usually not the way to do things in Rust.

mmMike:
What's wrong in the failed test? I don't get it.

ahayzen-kdab:
That is an odd test failure, it seems to suggest cargo doc failed to document proc-macro2

```
error: couldn't generate documentation: stream did not contain valid UTF-8
  |
  = note: failed to create or modify "/Users/runner/cxx-qt/build/./cargo/build/doc/implementors/core/cmp/trait.Eq.js"

error: could not document `proc-macro2`
```

Lets try a re-run...

ahayzen-kdab:
Now we are getting failures on macOS all the time, even with `main` so something has changed in the CI environment.

ahayzen-kdab:
`cc` version 1.0.84 has been yanked as it has caused a load of regressions on macOS, trying a rerun again.

LeonMatthesKDAB (commented):
Hm, I'm a bit on the fence about exposing this as public API right now :thinking: 

As then if we ever do introduce quintptr, it would be a breaking change to change the return type.
It could be useful to have a type like quintptr, which represents both a usize or a pointer, depending on the interpretation.
And representing this with a corresponding Rust type that allows conversion to both usize and `*mut T`/`*const T` may be useful.
That would also making wrapping other functions that use this type a lot easier as we don't have to convert in C++.
Though doing this kind of pointer magic is definitely not very Rust-y, but likely existing applications depend on it.

A good way forward I'm like to agree on:
* Merge this PR as-is, for most places usize should be the most useful Rust type anyway.
* Later if we ever want to support quintptr, add a `_as_ptr` suffix (e.g. `fn internal_id_as_ptr(&self) -> quintptr`
    * We should then use this `_as_ptr` suffix for all other places where we use `quintptr` as well.
 
@ahayzen-kdab If you're okay with this, feel free to approve :+1:

ahayzen-kdab:
I think for now we land as `usize` then we can either change it to `quintptr` when available in a future release or as you suggest have another method to get the ptr etc. But this is useful enough for now to have to unblock folks implementing models.

ahayzen-kdab (approved):
LGTM, thanks for the contribution!

PR #729, opened 2023-11-09
Add info about how ready last documentation.

Montel:
(We need to open it in browser at specific url)

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Rust both support, however, these are not tested regularly.
 
 CXX-Qt is in early development and the API changes frequently. For the latest documentation between releases, [install mdBook](https://rust-lang.github.io/mdBook/guide/installation.html)
 and run `mdbook serve --open` in the [book folder](./book). It will open your own browser.
-If you need to open in another brower open url http://localhost:3000.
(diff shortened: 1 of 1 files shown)


LeonMatthesKDAB:
Ah, you bring up a good point.
Do note that the best way to do this is through `mdbook serve --open`, which will automatically open your default browser.
So please add that and mention connecting to `localhost` as a fallback :)

README.md
```diff
@@ -41,7 +41,8 @@ CXX-Qt is tested on CI on Linux, Windows, and macOS (all on x86_64). It should w
 Rust both support, however, these are not tested regularly.
 
 CXX-Qt is in early development and the API changes frequently. For the latest documentation between releases, [install mdBook](https://rust-lang.github.io/mdBook/guide/installation.html)
-and run `mdbook serve` in the [book folder](./book).
+and run `mdbook serve --open` in the [book folder](./book). It will open your own browser.
+If you need to open in another brower open url http://localhost:3000.
```
LeonMatthesKDAB, line 45:
```suggestion
If you need to open it in another browser goto url [http://localhost:3000](http://localhost:3000).
```

LeonMatthesKDAB (approved):

PR #731, opened 2023-11-09
Remove space in include

Montel:


ahayzen-kdab:
Merged as part of #732

PR #733, opened 2023-11-13
WIP: cargo: ensure cc crate is less than 1.0.84

ahayzen-kdab:


crates/cxx-qt-lib/build.rs
```diff
@@ -270,6 +270,9 @@ fn main() {
     builder.flag_if_supported("/permissive-");
     builder.flag_if_supported("/bigobj");
 
+    // Clang macOS
+    // builder.flag_if_supported("-stdlib=libc++");
```
ahayzen-kdab, line 274: test run to check that `main` is also failing

crates/cxx-qt-build/src/lib.rs
```diff
@@ -506,6 +506,8 @@ impl CxxQtBuilder {
             builder.flag_if_supported("/Zc:__cplusplus");
             builder.flag_if_supported("/permissive-");
             builder.flag_if_supported("/bigobj");
+            // Clang macOS
+            builder.flag_if_supported("-stdlib=libc++");
```
ahayzen-kdab, line 510: not sure this is the correct thing to do, do we want to always force `libc++` ?

ahayzen-kdab: And it doesn't help anyway :sweat_smile:

ahayzen-kdab: Appears the issue is due to the `cc` crate

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -113,6 +113,10 @@ jobs:
     runs-on: ${{ matrix.os }}
     name: ${{ matrix.name }}
     env:
+        # With macOS we need
+        # - std::filesystem::path from 10.15
+        # - std::shared_mutex from 10.12
+        MACOSX_DEPLOYMENT_TARGET: 10.15
```
ahayzen-kdab, line 119: wonder if we just always set this in CI

As the minimum version for rustc will be 10.12 but we need 10.15

https://blog.rust-lang.org/2023/09/25/Increasing-Apple-Version-Requirements.html

ahayzen-kdab:
CI is now working again so this is not required.

.github/workflows/github-cxx-qt-tests.yml
Be-ing: Setting this on CI is fine, but it doesn't solve the problem of new versions of cc breaking the build when downstream users haven't set the environment variable.

PR #734, opened 2023-11-14
github: move to macOS 12

ahayzen-kdab:
This is the latest stable image for Github actions https://github.com/actions/runner-images


ahayzen-kdab:
We should move to macOS 12 as this is the stable release on Github actions, but appears this suffers the same libstd issue.

ahayzen-kdab:
Landed elsewhere.

PR #737, opened 2023-11-17
Getting-Started: Move Cargo build before CMake

LeonMatthesKDAB:
Building with Cargo has advanced to the point where it is far simpler than building with CMake 🥳.


book/src/getting-started/5-cmake-integration.md
```diff
@@ -47,24 +59,35 @@ edition = "2021"
...
 
 ```toml,ignore
-{{#include ../../../examples/qml_minimal/rust/Cargo.toml:book_all}}
-```
+{{#include ../../../examples/qml_minimal/rust/Cargo.toml:book_static_lib}}
 
-> Note that instead of the `*.workspace = true` arguments for the CXX-Qt crates, you should instead use the versions from [crates.io](https://crates.io/search?q=cxx-qt).
-> As described in the code comment above each dependency.
+[dependencies]
+cxx = "1.0.95"
```
ahayzen-kdab, line 75: we somehow need to remember to bump this CXX one ... Can you add a note next to the version in the workspace Cargo.toml ?

Cargo.toml
```diff
@@ -42,8 +42,11 @@ qt-build-utils = { path = "crates/qt-build-utils", version = "0.5.3" }
 
 cc = { version = "1.0.79", features = ["parallel"] }
 # Ensure that the example comments are kept in sync
-# examples/cargo_without_cmake/Cargo.toml
-# examples/qml_minimal/rust/Cargo.toml
+# ./examples/cargo_without_cmake/Cargo.toml
+# ./examples/qml_minimal/rust/Cargo.toml
+# ./book/src/getting-started/4-cargo-executable.md
+# ./book/src/getting-started/5-cmake-integration.md
+# TODO: Can we re-export cxx from cxx-qt, so people don't need to manually add this anymore?
```
ahayzen-kdab, line 49: probably yes, question will be does a `#[cxx_qt::cxx::bridge]` macro work in our expansion, or would we put a `use cxx_qt::cxx;` or something? :thinking: something to play about with in the next cycle, add an issue so we don't forget

ahayzen-kdab (approved):
LGTM :-)

ahayzen-kdab:
I'll rebase this one after the next one in line goes in

PR #738, opened 2023-11-17
Improve clangd support when developing

LeonMatthesKDAB:
This enables more features in Clangd.
Otherwise anything in the GUI C++ code may be greyed out, depending on the editor.

Also ignore the Clangd cache.


.clangd
```diff
@@ -0,0 +1,6 @@
+# SPDX-FileCopyrightText: 2023 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+# SPDX-FileContributor: Leon Matthes <leon.matthes@kdab.com>
+#
+# SPDX-License-Identifier: MIT OR Apache-2.0
+CompileFlags:
+  Add: [-DCXX_QT_GUI_FEATURE=1]
```
ahayzen-kdab, line 6: should you also enable `-DCXX_QT_QML_FEATURE=1` ?

ahayzen-kdab (approved):
LGTM, can add QML compiler flag in a further commit if required.

PR #741, opened 2023-11-17
QByteArrayCursor

LeonMatthesKDAB:
This is similar to Rusts Cursor<T> type.
It allows access to a QByteArray through the `Write` + `Seek` traits.
This is quite useful for interaction with the wider Rust ecosystem.


LeonMatthesKDAB:
@ahayzen-kdab @Montel I'm wondering whether we should just abandon this PR.
It's useful, yes, but also something that users could easily add themselves.
If we want to support this, I think it can go into `cxx-qt-lib-extra`, as it's a bit of a maintenance burden.

Any opinions

LeonMatthesKDAB:
See also: #766

PR #743, opened 2023-11-20
Initialize model to nullptr

Montel:


examples/demo_threading/cpp/helpers/sensor.h
```diff
@@ -49,6 +49,6 @@ private Q_SLOTS:
 
 private:
   ::std::optional<int> m_index;
-  EnergyUsageProxyModel* m_model;
+  EnergyUsageProxyModel* m_model = nullptr;
```
ahayzen-kdab, line 52: Can you do this in the constructor implementation instead, to keep consistency with elsewhere.

```cpp
Sensor::Sensor(QObject* parent)
   : QObject(parent)
   , m_model(nullptr)
{
...
}
```

Montel: Ok I will change it.

ahayzen-kdab (approved):
LGTM

LeonMatthesKDAB:
This is why we need memory-safe languages :sweat_smile:

ahayzen-kdab:
Even just smart pointers or sane initialise values :-)

PR #746, opened 2023-11-20
Improve QImage support

Montel:


crates/cxx-qt-lib/src/gui/qimage.rs
```diff
@@ -11,12 +11,65 @@ mod ffi {
...
+
+        /// Returns the number of pixels that fit horizontally in a physical meter.
+        #[rust_name = "dots_per_meterx"]
+        fn dotsPerMeterX(self: &QImage) -> i32;
+
+        /// Returns the number of pixels that fit vertically in a physical meter.
+        #[rust_name = "dots_per_metery"]
+        fn dotsPerMeterY(self: &QImage) -> i32;
+
+        fn valid(self: &QImage, x: i32, y: i32) -> bool;
```
ahayzen-kdab, line 72: don't forget doc lines :-) 

there is also the fun of Rust not supporting overloads, so for this method we could support QPoint or x,y. Either we support both with a suffix on one of them (eg `_with_point`) or support one of them and just use `point.x, point.y`. Depends how often we think this method would be used. `x, y` is probably enough.

Montel: yep I saw that we can't provide overloads methods :( 
For the moment x,y is enough I think.

crates/cxx-qt-lib/src/gui/qimage.rs
```diff
@@ -11,12 +11,100 @@ mod ffi {
...
+        /// Sets the number of pixels by which the image is intended to be offset by when positioning relative to other images, to offset.
+        #[rust_name = "set_offset"]
+        fn setOffset(self: &mut QImage, point: &QPoint);
+
+        /// Returns the pixel index at (x, y).
+        #[rust_name = "pixel_index"]
+        fn pixelIndex(self: &QImage, x: i32, y: i32) -> i32;
+
+        /// Mirrors of the image in the horizontal and/or the vertical direction depending on whether horizontal and vertical are set to true or false.
+        fn mirror(self: &mut QImage, horizontal: bool, vertical: bool);
```
ahayzen-kdab, line 103: mirror doesn't exist in Qt 5.

We do have a cfg option for the Qt major version we use elsewhere

```rust
#[cfg(qt_version_major = "5")]
```

However I'm not sure if this works within a CXX bridge.

So what you might need to do is create a second bridge for the Qt 6+ features

```rust
#[cfg(not(qt_version_major = "5"))]
#[cxx::bridge]
mod ffi_qt_6 {
   ...  // Qt 6 specific methods
}
```

crates/cxx-qt-lib/src/gui/qimage.rs
```diff
@@ -11,12 +11,100 @@ mod ffi {
...
 
         /// Whether the QImage is null.
         ///
         /// This means that the QImage has all parameters set to zero and no allocated data.
         #[rust_name = "is_null"]
         fn isNull(self: &QImage) -> bool;
+
+        /// Returns true if all the colors in the image are shades of gray
+        #[rust_name = "all_gray"]
+        fn allGray(self: &QImage) -> bool;
```
ahayzen-kdab, line 31: personally I would try to keep these methods ordered either
- alphabetically
- in the order they are in the Qt docs

(i think the Qt docs are mainly alphabetical so it's the same either way)

crates/cxx-qt-lib/src/gui/qimage.rs
```diff
@@ -30,9 +132,26 @@ mod ffi {
...
+
+        #[doc(hidden)]
+        #[rust_name = "qimage_cache_key"]
+        fn qimageCacheKey(image: &QImage) -> i64;
     }
 }
+/*
+#[cfg(qt_version_major = "6")]
+#[cxx::bridge]
+mod ffi_qt_6 {
```
ahayzen-kdab, line 144: either need to figure out what is happening or remove the commented code :-)

Montel: Yep for the moment it failed to compile :( No idea yet.

Montel: So perhaps I can remove commented code and qimage support is improved :)

crates/cxx-qt-lib/src/gui/qimage.rs
LeonMatthesKDAB: Another way to go about this is to template the method:
```rust
fn valid(&self, point: Into<QPoint>) -> bool { ... }
```
That way you get sort-off overloading in Rust.
Assume `(i32, i32)` supported `Into<QPoint>`.

Then you could do either:
```
qimage.valid(my_qpoint);
qimage.valid((x, y)); // <-- Note the double-parens to create a tuple.
```

Though I guess if you didn't template the method it could be:
```
qimage.valid((x, y).into())
```
Which isn't terrible either.

I guess the takeaway here is that QPoint should implement `From<(i32, i32)>`

crates/cxx-qt-lib/src/gui/qimage.rs
ahayzen-kdab: Can you split this into a separate issue and link how the `cfg` stuff should work

LeonMatthesKDAB (commented):
There are some small things we should still add.
Apart from that, looks great :blush: 

It's really cool how easy it is to add new types with CXX :partying_face: 

One thing we'll want to figure out **in another PR** is how to support the additional constructors specified by `QImage`. Especially the ones that take raw data and don't do any conversions.
Those may be performance-critical for some applications.
However, we should make sure that they are lifetime-safe and format-safe.
Introducing something like `QImageRef<'a>` may work.
Or maybe just taking the data (e.g. `Vec<u8>`/`Box<[u8]>`) by-value is good enough if we use the clean-up functions to clean up later.

LeonMatthesKDAB:
Oh, I also forgot to mention; we're completely missing tests for QImage atm!

With proper support in this PR, we should at least have some tests to ensure the basics actually work and we don't i.e. crash when moving a QImage around!

Then you can also remove this warning in the documentation :wink: 
```rust
/// > ⚠ **Warning**: The QImage API in CXX-Qt-lib is not yet complete and subject to change.
///
```

LeonMatthesKDAB:
When writing the 0.6 release notes, we just remembered: QImage isn't in the changelog, please add it as you're now making it a proper feature :)

ahayzen-kdab:
@Montel while adding QImage into the changelog, you could also mention QLine / QLineF ;-)

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `extern "C++Qt"` block support for declaring existing types with methods and signals
 - `#[qenum]` attribute for `Q_ENUM` and `Q_ENUM_NS` support
 - `qnamespace!` macro to support exposing namespaced enums to QML
+- Support for further types: `QLine`, `QLineF`, `QImage`
```
ahayzen-kdab, line 32: 0.6.0 is already released, so this needs to go in a new `### Added` section under the unreleased section ;-)

Montel: Ah ok !

crates/cxx-qt-lib/src/gui/qimage.rs
```diff
@@ -8,20 +8,189 @@ use std::mem::MaybeUninit;
...
+    #[namespace = "Qt"]
+    unsafe extern "C++" {
+        include!("cxx-qt-lib/qt.h");
+        type TransformationMode = crate::TransformationMode;
+        type AspectRatioMode = crate::AspectRatioMode;
+    }
+
+    /// The type of image format available in Qt.
+    #[repr(i32)]
+    #[namespace = "rust::cxxqtlib1"]
```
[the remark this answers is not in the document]
Montel: We aleardy have "using QImageFormat = QImage::Format;" in crates/cxx-qt-lib-headers/include/gui/qimage.h it was added when qimage was added first in source.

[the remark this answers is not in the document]
ahayzen-kdab: aha i missed that :-D

crates/cxx-qt-lib/src/gui/qimage.rs
Montel: https://github.com/KDAB/cxx-qt/issues/765

LeonMatthesKDAB (approved):
LGTM

PR #749, opened 2023-11-22
Add qline support

Montel:


crates/cxx-qt-lib/src/core/qline.rs
```diff
@@ -0,0 +1,114 @@
...
+
+        /// Sets this line to the start in x1, y1 and end in x2, y2.
+        #[rust_name = "set_line"]
+        fn setLine(self: &mut QLine, x1: i32, y1: i32, x2: i32, y2: i32);
+
+        /// Sets the start point of this line to p1 and the end point of this line to p2.
+        #[rust_name = "set_points"]
+        fn setPoints(self: &mut QLine, p1: &QPoint, p2: &QPoint);
+
+        //QLineF 	toLineF() const
```
ahayzen-kdab, line 64: do you add QLineF in a second commit ?

ahayzen-kdab:
Looks good, either remove the commented line for toLineF.  Or add a second commit with QLineF, then we rebase merge rather than squash merge to keep as them as two commits in main.

crates/cxx-qt-lib/src/core/qline.rs
Montel: Yep I will add QLineF in an another PR.
I want to be sure that all is ok for creating a new class :)

I though that I removed this line.

crates/cxx-qt-lib/src/core/qline.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    #[namespace = "rust::cxxqtlib1"]
+    unsafe extern "C++" {
+        include!("cxx-qt-lib/common.h");
+
+        #[doc(hidden)]
+        #[rust_name = "qline_default"]
+        fn construct() -> QLine;
+
+        #[rust_name = "qline_new"]
```
ahayzen-kdab, line 79: this should be `#[doc(hidden)]` as it's an internal method

ahayzen-kdab (approved):
Awesome looks great :-)

PR #750, opened 2023-11-22
CXX-Qt-lib: Add QModelIndex::internalPointer

### LINKED ISSUE #719: QModelIndex::internalId() needed for trees

LeonMatthesKDAB:
Closes #719 

Depends on #738

### DIFF
--- a/crates/cxx-qt-lib/src/core/qmodelindex.rs
+++ b/crates/cxx-qt-lib/src/core/qmodelindex.rs
@@ -34,11 +34,16 @@ mod ffi {
         /// Returns the sibling at row for the current column. If there is no sibling at this position, an invalid QModelIndex is returned.
         #[rust_name = "sibling_at_row"]
         fn siblingAtRow(self: &QModelIndex, row: i32) -> QModelIndex;
+
+        /// Returns a `*mut c_void` pointer used by the model to associate the index with the internal data structure.
+        #[rust_name = "internal_pointer_mut"]
+        fn internalPointer(self: &QModelIndex) -> *mut c_void;
     }
 
     #[namespace = "rust::cxxqtlib1"]
     unsafe extern "C++" {
         include!("cxx-qt-lib/common.h");
+        type c_void = crate::c_void;
 
         #[doc(hidden)]
         #[rust_name = "qmodelindex_init_default"]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Support for further types: `QLine`, `QLineF`, `QImage`
+- `internal_pointer_mut()` function on `QModelIndex`
+- `c_void` in CXX-Qt-lib for easy access to `void *`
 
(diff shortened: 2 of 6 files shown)


crates/cxx-qt-lib/src/core/qmodelindex.rs
```diff
@@ -34,11 +34,21 @@ mod ffi {
...
         fn siblingAtRow(self: &QModelIndex, row: i32) -> QModelIndex;
+
+        /// Returns a `*mut c_void` pointer used by the model to associate the index with the internal data structure.
+        #[rust_name = "internal_pointer_mut"]
+        fn internalPointer(self: &QModelIndex) -> *mut c_void;
+
+        /// Returns a `*const c_void` pointer used by the model to associate the index with the
+        /// internal data structure.
+        #[rust_name = "internal_pointer"]
+        fn constInternalPointer(self: &QModelIndex) -> *const c_void;
```
ahayzen-kdab, line 45: this only exists in Qt 6...

LeonMatthesKDAB: Yeah, CI noticed as well :see_no_evil: 

So I guess we should really try getting the `#[cfg()]` attributes working in CXX then I guess.
Has there been any progress on this for #746 ?

ahayzen-kdab: Not yet, we need to see if it works in a mini example then try and get it working with cxx-qt-lib

LeonMatthesKDAB: I've removed the `const` version of this function for now.
It's not all that useful, as you can always cast to a `const*` from a `mut*` anyway.

We can always readd it if people ask for it.

ahayzen-kdab (approved):
LGTM :-)

PR #752, opened 2023-11-23
Add missing qpersistentmodelindex swap method

Montel:


ahayzen-kdab (approved):
LGTM :-)

LeonMatthesKDAB (commented):
Do we even need functions like this for a trivial type?
We have `std::mem::swap`, which should do the trick.

ahayzen-kdab:
@LeonMatthesKDAB Hmm good point, feel free to test that works and submit a pull request reverting. Then maybe add a comment where the swap method would be explaining that `mem::swap` could be used. Looks like all Qt is doing internally is a pointer swap https://codebrowser.dev/qt6/qtbase/src/corelib/global/qswap.h.html#_Z11qt_ptr_swapRPT_S1_

PR #755, opened 2023-11-25
qt-build-utils: use QML module name in .qrc file name

Be-ing:
This is required to link together multiple QML modules in one project without duplicate symbol errors. This fixes the link error in #598.


ahayzen-kdab (approved):
Awesome thanks for spotting that :-) :tada:

PR #757, opened 2023-11-27
Add missing QLineF methods

Montel:


crates/cxx-qt-lib/src/core/qlinef.rs
```diff
@@ -64,11 +92,19 @@ mod ffi {
         #[rust_name = "set_points"]
         fn setPoints(self: &mut QLineF, p1: &QPointF, p2: &QPointF);
 
+        /// Returns an integer based copy of this line.
+        #[rust_name = "to_line"]
+        fn toLine(self: &QLineF) -> QLine;
```
ahayzen-kdab, line 97: could you add a `impl From<&QLineF> for QLine` as well ?

We need to check/decide what we do in these cases, do we keep the Qt style API of `toLine()` or do we only have `.into()` / `QLine::from` aka Rust style.

Looks like QPointF has both contained in it's file

```rust
        #[rust_name = "to_point"]
        fn toPoint(self: &QPointF) -> QPoint;

...

impl From<&ffi::QPoint> for QPointF {
    /// Constructs a copy of the given point.
    fn from(point: &ffi::QPoint) -> Self {
        ffi::qpointf_from_qpoint(point)
    }
}

impl From<QPointF> for ffi::QPoint {
    /// Rounds the coordinates of this point to the nearest integer,
    /// and returns a QPoint object with the rounded coordinates.
    fn from(value: QPointF) -> Self {
        value.to_point()
    }
}
```

So for now just copy that and add the extra From below in this file

crates/cxx-qt-lib/src/core/qlinef.rs
```diff
@@ -117,6 +153,14 @@ impl From<&ffi::QLine> for QLineF {
     }
 }
 
+impl From<QLineF> for ffi::QLine {
```
ahayzen-kdab, line 156: wonder why we did `QPointF` and not `&QPointF` before as this creates a copy anyway. Maybe it just felt more Rust like to do by-value. We have a load of places where we need to decide on by-ref or by-value, cc @LeonMatthesKDAB , this is another one.

PR #758, opened 2023-11-27
Add missing qcoreapplication remove path lib

Montel:


crates/cxx-qt-lib/src/core/qcoreapplication.rs
```diff
@@ -55,6 +55,9 @@ mod ffi {
         #[rust_name = "qcoreapplication_organization_name"]
         fn qapplicationOrganizationName(app: &QCoreApplication) -> QString;
         #[doc(hidden)]
+        #[rust_name = "qcoreapplication_remove_library_path"]
```
ahayzen-kdab, line 58: shouldn't this be added to QGuiApplication as well ?

crates/cxx-qt-lib-headers/include/core/qcoreapplication.h
```diff
@@ -120,5 +120,12 @@ qapplicationSetOrganizationName(T& app, const QString& name)
   app.setOrganizationName(name);
 }
 
+template<typename T>
+void
+qapplicationRemoveLibraryPath(T& app, const QString& path)
+{
+  app.removeLibraryPath(path);
```
ahayzen-kdab, line 127: in theory this could be `T::removeLibraryPath` as well but then `app` would be unused, so fine to do this :-D

ahayzen-kdab (approved):
LGTM :-)

PR #760, opened 2023-11-28
Add QCommandLineParser support

Montel:


crates/cxx-qt-lib-headers/include/core/qcommandlineparser.h
```diff
@@ -0,0 +1,43 @@
...
+                        const QString& optionName);
+
+QStringList
+qcommandlineparserValues(const QCommandLineParser& parser,
+                         const QString& optionName);
+
+bool
+qcommandlineparserIsSetFromQString(const QCommandLineParser& parser,
+                                   const QString& optionName);
+}
```
ahayzen-kdab, line 42: could have `// namespace cxxqtlib1` etc on the closing brackets

crates/cxx-qt-lib/src/core/qcommandlineparser.cpp
```diff
@@ -0,0 +1,49 @@
...
+// https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/tools/qcommandlineparser.h?h=v6.2.4#n109
+assert_alignment_and_size(QCommandLineParser,
+                          alignof(::std::size_t),
+                          sizeof(::std::size_t));
+
+static_assert(!::std::is_trivially_copy_assignable<QCommandLineParser>::value);
+static_assert(
+  !::std::is_trivially_copy_constructible<QCommandLineParser>::value);
+
+static_assert(!::std::is_trivially_destructible<QCommandLineParser>::value);
```
ahayzen-kdab, line 23: i think you need a `impl Drop` in Rust for this?

crates/cxx-qt-lib/src/core/qcommandlineparser.cpp
```diff
@@ -0,0 +1,49 @@
...
+// QCommandLineParser has a single pointer as it's member
+//
+// https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/tools/qcommandlineparser.h?h=v5.15.6-lts-lgpl#n107
+// https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/tools/qcommandlineparser.h?h=v6.2.4#n109
+assert_alignment_and_size(QCommandLineParser,
+                          alignof(::std::size_t),
+                          sizeof(::std::size_t));
+
+static_assert(!::std::is_trivially_copy_assignable<QCommandLineParser>::value);
+static_assert(
```
ahayzen-kdab, line 20: did you intentionally skip implementing Clone in Rust ? eg do we want to be able to copy this in Rust?

Montel:
CommandLineParser allows to add some specific Qt option directly.
I think it can be useful for Qt user which wants to use cxx-qt

Be-ing:
I don't really understand the point of binding this. A Rust CLI argument parsing library is probably going to be easier to use in any case. If the `main` function is in C++, there's no need for the Rust binding because QCommandLineParser can easily be used directly. It doesn't hurt to merge this except for the maintenance burden, so if @ahayzen-kdab and @LeonMatthesKDAB are fine with that, go ahead, but there are lots more Qt APIs that would be more useful to work on binding.

Be-ing:
> Is there any usecase for things like -platform xcb does QCommandLineParser automatically handle these ? Or is that handled elsewhere ?

We already take care of standard Qt CLI arguments like `-platform xcb` appropriately by passing CLI args from Q[Core/Gui]Application::new in Rust to the respective C++ constructors. There's no need for QCommandLineParse to interface with that.

Montel:
When we look at all option we can see
```
  --qmljsdebugger <value>              Activates the QML/JS debugger with a
                                       specified port. The value must be of
                                       format port:1234[,block]. "block" makes
                                       the application wait for a connection.
  --platform <platformName[:options]>  QPA plugin. See QGuiApplication
                                       documentation for available options for
                                       each plugin.
  --platformpluginpath <path>          Path to the platform plugins.
  --platformtheme <theme>              Platform theme.
  --plugin <plugin>                    Additional plugins to load, can be
                                       specified multiple times.
  --qwindowgeometry <geometry>         Window geometry for the main window,
                                       using the X11-syntax, like 100x100+50+50.
  --qwindowicon <icon>                 Default window icon.
  --qwindowtitle <title>               Title of the first window.
  --reverse                            Sets the application's layout direction
                                       to Qt::RightToLeft (debugging helper).
  --session <session>                  Restores the application from an earlier
                                       session.
  --display <display>                  Display name, overrides $DISPLAY.
  --name <name>                        Instance name according to ICCCM
                                       4.1.2.5.
  --nograb                             Disable mouse grabbing (useful in
                                       debuggers).
  --dograb                             Force mouse grabbing (even when running
                                       in a debugger).
  --visual <id>                        ID of the X11 Visual to use.
  --geometry <geometry>                Alias for --qwindowgeometry.
  --icon <icon>                        Alias for --qwindowicon.
  --title <title>                      Alias for --qwindowtitle.

```

It's provided automatically by Qt directly.
Not sure if we can see it when we use "<appname> --help-all"

Montel:
It depends if we think that cxx-qt is only for rust user which wants to use Qt/QML interface so for sure it can be easy to add "clap" rust module directly.
For me for Qt user which wants to use it, it can be more easy to use known api.

But indeed it depends about the target of cxx-qt.

Be-ing:
If you prefer to use QCommandLineParser over Rust CLI argument parsing libraries, then I think it would be easier to have your `main` function in C++ and use QCommandLineParser directly from C++ than have your `main` function in Rust and use Rust bindings to QCommandLineParser.

LeonMatthesKDAB:
I've added an issue regarding CXX-Qt-lib-extra here: #766 

As we seem to venture into this territory now reasonably often.

Montel:
Merge in https://github.com/KDAB/cxx-qt/pull/860

PR #761, opened 2023-11-28
Add QStringList::isEmpty missing method

Montel:


crates/cxx-qt-lib/src/core/qstringlist.rs
```diff
@@ -30,6 +30,10 @@ mod ffi {
         /// Returns a list of all the strings containing the substring str.
         fn filter(self: &QStringList, str: &QString, cs: CaseSensitivity) -> QStringList;
 
+        /// Returns true if the list has size 0; otherwise returns false.
+        #[rust_name = "is_empty"]
+        fn isEmpty(self: &QStringList) -> bool;
```
ahayzen-kdab, line 35: i think the reason we didn't have it before is that `QStringList` is effectively a `QList<QString>`, so the idea was that you would do `QList<QString>::from(&list).is_empty()`.

As the question is do we want to duplicate the whole QList API again for QStringList here, it's unfortunate we can't inherit easily :-/ Initially we were thinking QStringList would have it's specific API and you get a reference to it casted to a QList if you need any of that API. (Guess maybe we need a mutable reference as well and it should be documented better?)

Montel: Yep we will duplicate it :(
If we don't duplicate it we need to add info about "QList<QString>::from(&list).is_empty()" for example.
Otherwise we don't know how to do it.
I had the problem when I wanted to implement test for qcommandlineparser

I used is_empty() but it was not defined :)

ahayzen-kdab: If we impl Deref for QStringList to QList<QString> then Rust might be able to infer that `.is_empty()` is actually `.as_ref().is_empty()` :thinking: Maybe this is the way to go ?

LeonMatthesKDAB: I agree that a `Deref` implementation would be best :thinking: 
That should hopefully be rather easy to do :crossed_fingers:

redstrate: jnbooth is doing this in #1151, so I'm closing this attempt

PR #762, opened 2023-11-28
Add qelapsedtimer support

Montel:


crates/cxx-qt-lib/src/core/qelapsedtimer.rs
```diff
@@ -0,0 +1,72 @@
...
+    unsafe extern "C++" {
+        include!("cxx-qt-lib/common.h");
+
+        #[doc(hidden)]
+        #[rust_name = "qelapsedtimer_init_default"]
+        fn construct() -> QElapsedTimer;
+    }
+}
+
+/// The QRect struct defines a rectangle in the plane using integer precision.
```
ahayzen-kdab, line 42: this documentation is wrong? :-)

crates/cxx-qt-lib/src/core/qelapsedtimer.rs
```diff
@@ -0,0 +1,72 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[repr(C)]
+pub struct QElapsedTimer {
```
ahayzen-kdab, line 45: assume we want a QElapsedTimer in Rust and you don't just use the `std` stuff, guess it's still useful to have the Qt ones?

Montel: I think it's useful to have Qt class for user which wants to use cxx-qt

LeonMatthesKDAB: Having not needed QElapsedTimer yet, I'd prefer if this was in `cxx-qt-lib-extra`.
As you've already wrapped it for this PR, there's no reason to throw this away, but I also don't know if it's worth adding it to our maintenance list.

In my opinion, CXX-Qt-lib should only wrap the most fundamental types without which Qt as a framework is hardly usable. The main benefit of CXX-Qt as a framework is not to provide the most exhaustive list of wrapped Qt API, but rather that it provides you with the bare minimum and makes it easy to wrap everything else you may need yourself.
Of course wrapping more in CXX-Qt is going to be easier for our users,, but I think not maintainable, given the enormous size of Qt.

Overall, we need a good guideline to decide which types to wrap in CXX-Qt-lib and which not.
I agree that stuff like QMargins is not really fundamental, so probably shouldn't have been added to CXX-Qt-lib. We could even consider moving it out of lib for 0.7.
A first draft of such a guideline:
- Default to adding anything you think **may** be useful to CXX-Qt-lib-extra
- Only add types to CXX-Qt-lib directly, if they're both:
    - Hard to wrap optimally (i.e. should be trivial/templates, etc.)
    - Are often used as parameters/return values within the Qt codebase
- We can then always promote types from `lib-extra` to `lib`, if/when people request this. 

I've added an issue about this, see: #766

Montel:
Merge in  #860

PR #764, opened 2023-12-06
cxx-qt-lib: require headers to be includes manually

ahayzen-kdab:
This allows for cxx-qt-lib to be optional later.

Related to #319


Cargo.toml
```diff
@@ -9,7 +9,6 @@ members = [
     "crates/cxx-qt-build",
     "crates/cxx-qt-gen",
```
ahayzen-kdab, line 10: still need to make cxx-qt-gen only depend on cxx-qt headers / types and move them into cxx-qt.

crates/cxx-qt-build/Cargo.toml
```diff
@@ -26,6 +25,6 @@ version_check = "0.9"
 
 [features]
 default = ["qt_gui", "qt_qml"]
-qt_gui = ["cxx-qt-lib-headers/qt_gui"]
-qt_qml = ["cxx-qt-lib-headers/qt_qml"]
+qt_gui = []
```
ahayzen-kdab, line 28: we still need to set the defines in cxx-qt-build, wonder if instead of `with_headers(prefix, files)` it should also have defines or something ? Then have `with_opts(CxxQtBuilderOpts)` and have `cxx_qt_lib` return a struct, but then we have a circular dependency ...

ahayzen-kdab: Done as Opts

ahayzen-kdab:
Not sure why Windows is failing with a missing dll :thinking:

crates/cxx-qt-lib/Cargo.toml
```diff
@@ -11,13 +11,11 @@ edition.workspace = true
 license.workspace = true
 description = "Qt types for integrating `cxx-qt` crate with `cxx`"
 repository.workspace = true
-# Fake linking so that we expose cargo data as env vars for downstream crates
-# https://github.com/rust-lang/cargo/issues/3544
-links = "cxx-qt-lib"
```
ahayzen-kdab, line 16: i was hoping removing this might help the Windows issue, but appears not. Something else is broken ...

ahayzen-kdab: All we have to go on is the following which suggests some dll is missing at some point in the qml-feature build script. But it's unclear if it's a Qt one or a system one or a cxx-qt one or if it's when building the build script or building the final binary.

`  process didn't exit successfully: `C:/cxx-qt/build/./cargo/build\release\build\qml-features-814177af2461bb34\build-script-build` (exit code: 0xc0000135, STATUS_DLL_NOT_FOUND)`

examples/qml_minimal/rust/build.rs
```diff
@@ -17,6 +17,7 @@ fn main() {
             ..Default::default()
         })
         // ANCHOR_END: book_qml_module
+        .with_opts(cxx_qt_lib::build_opts())
```
ahayzen-kdab, line 20: should we have a cxx-qt-build feature that defaults to doing this ? then allow developers to opt out ? As cxx_qt_lib is probably used by most apps ?

crates/cxx-qt-lib/Cargo.toml
ahayzen-kdab: Appears when you have `cxx-qt-lib` as a build dependency the issue occurs so keeping the cxx-qt-lib-headers split works for now...

examples/qml_minimal/rust/build.rs
ahayzen-kdab: For now this will be opt-in

crates/cxx-qt-build/src/lib.rs
```diff
@@ -90,7 +84,18 @@ impl GeneratedCpp {
                             rust_file_path.display());
                     }
 
-                    file_ident = m.ident.to_string().to_case(Case::Snake);
+                    // Match upstream where they use the file name as the ident
+                    //
+                    // TODO: what happens if there are folders?
```
ahayzen-kdab, line 89: need to check what CXX does with folders

crates/cxx-qt-build/src/lib.rs
```diff
@@ -291,6 +305,16 @@ pub struct CxxQtBuilder {
     qt_modules: HashSet<String>,
     qml_modules: Vec<OwningQmlModule>,
     cc_builder: cc::Build,
+    extra_defines: HashSet<String>,
+}
+
+/// Options for external crates to use
+#[derive(Default)]
+pub struct CxxQtBuildersOpts {
```
LeonMatthesKDAB, line 313: As this is likely going to be used by other downstream dependencies as well, we should properly encapsulate this struct, so that we can extend it in future.

I suggest to not make the members public, and add functions to add defines and headers in a builder style.

crates/cxx-qt-build/Cargo.toml
LeonMatthesKDAB: Then why do we still need this feature in cxx-qt-build?

crates/cxx-qt-lib-headers/Cargo.toml
```diff
@@ -13,5 +13,8 @@ repository.workspace = true
 
 [features]
 default = []
-qt_gui = []
-qt_qml = []
+qt_gui = ["cxx-qt-build/qt_gui"]
```
LeonMatthesKDAB, line 16: The `cxx-qt-build/qt_gui` feature isn't necessary anymore, no?

crates/cxx-qt-lib/build.rs
```diff
@@ -257,40 +252,46 @@ fn main() {
...
+    builder = builder.cc_builder(|cc| {
+        // Load the include path
+        cc.include(&header_root);
+
+        // Enable Qt Gui in C++ if the feature is enabled
+        if feature_qt_gui_enabled {
+            cc.define("CXX_QT_GUI_FEATURE", None);
+        }
 
-    // Note, ensure our settings stay in sync across cxx-qt-build and cxx-qt-lib
```
LeonMatthesKDAB, line 283: Nice that we don't have to keep these in sync anymore :partying_face:

LeonMatthesKDAB (commented):
Overall direction looks good.

I think the Opts struct needs more encapsulation though.
Some of the features can now also be removed maybe?

crates/cxx-qt-build/Cargo.toml
ahayzen-kdab: Because `cxx-qt-build` also has the following which defaults to enabling the `gui` and `qml`  Qt modules in the Qt builders.

```rust
        let mut qt_modules = HashSet::new();
        qt_modules.insert("Core".to_owned());
        #[cfg(feature = "qt_gui")]
        qt_modules.insert("Gui".to_owned());
        #[cfg(feature = "qt_qml")]
        qt_modules.insert("Qml".to_owned());
```

Wonder how we should handle those, should our `CxxQtBuilderOpts` have the ability to pass in Qt modules that are required too ? Probably yes as that would be quite useful if you made a Bluetooth library to be able to then say i need the Bluetooth module as well.

CHANGELOG.md
```diff
@@ -27,6 +27,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - `VCPKG` is now set to off by default and packages are only built in release mode in the cache
+- `cxx-qt-lib` headers must be given to `cxx-qt-build` with `.with_opts(cxx_qt_lib_headers::build_opts())`
```
ahayzen-kdab, line 30: also mention that `cxx-qt-gen` doesn't depend on `cxx-qt-lib` anymore

crates/cxx-qt-build/src/lib.rs
ahayzen-kdab: Done :-)

crates/cxx-qt-lib-headers/Cargo.toml
ahayzen-kdab: Done :-)

crates/cxx-qt-build/src/lib.rs
ahayzen-kdab: Yes, if you have a `CRATE/src/folder/binding.rs` CXX ends up building it as `target/cxxbridge/CRATE/src/folder/binding.rs.{cc,h}` so we should follow this.

ahayzen-kdab: This breaks a lot of things as cxx-qt-lib assumes everything ends up without folders and our other tests are considering that the `src` folder would be in the path too.

Maybe I should do this as a separate change with the cxx_file_stem becoming mandatory as well ?

ahayzen-kdab: Split into #855

PR #769, opened 2023-12-26
Remove clang-format for autotests github action. It takes too many time

Montel:


.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -65,7 +65,7 @@ jobs:
             # once Ubuntu 22.04 is we can move to clang-format-14 everywhere
             # for now we need at least clang-format-12 otherwise include reordering fails after clang-format off
             # https://github.com/KDAB/cxx-qt/issues/121
-            ctest_args: --exclude-regex '^(example_qml_features_test|example_qml_minimal_myobject_test|reuse_lint|cpp_clang_format|cargo_build_rerun|.*valgrind)$'
+            ctest_args: --exclude-regex '^(cargo_tests|example_qml_features_test|example_qml_minimal_myobject_test|reuse_lint|cpp_clang_format|cargo_build_rerun|.*valgrind)$'
```
ahayzen-kdab, line 68: why is `cargo_tests` being added here?

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -184,7 +184,7 @@ jobs:
       # https://github.com/KDAB/cxx-qt/issues/121
       #
       # automake is needed for building libicu which is a dependency of Qt
-      run: brew install automake autoconf-archive ninja clang-format
```
ahayzen-kdab, line 187: we should instead be removing clang-format from the ctests as well and running clang-format in it's own job, see clang-format-rs for ideas https://github.com/KDAB/clang-format-rs/blob/main/.github/workflows/github-clang-format-tests.yml

.github/workflows/github-cxx-qt-tests.yml
Montel: because it tests clang-format => if we remove clang-format support this test will failed

.github/workflows/github-cxx-qt-tests.yml
ahayzen-kdab, line 187: In the end, moving to macOS 12 has made installing clang-format fast again ? So do we need to worry about this ? As we'll still want `cargo test` on all platforms i think ?

LeonMatthesKDAB:
I agree with @ahayzen-kdab , Installing clang-format is luckily fast again now.

I think we can close this. We do want to run cargo test on all platforms if we can.
It's still slow, but can hopefully be sped up by using `sccache`, which should push it to <10 min.

PR #771, opened 2023-12-26
Ignore CMakeLists.txt.user

Montel:


ahayzen-kdab (approved):
LGTM, assume we don't just do `*.user` ?

PR #772, opened 2023-12-26
Disable cargo_clippy and cargo_doc on macos

Montel:
It takes long time + it's already executes in linux/windows

### DIFF
--- a/.github/workflows/github-cxx-qt-tests.yml
+++ b/.github/workflows/github-cxx-qt-tests.yml
@@ -65,7 +65,7 @@ jobs:
             # once Ubuntu 22.04 is we can move to clang-format-14 everywhere
             # for now we need at least clang-format-12 otherwise include reordering fails after clang-format off
             # https://github.com/KDAB/cxx-qt/issues/121
-            ctest_args: --exclude-regex '^(example_qml_features_test|example_qml_minimal_myobject_test|reuse_lint|cpp_clang_format|cargo_build_rerun|.*valgrind)$'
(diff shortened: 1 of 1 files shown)


.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -65,7 +65,7 @@ jobs:
             # once Ubuntu 22.04 is we can move to clang-format-14 everywhere
             # for now we need at least clang-format-12 otherwise include reordering fails after clang-format off
             # https://github.com/KDAB/cxx-qt/issues/121
-            ctest_args: --exclude-regex '^(example_qml_features_test|example_qml_minimal_myobject_test|reuse_lint|cpp_clang_format|cargo_build_rerun|.*valgrind)$'
+            ctest_args: --exclude-regex '^(cargo_clippy|cargo_doc|example_qml_features_test|example_qml_minimal_myobject_test|reuse_lint|cpp_clang_format|cargo_build_rerun|.*valgrind)$'
```
ahayzen-kdab, line 68: instead remove from ctests and create individual jobs that then just run on Linux. See clang-format-rs where it does this https://github.com/KDAB/clang-format-rs/blob/main/.github/workflows/github-clang-format-tests.yml

LeonMatthesKDAB: I think we can still merge this right now.
We can refactor later, let's speed CI up first, as that makes refactoring at lot easier.

LeonMatthesKDAB (approved):

PR #774, opened 2023-12-27
Add QByteArray::prepend method

Montel:


crates/cxx-qt-lib-headers/include/core/qbytearray.h
```diff
@@ -63,5 +63,7 @@ qbytearrayToUpper(const QByteArray& byteArray);
 QByteArray
 qbytearrayTrimmed(const QByteArray& byteArray);
```
ahayzen-kdab, line 65: remove the empty line here i think, as these were grouped into methods that are just Qt wrappers and ones that are helpers to Rust

ahayzen-kdab: also keep the order A->Z so this should be after Len and before Remove

crates/cxx-qt-lib/src/core/qbytearray.cpp
```diff
@@ -92,6 +92,12 @@ qbytearrayAppend(QByteArray& byteArray, ::std::uint8_t ch)
   byteArray.append(static_cast<char>(ch));
 }
 
+void
+qbytearrayPrepend(QByteArray& byteArray, ::std::uint8_t ch)
```
ahayzen-kdab, line 96: keep the order the same as the header, so this will move to after Len and before Remove

crates/cxx-qt-lib/src/core/qbytearray.rs
```diff
@@ -99,6 +99,10 @@ mod ffi {
         #[doc(hidden)]
         #[rust_name = "qbytearray_trimmed"]
         fn qbytearrayTrimmed(bytearray: &QByteArray) -> QByteArray;
+
+        #[doc(hidden)]
+        #[rust_name = "qbytearray_prepend"]
```
ahayzen-kdab, line 104: keep A->Z

ahayzen-kdab (approved):
LGTM :-)

PR #775, opened 2023-12-27
Add QTime::isValid(...) static method

Montel:


crates/cxx-qt-lib/src/core/qtime.rs
```diff
@@ -157,6 +161,9 @@ impl QTime {
     pub fn secs_to(&self, t: Self) -> i32 {
         ffi::qtime_secs_to(self, t)
     }
+    pub fn is_valid_time(h: i32, m: i32, s: i32, ms: i32) -> bool {
```
ahayzen-kdab, line 164: add space above

ahayzen-kdab: and add documentation line

crates/cxx-qt-lib/src/core/qtime.rs
```diff
@@ -157,6 +161,12 @@ impl QTime {
     pub fn secs_to(&self, t: Self) -> i32 {
         ffi::qtime_secs_to(self, t)
     }
+
+    /// Returns true if the specified time is valid; otherwise returns false.
+    /// The time is valid if h is in the range 0 to 23, m and s are in the range 0 to 59, and ms is in the range 0 to 999.
+    pub fn is_valid_time(h: i32, m: i32, s: i32, ms: i32) -> bool {
```
ahayzen-kdab, line 167: why is this called `is_valid_time` and not `is_valid` ?

Montel: Because there is already a is_valid() method (without arguments)

ahayzen-kdab: Ah i see this is a static method, wonder if there is a common way of naming this in Rust, or if `_time` is enough

LeonMatthesKDAB (approved):
LGTM

PR #779, opened 2023-12-28
Add QPainter support

Montel:


crates/cxx-qt-lib/src/gui/qpainter.cpp
```diff
@@ -0,0 +1,27 @@
...
+// https://code.qt.io/cgit/qt/qtbase.git/tree/src/gui/painting/qpainter.h?h=v6.2.4#n451
+assert_alignment_and_size(QPainter,
+                          alignof(::std::size_t),
+                          sizeof(::std::size_t));
+
+static_assert(!::std::is_trivially_copy_assignable<QPainter>::value);
+static_assert(!::std::is_trivially_copy_constructible<QPainter>::value);
+
+static_assert(!::std::is_trivially_destructible<QPainter>::value);
+
```
ahayzen-kdab, line 26: Is `static_assert(QTypeInfo<QPainter>::isRelocatable);` true otherwise this type cannot be trivial in Rust ?

examples/qml_features/rust/src/custom_parent_class.rs
```diff
@@ -20,17 +20,10 @@ pub mod qobject {
...
-
-    // Define the API from QPainter that we need
-    unsafe extern "C++" {
-        /// QPainter from Qt
-        type QPainter;
-        include!(<QtGui/QPainter>);
 
-        /// QPainter::fillRect from Qt
-        #[rust_name = "fill_rect"]
-        fn fillRect(self: Pin<&mut QPainter>, rectangle: &QRectF, color: &QColor);
```
ahayzen-kdab, line 33: note here it was using QPainter as an opaque type rather than a trivial

crates/cxx-qt-lib/src/gui/qpainter.rs
```diff
@@ -0,0 +1,187 @@
...
+
+        /// Sets the background mode of the painter to the given mode
+        #[rust_name = "set_background_mode"]
+        fn setBackgroundMode(self: Pin<&mut QPainter>, mode: BGMode);
+
+        /// Enables clipping if enable is true, or disables clipping if enable is false.
+        #[rust_name = "set_clipping"]
+        fn setClipping(self: Pin<&mut QPainter>, enable: bool);
+
+        /// Enables clipping, and sets the clip region to the given rectangle using the given clip operation.
```
LeonMatthesKDAB, line 127:
```suggestion
        /// Enables clipping, and sets the clip region to the given rectangle using the given clip operation.
        ///
        /// Note that the clip rectangle is specified in logical (painter) coordinates.
```

crates/cxx-qt-lib/src/gui/qpainter.rs
```diff
@@ -0,0 +1,187 @@
...
+        fn backgroundMode(self: &QPainter) -> BGMode;
+
+        /// Returns the currently set brush origin.
+        #[rust_name = "brush_origin"]
+        fn brushOrigin(self: &QPainter) -> QPoint;
+
+        /// Returns the bounding rectangle of the current clip if there is a clip;
+        /// otherwise returns an empty rectangle. Note that the clip region is given in logical coordinates.
+        #[rust_name = "clip_bounding_rect"]
+        fn clipBoundingRect(self: &QPainter) -> QRectF;
```
LeonMatthesKDAB, line 44: In cases where the Qt API returns an uninitialized or otherwise "empty" object if something is not set, we usually opt to wrap the function for a more Rust-y API (i.e. returning `Option`).

This would make sense here as well. So please use: `clip_bound_rect_or_empty` as the Rust name and create a small wrapper function:
```rust
impl QPainter {
    fn clip_bounding_rect(&self) -> Option<QRectF> {
        let result = self.clip_bounding_rect_or_empty();
        if result.is_valid() {
            Some(result)
        } else {
            None
        }
    }
}
```

LeonMatthesKDAB (commented):
Hi @Montel , really nice change! QPainter is one of those APIs that we seem to need pretty often.

Only some smaller things left, then I'm happy to approve :)

crates/cxx-qt-lib/src/gui/qpainter.rs
```diff
@@ -0,0 +1,196 @@
...
+pub use ffi::QPainter;
+
+impl QPainter {
+    /// Create a QPainter
+    pub fn new() -> cxx::UniquePtr<Self> {
+        ffi::qpainter_init_default()
+    }
+
+    /// Returns the bounding rectangle of the current clip if there is a clip;
+    /// otherwise returns an empty rectangle. Note that the clip region is given in logical coordinates.
```
LeonMatthesKDAB, line 187:
```suggestion
    /// otherwise returns `None`. Note that the clip region is given in logical coordinates.
```

LeonMatthesKDAB (commented):
Just a tiny copy-pasta mistake in the documentation left.
Then we can merge :partying_face:

PR #781, opened 2023-12-30
Fix usability of QImage construction

alois31:
* Rename the method from_height_width_and_format to match its signature. This is a breaking change, but the method has not appeared in a release so far.
* Expose QImageFormat so that the method can be called without ugly hacks.

### DIFF
--- a/crates/cxx-qt-lib/src/gui/mod.rs
+++ b/crates/cxx-qt-lib/src/gui/mod.rs
@@ -19,4 +19,4 @@ mod qvector4d;
 pub use qvector4d::QVector4D;
 
 mod qimage;
-pub use qimage::QImage;
+pub use qimage::{QImage, QImageFormat};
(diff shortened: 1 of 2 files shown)


ahayzen-kdab (approved):
Thanks for the contribution, looks good to me :-)

Note that our CI is failing on macOS so we'll need to fix that before we can land this.

alois31:
Actually the CI seems to be failing everywhere now. Unfortunately, I do not understand how the changes from this PR would cause the errors I see. I suspect some non-reproducibility in the CI would need to be fixed first.

ahayzen-kdab:
@alois31 Right :-/ I think bumping corrosion #782 has helped with it failing everywhere in the configure stage. Then I assume it'll be back to the macOS issue. I suspect a dependency has changed and broken something (eg the cc crate changed before and broke many things), we'll investigate this separately :-)

PR #782, opened 2024-01-02
corrosion: bump to v0.4.5

ahayzen-kdab:


ahayzen-kdab:
One difference between when it worked on macOS Qt 6 and then didn't is CMake 3.27.8 vs 3.28.0 which is curious.

ahayzen-kdab:
Use this #783

PR #783, opened 2024-01-03
ci: bump versions

ahayzen-kdab:


crates/qt-build-utils/src/lib.rs
```diff
@@ -402,6 +402,14 @@ impl QtBuild {
             Err(_) => "lib",
         };
 
+        if let Ok(target) = &target {
+            if target.contains("apple") {
+                // Tell clang link to clang_rt as we build with -nodefaultlibs
+                // otherwise we cannot use helpers from the compiler in Qt
+                println!("cargo:rustc-link-arg=-fapple-link-rtlib");
```
ahayzen-kdab, line 409: can we tell that we are building with clang here? rather than just assuming "apple" == "clang" ?

ahayzen-kdab: Or do we always use a clang related linker on macOS ? I think yes as it's lld ?

ahayzen-kdab: Should this be in `qt_build_utils::setup_linker` instead of `QtBuild::cargo_link_libraries` ?

ahayzen-kdab: done

Montel (approved):

PR #784, opened 2024-01-03
Deduplicate CI Builds

LeonMatthesKDAB:
This deduplicates builds by:
- Testing everything in one go, no need for separate cargo_test_doc anymore
- Not building cargo_without_cmake in cmake (this is done by cargo test anyway).


ahayzen-kdab:
This needs further fixes which are in #788 for tests that were failing under `--all-features` :-)

PR #786, opened 2024-01-04
Add QPainterPath

Montel:


crates/cxx-qt-lib/src/gui/qpainterpath.rs
```diff
@@ -0,0 +1,87 @@
...
+    _cspec: MaybeUninit<i32>,
+}
+
+impl Default for QPainterPath {
+    fn default() -> Self {
+        ffi::qpainterpath_init_default()
+    }
+}
+
+impl Drop for QPainterPath {
```
ahayzen-kdab, line 75: could add a Clone implementation as well that uses the copy constructor?

crates/cxx-qt-lib/src/gui/qpainterpath.rs
```diff
@@ -0,0 +1,87 @@
...
+        /// Returns true if either there are no elements in this path, 
+        /// or if the only element is a MoveToElement; otherwise returns false.
+        #[rust_name = "is_empty"]
+        fn isEmpty(self: &QPainterPath) -> bool;
+
+        /// Returns a simplified version of this path.
+        fn simplified(self: &QPainterPath) -> QPainterPath;
+
+        /// Translates all elements in the path by (dx, dy).
+        fn translate(self: &mut QPainterPath, dx: f64, dy: f64);
```
ahayzen-kdab, line 46: what about `translated` ?

PR #787, opened 2024-01-04
Add QFont support

Montel:


crates/cxx-qt-lib/src/gui/qfont.rs
```diff
@@ -0,0 +1,213 @@
...
+    }
+}
+
+#[derive(Clone)]
+#[repr(C)]
+pub struct QFont {
+    _cspec: MaybeUninit<i32>,
+    _resolve_mask: MaybeUninit<u16>,
+}
+
```
ahayzen-kdab, line 194: clone ?

ahayzen-kdab: As in i meant you need to add a `impl Clone for QFont` because of the static assert on the C++ side
```cpp
static_assert(!::std::is_trivially_copy_constructible<QFont>::value);
```

ahayzen-kdab: (you can't use `derive(Clone)` because it's not trivial)

Montel: oops ! Ah ok ! :) 
I will do it

crates/cxx-qt-lib-headers/include/gui/qfont.h
```diff
@@ -0,0 +1,31 @@
+// clang-format off
+// SPDX-FileCopyrightText: 2024 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// clang-format on
+// SPDX-FileContributor: Laurent Montel <laurent.montel@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+#pragma once
+
+#include <cinttypes>
```
ahayzen-kdab, line 9: do you need this include? guess std::true_type might need type_traits? and then this might need to be moved to the cpp file ?

crates/cxx-qt-lib/src/gui/qfont.rs
```diff
@@ -0,0 +1,222 @@
+// SPDX-FileCopyrightText: 2024 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Laurent Montel <laurent.montel@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+use cxx::{type_id, ExternType};
+use std::mem::MaybeUninit;
+
+#[cxx::bridge]
+mod ffi {
+
```
ahayzen-kdab, line 10: can you document this enum and remove the empty line?

crates/cxx-qt-lib/src/gui/qfont.rs
```diff
@@ -0,0 +1,222 @@
...
+        StyleOblique,
+    }
+
+    /// This enum describes the different levels of hinting that can be applied to glyphs
+    /// to improve legibility on displays where it might be warranted by the density of pixels.
+    #[repr(i32)]
+    #[namespace = "rust::cxxqtlib1"]
+    #[derive(Debug)]
+    enum QFontHintingPreference {
+        PreferDefaultHinting,
```
ahayzen-kdab, line 26: can you add doc lines to these too (just copy Qt docs). As eventually we want to turn missing docs into errors in cxx-qt-lib

ahayzen-kdab (approved):
Looks good, lets see if Windows is happy as sometimes it needs specific includes for std stuff...

PR #788, opened 2024-01-04
Further ci fixes

ahayzen-kdab:


LeonMatthesKDAB (approved):
More & faster testing at the same time?
Sounds good to me :partying_face:

PR #789, opened 2024-01-05
Don't run cargo clippy & cargo doc on Windows

LeonMatthesKDAB:
Cargo doc is somewhat redundant to run on multiple platforms.

Same with cargo clippy, though we'll loose clippy lints for code that uses #[cfg(target_os="...")].
We expect this change to speed up CI by 5-10 min. though, which is likely worth it.


ahayzen-kdab (approved):
LGTM :-)

PR #792, opened 2024-01-05
github: run cargo fmt and reuse in separate runners

ahayzen-kdab:


.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -85,7 +108,7 @@ jobs:
             vcpkg_triplet: x64-osx-release
             # FIXME: qmltestrunner fails to import QtQuick module
             # https://github.com/KDAB/cxx-qt/issues/110
-            ctest_args: --exclude-regex '^(cargo_clippy|cargo_doc|example_qml_features_test|example_qml_minimal_myobject_test|reuse_lint|cpp_clang_format|cargo_build_rerun|.*valgrind)$'
+            ctest_args: --exclude-regex '^(cargo_clippy|cargo_doc|example_qml_features_test|example_qml_minimal_myobject_test|cpp_clang_format|cargo_build_rerun|.*valgrind)$'
```
LeonMatthesKDAB, line 111: `cpp_clang_format` can go as well.
Same for the other `ctest_args`

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -183,7 +206,6 @@ jobs:
         clang-format-12
```
LeonMatthesKDAB, line 206: This can go as well I think

LeonMatthesKDAB (commented):
Little bits of cleanup still left.

I'm unsure whether making these pre-checks is best, or whether having one separate runner with `fail_fast: true` is better. Happy to try this approach now, we can always change it again, if spinning up the runners isn't fast enough

.github/workflows/github-cxx-qt-tests.yml
LeonMatthesKDAB: Or is it needed for cargo_test?

ahayzen-kdab: required for cxx-qt-gen tests :-)

.github/workflows/github-cxx-qt-tests.yml
ahayzen-kdab: done

ahayzen-kdab:
> I'm unsure whether making these pre-checks is best, or whether having one separate runner with `fail_fast: true` is better. Happy to try this approach now, we can always change it again, if spinning up the runners isn't fast enough

Right i was wondering that as well, so far they seem to spin up/down quite quickly

ahayzen-kdab:
We could also move to a pip installed clang-format rather than apt installed, so that we can pin the version to the latest we want. Then start looking at pre-commit to match that version for developers.

LeonMatthesKDAB:
> We could also move to a pip installed clang-format rather than apt installed, so that we can pin the version to the latest we want. Then start looking at pre-commit to match that version for developers.

Yup, I think that's something we can do in a follow-up PR though :)

LeonMatthesKDAB (approved):
LGTM

PR #796, opened 2024-01-10
Qt-Build-Utils: Expand Moc arguments

LeonMatthesKDAB:
Supersedes #680 .

Doesn't yet add all possible moc arguments, but leaves us with an easy way to expand later, so we can merge this now.


crates/cxx-qt-build/src/lib.rs
```diff
@@ -417,6 +419,13 @@ impl CxxQtBuilder {
         self
     }
 
+    /// Specify additional include paths when running the moc compiler (qobject_header).
+    pub fn qobject_header_include(mut self, path: impl AsRef<Path>) -> Self {
```
ahayzen-kdab, line 423: could we copy the API we do for passing extra args cc builder ?

eg

```rust
.moc(|moc| {
   moc.include(...);
   moc.url(...);
})
```

Are these only specific to qobject_headers ? I assume things in qml_modules are separate ?

So maybe we instead need to wrap the qobject headers into it's own block thing like qml_modules ?

LeonMatthesKDAB: Ah, good point, we want this to use either MocArguments, or the builder patterns as well!

Maybe we should just add a `arguments: MocArguments` to `fn qobject_header`?
Then we could even support different arguments per-compilation. No idea if that's useful at all, but shouldn't hurt much either.

And we could do the same with qml_module I guess.

The only question that remains: do we want to use a sort of builder pattern like you describe, or just let people use the normal struct initialization with `..Default::default()`?
The only problematic part with that is currently that I couldn't mark the struct as `#[non_exhaustive]`, as then it wasn't initializable anymore.

LeonMatthesKDAB: I've added an optional function to add MocArguments to `qobject_header`.

Note that you can't yet pass MocArguments to the `qml_module` or `rust_file` functions.
I'm unsure whether we want to support this, as there moc runs on our own generated code, so hopefully we know the correct arguments.

LeonMatthesKDAB: Hm, I'll try the `Into` route and see where it goes. :thinking:

LeonMatthesKDAB: Done

crates/cxx-qt-build/src/opts.rs
```diff
@@ -42,3 +47,40 @@ impl CxxQtBuildersOpts {
         self
     }
 }
+
+/// Options for qobject_headers
+///
+/// QObjectHeaderOpts can be created using the `From<impl AsRef<Path>>` trait.
+/// ```
```
ahayzen-kdab, line 54: should this be `rust` so that it is tested or `rust,ignore` if not ?

LeonMatthesKDAB: Doc comments are `rust` by default.
They are already tested, as cargo check errored out on them when I forgot to add the `use` statements :)

ahayzen-kdab (approved):
LGTM :-)

PR #802, opened 2024-01-12
CI: markdown-lint

mattkdab:
only lints in book for now, the rules may have to be different for changelog & readme or they may need to be edited to conform or just have comments turning certain rules off

resolves #800

PR #803, opened 2024-01-13
CI markdown-lint whole repo

mattkdab:
followup PR to #802
small fixes in README.md and CHANGELOG.md to both adhere to style rules and disable certain rules for specific cases
moved markdownlint config to root dir & changed markdown_lint action job to parse entire repo instead of only book


README.md
```diff
@@ -7,6 +7,9 @@ SPDX-FileContributor: Leon Matthes <leon.matthes@kdab.com>
 SPDX-License-Identifier: MIT OR Apache-2.0
 -->
 
+<!-- markdownlint-disable MD034 -->
+<!-- ^^ MD034 complains about bare link in 'About KDAB' -->
```
ahayzen-kdab, line 11: isn't this one correct ? shouldn't it be `<https://www.kdab.com>` instead of `https://www.kdab.com` ?

mattkdab: I can change, I put this here because I think it's standard KDAB copypaste, but yes that's the right way.
There's another link somewhere in the repo that uses `[]()` but just has the same text in both. That should also be switched to `<>`

ahayzen-kdab (approved):
LGTM :-)

PR #807, opened 2024-01-16
Add QFont support

Montel:


crates/cxx-qt-lib-headers/include/core/qcoreapplication.h
```diff
@@ -14,6 +14,7 @@
 #include <QtCore/QCoreApplication>
 #include <QtCore/QStringList>
 #include <QtCore/QVector>
+#include <QtGui/QFont>
```
ahayzen-kdab, line 17: does this not need the if CXX_QT_GUI_FEATURE ?

crates/cxx-qt-lib-headers/include/core/qcoreapplication.h
```diff
@@ -127,5 +128,19 @@ qapplicationRemoveLibraryPath(T& app, const QString& path)
   app.removeLibraryPath(path);
 }
 
+template<typename T>
+void
+qapplicationSetFont(T& app, const QFont& font)
```
ahayzen-kdab, line 133: ah seems setFont is only available on QGuiApplication not QCoreApplication

crates/cxx-qt-lib-headers/include/core/qcoreapplication.h
Montel: ah yep indeed.

crates/cxx-qt-lib-headers/include/core/qcoreapplication.h
```diff
@@ -127,5 +130,14 @@ qapplicationRemoveLibraryPath(T& app, const QString& path)
   app.removeLibraryPath(path);
 }
 
+#ifdef CXX_QT_GUI_FEATURE
+template<typename T>
+QFont
+qapplicationFont(const T& app)
```
ahayzen-kdab, line 136: why is this still a template and not in qguiapplication like setFont (or the opposite).

crates/cxx-qt-lib/src/gui/qguiapplication.rs
```diff
@@ -67,6 +69,12 @@ mod ffi {
         #[rust_name = "qguiapplication_set_application_version"]
         fn qapplicationSetApplicationVersion(app: Pin<&mut QGuiApplication>, version: &QString);
         #[doc(hidden)]
+        #[rust_name = "qguiapplication_set_font"]
```
ahayzen-kdab, line 72: if these aren't templated anymore can we have direct CXX bindings rather than a C++ trampoline ?

eg

```rust
fn font(self: &QGuiApplication) -> QFont;
fn setFont(self: Pin<&mut QGuiApplication>, font: &QFont);

File: crates/cxx-qt-lib-headers/include/gui/qguiapplication.h [hunk 5]
```diff
@@ -18,5 +19,11 @@ namespace cxxqtlib1 {
 ::std::unique_ptr<QGuiApplication>
 qguiapplicationNew(const QVector<QByteArray>& args);
 
+void
+qapplicationSetFont(QGuiApplication& app, const QFont& font)
```
ahayzen-kdab, line 23: implementation should be in .cpp file as it's not a template, but also see comment below probably can remove all the C++ code related to font() / setFont()

crates/cxx-qt-lib/src/gui/qguiapplication.rs
ahayzen-kdab: (won't work because these are static methods)

crates/cxx-qt-lib-headers/include/gui/qguiapplication.h
```diff
@@ -9,6 +9,7 @@
 #include <memory>
 
 #include <QtGui/QGuiApplication>
+#include <QtGui/QFont>
```
ahayzen-kdab, line 12: needs to be before QGuiApplication import i think?

crates/cxx-qt-lib/src/gui/qguiapplication.rs
```diff
@@ -110,6 +118,10 @@ impl QGuiApplication {
         ffi::qguiapplication_exec(self)
     }
 
+    pub fn font(&self) -> QFont {
```
ahayzen-kdab, line 121: ideally add doc line

crates/cxx-qt-lib-headers/include/gui/qguiapplication.h
```diff
@@ -18,5 +19,11 @@ namespace cxxqtlib1 {
 ::std::unique_ptr<QGuiApplication>
 qguiapplicationNew(const QVector<QByteArray>& args);
 
+void
+qapplicationSetFont(QGuiApplication& app, const QFont& font);
```
ahayzen-kdab, line 23: these should probably be called `qguiapplication` rather than `qapplication` (sorry)

ahayzen-kdab (approved):
LGTM :-)

PR #809, opened 2024-01-16
Add QPen method in QPainter class

Montel:


crates/cxx-qt-lib/src/gui/qpainter.rs
```diff
@@ -158,6 +163,10 @@ mod ffi {
         #[rust_name = "set_opacity"]
         fn setOpacity(self: Pin<&mut QPainter>, opacity: f64);
 
+        /// Draws the outline (strokes) the path path with the pen specified by pen
+        #[rust_name = "stroke_path"]
+        fn strokePath(self: Pin<&mut QPainter>, path: &QPainterPath, pen: &QPen);
```
ahayzen-kdab, line 168: try to keep the ordering A->Z as it becomes increasingly hard to find methods :-)

crates/cxx-qt-lib/src/gui/qpainter.rs
```diff
@@ -123,6 +125,9 @@ mod ffi {
         /// Returns the opacity of the painter. The default value is 1.
         fn opacity(self: &QPainter) -> f64;
 
+        /// Returns the painter's current pen.
+        fn pen(self: &QPainter) -> &QPen;
```
ahayzen-kdab, line 129: hmm what about `setPen(const QPen&)` we need some way of solving / naming the overloads here :thinking: Or whether we always go to the most basic like QColor in this case (i think we have done this elsewhere)

LeonMatthesKDAB (approved):
LGTM

ahayzen-kdab:
Looks good, we need to figure out a plan for overloads at some point...

PR #811, opened 2024-01-17
qt-build-utils: only run qmlcachegen when there are QML files

ahayzen-kdab:
Otherwise we have a build failure as qmlcache_loader.cpp file does not exist.


LeonMatthesKDAB (approved):
LGTM

PR #814, opened 2024-01-17
Add bindings for QtMsgType, QMessageLogContext and qt_message_output

redstrate:
This allows Rust and CXX-Qt applications to send messages to the Qt logger.


crates/cxx-qt-lib/src/core/qtlogging.rs
```diff
@@ -0,0 +1,138 @@
...
+        unsafe {
+            CStr::from_ptr(ffi::category(self))
+        }
+    }
+
+    pub fn set_category(&mut self, category: &CStr) {
+        unsafe {
+            ffi::set_category(self, category.as_ptr());
+        }
+    }
```
redstrate, line 124: I'm not really happy with the string handling here, but I think this might be the best we can do. The string must live long enough for it to be used later by the message handler, so it can't be a temporary object (or else the memory is pointing to nothing.)

Might need to add some documentation clarifying this?

examples/cargo_without_cmake/src/main.rs
```diff
@@ -28,6 +32,14 @@ fn main() {
...
     }
 
+    let str = CString::new("test").unwrap();
+
+    let mut context: QMessageLogContext = QMessageLogContext::default();
+    context.set_line(5);
+    context.set_file(&str);
+
+    qt_message_output(QtMsgType::QtInfoMsg, &context, &QString::from("Test!"));
+
```
redstrate, line 42: Accidentally added this, will remove/move later

crates/cxx-qt-lib/src/core/qtlogging.rs
```diff
@@ -0,0 +1,138 @@
...
+        type QString = crate::QString;
+
+        include!("cxx-qt-lib/qtlogging.h");
+        type QMessageLogContext = crate::QMessageLogContext;
+        type QtMsgType;
+
+        fn qt_message_output(msgType: QtMsgType, context: &QMessageLogContext, string: &QString);
+
+        #[cxx_name = "qmessagelogcontext_line"]
+        fn line(context: &QMessageLogContext) -> i32;
```
ahayzen-kdab, line 34: you should be able to avoid the C++ proxy for types that CXX can express here and just do

```rust
fn line(self: &QMessageContext) -> i32;
```

Then you can remove the ffi::line() call below.

crates/cxx-qt-lib/src/core/qtlogging.rs
```diff
@@ -0,0 +1,138 @@
...
+
+        #[cxx_name = "qmessagelogcontext_set_line"]
+        fn set_line(context: &mut QMessageLogContext, line: i32);
```
ahayzen-kdab, line 37: same here

crates/cxx-qt-lib/src/core/qtlogging.rs
ahayzen-kdab: If you take `&str` here as the type you can use that in CXX, then in C++ you could use `rust::Str::data()` https://cxx.rs/binding/str.html to get the `const char*`

ahayzen-kdab: Or if it does need to live long enough can it be a `String` and take ownership of it somehow? then use rust::String::data() in C++

examples/cargo_without_cmake/src/main.rs
ahayzen-kdab: good for testing for now :-)

examples/cargo_without_cmake/src/main.rs
```diff
@@ -28,6 +32,14 @@ fn main() {
         engine.load(&QUrl::from("qrc:/qt/qml/com/kdab/cxx_qt/demo/qml/main.qml"));
     }
 
+    let str = CString::new("test").unwrap();
+
+    let mut context: QMessageLogContext = QMessageLogContext::default();
+    context.set_line(5);
+    context.set_file(&str);
```
ahayzen-kdab, line 39: wonder if we want CString or just a normal Rust string here then convert it behind the scenes?

crates/cxx-qt-lib/src/core/qtlogging.rs
ahayzen-kdab: Actually `&str` might work quite well, as with the lifetimes specified to CXX as well (`fn set_category<'a>(self: &'a mut QMessageLogContext, category: &'a str)` ) to bind it to QMessageLogContext it would force you to do the following or use a static str.
```rust
let string = ...;
context.set_category(&string);
```

wonder if the lifetimes would be needed or if it'd work without, something to experiment with.

crates/cxx-qt-lib-headers/include/core/qtlogging.h
```diff
@@ -0,0 +1,54 @@
+// clang-format off
```
ahayzen-kdab, line 1: (also FYI best to fork the repository and manage your branches there rather than adding them to the kdab repository)

crates/cxx-qt-lib/src/core/qtlogging.cpp
```diff
@@ -0,0 +1,7 @@
+// clang-format off
+// SPDX-FileCopyrightText: 2024 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// clang-format on
+// SPDX-FileContributor: Joshua Goins <joshua.goins@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+#include "cxx-qt-lib/qtlogging.h"
```
ahayzen-kdab, line 7: we need the asserts to be put here for checking the size, checking if copy / destructors are trivial (and if they aren't then we need to impl Clone / Drop). And then implementation from the header should be placed here.

crates/cxx-qt-lib/src/core/qtlogging.rs
ahayzen-kdab: ah i see there are no actual member functions in QMessageLogContext :see_no_evil: so you can't :sweat_smile:

crates/cxx-qt-lib/src/core/qtlogging.cpp
redstrate: Done, I think I added the correct check? I need to dig through some of the other trivial types, but I'm not sure what's exactly needed

examples/cargo_without_cmake/src/main.rs
redstrate: It should be CString, I would like to use String but there's numerous issues. Most notably it doesn't have a null terminator, and there's no good way to get around that. For now a CString is good enough though :)

examples/cargo_without_cmake/src/main.rs
redstrate: Removed now that it's ready for review

crates/cxx-qt-lib/src/core/qtlogging.rs
redstrate: Managed to work through this, now it's working as expected and won't allow you to create temporaries.

crates/cxx-qt-lib-headers/include/core/qtlogging.h
```diff
@@ -0,0 +1,38 @@
...
+// SPDX-FileContributor: Joshua Goins <joshua.goins@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+#pragma once
+
+#include <QtCore/QtLogging>
+#include <QDebug>
+#include "rust/cxx.h"
+
+int qmessagelogcontext_line(const QMessageLogContext &context);
```
ahayzen-kdab, line 13: this file needs clang-format running on it

crates/cxx-qt-lib/src/core/qtlogging.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include "../assertion_utils.h"
+
+// QMessageLogContext has three "const char*" members for line, category, etc
+// https://codebrowser.dev/qt5/qtbase/src/corelib/global/qlogging.h.html#QMessageLogContext
+assert_alignment_and_size(QMessageLogContext, alignof(intptr_t), sizeof(intptr_t) * 4);
+
+static_assert(::std::is_trivially_copyable<QMessageLogContext>::value,
+              "QMessageLogContext must be trivially copyable");
+
+int qmessagelogcontext_line(const QMessageLogContext &context) {
```
ahayzen-kdab, line 18: this file needs clang-format running on it :-)

crates/cxx-qt-lib/src/core/qtlogging.rs
```diff
@@ -0,0 +1,139 @@
...
+    #[namespace = "rust::cxxqtlib1"]
+    unsafe extern "C++" {
+        #[doc(hidden)]
+        #[rust_name = "qmessagelogcontext_default"]
+        fn construct() -> QMessageLogContext<'static>;
+    }
+}
+
+#[repr(C)]
+pub struct QMessageLogContext<'a> {
```
ahayzen-kdab, line 68: documentation lines would be useful

crates/cxx-qt-lib/src/core/qtlogging.rs
```diff
@@ -0,0 +1,139 @@
...
+}
+
+impl Default for QMessageLogContext<'_> {
+    fn default() -> Self {
+        ffi::qmessagelogcontext_default()
+    }
+}
+
+impl<'a> QMessageLogContext<'a> {
+    pub fn line(&self) -> i32 {
```
ahayzen-kdab, line 83: documentation lines to each of these would be useful

crates/cxx-qt-lib/src/core/qtlogging.rs
```diff
@@ -0,0 +1,139 @@
...
+        include!("cxx-qt-lib/qstring.h");
+        type QString = crate::QString;
+
+        include!("cxx-qt-lib/qtlogging.h");
+        type QMessageLogContext<'a> = crate::QMessageLogContext<'a>;
+        type QtMsgType;
+
+        fn qt_message_output(msgType: QtMsgType, context: &QMessageLogContext, string: &QString);
+
+        #[cxx_name = "qmessagelogcontext_line"]
```
ahayzen-kdab, line 34: add `#[doc(hidden)]` to all of these methods that are used internally

crates/cxx-qt-lib/src/core/qtlogging.rs
```diff
@@ -0,0 +1,139 @@
...
+        include!("cxx-qt-lib/common.h");
+
+        include!("cxx-qt-lib/qstring.h");
+        type QString = crate::QString;
+
+        include!("cxx-qt-lib/qtlogging.h");
+        type QMessageLogContext<'a> = crate::QMessageLogContext<'a>;
+        type QtMsgType;
+
+        fn qt_message_output(msgType: QtMsgType, context: &QMessageLogContext, string: &QString);
```
ahayzen-kdab, line 32: add a doc line to this method as it's public

crates/cxx-qt-lib/src/core/qtlogging.rs
```diff
@@ -0,0 +1,139 @@
...
+// SPDX-License-Identifier: MIT OR Apache-2.0
+use cxx::{type_id, ExternType};
+use std::ffi::c_char;
+use std::ffi::CStr;
+use std::marker::PhantomData;
+
+#[cxx::bridge]
+mod ffi {
+    #[repr(i32)]
+    enum QtMsgType {
```
ahayzen-kdab, line 13: add doc line to the enum and to each value

crates/cxx-qt-lib/src/core/qtlogging.rs
```diff
@@ -0,0 +1,139 @@
...
+        QtDebugMsg = 0,
+        QtInfoMsg = 4,
+        QtWarningMsg = 1,
+        QtFatalMsg = 3,
+        QtCriticalMsg = 2,
+        QtSystemMsg = 2
+    }
+
+    unsafe extern "C++" {
+        include!("cxx-qt-lib/common.h");
```
ahayzen-kdab, line 23: while here, best to move the common include into the next block below as that is the only ( ? ) place it is used for the `construct` method?

```rust
    unsafe extern "C++" {
```

crates/cxx-qt-lib/src/core/mod.rs
```diff
@@ -98,6 +98,9 @@ pub use qvariant::{QVariant, QVariantValue};
 mod qvector;
 pub use qvector::{QVector, QVectorElement};
 
+mod qtlogging;
+pub use qtlogging::{QMessageLogContext, QtMsgType, qt_message_output};
```
ahayzen-kdab, line 102: need `cargo fmt`

crates/cxx-qt-lib/src/core/qtlogging.rs
```diff
@@ -0,0 +1,139 @@
...
+    }
+}
+
+#[repr(C)]
+pub struct QMessageLogContext<'a> {
+    version: i32,
+    line: i32,
+    file: &'a *const c_char,
+    function: &'a *const c_char,
+    category: &'a *const c_char
```
ahayzen-kdab, line 73: needs `cargo fmt` running on this file

crates/cxx-qt-lib/src/core/qtlogging.cpp
```diff
@@ -0,0 +1,48 @@
...
+// SPDX-License-Identifier: MIT OR Apache-2.0
+#include "cxx-qt-lib/qtlogging.h"
+
+#include "../assertion_utils.h"
+
+// QMessageLogContext has three "const char*" members for line, category, etc
+// https://codebrowser.dev/qt5/qtbase/src/corelib/global/qlogging.h.html#QMessageLogContext
+assert_alignment_and_size(QMessageLogContext, alignof(intptr_t), sizeof(intptr_t) * 4);
+
+static_assert(::std::is_trivially_copyable<QMessageLogContext>::value,
```
ahayzen-kdab, line 15: is it actually trivially_copyable ? types that have that don't need `rust::IsRelocatable` definition.


Normally you either have to check all of these, if copy is false then you need to impl `Clone` manually and if deconstructible is failse then you need to manually implement `Drop` (which tends to be the case for these types.
```cpp
// in the header have the isrelocatable
rust::IsRelocatable<T> ...

// in the source
static_assert(!::std::is_trivially_copy_assignable<T>::value);
static_assert(!::std::is_trivially_copy_constructible<T>::value);
static_assert(!::std::is_trivially_destructible<T>::value);

static_assert(QTypeInfo<T>::isRelocatable);
```

or if the type is actually trivial then you only have the following and then derive the rest

```cpp
static_assert(::std::is_trivially_copyable<T>::value);
```

As this type uses pointers internally it suggests it's probably the former (or is it actually trivially copyable as it's just `const char*` ?) but it looks like the type doesn't have any custom deconstructors or constructors so you probably don't implement Clone and Drop, fun :-)

crates/cxx-qt-lib/src/core/qtlogging.rs
```diff
@@ -0,0 +1,139 @@
...
+        fn line(context: &QMessageLogContext) -> i32;
+
+        #[cxx_name = "qmessagelogcontext_set_line"]
+        fn set_line(context: &mut QMessageLogContext, line: i32);
+
+        #[cxx_name = "qmessagelogcontext_file"]
+        unsafe fn file(context: &QMessageLogContext) -> *const c_char;
+
+        #[cxx_name = "qmessagelogcontext_set_file"]
+        unsafe fn set_file(context: &mut QMessageLogContext, file: *const c_char);
```
LeonMatthesKDAB, line 44: I'm not a big fan of all the `setters` here.

For QMessageLogContext to be safe to use in Rust, it's lifetime needs to be bound to the lifetime of the `c_char*` it refers to.
The setters complicate this a lot, as they actually change the lifetime of the QMessageLogContext, which I don't even know how to express in Rust.
I'd recommend just providing a single constructor that takes `&'a CStr` for all the texts and binding QMessageLogContext to that lifetime.
For most use-cases it's probably fine if it's just immutable after construction, right?

crates/cxx-qt-lib/src/core/qtlogging.rs
```diff
@@ -0,0 +1,139 @@
...
+        fn construct() -> QMessageLogContext<'static>;
+    }
+}
+
+#[repr(C)]
+pub struct QMessageLogContext<'a> {
+    version: i32,
+    line: i32,
+    file: &'a *const c_char,
+    function: &'a *const c_char,
```
LeonMatthesKDAB, line 72: This is not the correct type!

`&*const c_char` is a reference to a pointer, so similar to a `const c_char**`.
So reading this from Rust after being assigned in C++ is undefined behavior, as C++ thinks it's a `const char*`.

This needs to just be a `*const c_char`.

To introduce the required lifetime, you can add a single `PhantomData<&'a c_char>` member to the struct.
I'm a bit uncertain how that interacts with `#[repr(C)]`. In `#[repr(Rust)]` the `PhantomData` won't change the size of the struct, but in `#[repr(C)]` it might add a single byte, which unfortunately wouldn't allow us to make this type trivial...
Please use static assertions to check what happens either way :)

LeonMatthesKDAB (changes requested):
Good first pass, unfortunately the lifetimes need sorting out.

crates/cxx-qt-lib/src/core/qtlogging.cpp
redstrate: I think the type is trivial, although you suggest to remove IsRelocatable in that case (IIUC) the CXX compilation fails with:

```
  cargo:warning=/home/josh/work/cxx-qt/target/debug/build/cxx-qt-lib-9ea1616c001b64fd/out/cxx-qt-gen/src/qtlogging.cxx.cpp:62:50: error: static assertion failed: type QMessageLogContext should be trivially move constructible and trivially destructible in C++ to be used as a return value of `qmessagelogcontext_default` or non-pinned mutable reference in signature of `set_line`, `set_file`, `set_function` in Rust
  cargo:warning=   62 |     ::rust::IsRelocatable<::QMessageLogContext>::value,
```

crates/cxx-qt-lib/src/core/qtlogging.rs
redstrate: Done

crates/cxx-qt-lib/src/core/mod.rs
redstrate: Done

crates/cxx-qt-lib/src/core/qtlogging.rs
redstrate: Done

crates/cxx-qt-lib/src/core/qtlogging.rs
redstrate: Done

crates/cxx-qt-lib/src/core/qtlogging.rs
redstrate: Done

crates/cxx-qt-lib/src/core/qtlogging.rs
redstrate: Done

crates/cxx-qt-lib/src/core/qtlogging.rs
redstrate: Done

crates/cxx-qt-lib/src/core/qtlogging.rs
redstrate: Done

crates/cxx-qt-lib/src/core/qtlogging.cpp
redstrate: Done

crates/cxx-qt-lib/src/core/qtlogging.rs
redstrate: Added, it seems like PhantomData does not change the memory layout of the struct - it's still 32 bytes. However, how do I add a static_assert for this? It looks like all of the ones in the .cpp files are checking the size of the Qt type, not the Rust type?

crates/cxx-qt-lib/src/core/qtlogging.rs
redstrate: Yeah the changing-lifetime-thing is something I only discovered after doing the other PR :)

I also agree the setters are unnecessary, everything can be thrown into a constructor. Done

crates/cxx-qt-lib/src/core/qtlogging.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include "../assertion_utils.h"
+
+// QMessageLogContext has three "const char*" members for line, category, etc
+// https://codebrowser.dev/qt5/qtbase/src/corelib/global/qlogging.h.html#QMessageLogContext
+assert_alignment_and_size(QMessageLogContext,
+                          alignof(intptr_t),
+                          sizeof(intptr_t) * 4);
+
+static_assert(::std::is_trivially_copyable<QMessageLogContext>::value,
+              "QMessageLogContext must be trivially copyable");
```
ahayzen-kdab, line 18: so looks like the CI failure on Windows at least might be coming from this?

```
  qstring.cpp
  src/core/qstring.cpp(46): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data
  qstringlist.cpp
  qurl.cpp
  qtlogging.cpp
  src/core/qtlogging.cpp(17): error C2338: static_assert failed: 'QMessageLogContext must be trivially copyable'
```

I also note that qstring warning that is unrelated but interesting :-)

redstrate: Meh I somehow missed the Q_DISABLE_COPY on the struct

redstrate: Fixed

crates/cxx-qt-lib/src/core/qtlogging.rs
LeonMatthesKDAB: The constructor still uses the setters.
Please add a single creation function that just takes the required arguments and returns the constructred QMessageLogContext, then we can remove the setters entirely.

crates/cxx-qt-lib/src/core/qtlogging.rs
LeonMatthesKDAB: `&'a *const c_char` is still not the correct type for the constructor argument!

I think it just needs to be `&'a CStr` (see: https://doc.rust-lang.org/std/ffi/struct.CStr.html ), which is the C-compatible equivalent to `&str`.

For the static assertions, we could use: https://crates.io/crates/static_assertions
We have tried to avoid unnecessary dependencies, but I think this crate should be small and stable enough to not bother anyone. @ahayzen-kdab agree?

crates/cxx-qt-lib/src/core/qtlogging.rs
```diff
@@ -0,0 +1,146 @@
...
+        context
+    }
+
+    /// The line number given to the message handler.
+    pub fn line(&self) -> i32 {
+        ffi::line(self)
+    }
+
+    /// The file path given to the message handler.
+    pub fn file(&self) -> &CStr {
```
LeonMatthesKDAB, line 123: This needs to return `&'a CStr`, as the lifetime is bound by the lifetime that is stored within the QMessageLogContext, not by the lifetime of the reference to the QMessageLogContext.

crates/cxx-qt-lib/src/core/qtlogging.rs
```diff
@@ -0,0 +1,146 @@
...
+        unsafe { CStr::from_ptr(ffi::file(self)) }
+    }
+
+    /// The name of the function given to the message handler.
+    pub fn function(&self) -> &CStr {
```
LeonMatthesKDAB, line 128: Also: `&'a CStr`

crates/cxx-qt-lib/src/core/qtlogging.rs
```diff
@@ -0,0 +1,146 @@
...
+        unsafe { CStr::from_ptr(ffi::function(self)) }
+    }
+
+    /// The category given to the message handler.
+    pub fn category(&self) -> &CStr {
```
LeonMatthesKDAB, line 133: Also: `&'a CStr`

crates/cxx-qt-lib/src/core/qtlogging.rs
redstrate: Done

crates/cxx-qt-lib/src/core/qtlogging.rs
redstrate: Done

crates/cxx-qt-lib/src/core/qtlogging.rs
redstrate: > I think it just needs to be &'a CStr (see: https://doc.rust-lang.org/std/ffi/struct.CStr.html ), which is the C-compatible equivalent to &str.

Oops yes, I should've known better. I use CStr all the time :sweat_smile: Will wait for andrew's opinion wrt/ static_assertions.

crates/cxx-qt-lib/src/core/qtlogging.rs
redstrate: Oops, why did I mark this as done? _Actually_ fixed now, the setters are gone. I also re-arranged the Rust constructor to match the C++ one.

LeonMatthesKDAB (commented):
I'm pretty happy with this PR now.
Having static_assertions would be a plus. Waiting for opinion from @ahayzen-kdab there.

crates/cxx-qt-lib/src/core/qtlogging.rs
```diff
@@ -0,0 +1,129 @@
...
+        /// A debug message.
+        QtDebugMsg = 0,
+        /// An info message.
+        QtInfoMsg = 4,
+        /// A warning message.
+        QtWarningMsg = 1,
+        /// A fatal message.
+        QtFatalMsg = 3,
+        /// A critical message.
+        QtCriticalMsg = 2,
```
redstrate, line 24: question while waiting: do we rename enum values to something more sensible? in this case the names are weird to avoid collisions, since in qtlogging.h these were just an `enum`. hence why there' prefixed with `Qt` and suffixed with `Msg`.

crates/cxx-qt-lib/src/core/qtlogging.rs
```diff
@@ -0,0 +1,129 @@
...
+        include!("cxx-qt-lib/common.h");
+
+        #[doc(hidden)]
+        #[rust_name = "construct_qmessagelogcontext"]
+        unsafe fn construct(
+            file_name: *const c_char,
+            line_number: i32,
+            function_name: *const c_char,
+            category_name: *const c_char,
+        ) -> QMessageLogContext<'static>;
```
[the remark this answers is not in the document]
redstrate: Done

crates/cxx-qt-lib/src/core/qtlogging.rs
redstrate: I have added a static assertion now, let me know if we should have some others

LeonMatthesKDAB: Looks good now.

LeonMatthesKDAB (approved):
LGTM, @redstrate please rebase, then we can merge.

LeonMatthesKDAB (approved):

PR #815, opened 2024-01-18
github: use 4 cores and 3 cores when available in CI

ahayzen-kdab:
https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/


ahayzen-kdab:
The test phase shows 4 jobs running at once, but there isn't much gain as they can still be blocked on the file lock between them. Building has minimal gains, i think the inner cargo build is likely already detecting the number of cores separate to CMake.

LeonMatthesKDAB (approved):
We're still bottlenecked by the macOS, but this still doesn't hurt :)

PR #820, opened 2024-01-22
Set RUSTC_WRAPPER to sccache to speed up building

LeonMatthesKDAB:
It turns out cc-rs falls back to `RUSTC_WRAPPER` to detect whether sccache is used.

So we can use this to speed up cxx-qt-lib and our other build scripts in CI.

However: This requires a small patch to cc-rs, as the fallback weirdly is broken if the full compiler path is provided.

This however, is a fairly easy alternative to https://github.com/corrosion-rs/corrosion/pull/474


CMakeLists.txt
```diff
@@ -218,6 +218,10 @@ add_test(NAME cargo_tests COMMAND cargo test --release --all-features --target-d
 add_test(NAME cargo_doc COMMAND cargo doc --release --all-features --target-dir ${CARGO_TARGET_DIR})
 add_test(NAME cargo_clippy COMMAND cargo clippy --release --all-features --target-dir ${CARGO_TARGET_DIR} -- -D warnings)
 
+if(CMAKE_RUSTC_WRAPPER)
```
ahayzen-kdab, line 221: so we already set `RUSTC_WRAPPER` as an env var in CI and devs might set it locally, should CMake read from this env var as expose it to targets directly ? Or is there a reason it needs to go through the `CompilerCaching.cmake` logic which forces either ccache or sccache ?

LeonMatthesKDAB: I've updated the CMake code to only set `CMAKE_RUSTC_WRAPPER` if it isn't defined in the environment.

As I think it's still valuable to have it auto-detected in our cmake build.
Then users can still overwrite it through the environment variable, but get it by default if it's just installed.

ahayzen-kdab: Nice :-)

ahayzen-kdab (approved):
LGTM :-)

PR #821, opened 2024-01-24
CI: Split up cargo test

LeonMatthesKDAB:
This should help investigate why Windows build is taking so long, even with caching.

### DIFF
--- a/.github/workflows/github-cxx-qt-tests.yml
+++ b/.github/workflows/github-cxx-qt-tests.yml
@@ -291,9 +291,17 @@ jobs:
       working-directory: ${{ matrix.workspace }}
       env:
         RUSTC_WRAPPER: sccache
-    - name: "Test"
+    - name: "CTest"
       run: ctest ${{ matrix.ctest_args }} -C Release -T test --output-on-failure --parallel ${{ matrix.cores }}
       working-directory: ${{ matrix.workspace }}/build
+      env:
+        RUSTC_WRAPPER: sccache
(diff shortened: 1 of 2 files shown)


.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -291,14 +291,22 @@ jobs:
...
       run: ctest ${{ matrix.ctest_args }} -C Release -T test --output-on-failure --parallel ${{ matrix.cores }}
       working-directory: ${{ matrix.workspace }}/build
       env:
         RUSTC_WRAPPER: sccache
         QT_QPA_PLATFORM: ${{ matrix.qt_qpa_platform }}
         QT_SELECT: qt${{ matrix.qt_version }}
 
+    - name: "Cargo test"
+      run: cargo test --release --all-features --target-dir ${{ matrix.workspace }}/build/cargo/build
+      working-directory: ${{ matrix.workspace }}
```
ahayzen-kdab, line 304: does this need to be `/build` like above or something ? as it's failing to find Qt on macOS / Windows aka vcpkg ...

LeonMatthesKDAB: I think it just needs the right `QMAKE`, but I don't really know how to get that out of CMake/Vcpkg :thinking: 

I think #822 does pretty much what I wanted, as it allows us to see the output of cargo test.

LeonMatthesKDAB:
#822 should be just as helpful.

PR #822, opened 2024-01-24
Enable verbose ctest output

LeonMatthesKDAB:
This could help investigate why CTest is slow on Windows

### DIFF
--- a/.github/workflows/github-cxx-qt-tests.yml
+++ b/.github/workflows/github-cxx-qt-tests.yml
@@ -283,19 +283,15 @@ jobs:
         -S . -B build
       working-directory: ${{ matrix.workspace }}
       env:
-        RUSTC_WRAPPER: sccache
         CC: ${{ matrix.cc }}
         CXX: ${{ matrix.cxx }}
     - name: "Build"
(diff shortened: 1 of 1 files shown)


LeonMatthesKDAB:
So it seems the issue with Windows is indeed that building before `cargo test` takes around 6 min.
This is very strange, as the build with CMake just takes 2 min. So why does building from within ctest all of a sudden take three times as long?
It almost seems as if compiler caching isn't working there anymore?

ahayzen-kdab:
> So it seems the issue with Windows is indeed that building before `cargo test` takes around 6 min. This is very strange, as the build with CMake just takes 2 min. So why does building from within ctest all of a sudden take three times as long? It almost seems as if compiler caching isn't working there anymore?

Are we hitting that `cargo` and `cmake` build into different directories so then you get duplicated `/target` folders and therefore it needs to build everything again ? (I have this problem locally sometimes)

ahayzen-kdab:
@LeonMatthesKDAB do you still need this ? or can it be closed ?

ahayzen-kdab:
Not needed anymore :-)

PR #825, opened 2024-01-30
cxx-qt-build: set minimum macOS version

Be-ing:
to override cc's default which is too low for CXX-Qt. Setting the MACOSX_DEPLOYMENT_TARGET environment variable in CI is fine, but that doesn't solve the problem for downstream users.

relies on a pending PR for cc https://github.com/rust-lang/cc-rs/pull/938

### DIFF
--- a/crates/cxx-qt-build/src/lib.rs
+++ b/crates/cxx-qt-build/src/lib.rs
@@ -510,6 +510,11 @@ impl CxxQtBuilder {
             builder.flag_if_supported("-std=c++17");
             // MinGW requires big-obj otherwise debug builds fail
             builder.flag_if_supported("-Wa,-mbig-obj");
+            // macOS
+            // We need
+            //   - std::filesystem::path from 10.15
+            //   - std::shared_mutex from 10.12
+            builder.apple_deployment_target(&"10.5");
             // Enable Qt Gui in C++ if the feature is enabled
             #[cfg(feature = "qt_gui")]
             builder.define("CXX_QT_GUI_FEATURE", None);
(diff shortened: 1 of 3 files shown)


ahayzen-kdab:
Fun issue on Linux
```console
   cargo:warning=Compiler version doesn't include clang or GCC: "/usr/bin/c++" "--version"
  cargo:warning=as: unrecognized option '-mbig-obj'

   exit status: 1
```

Be-ing:
Yeah I saw that... pretty puzzled why this change would make that show up.

crates/cxx-qt-build/src/lib.rs
```diff
@@ -510,6 +510,11 @@ impl CxxQtBuilder {
             builder.flag_if_supported("-std=c++17");
             // MinGW requires big-obj otherwise debug builds fail
             builder.flag_if_supported("-Wa,-mbig-obj");
+            // macOS
+            // We need
+            //   - std::filesystem::path from 10.15
+            //   - std::shared_mutex from 10.12
+            builder.apple_deployment_target(&"10.5");
```
ahayzen-kdab, line 517: does the upstream change mean we still need this ? https://github.com/rust-lang/cc-rs/pull/943 Sounds like it reads the SDK's `DefaultDeploymentTarget` again ?

Be-ing: nah let's close this

Be-ing:
Superceded by https://github.com/KDAB/cxx-qt/pull/848

Issue #823, opened 2024-01-26
CXX-Qt-Lib: Add CXX-compatible templates for make_unique, make_shared, etc.


LeonMatthesKDAB:
To complete our list of templated constructors in `cxx-qt-lib/common.h`.

We probably want to have at least:
- [x] `construct`
- [x] `make_unique` (added in #779 - to be released with 0.7)
- [ ]  `make_shared`
- [ ] `new`/`new_ptr` (as requested in #819 by @knoxfighter)

ahayzen-kdab:
Note this is also something that if cxx-qt-lib is split out could go into cxx-qt or a -common thing later :thinking:

ahayzen-kdab:
See comments in #840, consider having an up and downcast helper

```cpp
template<typename Base, typename T>
Base & upcast(T& t)
{
   // Maybe add some static_assert that Base is indeed base class of T
    return static_cast<Base&>(t);
}
```

BenFordTytherington:
#1068 Adds casting helpers :)

BenFordTytherington:
Documentation added in https://github.com/KDAB/cxx-qt/pull/1229

PR #830, opened 2024-02-02
Add make_shared template

Montel:


LeonMatthesKDAB:
See: #823

LeonMatthesKDAB (approved):
LGTM. I assume you tested this works as expected?

LeonMatthesKDAB:
@Montel I think this can be merged now?

ahayzen-kdab:
Once we have #843 I wonder if we move these into a `cxx-qt` header, so `cxx-qt-common` / `cxx-qt`

PR #832, opened 2024-02-02
Add missing qfont methods

Montel:


crates/cxx-qt-lib/src/gui/qfont.rs
```diff
@@ -89,10 +91,28 @@ mod ffi {
...
         fn defaultFamily(self: &QFont) -> QString;
 
+        /// Returns true if a window system font exactly matching
+        /// the settings of this font is available.
+        #[rust_name = "exact_match"]
+        fn exactMatch(self: &QFont) -> bool;
+
+        /// Returns the requested font family name, i.e. the name
+        /// set in the constructor or the last setFont() call.
+        fn family(self: &QFont) -> QString;
```
LeonMatthesKDAB, line 101: At least for Qt 6, the documentation states that this is always the first family in the `families` call, which may be an empty array.

Maybe we should have this function return an `Option<QString>` and just be implemented in Rust by calling `families.first()`?

crates/cxx-qt-lib/src/gui/qfont.rs
```diff
@@ -89,10 +91,28 @@ mod ffi {
...
+        fn families(self: &QFont) -> QStringList;
+
         /// Returns true if fixed pitch has been set; otherwise returns false.
         #[rust_name = "fixed_pitch"]
         fn fixedPitch(self: &QFont) -> bool;
 
+        /// Sets this font to match the description descrip. The description is a comma-separated
+        /// list of the font attributes, as returned by toString().
+        #[rust_name = "from_string"]
+        fn fromString(self: &mut QFont, descrip: &QString) -> bool;
```
LeonMatthesKDAB, line 114: I wonder if we should make this return a `Result` of some sort?

Don't know what the error type would be though...

crates/cxx-qt-lib/src/gui/qfont.rs
LeonMatthesKDAB: If you do this refactor, you can also keep this method and set `#[rust_name="family_or_default"]`, as we have done with most other such APIs

LeonMatthesKDAB (commented):
Mostly looks good.
We could maybe make `family` and `fromString` more rusty.

I don't have a strong opinion on those though and would be okay to merge as-is...

crates/cxx-qt-lib/src/gui/qfont.rs
Montel: I think it return false when we can't create font from string.
I don't see if it's necessary to return a Result here.

LeonMatthesKDAB: `Result` just communicates better that this is an operation that may fail.

The bool indicates this as well, but it can just be ignored. A `Result` would be `#[must_use]`.
But I don't know of a good type to use, `Result<(),()>` I guess, as we don't have more information...
Would be a bit silly though I guess.
We can also merge as-is.

LeonMatthesKDAB (approved):
Still not a great fan of the `bool` to indicate failure, but I guess it's good enough.
Nothing will explode if the operation fails.

PR #834, opened 2024-02-02
Add missing qpainter methods

Montel:


crates/cxx-qt-lib/src/gui/qpainter.rs
```diff
@@ -44,6 +44,11 @@ mod ffi {
         #[rust_name = "background_mode"]
         fn backgroundMode(self: &QPainter) -> BGMode;
 
+        /// Flushes the painting pipeline and prepares for the user issuing commands directly
+        /// to the underlying graphics context. Must be followed by a call to endNativePainting().
+        #[rust_name = "begin_native_painting"]
+        fn beginNativePainting(self: Pin<&mut QPainter>);
```
LeonMatthesKDAB, line 50: Oh, warning! This could lead to Undefined Behavior and isn't safe.

The Documentation warns that this call **must** be followed by a call to endNativePainting().

Also, I'm pretty sure you shouldn't use the QPainter again before calling `endNativePainting()`.

I see two ways forward for this:
- We don't know the right invariants, make this function `unsafe` (maybe `endNativePainting()` too).
- Use RAII and return a guard that is mutable-lifetime-bound to the QPainter (so you can't call anything mutable on the QPainter) and that automatically calls `endNativePainting()` when it goes out of scope.

For now it might be easiest to just make both functions `unsafe` and think about adding an RAII guard later.

LeonMatthesKDAB: We should maybe also link to the Qt documentation for this specific function, as it includes more details about what exactly is reset when calling this function.

crates/cxx-qt-lib/src/gui/qpainter.rs
```diff
@@ -119,10 +124,19 @@ mod ffi {
         #[rust_name = "draw_polyline"]
         fn drawPolyline(self: Pin<&mut QPainter>, points: &QPolygon);
 
+        /// Draws the current rectangle with the current pen and brush.
+        #[rust_name = "draw_rect"]
```
LeonMatthesKDAB, line 128: Maybe rename this to `draw_rect_f`?
We probably don't need `draw_rect`, but that leaves us some naming wiggle room to add it.

LeonMatthesKDAB (changes requested):
I think `beginNativePainting()` could lead to undefined behavior, so it needs to be `unsafe`.

crates/cxx-qt-lib/src/gui/qpainter.rs
ahayzen-kdab: I would do `unsafe` in cxx-qt-lib / cxx-qt-lib-extras.

Then RAII route should probably be part of a separate crate as this is similar to the `begin_reset_model()` + `end_reset_model()` of abstract item models. We originally thought we would have these wrapped / as helpers that are RAII based in another crate that could be built ontop of cxx-qt / cxx-qt-lib.

Montel: ok I remove them for the moment.

LeonMatthesKDAB: You didn't have to remove them, they could just be `unsafe` :)

But anyway, we can merge this now and re-add them back in later.

LeonMatthesKDAB (approved):

PR #836, opened 2024-02-06
github: always use same cache rather than creating new ones

ahayzen-kdab:
Before we had the problem that we tried to always restore the same cache but only save it when a commit reached `main`, this had the problem that if nothing reached `main` the cache would expire after 1 week.

Instead always save the cache, this will cause the cache to have some items from branches, but is probably still more beneficial and allows us to avoid evictions.

### DIFF
--- a/.github/workflows/github-cxx-qt-tests.yml
+++ b/.github/workflows/github-cxx-qt-tests.yml
@@ -198,7 +198,7 @@ jobs:
           ${{ matrix.cargo_dir }}/bin/sccache${{ matrix.exe_suffix }}
           ${{ matrix.cargo_dir }}/bin/mdbook${{ matrix.exe_suffix }}
           ${{ matrix.cargo_dir }}/bin/mdbook-linkcheck${{ matrix.exe_suffix }}
-        key: ${{ matrix.os }}_sccache-0.4.1_mdbook-0.4.28_mdbook-linkcheck-0.7.7
+        key: ${{ matrix.os }}_sccache-0.7.6_mdbook-0.4.36_mdbook-linkcheck-0.7.7
     - name: "Build Rust tools"
       if: steps.rust-tools-cache.outputs.cache-hit != 'true'
       run: cargo install --locked sccache mdbook mdbook-linkcheck
--- a/.github/workflows/github-cxx-qt-tests.yml
+++ b/.github/workflows/github-cxx-qt-tests.yml
@@ -201,7 +201,10 @@ jobs:
         key: ${{ matrix.os }}_sccache-0.7.6_mdbook-0.4.36_mdbook-linkcheck-0.7.7
     - name: "Build Rust tools"
       if: steps.rust-tools-cache.outputs.cache-hit != 'true'
-      run: cargo install --locked sccache mdbook mdbook-linkcheck
+      # Do not build with storage backends enabled, we only need local
+      run: |
+        cargo install --no-default-features sccache
+        cargo install mdbook mdbook-linkcheck
 
     - name: "Compiler cache"
       # Always cache and save all branches
--- a/.github/workflows/github-cxx-qt-tests.yml
+++ b/.github/workflows/github-cxx-qt-tests.yml
@@ -169,9 +169,10 @@ jobs:
         # Match the deployment target that Qt was built with via vcpkg, otherwise the following error occurs
         # ld: warning: object file (LIB) was built for newer macOS version (12.7) than being linked (12.0)
         MACOSX_DEPLOYMENT_TARGET: 12.7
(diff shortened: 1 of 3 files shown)


.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -190,26 +191,27 @@ jobs:
...
     - name: "Build Rust tools"
       if: steps.rust-tools-cache.outputs.cache-hit != 'true'
       run: cargo install --locked sccache mdbook mdbook-linkcheck
 
     - name: "Compiler cache"
-      uses: actions/cache@v3
+      # Always cache and save all branches
+      # otherwise if we only restore here and only save the main branch
+      # we don't update the cache until something reaches main which
+      # causes an eviction of the cache if the cache isn't accessed for 1 week
```
ahayzen-kdab, line 210: should we schedule a "nightly" run of the CI every few days, this would ensure that things don't break from newer versions of packages (as we have had before) and cause the cache not to expire.

Looks like it'd be adding an `on` block to this file https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

```
on:
  schedule:
    # * is a special character in YAML so you have to quote this string
    - cron:  '30 5,17 * * *'
```

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -190,26 +191,27 @@ jobs:
...
       with:
         path: |
           ${{ matrix.cargo_dir }}/bin/sccache${{ matrix.exe_suffix }}
           ${{ matrix.cargo_dir }}/bin/mdbook${{ matrix.exe_suffix }}
           ${{ matrix.cargo_dir }}/bin/mdbook-linkcheck${{ matrix.exe_suffix }}
-        key: ${{ matrix.os }}_sccache-0.4.1_mdbook-0.4.28_mdbook-linkcheck-0.7.7
+        key: ${{ matrix.os }}_sccache-0.7.6_mdbook-0.4.36_mdbook-linkcheck-0.7.7
     - name: "Build Rust tools"
       if: steps.rust-tools-cache.outputs.cache-hit != 'true'
       run: cargo install --locked sccache mdbook mdbook-linkcheck
```
ahayzen-kdab, line 204: sccache has a load of features that we don't need enabled by default, can we install with only the features we require (eg none of the remote stuff)? As this would cause installation to be much faster as it doesn't require a load of web dependencies ...

```
all = [
  "dist-client",
  "redis",
  "s3",
  "memcached",
  "gcs",
  "azure",
  "gha",
  "webdav",
  "oss",
]
```

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -190,26 +191,30 @@ jobs:
...
           ${{ matrix.cargo_dir }}/bin/mdbook${{ matrix.exe_suffix }}
           ${{ matrix.cargo_dir }}/bin/mdbook-linkcheck${{ matrix.exe_suffix }}
-        key: ${{ matrix.os }}_sccache-0.4.1_mdbook-0.4.28_mdbook-linkcheck-0.7.7
+        key: ${{ matrix.os }}_sccache-0.7.6_mdbook-0.4.36_mdbook-linkcheck-0.7.7
     - name: "Build Rust tools"
       if: steps.rust-tools-cache.outputs.cache-hit != 'true'
-      run: cargo install --locked sccache mdbook mdbook-linkcheck
+      # Do not build with storage backends enabled, we only need local
+      run: |
+        cargo install --no-default-features sccache
```
ahayzen-kdab, line 206: there is a GitHub cache storage backend, wonder if this is better or worse than using local ?

https://github.com/mozilla/sccache/blob/main/docs/GHA.md

ahayzen-kdab: But looks like that would download on the fly, so wonder if that is slower than downloading a compressed local blob then querying that

.github/workflows/github-cxx-qt-tests.yml
LeonMatthesKDAB: If I read the current PR correctly, you already disabled those features :+1:

.github/workflows/github-cxx-qt-tests.yml
LeonMatthesKDAB: I think this is something we can just try out in a different PR.

.github/workflows/github-cxx-qt-tests.yml
LeonMatthesKDAB: Sounds like a good idea.

But that can be a separate PR I think.

LeonMatthesKDAB (approved):

PR #838, opened 2024-02-07
Extract former CXXMappings into TypeNames struct

LeonMatthesKDAB:
This is the first stage of the Naming phase (#835) refactor. All associations between types and their fully qualified names are now stored in one central database, the TypeNames struct.

The final goal for this struct is to include all fully qualified names for each nameable type in the bridge.
Therefore the generator can use this to lookup any references from e.g. arguments.

This is not quite done yet, as this isn't yet used everywhere, but it's a good first step


LeonMatthesKDAB:
Information for the reviewer:
The important part here is in the `parser/mod.rs`, `parser/naming.rs` and `parser/cxxqtdata.rs`.
All of the other files basically just include renaming from CXXMappings -> TypeNames

crates/cxx-qt-gen/src/parser/mappings.rs
```diff
@@ -1,225 +0,0 @@
...
-
-use std::collections::BTreeMap;
-
-use quote::format_ident;
-use syn::{spanned::Spanned, Attribute, Ident, Path, Result};
-
-use crate::syntax::{attribute::attribute_find_path, expr::expr_to_string, path::path_from_idents};
-
-#[derive(Default, Debug)]
-pub struct ParsedCxxMappings {
```
ahayzen-kdab, line 14: renaming / moving this struct could have been done in a separate commit to ease review :-)

crates/cxx-qt-gen/src/parser/naming.rs
```diff
@@ -0,0 +1,507 @@
...
+    /// This is used in the Rust generation to map the bridge type A to ffi::B
+    pub qualified: BTreeMap<Ident, Path>,
+}
+
+impl TypeNames {
+    /// The "Naming" phase.
+    /// Extract all nameable types from the CXX-Qt data and the CXX items.
+    ///
+    /// This allows the generator to fully-qualify all types in the generated code.
+    pub fn from_parsed_data(
```
ahayzen-kdab, line 46: ideally later if we did go the `syn::Parse` route to a `CxxQtMod` this would then take just that as the input.

```rust
TypeNames::from_parsed_data(module: &CxxQtMod)
```

This would then mean not needing to pass in the bridge namespace or module ident or items as everything would be contained in `CxxQtMod`.

Then this struct could iterate over the required blocks to build up the cxx_names, namespaces, qualified names etc.

ahayzen-kdab (approved):
Looks good overall I think

crates/cxx-qt-gen/src/parser/naming.rs
LeonMatthesKDAB: That's the basic idea.
Parse everything first, then extract the names from this parsed info, whatever that looks like.

crates/cxx-qt-gen/src/parser/mappings.rs
LeonMatthesKDAB: Sorry about that :see_no_evil:

PR #839, opened 2024-02-07
Refactor syn_type_to_cpp_type

LeonMatthesKDAB:
This should be a responsibility of the TypeNames struct in the future.

### DIFF
--- a/crates/cxx-qt-gen/src/generator/utils/rust.rs
+++ b/crates/cxx-qt-gen/src/generator/utils/rust.rs
@@ -3,28 +3,9 @@
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
-use std::collections::BTreeMap;
+use syn::{GenericArgument, PathArguments, PathSegment, ReturnType, Type, TypeReference};
 
-use syn::{
-    GenericArgument, Ident, Path, PathArguments, PathSegment, ReturnType, Type, TypeReference,
-};
-
-/// Return a qualified version of the ident that can by used for impl T outside of a CXX bridge
-///
-/// Eg MyObject -> ffi::MyObject
-///
-/// Note that this does not handle CXX types such as UniquePtr -> cxx::UniquePtr.
-/// This is just for resolving impl T {} -> impl module::T {}
-pub(crate) fn syn_ident_cxx_bridge_to_qualified_impl(
-    ident: &syn::Ident,
-    qualified_mappings: &BTreeMap<Ident, Path>,
-) -> syn::Path {
-    if let Some(qualified_path) = qualified_mappings.get(ident) {
-        qualified_path.clone()
-    } else {
-        Path::from(ident.clone())
-    }
-}
+use crate::parser::naming::TypeNames;
 
 /// Return a qualified version of the type that can by used outside of a CXX bridge
 ///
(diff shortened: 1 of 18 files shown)


ahayzen-kdab (commented):
Second commit looks good, but first one needs to land...

LeonMatthesKDAB:
This isn't entirely done yet, but can already be merged so that we have continuous integration.

The current state is already better and shouldn't break anything :)

crates/cxx-qt-gen/src/parser/naming.rs
```diff
@@ -214,6 +214,20 @@ impl TypeNames {
             .cloned()
     }
 
+    /// Return a qualified version of the ident that can by used to refer to the type T outside of a CXX bridge
+    ///
+    /// Eg MyObject -> ffi::MyObject
+    ///
+    /// Note that this does not handle CXX types such as UniquePtr -> cxx::UniquePtr.
+    /// This is just for resolving impl T {} -> impl module::T {}
+    pub fn rust_qualified(&self, ident: &Ident) -> syn::Path {
```
ahayzen-kdab, line 223: this was originally called `syn_ident_cxx_bridge_to_qualified_impl` and was specifically for the impl T type, so I think we should keep the `impl` name in the method otherwise it could easily be confused with `syn_type_cxx_bridge_to_qualified`

crates/cxx-qt-gen/src/generator/utils/rust.rs
```diff
@@ -3,28 +3,9 @@
...
-    GenericArgument, Ident, Path, PathArguments, PathSegment, ReturnType, Type, TypeReference,
-};
-
-/// Return a qualified version of the ident that can by used for impl T outside of a CXX bridge
-///
-/// Eg MyObject -> ffi::MyObject
-///
-/// Note that this does not handle CXX types such as UniquePtr -> cxx::UniquePtr.
-/// This is just for resolving impl T {} -> impl module::T {}
-pub(crate) fn syn_ident_cxx_bridge_to_qualified_impl(
```
ahayzen-kdab, line 18: I'm not sure that removing methods from here is a good idea, the original point of `generator/utils` was that these are helper methods for CXX in general that could be split into a separate crate (see upstream discussion here [#1144](https://github.com/dtolnay/cxx/issues/1144) ).

Instead ideally we would have something like `cxx-qt-gen-utils` with these independent methods, then have `cxx-qt-gen` use these methods in the naming module. For now they should ideally remain at minimum in their own independent modules so that they could be split later if they are useful to others.

This also leads to the qualified_mappings type change of `BTreeMap<Ident, Path>` to `TypeNames` being an issue as it makes these methods dependent on cxx-qt types when they should be independent.

ahayzen-kdab: We also have created the same problem with utils/cpp with #838, so maybe we need to split this out to `cxx-qt-gen-utils` before moving to naming module as it's combining two different disciplines (cxx-qt specific naming and generate cxx type conversions) and is going to lose the nicely isolated code.

To do this have the naming module of cxx-qt-gen use the generic of cxx-qt-gen-utils internally, then it can pass in the BTreeMap or use whichever generic types are in cxx-qt-gen-utils

LeonMatthesKDAB: As discussed in our meeting, the idea here is to make the TypeNames struct independent from CXX-Qt.
It would just include all nameable types inside the bridge, with helpers like these also independent.

The only thing dependent on CXX-Qt then would be populating the TypeNames struct, which can be done by CXX-Qt.
Once it's populated, access is independent.

This may be a good time to extract this into its own crate then, maybe as `cxx-qt-naming`.

crates/cxx-qt-gen/src/parser/naming.rs
LeonMatthesKDAB: As discussed in our meeting, this should be good to stay here.
I'll update the documentation on the function though.

LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/generator/utils/rust.rs
ahayzen-kdab: Assume this will be in a future pull request?

ahayzen-kdab (approved):
Looks good I think, heading in the right direction...

crates/cxx-qt-gen/src/generator/utils/rust.rs
LeonMatthesKDAB: Yes, definitely.
I want to get this in piece-by-piece so the diff doesn't become too big.

PR #837, opened 2024-02-06
WIP: cxx-qt-lib: use cxx-qt-build to build rather than cxx

ahayzen-kdab:
Requires #764

PR #840, opened 2024-02-08
WIP: cxx-qt-lib: use cxx_qt::bridge and #[qsignal]

ahayzen-kdab:
Requires #837


crates/cxx-qt-lib/src/qml/qqmlengine.rs
```diff
@@ -3,8 +3,26 @@
...
 
-#[cxx::bridge]
+#[cxx_qt::bridge(cxx_file_stem = "qqmlengine")]
 mod ffi {
+    unsafe extern "C++Qt" {
+        include!("cxx-qt-lib/qqmlengine.h");
+        type QQmlEngine;
+
+        /// This signal is emitted when the QML loaded by the engine would like to exit from the event loop with the specified return code retCode.
+        #[qsignal]
```
ahayzen-kdab, line 13: this requires `cxx_qt_lib::ConnectionType` but we are `cxx_qt_lib` so it then fails.

This will work after moving anything that `cxx-qt-gen` uses to `cxx-qt` types

ahayzen-kdab: #843 moves those types into `cxx-qt` :tada:

examples/cargo_without_cmake/src/main.rs
```diff
@@ -28,6 +28,13 @@ fn main() {
         engine.load(&QUrl::from("qrc:/qt/qml/com/kdab/cxx_qt/demo/qml/main.qml"));
     }
 
+    if let Some(engine) = engine.as_mut() {
```
ahayzen-kdab, line 31: do we want this in the actual examples?

crates/cxx-qt-lib-headers/include/qml/qqmlapplicationengine.h
```diff
@@ -11,13 +11,17 @@
...
+#include <QtQml/QQmlEngine>
 
 namespace rust {
 namespace cxxqtlib1 {
 
 ::std::unique_ptr<QQmlApplicationEngine>
 qqmlapplicationengineNew();
 
+QQmlEngine&
+qqmlapplicationengineAsQQmlEngine(QQmlApplicationEngine&);
```
LeonMatthesKDAB, line 23: This sounds like a good candidate for another template in common.h:
```cpp
template<typename Base, typename T>
Base & upcast(T& t)
{
   // Maybe add some static_assert that Base is indeed base class of T
    return static_cast<Base&>(t);
}
```

examples/cargo_without_cmake/src/main.rs
LeonMatthesKDAB: Doesn't hurt anything I guess and is nice to show off.

crates/cxx-qt-lib/src/qml/qqmlapplicationengine.rs
```diff
@@ -75,9 +86,17 @@ mod ffi {
...
 }
 
+use crate::QQmlEngine;
+use core::pin::Pin;
+
 pub use ffi::QQmlApplicationEngine;
 
 impl QQmlApplicationEngine {
+    /// Convert the existing [QQmlApplicationEngine] to a [QQmlEngine]
+    pub fn as_qqmlengine<'a>(self: Pin<&'a mut Self>) -> Pin<&'a mut QQmlEngine> {
```
LeonMatthesKDAB, line 96: Could this be a `DerefMut` implementation somehow?

LeonMatthesKDAB: Hm, I guess not, as the DerefMut of `Pin` conflicts with this.

LeonMatthesKDAB (approved):
Looks good for now.
We should investigate adding some kind of `Inherits` trait later that generalizes the concept of `upcasting`.

crates/cxx-qt-lib-headers/include/qml/qqmlapplicationengine.h
ahayzen-kdab: yup, lets add it to #823

PR #842, opened 2024-02-08
WIP: cxx-qt-lib: remove cxx-qt-lib-headers and store in cxx-qt-lib

ahayzen-kdab:
This reverts commit c35772277971559e522b0168518bc1514de9cd2d.
    
But it fails to build on Windows due to a missing dll.
It appears to happen when cxx-qt-lib is a build dependency.

Requires #764


examples/qml_features/rust/Cargo.toml
```diff
@@ -22,6 +22,7 @@ serde_json.workspace = true
 
 [build-dependencies]
 cxx-qt-build.workspace = true
+cxx-qt-lib.workspace = true
```
ahayzen-kdab, line 25: this causes Windows builds to fail with a missing dll :thinking:

ahayzen-kdab: (note it works when using cxx-qt-lib-headers, so is something specific to the build.rs and note it happened when using cxx_build instead of cxx-qt-build too)

crates/cxx-qt-lib/build.rs
```diff
@@ -251,5 +251,52 @@ fn main() {
     });
     println!("cargo:rerun-if-changed=src/assertion_utils.h");
 
-    builder.with_opts(cxx_qt_lib_headers::build_opts()).build();
```
ahayzen-kdab, line 254: this makes `cxx-qt-lib` worse to build as we need to duplicate the header copy phase :-/

ahayzen-kdab:
`cxx-qt-lib-headers` has now gone due to other refactorings.

PR #843, opened 2024-02-08
cxx-qt-gen: move headers to cxx-qt crate

### LINKED ISSUE #844: Investigate moving signals to return a normal type then use a manual guard
Instead of having our QMetaObjectConnection automatically be RAII, instead just return the normal type in cxx-qt-gen.

Then in cxx-qt-lib change QMetaObjectConnection to be QMetaObjectConnectionGuard.

This could allow for cxx-qt-gen to not depend on cxx-qt-lib (helping #319 ).

```rust
my_qobject.on_signal(...).release();
// becomes
my_qobject.on_signal(...);

let connection = my_object.on_signal(...);
connection.release();
//becomes
let guard = cxx_qt_lib::QMetaObjectConnectiongGuard::from(my_qobject.on_signal(...));
guard.release();

{
  let _ = my_object.on_signal(...);
}  // disconnection
// becomes
{
  let _ = cxx_qt_lib::QMetaObjectConnectiongGuard::from(my_qobject.on_signal(...));
}
```

ahayzen-kdab:
Closes #844


crates/cxx-qt/src/lib.rs
```diff
@@ -357,3 +359,36 @@ where
         Self::initialize(self);
     }
 }
+
+/// Write the cxx-qt headers to the specified directory.
+pub fn write_headers(directory: impl AsRef<Path>) {
```
ahayzen-kdab, line 364: would we want this documented?

ahayzen-kdab: do we instead want a `cxx-qt-headers` as this then means we only have to list `cxxqt_connection` once

crates/cxx-qt-build/src/lib.rs
```diff
@@ -456,7 +456,7 @@ impl CxxQtBuilder {
         qtbuild.cargo_link_libraries(&mut self.cc_builder);
 
         // Write cxx-qt-gen, cxx-qt-lib and cxx headers
-        cxx_qt_gen::write_headers(format!("{header_root}/cxx-qt-common"));
+        cxx_qt::write_headers(format!("{header_root}/cxx-qt-common"));
```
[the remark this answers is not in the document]
ahayzen-kdab: We need to move it if we are moving types to `cxx-qt`

crates/cxx-qt/include/cxxqt_locking.h
ahayzen-kdab, line 1: wonder if we should change the namespace of these to `cxxqtcommon` ?

crates/cxx-qt/include/cxxqt_connection.h
```diff
@@ -24,12 +23,15 @@ struct IsRelocatable<::QMetaObject::Connection> : ::std::true_type
 } // namespace rust
 
 namespace rust {
-namespace cxxqtlib1 {
+namespace cxxqtcommon1 {
```
ahayzen-kdab, line 26: should we use `cxxqtcommon` and `cxx-qt-common` (directory) or just `cxxqt` and `cxx-qt` (directory) ?

crates/cxx-qt-lib/src/core/qmetaobjectconnectionguard.rs
```diff
@@ -0,0 +1,53 @@
...
+    fn drop(&mut self) {
+        if let Some(inner) = self.inner.as_ref() {
+            inner.disconnect();
+        }
+    }
+}
+
+impl QMetaObjectConnectionGuard {
+    /// Release the connection without disconnecting
+    pub fn release(mut self) {
```
LeonMatthesKDAB, line 44: would it make sense to have this method return the QMetaObjectConnection?
Then maybe we could get around using an `Option`?

If you wanted to release the guard but still keep the connection handle around anyway, this could be handy.
And if you don't use it it is just deconstructed as it should.

crates/cxx-qt/include/cxxqt_connection.h
LeonMatthesKDAB: Hm, I think just `cxxqt`/`cxx-qt` is a bit nicer.
Doesn't really matter much though.

crates/cxx-qt/include/cxxqt_locking.h
LeonMatthesKDAB: As they move into cxx-qt instead of lib, then yes.

crates/cxx-qt-lib/src/core/qmetaobjectconnectionguard.rs
ahayzen-kdab: Done :-)

crates/cxx-qt/include/cxxqt_connection.h
ahayzen-kdab: Done as `cxxqt1`

crates/cxx-qt-lib/src/core/qmetaobjectconnectionguard.rs
```diff
@@ -0,0 +1,45 @@
+// SPDX-FileCopyrightText: 2023 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+#[cxx::bridge]
+mod ffi {
+    extern "Rust" {
+        #[doc(hidden)]
+        type QMetaObjectConnectionGuard;
```
ahayzen-kdab, line 10: move to cxx-qt

crates/cxx-qt/src/lib.rs
ahayzen-kdab: done

crates/cxx-qt-lib/src/core/qmetaobjectconnectionguard.rs
ahayzen-kdab: done

book/src/bridge/extern_rustqt.md
```diff
@@ -225,17 +225,9 @@ using the signal name `<property>Changed` with no parameters.
 {{#include ../../../examples/qml_features/rust/src/signals.rs:book_signals_connect}}
 ```
 
-Each connection returns a [`QMetaObject::Connection`](https://doc.qt.io/qt-6/qmetaobject-connection.html) which is used to manage the connection.
+Each connection returns a [`QMetaObject::Connection`](https://doc.qt.io/qt-6/qmetaobject-connection.html) which can be used to manage the connection.
 
-Note that the `QMetaObjectConnection` returned by CXX-Qt behaves a bit different from the Qt C++ implementation.
```
ahayzen-kdab, line 230: need to revert this as we want to return guard by default again

ahayzen-kdab:
Windows CI broke on this branch this time with the missing DLL, hoping it was related to add `cxx-qt` as a dependency to `qt_types_standalone` ...

ahayzen-kdab:
> Windows CI broke on this branch this time with the missing DLL, hoping it was related to add `cxx-qt` as a dependency to `qt_types_standalone` ...

Nope still fails, on my CI instance https://github.com/ahayzen-kdab/cxx-qt/actions it can be seen that when the commit chain including rename to cxx-qt-common to cxx-qt is when it started to occur. So will need to diff between https://github.com/ahayzen-kdab/cxx-qt/commit/7c821a1d4405aa8699af532c72f9bfcae3da28c4 and https://github.com/ahayzen-kdab/cxx-qt/commit/54559b774f3b61d919a31f4b28ab5585ce6b34e1

ahayzen-kdab:
The diff is quite small, this is when the guard was moved to cxx-qt and added as a bridge that we might not require. So is the whole issue of the missing DLL on Windows something to do with `extern "Rust"` blocks ? (Note we don't even potentially require this)

ahayzen-kdab:
So yes turns out that the `extern "Rust"` in `cxx-qt` was causing the build failures of a missing DLL on Windows.

LeonMatthesKDAB (approved):
Looks good :)

Great change!

crates/cxx-qt/include/cxxqt_connection.h
```diff
@@ -0,0 +1,40 @@
+// clang-format off
```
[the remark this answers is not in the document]
ahayzen-kdab: Right, lets solve in another pull request..

crates/cxx-qt/include/cxxqt_connection.h
```diff
@@ -0,0 +1,40 @@
...
+
+} // namespace rust
+
+namespace rust {
+namespace cxxqt1 {
+
+using QMetaObjectConnection = ::QMetaObject::Connection;
+
+::QMetaObject::Connection
+qmetaobjectconnectionDefault();
```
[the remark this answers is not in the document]
ahayzen-kdab: Yup that was my comment before, we should move the basic construct / drop etc into cxx-qt rather than lib. See #823 and i'll add a task there

crates/cxx-qt/include/cxxqt_connection.h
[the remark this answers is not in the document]
ahayzen-kdab: #854

PR #846, opened 2024-02-09
We can use Self, as long as the surrounding extern "..."

Montel:
block only has a single type declaration in it!


LeonMatthesKDAB (approved):
I want to be clear that I'm fine with using this feature, but I don't think we should add it to all of our codebase in retrospect, just for the sake of using it.

For new code, feel free to use `&Self`, but I don't see the point of re-doing all of our existing code!

Montel:
Yep my plan was not to replace all in old code, only testing it to be sure that I understood correctly it :)

PR #851, opened 2024-02-16
Extract naming module

LeonMatthesKDAB:
This naming module could later be extracted into another crate that may be shared with other crates that generate code for CXX.


crates/cxx-qt-gen/src/generator/cpp/mod.rs
```diff
@@ -73,6 +73,55 @@ impl GeneratedCppBlocks {
     }
 }
 
+mod utils {
```
ahayzen-kdab, line 76: any reason for this not being in a `cpp/utils.rs` ?

LeonMatthesKDAB: Just thought it's a bit too small to deserve it's own file, but don't mind either way :)
Will move into a file :+1:

LeonMatthesKDAB: Done

ahayzen-kdab (approved):
LGTM

Issue #855, opened 2024-02-21
Follow the cxx build for output file names


ahayzen-kdab:
At the moment we use the file stem or the module ident or the cxx_file_stem depending on the situation in cxx-qt-build.

With CXX if the bridge is in `src/folder/binding.rs` they generate `src/folder/binding.rs.h`.

In this situation should we do the following
| bridge type | proposed path | current path |
|-------------------|-----------------------|--------------------|
| `#[cxx::bridge]` | `src/folder/binding.rs.h` | `binding.cxx.h` |
| `#[cxx_qt::bridge]` | `src/folder/binding.rs.qt.h` | `binding.cxxqt.h` |
| `#[cxx_qt::bridge(cxx_file_stem = "binding"]` | `binding.rs.qt.h` | `binding.cxxqt.h` |

Note that due missing support in macros we cannot find the path in the second case (as seen in #200 ) so ...

- [ ] For now disable having a `cxx_qt::bridge` without a `cxx_file_stem` until span `source_file()` API is stable

LeonMatthesKDAB:
New idea:
CXX probably only needs the `include!` in the build script expansion, where it generates the C++ code.
So could we just emit the `include!` to our generated header in the build script, where we know all the paths?

If that doesn't work, we could also try hashing the tokenstream to get a unique filename that we could reference when expanding the macro.
The build script could then generate symlinks from the hashed file name to the actual path, as the build script knows the paths.

mystchonky:
now that cxx_file_stem is gone, how do you use prevent the linking error because of same module name ?

ahayzen-kdab:
What is the linker error you get ? As now we are following the path similar to CXX, i also wonder what happens if you do the same with CXX ?

mystchonky:
It's a linker error. it cannot find the objects for linking the library. I have cxx-qt based Library which builds fine but when I consume it another project, it fails with linker error.

ahayzen-kdab:
Don't know how much this is possibly related to #1065 which we hope to sort out in the next cycle.

Do you have an example project and the errors or an easy way to reproduce ?

mystchonky:
I do. I will send it along with proper logs after I am done with work.

mystchonky:
I did a clean update+build and it seems to start working automatically. so for now it seems to be resolved. Thanks for help!

PR #856, opened 2024-02-21
cxx-qt-build: consider the folders in rust path

ahayzen-kdab:
Related to #855


crates/cxx-qt-build/src/lib.rs
```diff
@@ -730,6 +751,31 @@ impl CxxQtBuilder {
...
+                    file.parent()
+                        .unwrap()
+                        .file_name()
+                        .unwrap()
+                        .to_string_lossy()
+                        .to_string()
+                })
+                .collect::<HashSet<String>>();
+            if dirs.len() > 1 {
+                panic!("Only one directory is support per QmlModule for rust_files");
```
ahayzen-kdab, line 776: link to Qt bug

crates/cxx-qt-gen/src/parser/mod.rs
```diff
@@ -80,13 +77,16 @@ impl Parser {
...
+                        Meta::NameValue(ref name_value) => {
                             // Parse any namespace in the cxx_qt::bridge macro
                             if name_value.path.is_ident("namespace") {
                                 namespace = Some(expr_to_string(&name_value.value)?);
                                 // Parse any custom file stem
                             } else if name_value.path.is_ident("cxx_file_stem") {
-                                cxx_file_stem = expr_to_string(&name_value.value)?;
+                                return Err(Error::new(
+                                    meta.span(),
+                                    "cxx_file_stem is unsupported: TODO new system message",
```
ahayzen-kdab, line 88: write proper error, report/find Qt bug

crates/cxx-qt-build/src/lib.rs
```diff
@@ -730,6 +751,35 @@ impl CxxQtBuilder {
...
+            // has a QML_ELEMENT the file must be in the same directory as the
+            // CMakeLists and cannot be a relative path to a sub directory.
+            let dirs = qml_module
+                .rust_files
+                .iter()
+                .map(|file| {
+                    file.parent()
+                        .unwrap()
+                        .file_name()
+                        .unwrap()
```
LeonMatthesKDAB, line 770: I think this is incorrect.

If we compare the file_name of the parent, we'll only be comparing the very last directory name.
So e.g. if we have this as rust_files:
```
src/generator/rust/property/mod.rs
src/generator/cpp/property/mod.rs
```
This will find `property` and `property` and discard the discrepancy in the earlier paths.

crates/cxx-qt-build/src/lib.rs
```diff
@@ -730,6 +751,35 @@ impl CxxQtBuilder {
...
+            //
+            // This can also be observed when using qt_add_qml_module, if a class
+            // has a QML_ELEMENT the file must be in the same directory as the
+            // CMakeLists and cannot be a relative path to a sub directory.
+            let dirs = qml_module
+                .rust_files
+                .iter()
+                .map(|file| {
+                    file.parent()
+                        .unwrap()
```
LeonMatthesKDAB, line 768: Prefer actually handling the error by providing a default.

It's theoretically okay if the user inputs e.g. "my_file.rs".
Even though it's probably not a real file, we also shouldn't just panic without explanation.
At least prefer using `expect` with a custom error message.

crates/cxx-qt-build/src/lib.rs
```diff
@@ -738,11 +788,21 @@ impl CxxQtBuilder {
                 self.cc_builder.file(files.plain_cpp);
                 if let (Some(qobject), Some(qobject_header)) = (files.qobject, files.qobject_header)
                 {
+                    // Ensure that the generated QObject header is in the include path
+                    // so that qmltyperegistar can include them later
+                    if let Some(dir) = qobject_header.parent() {
+                        self.cc_builder.include(dir);
```
LeonMatthesKDAB, line 794: Hm, I have two concerns with this:

1.: This should only be done once per QML module, not for every file.

2.: More general: We'll end up adding every qml modules root folder as an include directory.
This sounds like it could lead to conflicts as soon as two QML modules have a file by the same name.
Can we fix this somehow by using a dedicated cc_builder? I believe it's not possible to only add the include path when building the qmltyperegistrar .cpp, right?

crates/cxx-qt-build/src/lib.rs
```diff
@@ -738,11 +788,21 @@ impl CxxQtBuilder {
...
+                    }
+
                     self.cc_builder.file(&qobject);
                     let moc_products = qtbuild.moc(
                         qobject_header,
                         MocArguments::default().uri(qml_module.uri.clone()),
                     );
+                    // Include the moc folder
+                    if let Some(dir) = moc_products.cpp.parent() {
+                        self.cc_builder.include(dir);
```
LeonMatthesKDAB, line 804: See comment about including the qobject_header.parent() path.

crates/cxx-qt-build/src/lib.rs
```diff
@@ -769,6 +829,12 @@ impl CxxQtBuilder {
                 // RCC file.
                 .file(qml_module_registration_files.rcc);
 
+            // Add any include paths the qml module registration needs
+            // this is most likely the moc folder for the plugin
+            if let Some(include_path) = qml_module_registration_files.include_path {
+                self.cc_builder.include(include_path);
```
LeonMatthesKDAB, line 835: Isn't this redundant with the include paths that are added in the loop?

LeonMatthesKDAB (changes requested):
Awesome that we can finally get this to work!

I'm slightly concerned that the current include path setup may lead to conflicts though, especially when adding multiple qml modules.

crates/cxx-qt-build/Cargo.toml
```diff
@@ -24,6 +24,7 @@ codespan-reporting = "0.11"
 version_check = "0.9"
 serde = { version = "1.0", features = ["default", "derive"] }
 serde_json = "1.0"
+syn.workspace = true
```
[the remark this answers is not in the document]
ahayzen-kdab: Aha i might not need that anymore now that i pass the include path as a PathBuf/String before i was building the extern "C++" block with parse_quote in cxx-qt-build and injecting it, so this can likely go...

crates/cxx-qt-build/src/lib.rs
ahayzen-kdab: ah yes it should be the whole of `parent()` not just the filename

crates/cxx-qt-build/src/lib.rs
ahayzen-kdab: i'll try falling back to an empty string if there is no parent so that in theory `"my_file.rs"` could work

crates/cxx-qt-build/src/lib.rs
ahayzen-kdab: I guess the fun is what if the different moc calls ended up in different directories? Maybe i can make a HashSet of all the possible directories (which are likely to be one at the moment but could change if we modified our moc call). Then it could add that HashSet of dirs to the include (avoiding duplicates)

ahayzen-kdab:
Weird it compiles locally for me in both release and debug :scream: and it did work before with CI :thinking:

ahayzen-kdab:
Found it was the cargo build failing, now that is fixed just cargo test now fails :-/

crates/cxx-qt-build/src/lib.rs
```diff
@@ -285,6 +302,15 @@ fn static_lib_name() -> String {
...
 }
 
+// TODO: useful once we have a cc_builder per qml_module
+// fn qml_module_static_lib_name(module_uri: &str) -> String {
+//     format!(
+//         "{}-qml-module-{}-cxxqt-generated",
+//         crate_name(),
+//         module_name_from_uri(module_uri)
+//     )
+// }
```
ahayzen-kdab, line 312: split into separate change

crates/cxx-qt-build/src/lib.rs
```diff
@@ -798,6 +881,24 @@ impl CxxQtBuilder {
...
+            //     // The linker argument order matters!
+            //     // We need to link the object file first, then link the static library.
+            //     // Otherwise, the linker will be unable to find the symbols in the static library file.
+            //     // See also: https://stackoverflow.com/questions/45135/why-does-the-order-in-which-libraries-are-linked-sometimes-cause-errors-in-gcc
+            //     if !dir::is_exporting() {
+            //         println!("cargo::rustc-link-arg=-l{}", &qml_library_name);
+            //     }
+
+            //     cc_builder.compile(&qml_library_name);
+            // }
```
ahayzen-kdab, line 901: split into separate change

crates/cxx-qt-build/src/lib.rs
ahayzen-kdab: Split into #1066 so it's not lost

crates/cxx-qt-build/src/lib.rs
ahayzen-kdab: Split into #1066 so it's not lost

LeonMatthesKDAB (approved):
LGTM :)

PR #858, opened 2024-02-21
Add Name type

LeonMatthesKDAB:
For now this is just used in the `TypeNames` data structure. In future, every nameable object should know its own name, as that includes all data necessary to resolve naming in the generator for every possible use-case.


crates/cxx-qt-gen/src/naming/mod.rs
```diff
@@ -29,24 +29,76 @@ use crate::syntax::{
 
 use crate::parser::{cxxqtdata::ParsedCxxQtData, qenum::ParsedQEnum, qobject::ParsedQObject};
 
+/// This struct contains all names a certain syntax element may have
+///
+/// This includes the rust_name, cxx_name, as well as qualifications like
+/// the rust module and C++ namespace.
+#[derive(Debug)]
+pub struct Name {
```
ahayzen-kdab, line 37: can we split Name into it's own file ? Personally I think we should be heading to almost one struct per file (this is how i see other large projects structured, like CXX tends to do this)

crates/cxx-qt-gen/src/naming/mod.rs
```diff
@@ -202,27 +253,37 @@ impl TypeNames {
...
-        if let Some(namespace) = self.namespace(ident) {
+        if let Some(namespace) = &name.namespace {
             format!("::{namespace}::{cxx_name}")
         } else {
-            cxx_name
+            cxx_name.clone()
+        }
+    }
+
+    pub fn cxx_unqualified(&self, ident: &Ident) -> Result<String> {
```
ahayzen-kdab, line 274: add a doc line saying what this is

crates/cxx-qt-gen/src/naming/mod.rs
```diff
@@ -29,24 +29,76 @@ use crate::syntax::{
...
-    cxx_names: BTreeMap<String, String>,
-    /// Map of the namespace of any types or methods defined in CXX extern blocks
-    ///
-    /// This is used in the C++ generation to map the Rust type name to the C++ name
-    namespaces: BTreeMap<String, String>,
-    /// Mappings for CXX types when used outside the bridge
-    ///
-    /// This is used in the Rust generation to map the bridge type A to ffi::B
-    qualified: BTreeMap<Ident, Path>,
+    names: BTreeMap<Ident, Name>,
```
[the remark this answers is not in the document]
LeonMatthesKDAB: We might want to do some sharing between names in the future (e.g. introduce a single `bridge_namespace`) or something like that for efficiency reasons.

Also, I think we can't add new methods to a type alias, right?

crates/cxx-qt-gen/src/naming/rust.rs
```diff
@@ -257,10 +258,10 @@ mod tests {
     #[test]
     fn test_syn_type_cxx_bridge_to_qualified_mapped() {
         let mut mappings = TypeNames::default();
-        mappings.insert("A", Some(parse_quote! { ffi::B }), None, None);
+        mappings.insert("A", Some(format_ident!("ffi")), None, None);
```
ahayzen-kdab, line 261: hmm I think the point of this test was to check that mapped names still get qualified.

eg

```rust
#[rust_name = "B"]
type A;
```

becomes `ffi::B` not `ffi:A` as it's been renamed. Is this possible to test still with the new mappings ?

crates/cxx-qt-gen/src/naming/mod.rs
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/naming/mod.rs
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/naming/mod.rs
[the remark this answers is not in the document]
ahayzen-kdab: I thought you could ... If you think there are going to be more fields then it's fine. There is also the new type route otherwise `struct TypeNames(BTreeMap<Ident, Name>)`

crates/cxx-qt-gen/src/generator/cpp/signal.rs
```diff
@@ -49,7 +49,7 @@ struct Parameters {
 fn parameter_types_and_values(
     parameters: &[ParsedFunctionParameter],
     type_names: &TypeNames,
-    self_ty: &str,
+    self_ty: &Ident,
```
ahayzen-kdab, line 52: wonder if this should be called something else, if it is an Ident then why is it a `_ty` ? Maybe `self_ident` but not sure what the actual values coming in here are without digging around.

Or is this simply just `qobject_ident` ?

crates/cxx-qt-gen/src/naming/rust.rs
LeonMatthesKDAB: Not on the Rust side of things.

It was also not correct behavior before.

If I have this:
``` rust
mod ffi {
extern "Rust" {
    #[rust_name="B"]
    type A;
}
}
```

The type can only be referred to by `A`, not `B`.

crates/cxx-qt-gen/src/generator/cpp/signal.rs
```diff
@@ -103,8 +102,7 @@ pub fn generate_cpp_signal(
     let free_connect_ident_cpp = idents_helper.connect_name.cpp;
 
     // Retrieve the parameters for the signal
-    let parameters =
-        parameter_types_and_values(&signal.parameters, type_names, &qobject_ident_namespaced)?;
+    let parameters = parameter_types_and_values(&signal.parameters, type_names, &qobject_ident)?;
```
ahayzen-kdab, line 105: why is this not namespaced when it was before?

crates/cxx-qt-gen/src/naming/rust.rs
LeonMatthesKDAB: The correct identifier to use to refer to a type in CXX is always the `rust_name`.

crates/cxx-qt-gen/src/generator/rust/mod.rs
```diff
@@ -70,7 +70,11 @@ impl GeneratedRustBlocks {
         Ok(GeneratedRustBlocks {
             cxx_mod: parser.passthrough_module.clone(),
             cxx_mod_contents,
-            namespace: parser.cxx_qt_data.namespace.clone(),
+            namespace: parser
+                .cxx_qt_data
+                .namespace
+                .clone()
+                .unwrap_or_else(|| "".to_owned()),
```
ahayzen-kdab, line 77: is `unwrap_or_default()` any better here? guess is it obvious enough it's a string

crates/cxx-qt-gen/src/generator/cpp/signal.rs
LeonMatthesKDAB: I honestly have no idea why it was before.
The `parameter_types_and_values` will namespace it using `cxx_qualified`.
It think before it only worked because cxx_qualified does a passthrough.

crates/cxx-qt-gen/src/naming/rust.rs
ahayzen-kdab: ah maybe it was this case

```rust
mod ffi {
  extern "C++" {
    #[rust_name="B"]
    type A;
  }
}
```

which becomes `ffi::B` to the outside and `B` inside the CXX bridge. `A` is never used ?

crates/cxx-qt-gen/src/naming/cpp.rs
```diff
@@ -205,15 +205,18 @@ fn path_argument_to_string(
 
 /// Convert a segment of a path to C++
 fn path_segment_to_string(segment: &PathSegment, type_names: &TypeNames) -> Result<String> {
-    let mut ident = segment.ident.to_string();
+    let ident = segment.ident.to_string();
 
     // If we are a Pin<T> then for C++ it becomes just T
-    let args = match ident.as_str() {
+    let args = match &*segment.ident.to_string() {
```
ahayzen-kdab, line 211: why not `&*ident` ?

crates/cxx-qt-gen/src/naming/name.rs
```diff
@@ -0,0 +1,72 @@
...
+use crate::syntax::{attribute::attribute_find_path, expr::expr_to_string};
+
+/// This struct contains all names a certain syntax element may have
+///
+/// This includes the rust_name, cxx_name, as well as qualifications like
+/// the rust module and C++ namespace.
+#[derive(Debug)]
+pub struct Name {
+    pub(super) rust: Ident,
+    pub(super) cxx: Option<String>,
```
ahayzen-kdab, line 18: does this refer to the C++ name or the CXX bridge name ? (it's the C++ name, but not initially obvious :thinking: )

But i guess `.cxx` is better than `.cpp` as it's come from `cxx_name`. Maybe just add a doc line onto these fields to say where they come from. Eg from the `cxx_name` attribute or the `namespace` attribute etc.

crates/cxx-qt-gen/src/parser/cxxqtdata.rs
```diff
@@ -147,7 +147,11 @@ impl ParsedCxxQtData {
     fn parse_namespaced_qenum(&mut self, item: ItemEnum) -> Result<()> {
         let mut qenum = ParsedQEnum::parse(item)?;
         if qenum.namespace.is_empty() {
-            qenum.namespace = self.namespace.clone();
+            qenum.namespace = self
+                .namespace
+                .as_ref()
+                .cloned()
+                .unwrap_or_else(|| "".to_owned());
```
ahayzen-kdab, line 154: `unwrap_or_default` ?

crates/cxx-qt-gen/src/naming/cpp.rs
LeonMatthesKDAB: Ah, right. I considered removing the ident variable, as I don't think the variable name is quite appropriate.
But I ended up in this weird in-between state :see_no_evil: 

Fixing it.

crates/cxx-qt-gen/src/parser/cxxqtdata.rs
LeonMatthesKDAB: Done.

crates/cxx-qt-gen/src/naming/name.rs
LeonMatthesKDAB: Yeah, the idea is that access would be something like: `qenum.name.cxx`.
Which should be rather obvious that that's the `cxx_name`.

LeonMatthesKDAB: Added additional doc comments

crates/cxx-qt-gen/src/naming/cpp.rs
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/naming/rust.rs
ahayzen-kdab, line 261: Even though it's for tests i wonder if it would be nicer to use a new type for these args, as it's not obvious what None, None means. Eg if there was `struct CxxName(Option<&str>)` etc this could instead read.

```rust
mappings.insert("A", QualifiedPath(Some(format_ident!("ffi"))), CxxName(Some("B")), Namespace(None));
```

I would also ask why is `ident` here a `&str` and not a `&Ident` ?

Something like the `Namespace` newtype could be generally useful elsewhere too ?

LeonMatthesKDAB: With your example the insert call would have to be:
```
mappings.insert("B", Some(format_ident!("ffi")), Some("A"), None);
```
As in this case the `cxx_name` would be "A".

I don't know if that makes sense to test in the `rust`  part of the naming module?

ahayzen-kdab: I just think if we had a test before that was checking something we should still test after, as it's easy to break things when refactoring. Whether the test is in this file or another doesn't matter. But we should ensure that having a cxx_name + the qualified method does the right thing

ahayzen-kdab: Yup, for the tests it doesn't matter much, so not much of a problem here. Just a thought for any non-test methods that we could have better types in the refactor too.

crates/cxx-qt-gen/src/naming/mod.rs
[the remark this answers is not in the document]
LeonMatthesKDAB: Right.
These are all pretty much equivalent options I'd say.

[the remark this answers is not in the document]
LeonMatthesKDAB: For now let's stick with what we have, it's encapsulated so doesn't matter which option we choose

crates/cxx-qt-gen/src/naming/rust.rs
LeonMatthesKDAB: Then in this case I think the right test would be:
```
mappings.insert("A", Some(format_ident!("ffi")), Some("B"), None);
```

As we want to test that a type that is mapped (i.e. has a `cxx_name` that is different from the `rust_name`) uses the right qualified access.

Will change to use that.

LeonMatthesKDAB: Right, we'll have to see which types benefit from more encapsulation when we get around to rewriting the name generation after we split the parsing...

Probably the Namespace. As that has quite a few constraints on it...
Maybe also a CXX identifier or something :thinking: 
As then a namespace would be basically just a `Vec<CxxIdent>`.
That CxxIdent type could make sure that the CXX type is actually a valid identifier in C++.

ahayzen-kdab (approved):
LGTM :-)

Issue #619, opened 2023-07-21
Fully qualify all C++ accesses


LeonMatthesKDAB:
Check what else can be fully-qualified in C++.
Currently we only fully-qualify some accesses, but not all.

ahayzen-kdab:
In #859 we had issues around fully qualifying in Qt 5 so we went with dropping the leading `::`.

A possible workaround to try with Qt 5 is the following.

> hmm what we could possibly do, any type that is registered in a C++ block we could do the required Qt registration in our C++ generation 🤔
>
> Eg if there is a type T = ... then we could generate the missing type declaration with the fully qualified version. Guess the only fun would be if the type is not in fact supposed to be registered?

PR #859, opened 2024-02-21
Fully qualify C++ types without explicit namespace

LeonMatthesKDAB:
Depends on #858

### DIFF
--- a/crates/cxx-qt-gen/src/naming/type_names.rs
+++ b/crates/cxx-qt-gen/src/naming/type_names.rs
@@ -178,6 +178,15 @@ impl TypeNames {
     }
 
     /// For a given rust ident return the CXX name with its namespace
+    ///
+    /// Ideally we'd want this type name to always be **fully** qualified, staring with `::`.
+    /// Unfortunately, this isn't always possible, as the Qt5 meta object system doesn't register
+    /// types with the fully qualified path :(
+    /// E.g. it will recognize `QString`, but not `::QString` from QML.
+    ///
+    /// This needs to be considered in many places (properties, signals, invokables, etc.)
+    /// Therefore, for now we'll use the qualified, but not fully qualified version of `namespace::type`.
+    /// This should work in most cases, but it's not perfect.
     pub fn cxx_qualified(&self, ident: &Ident) -> String {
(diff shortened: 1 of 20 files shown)


ahayzen-kdab:
@LeonMatthesKDAB leave as draft if a pull request depends / includes another, otherwise folks may review the wrong one first :-)

LeonMatthesKDAB:
Hm, I kind of hate how inconsistent this has become...
Whenever we hard-code C++ into our generator we still tend to use fully-qualified paths (which I think is still the best approach).
But for every generated path, we no use paths that are not quite fully qualified...

That's really unfortunate :(
But if it works I guess it must do...

crates/cxx-qt-gen/src/naming/type_names.rs
```diff
@@ -178,6 +178,15 @@ impl TypeNames {
     }
 
     /// For a given rust ident return the CXX name with its namespace
+    ///
+    /// Ideally we'd want this type name to always be **fully** qualified, staring with `::`.
+    /// Unfortunately, this isn't always possible, as the Qt5 meta object system doesn't register
+    /// types with the fully qualified path :(
+    /// E.g. it will recognize `QString`, but not `::QString` from QML.
```
ahayzen-kdab, line 185: hmm what we could possibly do, any type that is registered in a C++ block we could do the required Qt registration in our C++ generation :thinking: 

Eg if there is a `type T = ...` then we could generate the missing type declaration with the fully qualified version. Guess the only fun would be if the type is not in fact supposed to be registered?

LeonMatthesKDAB: Interesting idea...
At least we could do that for all QObject types that we found in `extern "C++Qt"` and `extern "RustQt"`.
Though as you said, unsure if there's ever a point where it **shouldn't** be registered :thinking: 

Maybe let's add this as a note into #619 and try this in another PR.

ahayzen-kdab: Commented :-)

ahayzen-kdab (approved):
Lets go with this to move forward and hope to tidy up the Qt 5 mess later.

PR #862, opened 2024-02-23
Add QBrush support

Montel:


crates/cxx-qt-lib/src/core/qt.rs
```diff
@@ -134,6 +134,29 @@ mod ffi {
         IntersectClip,
     }
 
+    #[repr(i32)]
+    enum BrushStyle {
```
ahayzen-kdab, line 138: doc line could be good

crates/cxx-qt-lib/src/gui/qbrush.rs
```diff
@@ -0,0 +1,102 @@
...
+        #[doc(hidden)]
+        #[rust_name = "qbrush_clone"]
+        fn construct(brush: &QBrush) -> QBrush;
+    }
+}
+
+#[repr(C)]
+pub struct QBrush {
+    #[cfg(qt_version_major = "5")]
+    _cspec: MaybeUninit<[i32; 2]>,
```
ahayzen-kdab, line 72: why is this a different size in Rust but in the C++ assertions it only does the following and passes? This suggests that it is just `MaybeUninit<usize>` ? Also note that isn't `size_t` -> `usize` in Rust ?

```cpp
assert_alignment_and_size(QBrush,
                          alignof(::std::size_t),
                          sizeof(::std::size_t));
```

crates/cxx-qt-lib/src/gui/qbrush.rs
```diff
@@ -0,0 +1,102 @@
...
+        fn drop(brush: &mut QBrush);
+
+        #[doc(hidden)]
+        #[rust_name = "qbrush_clone"]
+        fn construct(brush: &QBrush) -> QBrush;
+    }
+}
+
+#[repr(C)]
+pub struct QBrush {
```
ahayzen-kdab, line 70: would be good to add this type (and any other new ones) to `qt_types_standalone` tests with the usual create, clone, drop stuff as this then runs through `valgrind` which would potentially pick up the below issue.

Montel: I started to create patch for adding test for news one.

crates/cxx-qt-lib/src/gui/qbrush.rs
```diff
@@ -0,0 +1,130 @@
...
+
+        #[doc(hidden)]
+        #[rust_name = "qbrush_eq"]
+        fn operatorEq(a: &QBrush, b: &QBrush) -> bool;
+    }
+}
+
+#[repr(C)]
+pub struct QBrush {
+    _cspec: MaybeUninit<usize>,
```
ahayzen-kdab, line 82: valgrind isn't happy about something, maybe something is still incorrect?

Montel: yep :) I try to make it working on local here.

Montel: No idea where is the problem...
we have "'using DataPtr = std::unique_ptr<QBrushData, QBrushDataPointerDeleter>;"

which is a _cspec: MaybeUninit<usize>,

but valgrind found a memory leak...

PR #863, opened 2024-02-23
CI: Always update the compiler cache

LeonMatthesKDAB:
Previously, it would skip updating, as the key is always the same. That meant the cache got progressively older.

### DIFF
--- a/.github/workflows/github-cxx-qt-tests.yml
+++ b/.github/workflows/github-cxx-qt-tests.yml
@@ -206,12 +206,24 @@ jobs:
         cargo install --no-default-features sccache
         cargo install mdbook mdbook-linkcheck
 
-    - name: "Compiler cache"
-      # Always cache and save all branches
-      # otherwise if we only restore here and only save the main branch
-      # we don't update the cache until something reaches main which
-      # causes an eviction of the cache if the cache isn't accessed for 1 week
-      uses: actions/cache@v4
+    # We want our compiler cache to always update to the newest state.
+    # The best way for us to achieve this is to **always** update the cache after every landed commit.
+    # That way it will closely follow our development.
+    # And if a PR diverges a lot with its cache that's not a big deal, as it will be merged eventually.
+    #
+    # This is a workaround for the fact that GH doesn't support updating existing caches.
+    # See: https://github.com/azu/github-actions-overwrite-cache-example
+    #
+    # Ideally we'd like to use this:
+    # - name: "Compiler cache"
+    #   uses: actions/cache@v4
+    #   with:
+    #     update: true <------- THIS DOESN'T EXIST YET
+    #     path: ${{ matrix.compiler_cache_path }}
+    #     key: ${{ matrix.name }}_compiler_cache
+    - name: "Restore Compiler Cache"
+      id: compiler-cache-restore
+      uses: actions/cache/restore@v3
       with:
         path: ${{ matrix.compiler_cache_path }}
         key: ${{ matrix.name }}_compiler_cache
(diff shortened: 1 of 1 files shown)


.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -307,6 +322,23 @@ jobs:
...
 
+
+    # This is a workaround for the fact that GH doesn't support updating existing caches.
+    # See: https://github.com/azu/github-actions-overwrite-cache-example
+    - name: "Delete previous compiler cache"
+      if: ${{ steps.compiler-cache-restore.outputs.cache-hit }}
+      continue-on-error: true
+      run: |
+        gh extension install actions/gh-actions-cache
+        gh actions-cache delete "${{ matrix.name }}_compiler_cache" --confirm
```
ahayzen-kdab, line 333:
```console

Run gh extension install actions/gh-actions-cache
Error: failed to run git: fatal: not a git repository (or any of the parent directories): .git
. error: exit status 128
Usage:
  gh-actions-cache delete <key> [flags]

Flags:
  -B, --branch string   Filter by branch
      --confirm         Delete the cache without asking user for confirmation.
  -h, --help            help for delete
  -R, --repo string     Select another repository for finding actions cache.

Error: Process completed with exit code 1.
```

LeonMatthesKDAB: Noticed that as well...
I'm a bit confused why that happens...
Maybe need to set the working_dir or something :thinking:

LeonMatthesKDAB:
```
  gh actions-cache delete "Ubuntu 22.04 (gcc) Qt5_compiler_cache" --confirm
  shell: /usr/bin/bash -e {0}
  env:
    MACOSX_DEPLOYMENT_TARGET: 12.7
    SCCACHE_CACHE_SIZE: 600M
    GH_TOKEN: ***
  
Error: Resource not accessible by integration
Error: Process completed with exit code 1.
```

Hm, this may be because it's a forked repository :thinking: 
As I'm adding the `permissions: write`...
We may have to test this on a branch within the main repo.
I'll try merging it into a branch there.

ahayzen-kdab: Assume it's not related to the custom path we use from...
```
    # On Windows, the D drive that the workspace is on by default runs out of space when
    # vcpkg builds Qt, so the repo needs to be cloned to the C drive. GitHub's checkout
    # action does not support cloning to a path outside GITHUB_WORKSPACE:
    # https://github.com/actions/checkout/issues/197
    - name: "Clone Git repository"
      uses: LebedevRI/checkout@issue197
      with:
        path: ${{ matrix.workspace }}
```

ahayzen-kdab: And on linux we use `workspace: /home/runner/cxx-qt`

ahayzen-kdab: Ah I see you solved that and now it has permission issues

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -41,6 +41,9 @@ jobs:
           globs: '**/*.md'
 
   build:
+    permissions:
```
ahayzen-kdab, line 44: vcpkg still seems to be rebuilding, wonder if it's related to this part or something

LeonMatthesKDAB: Yes, this seems to be the case, I've set permissions back to default, which is working again :)

ahayzen-kdab (approved):
LGTM :-)

LeonMatthesKDAB:
Before:
```
Run sccache --show-stats
Compile requests                   2308
Compile requests executed          1959
Cache hits                           47 <------------- YIKES!
Cache hits (C/C++)                   47
Cache misses                       1853
Cache misses (C/C++)               1438
Cache misses (Rust)                 415
```
After:
```
Run sccache --show-stats
Compile requests                   2308
Compile requests executed          1959
Cache hits                         1232 <-------------- MUCH BETTER!
Cache hits (C/C++)                  826
Cache hits (Rust)                   406
Cache misses                        668
Cache misses (C/C++)                659
```
:partying_face:

PR #866, opened 2024-02-26
Export qfont enum

Montel:


crates/cxx-qt-lib/src/gui/qfont.rs
```diff
@@ -265,6 +265,11 @@ mod ffi {
     }
 }
 
+pub use ffi::QFontCapitalization;
```
ahayzen-kdab, line 268: why not `pub use ffi::{QFontCapitalization, QFontHintingPreference, QFontSpacingType, QFontStyle};` ?

Montel: Yep why not :)
I will change it.

ahayzen-kdab (approved):
LGTM, slightly surprised that the compiler doesn't pick up that we are creating methods on a type with arguments that aren't accessible :thinking:

Issue #562, opened 2023-05-29
Traits for CxxQtType/QtObject


LeonMatthesKDAB:
We could move some of our "magic" methods into traits, which would make it a lot more obvious where they are coming from, and would allow us to e.g. "inherit" methods from QObject, reason about types, etc.

``` rust
// cxx-qt crate - may also need to go in cxx-qt-lib
mod ffi {
    extern "C++" {
        type QObject;

        fn children(self: &QObject) -> QList<*mut QObject>;
        fn objectName(self: &QObject) -> QString;
    }
}

// Note: QtObject vs QObject
trait QtObject {
    fn as_qobject(&self) -> &cxx_qt_lib::QObject;
    fn as_qobject_mut(&mut self) -> Pin<&mut cxx_qt_lib::QObject>;

    fn children(&self) -> QList<*mut QObject> {
        as_qobject().children()
    }

    fn objectName(&self) -> Pin<&mut cxx_qt_lib::QObject> {
        as_qobject().objectName()
    }

    // ... other methods wrapped as needed.
}

trait CxxQtType {
    type Inner;

    fn rust(&self) -> &Self::Inner;
    fn rust_mut(self: Pin<&mut self>) -> &mut Self::Inner;
}
```

Then in C++, we could generate a generic cast to QObject, similar to how we do constructors.
``` cpp
template<typename T>
QObject *as_qobject(T* t) {
    return static_cast<QObject*>(t);
}
```

ahayzen-kdab:
I also wonder with this if we have an existing C++ QObject if we should allow for tagging it to get the trait implemented on it ?

Eg if we have

```rust
unsafe extern "C++Qt" {
   type MyQObject;
}
```

This won't have the `QtObject` trait implemented on it, but instead we could use the same syntax as `RustQt` to to do this

Eg

```rust
unsafe extern "C++Qt" {
   #[qobject] 
   type MyQObject;
}
```

LeonMatthesKDAB:
Hm, shouldn't all classes in `"C++Qt"` be QObjects?
In that case we should just always implement the trait.

Though that brings up the question of whether all types within C++Qt are actually QObjects :thinking: 
Could they also be QGadgets? And would there be any use for it?

But maybe we should start requiring `#[qobject]` on every type in "C++Qt" to make sure the behavior is the same in C++Qt as in RustQt.

LeonMatthesKDAB:
Also another suggestion regarding naming, we could use `AsX` here.
[...]

PR #869, opened 2024-02-28
Add qapplication in add cxx qt lib extras

Montel:


crates/cxx-qt-lib-extras/src/gui/qapplication.rs
```diff
@@ -0,0 +1,200 @@
...
+impl QApplication {
+    /// Prepends path to the beginning of the library path list,
+    /// ensuring that it is searched for libraries first.
+    /// If path is empty or already in the path list, the path list is not changed.
+    pub fn add_library_path(self: Pin<&mut Self>, path: &QString) {
+        ffi::qapplication_add_library_path(self, path);
+    }
+
+    /// The name of this application
+    pub fn application_name(&self) -> QString {
```
ahayzen-kdab, line 105: don't some of these come from `QCoreApplication` or `QGuiApplication`.

It would be good if we figured out a plan with @LeonMatthesKDAB around how to handle inheritance, we were considering if we just implement From/Into on these Qt classes. Eg QStringList -> QList<QString> too so that we don't have to implement the methods again for all types.

ahayzen-kdab: Up and downcast C++ helpers was in #823 not sure we have an issue to track From/Into or a trait for Deref or something yet.

Montel: indeed allow to handle inheritance will be good. 
I don't know how we can do at the moment.
but indeed better !

ahayzen-kdab: Right, this needs to be discussed, we have a few ideas :-) Basically have some kind of trait or use From/Into on the Rust side to call the cast on the C++ side. Then it'll return the other type. (and use TryFrom/TryInto where it can be null etc). But lets figure out a proper plan with @LeonMatthesKDAB :-)

crates/cxx-qt-lib-extras/src/gui/qapplication.rs
```diff
@@ -0,0 +1,200 @@
...
+        type QString = crate::QString;
+        include!("cxx-qt-lib/qstringlist.h");
+        type QStringList = crate::QStringList;
+        include!("cxx-qt-lib/qvector.h");
+        type QVector_QByteArray = crate::QVector<QByteArray>;
+        include!("cxx-qt-lib/qfont.h");
+        type QFont = crate::QFont;
+
+        include!("cxx-qt-lib-extras/qapplication.h");
+        type QApplication;
```
ahayzen-kdab, line 24: So in #622 i needed to create a `QApplication` to do a widgets example. I also made a `QPushButton`.

Wonder if we add `QPushButton` to `cxx-qt-lib-extras` and then have an `examples/` folder in `cxx-qt-lib-extras` with the widgets example ? This could also then be a simple cargo-only thing to test building and CI ?

Montel: For me it will be a good idea to add it + test

Montel:
on windows:

unresolved external symbol "__declspec(dllimport) public: static int __cdecl QApplication::exec(void)" (__imp_?exec@QApplication@@SAHXZ) referenced in function "int __cdecl rust::cxxqtlib1::qapplicationExec<class QApplication>(class QApplication &)" (??

...

crates/cxx-qt-lib-extras-headers/include/gui/qapplication.h
```diff
@@ -0,0 +1,29 @@
...
+// clang-format on
+// SPDX-FileContributor: Laurent Montel <laurent.montel@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+#pragma once
+
+#include <memory>
+
+#include <QtGui/QFont>
+#include <QtWidgets/QApplication>
```
ahayzen-kdab, line 12: wonder if you are simply missing the QtWidgets module when building ?

In the `cxx-qt-lib-extras-headers/lib.rs` can you try changing the last `opts` to the following so that the Gui and Widgets module are imported ?  ...

```rust
opts.qt_module("Gui").qt_module("Widgets")
```

ahayzen-kdab: Seems to have worked :-)

crates/cxx-qt-lib-extras/src/gui/qapplication.rs
ahayzen-kdab: Lets investigate this in a later change as part of #562

ahayzen-kdab (approved):
LGTM :-)

PR #870, opened 2024-02-28
Try increasing CI cache size.

LeonMatthesKDAB:
Now that https://github.com/KDAB/cxx-qt/pull/863 has been merged, we only ever have 9 caches in use:
- 6 compiler caches (macOS, Windows, Linux * Qt5 & Qt6)
- 3 tools caches (macOS, Windows, Linux)

This should allow for ~1G in size per cache without issue. So let's use that.

### DIFF
--- a/.github/workflows/github-cxx-qt-tests.yml
+++ b/.github/workflows/github-cxx-qt-tests.yml
@@ -170,10 +170,10 @@ jobs:
         # ld: warning: object file (LIB) was built for newer macOS version (12.7) than being linked (12.0)
         MACOSX_DEPLOYMENT_TARGET: 12.7
         # sccache is around 180-300M in size for a normal build
-        # With GitHub caches we have a 10 GB limit / 6 conditions = 1666 MB
+        # With GitHub caches we have a 10 GB limit / 6 compiler caches = 1666 MB
         # Allow a larger cache size so that code in branches can be cached
         # but still leave room for the tools cache
-        SCCACHE_CACHE_SIZE: 600M
+        SCCACHE_CACHE_SIZE: 1000M
 
     steps:
     # On Windows, the D drive that the workspace is on by default runs out of space when


.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -170,10 +170,10 @@ jobs:
         # ld: warning: object file (LIB) was built for newer macOS version (12.7) than being linked (12.0)
         MACOSX_DEPLOYMENT_TARGET: 12.7
         # sccache is around 180-300M in size for a normal build
-        # With GitHub caches we have a 10 GB limit / 6 conditions = 1666 MB
+        # With GitHub caches we have a 10 GB limit / 6 compiler caches = 1666 MB
         # Allow a larger cache size so that code in branches can be cached
         # but still leave room for the tools cache
-        SCCACHE_CACHE_SIZE: 600M
+        SCCACHE_CACHE_SIZE: 1000M
```
ahayzen-kdab, line 176: why does the cache size need to be increased? If a blank cache is around 250-350MB ? Won't this just be adding things to the cache that aren't used and slow down the download + upload?

LeonMatthesKDAB: The thing is that I don't know if that size (~300MB) is still at all accurate :thinking: 
And unfortunately I don't thinks sccache tells us how much of the cache was used in the last build.

Also it may not be accurate on all platforms, so I'd just like to use the space we have to try and see if that helps anything.

LeonMatthesKDAB: Basically I'm wondering how it can be that the CI build of this PR can still have cache misses, even though I didn't change any code.
So maybe the cache is too small after all...

LeonMatthesKDAB: See: https://github.com/KDAB/cxx-qt/actions/runs/8079714667/job/22074722538

Cache misses: 411

LeonMatthesKDAB: Hm, I guess the cache size is still ~300MB, even on the slower Windows builds :see_no_evil:

LeonMatthesKDAB:
Checking the output, the build size is apparently still ~300MB.
So there's no point to this

PR #872, opened 2024-02-28
Fix CI conditional cache deletion.

LeonMatthesKDAB:
The `&&` has to be *within* the curly braces


LeonMatthesKDAB:
Let's check the CI output first to see that it actually works as expected.

ahayzen-kdab (approved):
LGTM :-)

PR #875, opened 2024-02-29
Add QPainterCompositionMode enum + method

Montel:


crates/cxx-qt-lib/src/gui/qpainter.rs
```diff
@@ -189,6 +289,10 @@ mod ffi {
         #[rust_name = "set_clip_rect"]
         fn setClipRect(self: Pin<&mut QPainter>, rectangle: &QRect, operation: ClipOperation);
 
+        /// Sets the composition mode to the given mode.
```
ahayzen-kdab, line 292: wonder if this warning should be somewhere?

```
Warning: Only a [QPainter](https://doc.qt.io/qt-6/qpainter.html) operating on a [QImage](https://doc.qt.io/qt-6/qimage.html) fully supports all composition modes. The RasterOp modes are supported for X11 as described in [compositionMode](https://doc.qt.io/qt-6/qpainter.html#compositionMode)().
```

and the https://doc.qt.io/qt-6/qpainter.html#CompositionMode-enum has a lot more description too

crates/cxx-qt-lib-headers/include/gui/qpainter.h
```diff
@@ -11,8 +11,13 @@
 
 #include <QtGui/QPainter>
 
+#include "rust/cxx.h"
+
+#include <cstdint>
```
ahayzen-kdab, line 16: why is cstdint and rust/cxx.h required here? I thought maybe for the shared enum but `qt.h` doesn't have these and those work ?

Montel: I think it's a copy/paste error.

ahayzen-kdab:
@Montel rebase this one after the other one lands as it needs manual fixing.

PR #888, opened 2024-03-12
Add toString() in QFont

Montel:


LeonMatthesKDAB:
From reading the Qt documentation, it seems the result looks a lot like it would fit well with the `Debug` trait.

Could you use a different name for the `to_string` function (e.g. `to_qstring`) and then `impl Debug` based on that?

`to_string` sounds a lot like it belongs to https://doc.rust-lang.org/std/string/trait.ToString.html , which would return a `String` instance, not a `QString`.
You could also add a `ToString` implementation based on `Debug`, though I'm unsure whether this is wise, as usually, the output of `Display` is used for this.

Montel:
Better I implement std::fmt::Display

ahayzen-kdab (approved):
LGTM :-)

PR #889, opened 2024-03-12
Add QFontStyleStrategy support

Montel:


crates/cxx-qt-lib/src/gui/qfont.rs
```diff
@@ -176,6 +217,10 @@ mod ffi {
         #[rust_name = "set_fixed_pitch"]
         fn setFixedPitch(self: &mut QFont, enable: bool);
 
+        /// Sets the style strategy for the font to s.
+        #[rust_name = "set_style_strategy"]
+        fn setStyleStrategy(self: &mut QFont, s: QFontStyleStrategy);
```
LeonMatthesKDAB, line 222: Can you please spell out the `s` -> `strategy` variable.

Just because Qt abbreviates this doesn't mean we have to :)

crates/cxx-qt-lib/src/gui/qfont.rs
```diff
@@ -73,6 +73,47 @@ mod ffi {
...
+        /// prefers device fonts.
+        PreferDevice = 0x0004,
+        /// prefers outline fonts (as opposed to bitmap fonts).
+        PreferOutline = 0x0008,
+        /// forces the use of outline fonts.
+        ForceOutline = 0x0010,
+        /// don't antialias the fonts.
+        PreferMatch = 0x0020,
+        /// avoid subpixel antialiasing on the fonts if possible
+        PreferQuality = 0x0040,
```
LeonMatthesKDAB, line 95: The Qt docs for PreferMatch and PreferQuality suggest that these are flags that may be `OR`-ed with the other values of this enum.

With this wrapping, they are only available themselves, not as an `OR`-ed combination of the other flags.

Are they still valid StyleStrategy values without `OR`ing them together?

LeonMatthesKDAB (commented):
Mostly looks good.

I don't know if `PreferQuality` and `PreferMatch` are valid values by themselves though.
They may well be, I'm just unsure. Please check.

crates/cxx-qt-lib/src/gui/qfont.rs
Montel: StyleStrategy is still valid without them.
I will comment them until we will support "OR"

LeonMatthesKDAB: Please use `//`, not `///` for commenting-out the PreferMatch, and PreferQuality.
Otherwise they end up as doc comments on PreferAntialias :sweat_smile:

Montel: oops :)

LeonMatthesKDAB (approved):

PR #891, opened 2024-03-13
TypeNames: Error on unknown type

LeonMatthesKDAB:


crates/cxx-qt-gen/src/naming/cpp.rs
```diff
@@ -242,7 +242,10 @@ fn path_segment_to_string(segment: &PathSegment, type_names: &TypeNames) -> Resu
             ));
         }
     } else {
-        type_names.cxx_qualified(&segment.ident)
+        if let Some(ty) = possible_built_in(&segment.ident.to_string()) {
```
ahayzen-kdab, line 245: which case is this hitting?

crates/cxx-qt-gen/src/naming/name.rs
```diff
@@ -88,4 +93,8 @@ impl Name {
             module: module.clone(),
         })
     }
+
+    pub(super) fn cxx_unqualified(&self) -> String {
```
ahayzen-kdab, line 97: doc line what this is and used for would be useful

crates/cxx-qt-gen/src/naming/type_names.rs
```diff
@@ -244,6 +246,12 @@ impl TypeNames {
         module_ident: &Ident,
     ) -> Result<()> {
         let name = Name::from_ident_and_attrs(ident, attrs, parent_namespace, module_ident)?;
+        if self.names.contains_key(&name.rust) {
```
ahayzen-kdab, line 249: instead use the result of the `insert` to determine if the key already existed, as this avoids a double lookup and as we are going to fail it doesn't matter that the value is overwritten

```
If the map did not have this key present, None is returned.

If the map did have this key present, the value is updated, and the old value is returned. The key is not updated, though; this matters for types that can be == without being identical. See the [module-level documentation](https://doc.rust-lang.org/stable/std/collections/index.html#insert-and-complex-keys) for more.
```

LeonMatthesKDAB: Hm, I did want to avoid overwriting in that case :thinking: 
It probably doesn't matter the way we use it, but it still feels a bit wrong...

Anyway, I can try using the `entry` function as well to avoid the double lookup.

ahayzen-kdab: Yup in that case use the entry API to query if something exists and if it doesn't then insert in-place

LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/naming/name.rs
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/naming/cpp.rs
LeonMatthesKDAB: Done

LeonMatthesKDAB: Also refactored the function a bit, I think it's a bit cleaner and more functional now :)

ahayzen-kdab (approved):
LGTM :-)

PR #898, opened 2024-03-15
Add QFontMetrics support

Montel:


crates/cxx-qt-lib/src/gui/qfontmetrics.rs
```diff
@@ -0,0 +1,129 @@
...
+        fn drop(fontmetics: &mut QFontMetrics);
+
+        #[doc(hidden)]
+        #[rust_name = "qfontmetrics_clone"]
+        fn construct(font: &QFontMetrics) -> QFontMetrics;
+    }
+}
+
+#[repr(C)]
+pub struct QFontMetrics {
```
ahayzen-kdab, line 107: a test in the standalones would be good too :-)

Montel: I will add it.

tests/qt_types_standalone/cpp/qfontmetrics.h
```diff
@@ -0,0 +1,34 @@
...
+#include <QtTest/QTest>
+
+#include "cxx-qt-gen/qfontmetrics.cxx.h"
+
+class QFontMetricsTest : public QObject
+{
+  Q_OBJECT
+
+private Q_SLOTS:
+  void construct()
```
ahayzen-kdab, line 20: clang-format is complaining around here

ahayzen-kdab: Could you use the From to construct in Rust here? eg

- Create a QFont (either C++ or Rust)
- Construct a QFontMetric from the QFont in Rust
- Then check the values are correct in C++

crates/cxx-qt-lib/src/gui/qfontmetrics.rs
```diff
@@ -0,0 +1,146 @@
...
+
+        #[doc(hidden)]
+        #[rust_name = "qfontmetrics_init_from_qfont"]
+        fn construct(font: &QFont) -> QFontMetrics;
+    }
+}
+
+#[repr(C)]
+pub struct QFontMetrics {
+    _cspec: MaybeUninit<i32>,
```
ahayzen-kdab, line 119: I think it crashes because `MaybeUninit<i32>` is not of the size `std::size_t` (which is what you are asserting the size to be in C++.

This should instead be `MaybeUninit<usize>` for the size of the memory to be correct.

tests/qt_types_standalone/cpp/qfontmetrics.h
```diff
@@ -0,0 +1,59 @@
...
+
+    // QFontDatabase needs a QGuiApplication first
+    QGuiApplication app(argc, args.data());
+    app.setApplicationName(QStringLiteral("kdab"));
+
+    QFont f;
+    const int pointSize = 40;
+    f.setPointSize(pointSize);
+    const auto m = constructor_qfontmetrics(f);
+    QCOMPARE(m.ascent(), 57);
```
Montel, line 36: value depend against font size etc...
=> we can't test it...

No idea what testing from constructor...

PR #899, opened 2024-03-15
TypeNames: Error on unknown rust_qualified type

LeonMatthesKDAB:


crates/cxx-qt-gen/src/naming/cpp.rs
```diff
@@ -242,45 +235,37 @@ fn path_segment_to_string(segment: &PathSegment, type_names: &TypeNames) -> Resu
...
-        "isize" => Some("::rust::isize"),
-        "f32" => Some("float"),
-        "f64" => Some("double"),
-        "CxxString" => Some("::std::string"),
-        "String" => Some("::rust::String"),
-        // TODO: handle pointer
-        _others => None,
-    }
-}
+// /// Convert any built in types to known C++ equivalents
```
ahayzen-kdab, line 242: what's happening here ? commented code should be either used or removed...

crates/cxx-qt-gen/src/naming/rust.rs
```diff
@@ -3,112 +3,103 @@
...
+
+    // Convert the first element if it matches
+    if let Some(segment) = ty_path.path.segments.first() {
+        let qualified_prefix = match segment.ident.to_string().as_str() {
+            // Note we need to fully qualify any types that CXX supports that aren't
+            // - primitive types https://doc.rust-lang.org/stable/std/primitive/index.html
+            // - prelude types https://doc.rust-lang.org/stable/std/prelude/index.html
+            //
+            // We could also fully qualify types primitive and prelude types for full macro hygiene.
+            "CxxString" | "CxxVector" | "SharedPtr" | "UniquePtr" | "WeakPtr" => Some(vec!["cxx"]),
```
ahayzen-kdab, line 42: this is stuff that would be good in the separate lib

ahayzen-kdab: but is that what naming has / is becoming what did we decide before ?

I thought we were thinking of having naming -> generic CXX helper lib

crates/cxx-qt-gen/src/naming/type_names.rs
```diff
@@ -21,9 +23,88 @@ use crate::parser::{cxxqtdata::ParsedCxxQtData, qenum::ParsedQEnum, qobject::Par
...
+            rust: format_ident!("{}", rust_name),
+            cxx: Some(cxx_name.to_owned()),
+            module: None,
+            namespace: Some(namespace.to_owned()),
+        };
+
+        // By default, CXX supports a set of types that don't need to be declared:
+        // See: https://github.com/dtolnay/cxx/blob/master/syntax/atom.rs
+        // and: https://github.com/dtolnay/cxx/blob/a6e1cd1e8d9d6df20e88e7443963dc4c5c8c4875/gen/src/write.rs#L1311
+        let cxx_types = [
```
ahayzen-kdab, line 49: this stuff would be better in the separate lib?

As these were helpers that other CXX tools could use for going between CXX Rust and C++ types.

crates/cxx-qt-gen/src/naming/rust.rs
LeonMatthesKDAB: If I recall correctly the idea is to eventually move the naming module into its own library.
So that's alright then?

crates/cxx-qt-gen/src/naming/type_names.rs
LeonMatthesKDAB: I think the idea was to extract the whole naming module into its own library if there's interest in using it.
So I think that's alright?

crates/cxx-qt-gen/src/generator/rust/constructor.rs
```diff
@@ -461,16 +461,19 @@ mod tests {
     }
 
     fn mock_namespace() -> NamespaceName {
-        NamespaceName::from_pair_str("ffi", &format_ident!("MyObject"))
+        NamespaceName::from_pair_str("qobject", &format_ident!("MyObject"))
```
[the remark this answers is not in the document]
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/naming/rust.rs
ahayzen-kdab: If we think that the naming module would not be dependant on CXX-Qt crate types, then i guess yup :thinking:

crates/cxx-qt-gen/src/naming/rust.rs
```diff
@@ -138,131 +129,52 @@ pub(crate) fn syn_type_is_cxx_bridge_unsafe(ty: &syn::Type) -> bool {
...
+        { [UniquePtr<T>; 1] } => { [cxx::UniquePtr<T>; 1] },
+        { fn(UniquePtr<T>) -> SharedPtr<T> } => { fn(cxx::UniquePtr<T>) -> cxx::SharedPtr<T> },
+        { Pin<UniquePtr<T>> } =>  { core::pin::Pin<cxx::UniquePtr<T>> },
+        { *mut UniquePtr<T> } => { *mut cxx::UniquePtr<T> },
+        { &UniquePtr<*mut T> } => { &cxx::UniquePtr<*mut T> },
+        { &[UniquePtr<T>] } => { &[cxx::UniquePtr<T>] },
+        { &mut [UniquePtr<T>] } => { &mut[cxx::UniquePtr<T>] },
+        { (UniquePtr<T>, ) } => { (cxx::UniquePtr<T>, ) },
+        // Mapped type
+        { MyObject } => { qobject::MyObject }
```
ahayzen-kdab, line 176: do we need to test any rust_name mappings ? eg where the cxx bridge type name is different to the qualified ?

ahayzen-kdab: OK after conversation turns out we figured out a plan here and that the rust name is always used.

ahayzen-kdab (approved):
Lets go with this, LGTM :-)

Issue #810, opened 2024-01-17
For overloads in cxx-qt-lib use impl AsRef


ahayzen-kdab:
Where we need to have an overload in cxx-qt-lib, eg `QPainter::setPen` can take QColor or QPen or an enum. Accept `impl AsRef<QPen>` then implement Into<QPen> from QColor and the enum. Then this should allow Rust code to have any of the types.

There are a few places already where we could have had overloads for QPoint, QSize etc consider doing this in those places.

Note however this will mean we need an extra impl block that then calls the direct CXX binding.

jnbooth:
I don't know if it makes sense for `QColor` to implement `AsRef<QPen>`, because a `QColor` reference is not and cannot be a reference to a `QPen`. You would need to construct the `QPen` object inside the `AsRef` implementation, which seems like misuse of the trait and a high risk of memory management issues.

One alternative would be to use enums, like this:

```rs
enum SetPenParam<'a> {
    PenStyle(ffi::PenStyle),
    QColor(&'a QColor),
    QPen(&'a QPen),
}

impl From<ffi::PenStyle> for SetPenParam<'static> {
    fn from(value: ffi::PenStyle) -> Self {
        Self::PenStyle(value)
    }
}
impl<'a> From<&'a QColor> for SetPenParam<'a> {
    fn from(value: &'a QColor) -> Self {
        Self::QColor(value)
    }
}
impl<'a> From<&'a QPen> for SetPenParam<'a> {
    fn from(value: &'a QPen) -> Self {
        Self::QPen(value)
    }
}

impl QPainter {
    pub fn set_pen<'a, T: Into<SetPenParam<'a>>>(self: Pin<&mut QPainter>, pen: T) {
        match pen.into() {
            SetPenParam::PenStyle(style) => ffi::qpainter_set_pen_penstyle(self, style),
            SetPenParam::QColor(color) => ffi::qpainter_set_pen_qcolor(self, color),
            SetPenParam::QPen(pen) => ffi::qpainter_set_pen_qpen(self, pen),
        }
    }
}
```

PR #901, opened 2024-03-20
Implement as_ref for overloads in cxx-qt-lib

Montel:
Closed: #810

### DIFF
--- a/crates/cxx-qt-lib/src/gui/qpainter.rs
+++ b/crates/cxx-qt-lib/src/gui/qpainter.rs
@@ -3,6 +3,8 @@
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
+use core::pin::Pin;
+
 #[cxx::bridge]
 mod ffi {
     #[namespace = "Qt"]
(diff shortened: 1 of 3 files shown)


Montel:
error[E0308]: mismatched types
   --> crates/cxx-qt-lib/src/gui/qpen.rs:186:9
    |
184 |     fn as_ref(&self) -> &ffi::QPen {
    |                         ---------- expected `&qpen::QPen` because of return type
185 |         let pen = ffi::qpen_init_from_qcolor(&self);
186 |         pen
    |         ^^^ expected `&QPen`, found `QPen`
    |
help: consider borrowing here
    |
186 |         &pen
    |         +

For more information about this error, try `rustc --explain E0308`.
The following warnings were emitted during compilation:


but if I use &pen if doesn't compile too.

Do you have an idea ?
thanks

Montel:
warning: cxx-qt-lib@0.6.0: Compiler version doesn't include clang or GCC: "c++" "--version"
error[E0515]: cannot return reference to local variable `pen`
   --> crates/cxx-qt-lib/src/gui/qpen.rs:186:9
    |
186 |         &pen
    |         ^^^^ returns a reference to data owned by the current function

For more information about this error, try `rustc --explain E0515`.
The following warnings were emitted during compilation:

ahayzen-kdab:
Fun :-) so because it wants a reference the lifetime is the fun part here. Either we would need to link up the lifetimes somehow (not sure if possible?) or maybe we'd need to go the `Into` route instead as that could be a value not a reference, eg accept `Into<QPen>` we also need to see what is the common way of doing this in the Rust ecosystem, so looking for examples in std would be good.

ahayzen-kdab:
Also note that `From<T>` is usually preferred over `Into<T>`

Montel:
@LeonMatthesKDAB  do you have an idea how to fix it ? :)

PR #904, opened 2024-03-28
qtypes: add support for Qt alias types that don't match

### LINKED ISSUE #882: Add `qsizetype` to `cxx-qt` crate
This means that when wrapping APIs we don't need to cast the type in C++.

Then once conditional code is working correctly #765 this could mean that `cxx-qt-lib` can remove a load of C++ code where it's just casting between `isize` and `qsizetype` or `int` depending on Qt 5 or Qt 6.

Consider which of these types from `<QtGlobal>` could be useful, eg `qint64` could remove a lot of `static_cast` in C++, probably remove the need for C++ trampolines so that there are direct CXX bindings and just have a into/from on the Rust side? (either in a Rust side wrapper or just accept Into<qint64> in methods not sure what happens with returns though?)

## Already map to std numerics (as in Rust/C++ std, so i32 ⇄ qint32 ⇄ std::int32_t)

- [x] ~`qint8`~
- [x] ~`qint16`~
- [x] ~`qint32`~
- [x] ~`quint8`~
- [x] ~`quint16`~
- [x] ~`quint32`~

## Most useful types

- [ ] `qint64`
- [ ] `qintptr`
- [ ] `qsizetype`
- [ ] `quint64`
- [ ] `quintptr`
- [ ] `qreal`

## Probably useful types?

- [ ] `qlonglong`
- [ ] `qptrdiff`
- [ ] `qulonglong`

ahayzen-kdab:
Closes #882


crates/cxx-qt-lib/src/core/qtypes.rs
```diff
@@ -0,0 +1,83 @@
...
+
+        #[rust_name = "qsizetype_from_isize"]
+        fn qsizetypeFromIsize(value: isize) -> QSizeType;
+        #[rust_name = "qsizetype_into_isize"]
+        fn qsizetypeIntoIsize(value: QSizeType) -> isize;
+    }
+}
+
+#[repr(C)]
+pub struct QInt64 {
```
ahayzen-kdab, line 36: should probably have things like Eq, Ord, Debug ?

Or do we change this to be a totally trivial type?

crates/cxx-qt-lib/src/core/qtypes.rs
```diff
@@ -0,0 +1,83 @@
+// SPDX-FileCopyrightText: 2023 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
```
Montel, line 1: 2024 :) new file

crates/cxx-qt-lib/src/core/qtypes.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+assert_alignment_and_size(::qsizetype,
+                          alignof(::std::size_t),
+                          sizeof(::std::size_t));
+
+assert_alignment_and_size(::qint64,
+                          alignof(::std::int64_t),
+                          sizeof(::std::int64_t));
+
+// static_assert(!::std::is_trivially_copy_assignable<qsizetype>::value);
```
Montel, line 19: remove it ?

crates/cxx-qt-lib/src/core/qtypes.rs
ahayzen-kdab: This then doesn't use `MaybeUninit`, do we need this as the memory might not be initialised and be random? But then it also is just a number :thinking:

crates/cxx-qt-lib/src/core/qtypes.rs
```diff
@@ -0,0 +1,189 @@
...
+use cxx::{type_id, ExternType};
+use std::mem::MaybeUninit;
+
+#[cxx::bridge]
+mod ffi {
+    unsafe extern "C++" {
+        include!("cxx-qt-lib/qtypes.h");
+
+        #[cxx_name = "qint64"]
+        type QInt64 = super::QInt64;
```
ahayzen-kdab, line 15: In Rust should we be calling these `QInt64` and `QUInt64` or should we copy Qt and use lowercase `qint64` and `quint64` ? :thinking:

crates/cxx-qt-lib/src/core/qtypes.rs
ahayzen-kdab: Gone with new type

crates/cxx-qt-lib/src/core/mod.rs
```diff
@@ -83,6 +83,9 @@ pub use qt::{
 mod qtime;
 pub use qtime::QTime;
 
+mod qtypes;
+pub use qtypes::{QInt64, QIntPtr, QSizeType, QUInt64, QUIntPtr};
```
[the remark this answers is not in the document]
ahayzen-kdab: Added `qreal` and only supporting as a `double`.

crates/cxx-qt-lib/src/core/qtypes.rs
```diff
@@ -0,0 +1,253 @@
...
+#[cxx::bridge]
+mod ffi {
+    unsafe extern "C++" {
+        include!("cxx-qt-lib/qtypes.h");
+    }
+}
+
+/// Typedef for long long int. This type is guaranteed to be 64-bit on all platforms supported by Qt.
+#[repr(transparent)]
+#[derive(Default, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
```
LeonMatthesKDAB, line 18: I would add `Clone` and `Copy` to this list as well...

crates/cxx-qt-lib/src/core/qtypes.rs
```diff
@@ -0,0 +1,253 @@
...
+pub struct QInt64(i64);
```
LeonMatthesKDAB, line 19: Let's go with `qint64` here. No need to rename the type. It's still a sort off "fundamental type" like i64, which is also lowercase.

crates/cxx-qt-lib/src/core/qtypes.rs
```diff
@@ -0,0 +1,253 @@
...
+//
+// Static checks on the C++ side to ensure the size is the same.
+unsafe impl ExternType for QInt64 {
+    type Id = type_id!("qint64");
+    type Kind = cxx::kind::Trivial;
+}
+
+/// Integral type for representing pointers in a signed integer (useful for hashing, etc.).
+#[repr(transparent)]
+#[derive(Default, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
```
LeonMatthesKDAB, line 57: Again, let's add `Copy` and `Clone` to that list :)

crates/cxx-qt-lib/src/core/qtypes.rs
```diff
@@ -0,0 +1,253 @@
...
+unsafe impl ExternType for QIntPtr {
+    type Id = type_id!("qintptr");
+    type Kind = cxx::kind::Trivial;
+}
+
+/// Typedef for double
+///
+/// Note that configuring Qt with -qreal float is not supported
+#[repr(transparent)]
+#[derive(Default, Debug, PartialEq, PartialOrd)]
```
LeonMatthesKDAB, line 98: Copy + Clone

crates/cxx-qt-lib/src/core/qtypes.rs
```diff
@@ -0,0 +1,253 @@
...
+//
+// Static checks on the C++ side to ensure the size is the same.
+unsafe impl ExternType for qreal {
+    type Id = type_id!("qreal");
+    type Kind = cxx::kind::Trivial;
+}
+
+/// Typedef for unsigned long long int. This type is guaranteed to be 64-bit on all platforms supported by Qt.
+#[repr(transparent)]
+#[derive(Default, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
```
LeonMatthesKDAB, line 138: Copy + Clone

crates/cxx-qt-lib/src/core/qtypes.rs
```diff
@@ -0,0 +1,253 @@
...
+//
+// Static checks on the C++ side to ensure the size is the same.
+unsafe impl ExternType for QUInt64 {
+    type Id = type_id!("quint64");
+    type Kind = cxx::kind::Trivial;
+}
+
+/// Integral type for representing pointers in an unsigned integer (useful for hashing, etc.).
+#[repr(transparent)]
+#[derive(Default, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
```
LeonMatthesKDAB, line 177: Copy + Clone

crates/cxx-qt-lib/src/core/qtypes.rs
```diff
@@ -0,0 +1,253 @@
...
+// Static checks on the C++ side to ensure the size is the same.
+unsafe impl ExternType for QUIntPtr {
+    type Id = type_id!("quintptr");
+    type Kind = cxx::kind::Trivial;
+}
+
+/// Integral type providing Posix' ssize_t for all platforms.
+///
+/// This type is guaranteed to be the same size as a size_t on all platforms supported by Qt.
+#[derive(Default, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
```
LeonMatthesKDAB, line 217: Copy + Clone

LeonMatthesKDAB (commented):
Overall this looks good.
Naming wise I would just copy Qts casing (so lowercase). These types are quiet "fundamental types", which even in Rust are lowercase (i64, isize, etc.).

A sort-off open question is still how many operations we want to support on these...
Do we basically just enable `From/Into` and call it a day, or do we want to have support for using the type more directly, like support for `+-*/` on qint64...

For now it's probably good enough if we just have them convertable to their inner type, but something to keep in mind.

On the same note: Maybe we *don't* want to support Deref on these types?
It may be weird if you call e.g. `abs` on qint64, but get an i64 back...
And now it would be a breaking change for us to add our own `abs` method that returns qint64...

So it seems like limiting the API to just From/To with some derives is maybe best?

crates/cxx-qt-lib/src/core/qtypes.rs
ahayzen-kdab: done

crates/cxx-qt-lib/src/core/qtypes.rs
ahayzen-kdab: done

crates/cxx-qt-lib/src/core/qtypes.rs
ahayzen-kdab: done

crates/cxx-qt-lib/src/core/qtypes.rs
ahayzen-kdab: done

crates/cxx-qt-lib/src/core/qtypes.rs
ahayzen-kdab: done

crates/cxx-qt-lib/src/core/qtypes.rs
ahayzen-kdab: done

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/KDAB/cxx-qt/compare/v0.7.0...HEAD)
 
+### Added
+
+- Support for further types: `qreal`, `qint64`, `qintptr`, `qsizetype`, `quint64`, `quintptr`
```
ahayzen-kdab, line 22: do we also want `qlonglong`, `qulonglong`, or `qptrdiff` ?

See https://doc.qt.io/qt-6/qttypes.html#types for full types in Qt

crates/cxx-qt-lib/src/core/qmodelindex.rs
```diff
@@ -35,6 +36,9 @@ mod ffi {
         #[rust_name = "sibling_at_row"]
         fn siblingAtRow(self: &QModelIndex, row: i32) -> QModelIndex;
 
+        /// Returns a `quintptr` used by the model to associate the index with the internal data structure.
+        #[rust_name = "internal_id"]
+        fn internalId(self: &QModelIndex) -> quintptr;
```
LeonMatthesKDAB, line 41: Note that this is a breaking change.
However, let's do this now while we are still pre-1.0

CHANGELOG.md
LeonMatthesKDAB: Let's just go with what we need at the moment and add new types as needed.

LeonMatthesKDAB (dismissed):

crates/cxx-qt-lib/src/core/qmodelindex.rs
LeonMatthesKDAB: Of course it's the same fo all other APIs that were changed during this PR.

LeonMatthesKDAB (approved):
Re-approved after rebase

PR #907, opened 2024-04-03
cxx-qt-build: set a cfg_evaluator so that we can use cfg attributes

### LINKED ISSUE #765: Allow to create #cfg for adding specific method in qt6 or qt5
Hi,
Qt5 and Qt6 must have new methods or method removed.
We must be able to use a #cfg(...) for adding it in specific qt_version_major

Idea is using specific mod

For example:
#[cfg(qt_version_major = "6")]
#[cxx::bridge]
mod ffi_qt_6 {
...
}

At the moment it doesn't build ad it seems that qt_version_major  is not forwarded by the build system.

We really need to add it if we want to support new qt6 methods.

ahayzen-kdab:
Closes #765

### DIFF
--- a/crates/cxx-qt-build/src/cfg_evaluator.rs
+++ b/crates/cxx-qt-build/src/cfg_evaluator.rs
@@ -1,0 +1,167 @@
+// SPDX-FileCopyrightText: CXX Authors
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+// SPDX-FileContributor: David Tolnay <dtolnay@gmail.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+// Modified from
+// https://github.com/dtolnay/cxx/blob/2193addc308c99512b8757431abfaf87cc54c086/gen/build/src/cargo.rs
+
+use cxx_gen::{CfgEvaluator, CfgResult};
+use std::borrow::Borrow;
+use std::cmp::Ordering;
+use std::collections::{BTreeMap as Map, BTreeSet as Set};
+use std::env;
+use std::sync::OnceLock;
+
+static ENV: OnceLock<CargoEnv> = OnceLock::new();
+
+struct CargoEnv {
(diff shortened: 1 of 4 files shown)


Montel:
Add info about cxx file (https://github.com/dtolnay/cxx/pull/1334)

crates/cxx-qt-build/src/cfg_evaluator.rs
```diff
@@ -0,0 +1,167 @@
+// SPDX-FileCopyrightText: 2024 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+// SPDX-FileContributor: David Tolnay <dtolnay@gmail.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+// Modified from
+// https://github.com/dtolnay/cxx/blob/2193addc308c99512b8757431abfaf87cc54c086/gen/build/src/cargo.rs
```
LeonMatthesKDAB, line 8: Do we want to include the original licenses from the CXX repository here?

I know CXX is also MIT & Apache dual-licensed, but maybe we can at least link to them to ensure we're doing things correctly :thinking:

LeonMatthesKDAB (commented):
Overall looks good, do we need to add additional licensing information?

crates/cxx-qt-build/src/cfg_evaluator.rs
ahayzen-kdab: We wouldn't need to link to them as they are already in the LICENSES folder and specified by the SPDX identifier. However the `SPDX-FileCopyrightText` could instead be CXX authors or a name they are using as it's primarily come from that source.

ahayzen-kdab: Updated to "CXX Authors" with the file contributors, so should be correct now.

LeonMatthesKDAB (approved):

PR #909, opened 2024-04-05
github: add basic dependabot to update github actions

ahayzen-kdab:


ahayzen-kdab:
Note this should in theory automatically create branches to update any github action usages if they are out of date eg `- uses: actions/checkout@v3` would be updated to `- uses: actions/checkout@v4` etc

LeonMatthesKDAB (approved):
LGTM, let's try it out.

We can also consider adding a dependabot for our Rust dependencies. Though that can get spammy quite quickly :thinking:

ahayzen-kdab:
Right, wonder how it compares to `cargo outdated` as we don't have a lock file so we don't care about updating the micro version if we are only depending on the minor version in the toml

PR #917, opened 2024-04-09
Add qpen autotest

Montel:


Montel:
Problem with configure on windows (it's a CI problem)
Is it possible to start windows build by hand ? 
or we need to push a new change ?

crates/cxx-qt-lib/src/gui/qpen.rs
```diff
@@ -129,10 +129,7 @@ mod ffi {
 
 #[repr(C)]
 pub struct QPen {
-    #[cfg(cxxqt_qt_version_major = "5")]
     _cspec: MaybeUninit<[i32; 2]>,
```
ahayzen-kdab, line 132: why is this the size of two i32s ? it looks to be a single raw pointer in C++ ? So this should be `MaybeUninit<usize>` ?

ahayzen-kdab (approved):
LGTM :-)

PR #926, opened 2024-04-16
WIP: experiment of using cfg options for differing Qt types

ahayzen-kdab:
This requires #904 and a CXX patch to work.


ahayzen-kdab:
this causes duplicate names in CXX, instead this should be handled by having different names in the CXX bridge then a wrapper in Rust to call the right one, or a wrapper in C++ like there is currently. Which means this change is irrelevant other than moving the casting of the type from C++ to Rust which could be done in a separate change.

PR #932, opened 2024-04-18
signals: use #[must_use] when returning a QMetaObjectConnectionGuard

### LINKED ISSUE #930: Connections don't have must_use when returning a QMetaObjectConnectionGuard (was: unable to make qsignal to call its closure)
I mapped quite a few Widgets for my project. Now i want to use callbacks on a QPushButton. Unfortunately i am unable to make the `clicked` (`on_clicked`) signal to work. It seems like the function is never called.

Example code:
```rs
fn main() {
    // Create the application and engine
    let mut app = QApplication::new();

    let mut main_window = QMainWindow_new_ptr();

    let mut button = unsafe { QPushButton_new_ptr_with_text_parent(&QString::from("test"), main_window as *mut _) };
    let mut button = unsafe {Pin::new_unchecked(&mut *button)};
    println!("test");
    button.as_mut().on_clicked(|btn, status| println!("button pressed! {status}"));
    // button.show();

    unsafe {Pin::new_unchecked(&mut *main_window)}.show();

    // Start the app
    if let Some(app) = app.as_mut() {
        app.exec();
    }
}
```

Definition of the signal:
```rs
#[cxx_qt::bridge]
pub mod qobject {
    unsafe extern "C++Qt" {
        include!(<QPushButton>);
        #[qobject]
        type QPushButton;

        #[qsignal]
        fn clicked(self: Pin<&mut QPushButton>, checked: bool);
    }
}
```

I pushed the full example to my fork of this project: https://github.com/binja-tools/cxx-qt/tree/test-button/examples/cargo_without_cmake_2

Am i doing something wrong or is this not implemented for C++Qt classes?

ahayzen-kdab:
Before we used to return a QMetaObjectConnection which used to disconnection now we have two types, so ensure #[must_use] is in the right place.

Closes #930

### DIFF
--- a/crates/cxx-qt-gen/src/generator/rust/property/mod.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/property/mod.rs
@@ -287,7 +287,6 @@ mod tests {
 
                     #[doc(hidden)]
                     #[namespace = "rust::cxxqtgen1"]
-                    #[must_use]
                     #[rust_name = "MyObject_connect_trivial_property_changed"]
                     fn MyObject_trivialPropertyChangedConnect(self_value: Pin<&mut MyObject>, signal_handler: MyObjectCxxQtSignalHandlertrivialPropertyChanged, conn_type: CxxQtConnectionType) -> CxxQtQMetaObjectConnection;
                 }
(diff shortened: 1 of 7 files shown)


greaka (commented):
It is easy to miss adding `#[must_use]` when adding a new signal.
By annotating [QMetaObjectConnectionGuard](https://github.com/KDAB/cxx-qt/blob/8b809e8b4102fdac0690161bf3264ff65864432e/crates/cxx-qt/src/connectionguard.rs#L15) instead, the must_use warning will automatically be applied to every function that returns that struct.

While both changing the return type and adding new functions is equally unlikely, putting must_use on the type also helps downstream users when working with this type.

LeonMatthesKDAB (approved):
Really good suggestion by @greaka

PR #933, opened 2024-04-18
Add qt6 rgbSwap method

Montel:


crates/cxx-qt-lib/src/gui/qimage.rs
```diff
@@ -147,6 +147,11 @@ mod ffi {
         /// Returns the enclosing rectangle (0, 0, width(), height()) of the image.
         fn rect(self: &QImage) -> QRect;
 
+        /// Swaps the values of the red and blue components of all pixels, effectively converting an RGB image to an BGR image.
+        #[cfg(cxxqt_qt_version_major = "6")]
```
ahayzen-kdab, line 151: guess this sortof thing will be better once we have the other change for `at_least_qt_version_6`

Montel: I will merge with other patch it's better :)

Montel:
merge in #934

PR #934, opened 2024-04-18
Add support for several qt version (minor patch)

Montel:


crates/cxx-qt-build/src/lib.rs
```diff
@@ -491,6 +491,18 @@ impl CxxQtBuilder {
...
+
+        for minor in 0..qtbuild.version().minor {
+            let at_least_qt_version = format!(
+                "cxxqt_at_least_qt_version{}_{}",
+                qtbuild.version().major,
+                minor
+            );
+            println!("cargo:rustc-cfg={}", at_least_qt_version);
+            let variable_cargo = format!("CARGO_CFG_{}", at_least_qt_version);
+            env::set_var(variable_cargo, at_least_qt_version);
```
ahayzen-kdab, line 503: maybe just set it to `true` or `1` something ? eg `env::set_var(variable_cargo, "true");`

crates/cxx-qt-build/src/lib.rs
```diff
@@ -491,6 +491,18 @@ impl CxxQtBuilder {
...
+        }
+
         // Also set to env so our cfg evaulator finds it
```
ahayzen-kdab, line 506: i'd keep this code with the ones above so that the `cxxqt_qt_version_major` code stays together.

Eg move the new code to be after this set_var

crates/cxx-qt-build/src/lib.rs
```diff
@@ -491,13 +491,25 @@ impl CxxQtBuilder {
             "cargo:rustc-cfg=cxxqt_qt_version_major=\"{}\"",
             qtbuild.version().major
         );
+
```
ahayzen-kdab, line 494: random new line :-)

crates/cxx-qt-build/src/lib.rs
```diff
@@ -491,13 +491,25 @@ impl CxxQtBuilder {
...
             qtbuild.version().major.to_string(),
         );
 
+        for minor in 0..qtbuild.version().minor {
+            let at_least_qt_version = format!(
+                "cxxqt_at_least_qt_version_{}_{}",
+                qtbuild.version().major,
+                minor
+            );
+            println!("cargo:rustc-cfg={}", at_least_qt_version);
```
ahayzen-kdab, line 508: wonder if this should be set to true as well so `println!("cargo:rustc-cfg={}=\"true\"", at_least_qt_version);` or maybe that would then mean that code outside of CXX bridges would require `#[cfg(cxxqt_qt_least_qt_version_6_5 = "true")]` rather than `#[cfg(cxxqt_qt_least_qt_version_6_5)]`.

Can you see what happens at the moment for a method outside of the CXX bridges (eg normal Rust code) does `#[cfg(cxxqt_qt_least_qt_version_6_5)]` work with the existing code?

Montel:
```
  --> examples/qml_features/rust/src/invokables.rs:62:12
   |
10 |   #[cxx_qt::bridge(cxx_file_stem = "rust_invokables")]
   |   ---------------------------------------------------- function or associated item `foo` not found for this struct
...
21 |           type RustInvokables = super::RustInvokablesRust;
   |  ______________-
22 | |     }
23 | |
24 | |     // ANCHOR: book_qnamespace
...  |
61 | |         #[cfg(cxxqt_qt_least_qt_version_6_9)]
62 | |         fn foo(self: Pin<&mut RustInvokables>);
   | |           -^^^ function or associated item not found in `RustInvokables`
   | |___________|
   | 
```
=> doesn't work.

ahayzen-kdab: Isn't this inside a CXX / CXX-Qt bridge macro, i was meaning outside the bridge, as that is where `cargo:rustc-cfg` end up

Montel:
```
diff --git a/crates/cxx-qt-build/src/lib.rs b/crates/cxx-qt-build/src/lib.rs
index 73be1ae0..10bbf51b 100644
--- a/crates/cxx-qt-build/src/lib.rs
+++ b/crates/cxx-qt-build/src/lib.rs
@@ -504,7 +504,7 @@ impl CxxQtBuilder {
                 qtbuild.version().major,
                 minor
             );
-            println!("cargo:rustc-cfg={}", at_least_qt_version);
+            println!("cargo:rustc-cfg={}=\true\"", at_least_qt_version);
             let variable_cargo = format!("CARGO_CFG_{}", at_least_qt_version);
             env::set_var(variable_cargo, "true");
         }

```

compile failed
```
   Compiling qml-features v0.1.0 (/opt/travail/git/cxx-qt/examples/qml_features/rust)
error[E0765]: unterminated double quote string
  |
  = note: this error occurred on the command line: `--cfg=cxxqt_at_least_qt_version_6_0=    rue"`

error: could not compile `cxx-qt-lib` (lib)
```

ahayzen-kdab: OK tested locally this doesn't need the true for the `cargo:rustc-cfg` so what you have currently is fine

ahayzen-kdab: Think you had a typo `cxxqt_qt_least_qt_version_6_5` should be `cxxqt_at_least_qt_version_6_5` :-)

Montel: ok

crates/cxx-qt-build/src/lib.rs
```diff
@@ -498,6 +498,17 @@ impl CxxQtBuilder {
             qtbuild.version().major.to_string(),
         );
 
+        for minor in 0..qtbuild.version().minor {
+            let at_least_qt_version = format!(
```
ahayzen-kdab, line 502: i guess one thing to note here is that it doesn't work across major versions.

Eg i can't do "at least 5.15", as it only loops through the minor version for the current major version.

I don't think that is a problem at the moment as we only care about 5.15+ for the 5.x series.

But i'm thinking here about if/when Qt 7 comes out then this is going to break ?

But to resolve this it'd need a manual loop ? as once we are eg 7.0 you can't loop over 6.x for the minor version.

Montel: indeed. Perhaps we need to add a loop when major > 6
For the moment it's not necessary.
For qt7
We need to make range for qt6 = > create for qt 6.15 or other...
Otherwise we will not create at_least_qt_6_99 etc :)
=> We must improve it when we will know which will be the last minor qt6 version

crates/cxx-qt-build/src/lib.rs
```diff
@@ -498,6 +498,26 @@ impl CxxQtBuilder {
             qtbuild.version().major.to_string(),
         );
 
+        for minor in 0..qtbuild.version().minor {
```
ahayzen-kdab, line 501: i think you were right these need to be +1 ?

As https://doc.rust-lang.org/std/ops/struct.Range.html says that `The range start..end contains all values with start <= x < end. It is empty if start >= end.`

So this should be `0..minor+1` ?

crates/cxx-qt-build/src/lib.rs
```diff
@@ -498,6 +498,26 @@ impl CxxQtBuilder {
...
+                qtbuild.version().major,
+                minor
+            );
+            println!("cargo:rustc-cfg={}", at_least_qt_version);
+            let variable_cargo = format!("CARGO_CFG_{}", at_least_qt_version);
+            env::set_var(variable_cargo, "true");
+        }
+
+        // We don't support Qt < 5
+        (5..qtbuild.version().major).for_each(|_: u32| {
```
ahayzen-kdab, line 513: this should be `5..major+1`

ahayzen-kdab, line 513: why is this one using `for_each` whereas above uses `for`, can these be consistent ?

crates/cxx-qt-build/src/lib.rs
```diff
@@ -498,6 +498,26 @@ impl CxxQtBuilder {
...
+            let at_least_qt_major_version =
+                format!("cxxqt_at_least_qt_version_{}", qtbuild.version().major);
```
ahayzen-kdab, line 515: this should not be `qtbuild.version().major` but the `u32` value in the for loop

crates/cxx-qt-build/src/lib.rs
Montel: Ah yep indeed :) 
I am an rust expert without knowing it :)

crates/cxx-qt-build/src/lib.rs
Montel: I think that for_each was better but indeed I can use for _ in 5..qtbuild.version().major
as you want.

crates/cxx-qt-build/src/lib.rs
```diff
@@ -498,6 +498,25 @@ impl CxxQtBuilder {
...
+                qtbuild.version().major,
+                minor
+            );
+            println!("cargo:rustc-cfg={}", at_least_qt_version);
+            let variable_cargo = format!("CARGO_CFG_{}", at_least_qt_version);
+            env::set_var(variable_cargo, "true");
+        }
+
+        // We don't support Qt < 5
+        for n in 5..=qtbuild.version().major {
```
ahayzen-kdab, line 513: tiny nit, could use `major` instead of `n`

ahayzen-kdab: suspect it'll make cargo fmt wrap the next line though :-) but it's easier to read

crates/cxx-qt-build/src/lib.rs
```diff
@@ -498,6 +498,25 @@ impl CxxQtBuilder {
             qtbuild.version().major.to_string(),
         );
 
+        for minor in 0..=qtbuild.version().minor {
+            let at_least_qt_version = format!(
+                "cxxqt_at_least_qt_version_{}_{}",
```
LeonMatthesKDAB, line 503: Overall I like the `at_least`, it's nice and explicit :)

I'd prefer, if it was `qt_version_at_least`, not `at_least_qt_version` though, as that would be in line with `qt_version_major` and `qt_version_minor`.

LeonMatthesKDAB (commented):
So to summarize, we can do:

### Qt Version >= 6.4
```rust
#[cfg(qt_version_at_least_6_4)]
```

### Qt Version < 6.4
``` rust
#[cfg(not(qt_version_at_least_6_4))]
```

### Qt Version == 6.4
```rust
#[cfg(all(qt_version_major="6",qt_version_minor="4"))]
```

That should cover all use cases, right?

So overall, nice to see this working, I've only got small nitpicks left.

crates/cxx-qt-lib/src/core/qline.rs
```diff
@@ -64,6 +67,12 @@ mod ffi {
         #[rust_name = "set_points"]
         fn setPoints(self: &mut QLine, p1: &QPoint, p2: &QPoint);
 
+        /// Returns this line as a line with floating point accuracy.
+        /// since Qt 6.4.
+        #[cfg(any(cxxqt_at_least_qt_version_7, cxxqt_at_least_qt_version_6_4))]
```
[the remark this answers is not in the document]
LeonMatthesKDAB: As discussed, doesn't hurt anything for now, so let's just keep it.

LeonMatthesKDAB (approved):

Issue #939, opened 2024-04-19
Use the `naming` module everywhere


LeonMatthesKDAB:
This module is more complete than our previous `CombinedIdent` approach.

Make sure to use it everywhere, especially in generation.

BenFordTytherington:
CombinedIdent has now been completely removed from the codebase as of PR https://github.com/KDAB/cxx-qt/pull/1005

LeonMatthesKDAB:
> CombinedIdent has now been completely removed from the codebase as of PR #1005

Awesome!
I'll still go through the codebase and check whether there's places where we need to replace some more `Ident` instances with `Name`, but after that we can close this issue :partying_face:

LeonMatthesKDAB:
I think this is done now.
@BenFordTytherington can you confirm and close the issue? 🤔

Issue #938, opened 2024-04-19
Split out structuring into its own phase


LeonMatthesKDAB:
See #937 

We want to avoid structuring while parsing, which leads to oddities like us having to parse all qobjects first, so we can add e.g. QEnums later.

So do this as its own phase before C++ generation, as we also probably don't need this for the Rust generation at all anyway.

What we need to associate:
- [x] QEnums (see #942 )
- [x] Signals
- [x] Methods (e.g. invokables)
- [ ] Traits (e.g. threading, constructor, etc.)
- [ ] ...

Issue #937, opened 2024-04-19
Tracking Issues: Current cxx-qt-gen Refactorings


LeonMatthesKDAB:
What originally started as a refactor of naming (see #835 ) has now become a plan for a larger internal refactoring, which we aim to be the baseline of the next 0.7 release.
From this improved, cleaner architecture, we should be able to stabilize more easily and quickly.

The planned structure of cxx-qt-gen:

- Frontend (Parser)
    - Parsing
        This phase should parse the syn tree into a list of declarations with all attributes resolved into usable data formats.
        However, all items should remain independent from each other and remain close to the way the user wrote it. **(new :star2: )**
        The result should be a list of extern blocks, each with a list of items, like QObject declarations, QEnum declarations, etc.
    - Type Name extraction (the "Naming phase") **(new :star2: )**
        Extracts all nameable types out of the parsed data.
        This is useful to resolve references and renamed types in the generation phase.
    - Checks phase??? **(new :star2: )**
        We also consider adding a separate phase for certain checks. Though it's not yet clear what should belong there and what not.
- Rust Backend
    - Generation
        We assume that for Rust code generation, most things can stay independent.
        Therefore we should be able to directly generate the Rust code from the input declarations without having to rearrange and associate more complex items like QObjects.
    - Writer
- C++ Backend
    - Structuring **(new :star2: )**
        C++ headers need to be structured, as all data within a QObject must be written inside the same pair of `{...}`.
        Therefore we need to associate the loosely parsed data from the Frontend.
        This should now be its own phase
    - Generation
    - Writer

Overall we should use the new functions from the `naming` module wherever possible, which should then solve our issues regarding `cxx_name`, `rust_name`, etc.

### Associated issues:
- [x] #828 
- [x] #938 
- [x] #939
- [x] #945

PR #942, opened 2024-04-24
QEnum: Separate Structuring and use `Name` struct.

LeonMatthesKDAB:
Part of #938 and #939

### DIFF
--- a/crates/cxx-qt-gen/src/generator/rust/qenum.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/qenum.rs
@@ -3,7 +3,7 @@
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
-use crate::{generator::rust::fragment::GeneratedRustFragment, parser::qenum::ParsedQEnum};
+use crate::parser::qenum::ParsedQEnum;
 use quote::quote;
 use syn::{parse_quote_spanned, spanned::Spanned, Item};
 
@@ -12,9 +12,17 @@ pub fn generate_cxx_mod_contents(qenums: &[ParsedQEnum]) -> Vec<Item> {
         .iter()
         .flat_map(|qenum| {
             let qenum_item = &qenum.item;
-            let qenum_ident = &qenum.ident;
-            let namespace = &qenum.namespace;
-            let namespace = if namespace.is_empty() {
+            let qenum_ident = &qenum.name.rust_unqualified();
+            let namespace = &qenum.name.namespace();
+
+            // If the QEnum has a namespace inherited from its QObject, it won't have the
+            // `#[namespace=...]` attribute, so we need to inject it.
+            let shared_namespace = if namespace.is_some() && qenum.qobject.is_some() {
+                quote! { #[namespace = #namespace] }
+            } else {
+                quote! {}
+            };
+            let cxx_namespace = if namespace.is_none() {
                 quote! {}
             } else {
                 quote! { #[namespace = #namespace ] }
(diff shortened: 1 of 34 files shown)


crates/cxx-qt-gen/src/generator/rust/qenum.rs
```diff
@@ -12,9 +12,17 @@ pub fn generate_cxx_mod_contents(qenums: &[ParsedQEnum]) -> Vec<Item> {
         .iter()
         .flat_map(|qenum| {
             let qenum_item = &qenum.item;
-            let qenum_ident = &qenum.ident;
-            let namespace = &qenum.namespace;
-            let namespace = if namespace.is_empty() {
+            let qenum_ident = &qenum.name.rust_unqualified();
+            let namespace = &qenum.name.namespace();
+
+            // If the QEnum has a namespace inherited from its QObject, it won't have the
```
ahayzen-kdab, line 18: can an enum have a different namespace to it's QObject if it's on a QObject ?

eg should we be blocking that case?

ahayzen-kdab: Unsure :-D Maybe we need to think what that looks like in C++, and could be solved in another pull request, just something i noticed. Maybe create an issue to discuss it

crates/cxx-qt-gen/src/generator/structuring/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+pub mod qobject;
+pub use qobject::StructuredQObject;
+
+use crate::parser::cxxqtdata::ParsedCxxQtData;
+use syn::{Error, Ident, Result};
+
+/// The naming phase may already associate QEnums with QObjects.
+/// In this case, we want to use the same error message, as in this module.
+pub(crate) fn error_unknown_qobject(ident: &Ident) -> Error {
+    Error::new_spanned(ident, format!("Unkown QObject: {ident}"))
```
ahayzen-kdab, line 24: Typo, should be Unknown

crates/cxx-qt-gen/src/generator/rust/qenum.rs
LeonMatthesKDAB: Done in #945

crates/cxx-qt-gen/src/parser/mod.rs
```diff
@@ -104,18 +110,53 @@ impl Parser {
...
+                } else {
+                    return Err(structuring::error_unknown_qobject(qobject_ident));
+                }
+            }
+        }
+
+        Ok(())
+    }
+
+    fn naming_phase(
```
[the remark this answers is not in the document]
LeonMatthesKDAB: At this point I think it's easiest to actually generate the `Name` of each item during parsing directly from the given attributes.
So most of what's left to do is to identify all nameable types, which is what remains for the "Naming Phase".

See: https://github.com/KDAB/cxx-qt/issues/937

crates/cxx-qt-gen/src/parser/mod.rs
```diff
@@ -104,18 +110,53 @@ impl Parser {
         Ok((cxx_qt_data, module))
     }
 
+    fn resolve_qenum_inherited_namespaces(cxx_qt_data: &mut ParsedCxxQtData) -> Result<()> {
```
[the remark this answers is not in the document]
LeonMatthesKDAB: Done

crates/cxx-qt-gen/src/generator/structuring/mod.rs
LeonMatthesKDAB: Fixed

ahayzen-kdab (approved):
LGTM :-)

Issue #943, opened 2024-04-24
Add `serde` support for cxx-qt-lib using `#[serde(into=...,from=...)]`


LeonMatthesKDAB:
Serde supports automatically converting types after/before (de-)serialization to a proxy type.

This way, we could more easily support Serde in cxx-qt-lib, after our previous failed attempts in #425 .
It would be somewhat less optimal, as e.g. serializing a QString would mean a conversion to/from String, but still better than not supporting it.

@Montel does this sound like a feature you'd be interested in adding?

ahayzen-kdab:
I think the question for some of them is what JSON format should we use, as eg QJsonValue doesn't have an implementation for QPoint, so should we just `{"x": 1, "y": 2}` or `[1, 2]` etc etc.

LeonMatthesKDAB:
The exact output format is decided by the serializer, which is unrelated to JSON, but the question remains what types to convert to which native Rust types.

In the case of QPoint, we can actually just use `#[derive(Serialize,Deserialize)]` directly, as we've implemented it as a trivial type.

In the case of JSON, this would result in `{"x": 1, "y": 2}`, which is the right way to implement this using serde.
If you want to parse your own custom format, you'll have to roll your own Deserialize implementation anyway.

ahayzen-kdab:
Right, I guess it's things like color which can be tricky and things where even the Qt internals aren't consistent like QRect vs QRectF.

ahayzen-kdab:
We should probably add support to QByteArray as well.

PR #944, opened 2024-04-24
feat: lib: Serde support for QString

LeonMatthesKDAB:
See also: #943

### DIFF
--- a/crates/cxx-qt-lib/Cargo.toml
+++ b/crates/cxx-qt-lib/Cargo.toml
@@ -22,6 +22,7 @@ http = { version = "1.0", optional = true }
 rgb = { version = "0.8", optional = true }
 time = { version = "0.3.20", optional = true }
 url = { version = "2.3", optional = true }
+serde = { version = "1", features=["derive"], optional = true }
 
 [build-dependencies]
 cxx-qt-build.workspace = true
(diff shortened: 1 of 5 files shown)


crates/cxx-qt-lib/Cargo.toml
```diff
@@ -22,6 +22,7 @@ http = { version = "1.0", optional = true }
 rgb = { version = "0.8", optional = true }
 time = { version = "0.3.20", optional = true }
 url = { version = "2.3", optional = true }
+serde = { version = "1", features=["default", "derive"], optional = true }
```
ahayzen-kdab, line 25: Why do you need "default" ?

The docs suggest that this should work `serde = { version = "1.0", features = ["derive"] }`

https://serde.rs/derive.html

examples/qml_features/rust/Cargo.toml
```diff
@@ -16,7 +16,7 @@ crate-type = ["staticlib"]
 [dependencies]
 cxx.workspace = true
 cxx-qt.workspace = true
-cxx-qt-lib.workspace = true
+cxx-qt-lib = { workspace = true, features=["default", "serde"] }
```
ahayzen-kdab, line 19: same here, do we need default?

crates/cxx-qt-lib/Cargo.toml
LeonMatthesKDAB: It seems we don't, done :)

examples/qml_features/rust/Cargo.toml
LeonMatthesKDAB: Done

LeonMatthesKDAB:
Let's also see what @Montel thinks of this :)

Montel:
It's a good idea ! I like it!

Montel (approved):

PR #946, opened 2024-04-24
refactor: Free namespacing on associated QEnum

### LINKED ISSUE #945: Independent namespacing of QEnum and QObject?
As per #942 , an associated QEnum inherits its namespace from the QObject it is associated to.

However, we're uncertain if this is indeed the right way to go.
In contrast to Qt, we generate the QEnum, as an entirely independent type from the QObject, which is just re-exported within the QObject class.

So we could allow the user to set their own namespace for the QEnum.
This could even be very useful if the QObject doesn't have a namespace, but we don't want to add the QEnum to the global namespace.

e.g.
```rust
#[cxx_qt::bridge]
mod qobject {
    #[qenum(MyObject)]
    #[namespace="my_object"]
    enum MyEnum { A, B };

    extern "Rust" {
        #[qobject]
        type MyEnum = super::MyEnumRust;
    }
}
```

Which would result in the types:
`::MyObject`, `::my_object::MyEnum`, `::MyObject::MyEnum`.

Especially when you have multiple enums for a single class, it could make sense to move all of the associated enums into a shared namespace, like we have with the `CustomBaseClass` in the qml_features example.

LeonMatthesKDAB:
There is really no reason the QEnum must inherit the QObject namespace.
Even makes our code generation a lot simpler.

Closes #945

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - File name is used for CXX bridges rather than module name to match upstream
 - `#[qobject]` attribute is now optional on types in `extern "RustQt"`
 - `#[qobject]` attribute is now required on types in `extern "C++Qt"`
+- `#[qenum]`s now resolve their namespace independently from their associated QObject
 
 ### Removed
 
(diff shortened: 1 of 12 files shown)


ahayzen-kdab (approved):
Nice! LGTM :-)

PR #947, opened 2024-04-29
CMake: option for WASM build

mattkdab:
Option `BUILD_WASM` to specify WebAssembly as target arch
Support building current main branch with emscripten

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,7 +125,7 @@ if(BUILD_WASM)
     # TODO: fix wasm test failures
     set(BUILD_TESTING OFF)
     # Ensure Rust build for the correct target
-    set(Rust_CARGO_TARGET wasm32-unknown-unknown)
+    set(Rust_CARGO_TARGET wasm32-unknown-emscripten)
     set(THREADS_PREFER_PTHREAD_FLAG ON)
     find_package(Threads REQUIRED)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -121,6 +121,15 @@ add_compile_definitions(
     QT_USE_QSTRINGBUILDER
 )
 
+if(BUILD_WASM)
+    # TODO: fix wasm test failures
+    set(BUILD_TESTING OFF)
+    # Ensure Rust build for the correct target
+    set(Rust_CARGO_TARGET wasm32-unknown-unknown)
+    set(THREADS_PREFER_PTHREAD_FLAG ON)
+    find_package(Threads REQUIRED)
+endif()
+
 # QMAKE environment variable is needed by qt-build-utils to ensure that Cargo
 # uses the same installation of Qt as CMake does.
 if(NOT USE_QT5)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,7 +127,12 @@ if(NOT USE_QT5)
     find_package(Qt6 COMPONENTS Core Gui Test QuickControls2)
 endif()
 if(NOT Qt6_FOUND)
-    find_package(Qt5 5.15 COMPONENTS Core Gui Test QuickControls2 REQUIRED)
+    if(BUILD_WASM)
+        message(FATAL_ERROR 
+            "CXX-Qt for WebAssembly only currently supports Qt 6 builds."
+        )
+    endif()
+    find_package(Qt5 5.15 COMPONENTS Core Gui Test REQUIRED)
 endif()
 
 if (NOT Qt5_FOUND)
--- a/book/src/getting-started/6-wasm-builds.md
+++ b/book/src/getting-started/6-wasm-builds.md
@@ -94,6 +94,26 @@ Example|Working
 `qml_features`|✅ working
 `qml_minimal`|✅ working
 
+## Compiling CXX-Qt Projects for WebAssembly
+
+When compiling a CXX-Qt project for wasm, the Rust target must be set to `wasm32-unknown-emscripten`, and the project must be configured to use POSIX threads.
+
+```cmake
+set(Rust_CARGO_TARGET wasm32-unknown-emscripten)
+set(THREADS_PREFER_PTHREAD_FLAG ON)
+find_package(Threads REQUIRED)
+```
+
+Using CMake, `add_executable` will not output an HTML file when targeting wasm. In order to render an HTML file, one must use `qt_add_executable` in its place. Assuming a project has a CMake flag `BUILD_WASM` to toggle wasm and native builds, one could write the following:
+
+```cmake
+if(BUILD_WASM)
+    qt_add_executable(${APP_NAME} ${SOURCE_FILES})
+else()
+    add_executable(${APP_NAME} ${SOURCE_FILES})
+endif()
+```
+
 ## Issues
 
 - CXX-Qt will currently not build with `wasm_multithread` versions of Qt:
--- a/book/src/getting-started/6-wasm-builds.md
+++ b/book/src/getting-started/6-wasm-builds.md
@@ -23,17 +23,21 @@ The version of Emscripten used to build CXX-Qt and programs using it should matc
 
 Here are the associated Qt and Emscripten versions, and whether they are currently working with CXX-Qt for WebAssembly:
 
-Qt|Emscripten|CXX-Qt Support: `wasm_32`|CXX-Qt Support: `wasm_singlethread`|CXX-Qt Support: `wasm_multithread`
--|-|-|-|-
-6.2|2.0.14|✅ working|*️⃣ N/A|*️⃣ N/A
-6.3|3.0.0|✅ working|*️⃣ N/A|*️⃣ N/A
-6.4|3.1.14|✅ working|*️⃣ N/A|*️⃣ N/A
-6.5|3.1.25|*️⃣ N/A|✅ working|❌ broken
-6.6|3.1.37|*️⃣ N/A|✅ working|❌ broken
+Qt|Emscripten
+-|-
+6.2|2.0.14
+6.3|3.0.0
+6.4|3.1.14
+6.5|3.1.25
+6.6|3.1.37
 
-Info about other versions can be found in the [Qt documentation](https://doc.qt.io/qt-6/wasm.html).
+For Qt 6.5 or later, use the `wasm_singlethread` toolchain. For versions earlier than 6.5 use `wasm_32`.
 
-Make sure you have a version of Qt for WebAssembly that will work and clone the `emsdk` repository if you do not already have it.
+The `wasm_multithread` toolchain available in 6.5 and later is currently not supported. For more information, see the [Known Issues](#known-issues) section at the bottom of this page.
+
+Info about other Qt and emscripten versions can be found in the [Qt documentation](https://doc.qt.io/qt-6/wasm.html).
+
+Make sure you have a version of Qt for WebAssembly that will work, and clone the `emsdk` repository if you do not already have it.
 
 ## Setting Up `emsdk`
 
@@ -94,6 +98,8 @@ Example|Working
 `qml_features`|✅ working
 `qml_minimal`|✅ working
 
+For more information, see the [Known Issues](#known-issues) section at the bottom of this page.
+
 ## Compiling CXX-Qt Projects for WebAssembly
 
 When compiling a CXX-Qt project for wasm, the Rust target must be set to `wasm32-unknown-emscripten`, and the project must be configured to use POSIX threads.
@@ -114,14 +120,35 @@ else()
 endif()
 ```
 
-## Issues
+## Known Issues
 
-- CXX-Qt will currently not build with `wasm_multithread` versions of Qt:
+### `wasm_multithread` toolchain
 
-    ```console
-    wasm-ld: error: --shared-memory is disallowed by qml_minimal-e6f36338b0e1fa5c.17g6vcid2nczsjj0.rcgu.o 
-                because it was not compiled with 'atomics' or 'bulk-memory' features.
-    ```
+CXX-Qt will currently not build with `wasm_multithread` versions of Qt.
 
-- The example `qml-minimal-no-cmake` will not build for WebAssembly with Cargo
-- The example `demo_threading` will not build for WebAssembly
+```console
+wasm-ld: error: --shared-memory is disallowed by qml_minimal-e6f36338b0e1fa5c.17g6vcid2nczsjj0.rcgu.o 
+            because it was not compiled with 'atomics' or 'bulk-memory' features.
+```
+
+This issue is related to `pthread` in the `libc` crate. It is possible that manually compiling `cxx` and `libc` crates with `-pthread` may solve this.
+
+### `cargo`-only builds
+
+The example `qml-minimal-no-cmake` will not build for WebAssembly with `cargo`, and attempts to build with `cargo` without `cmake` will not work. This is due to an upstream issue with the `libc` crate, which does not support wasm and can cause breakage.
+
+```console
+cannot find function `pthread_kill` in crate `libc`
+```
+
+### `demo_threading` example
+
+The example `demo_threading` will not build for WebAssembly due to an upstream issue with `async-std`, which does not support wasm. On Linux, the observed breakage is due to `socket2` using its `unix.rs` file to target a Unix rather than wasm environment, resulting in error messages from `unix.rs` like the following:
+
+```console
+error[E0433]: failed to resolve: use of undeclared type `IovLen`
+```
+
+`socket2` is a dependency of `async-io`, which is a dependency of `async-std`.
+
+There is discussion around supporting wasm in the GitHub repository for `async-std`, and the progress is being tracked [here](https://github.com/async-rs/async-std/issues/220).
--- a/book/src/getting-started/6-wasm-builds.md
+++ b/book/src/getting-started/6-wasm-builds.md
@@ -9,11 +9,7 @@ SPDX-License-Identifier: MIT OR Apache-2.0
 
 CXX-Qt and applications written with it can be compiled for WebAssembly, with a few limitations. Below you will find detailed instructions regarding how to build for the WASM target.
 
-## Additional Requirements
-
 You will need to have Qt for WebAssembly installed. The next section shows versions that have been tested.
-
-Make sure you have the Emscripten target for `rustc` with `rustup target add wasm-unknown-emscripten`.
 
 Additionally, if you haven't already, clone the `emsdk` git repo from [here](https://github.com/emscripten-core/emsdk).
 
@@ -31,13 +27,7 @@ Qt|Emscripten
 6.5|3.1.25
 6.6|3.1.37
 
-For Qt 6.5 or later, use the `wasm_singlethread` toolchain. For versions earlier than 6.5 use `wasm_32`.
-
-The `wasm_multithread` toolchain available in 6.5 and later is currently not supported. For more information, see the [Known Issues](#known-issues) section at the bottom of this page.
-
 Info about other Qt and emscripten versions can be found in the [Qt documentation](https://doc.qt.io/qt-6/wasm.html).
-
-Make sure you have a version of Qt for WebAssembly that will work, and clone the `emsdk` repository if you do not already have it.
 
 ## Setting Up `emsdk`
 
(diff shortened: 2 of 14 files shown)


mattkdab:
Note, WASM builds are still requiring two ninja runs -- Not 100% sure why this is

CMakeLists.txt
```diff
@@ -171,39 +173,49 @@ function(add_valgrind_test NAME_WITH_PREFIX BINARY WORKING_DIRECTORY)
     endif()
 endfunction()
 
-get_target_property(QMAKE Qt::qmake IMPORTED_LOCATION)
-set(CARGO_ENV "QMAKE=set:${QMAKE}")
-set(RUNTIME_ENV "")
```
ahayzen-kdab, line 176: why are these still not set all the time?

This shouldn't "break" wasm builds by setting the QMAKE path ? As we are still seeing it use the wrong one anyway aren't we ?

CMakeLists.txt
```diff
@@ -171,39 +173,49 @@ function(add_valgrind_test NAME_WITH_PREFIX BINARY WORKING_DIRECTORY)
...
-    list(
-        APPEND
-        RUNTIME_ENV
-        "PATH=path_list_append:${QT_INSTALL_BINS}"
-        "QT_PLUGIN_PATH=path_list_append:${QT_INSTALL_PLUGINS}"
-        "QML_IMPORT_PATH=path_list_append:${QT_INSTALL_QML}"
-        "QML2_IMPORT_PATH=path_list_append:${QT_INSTALL_QML}"
-    )
-    list(APPEND CARGO_ENV ${RUNTIME_ENV})
+if(BUILD_WASM)
```
ahayzen-kdab, line 176: did `if(EMSCRIPTEN)` not work?

CMakeLists.txt
```diff
@@ -171,39 +173,49 @@ function(add_valgrind_test NAME_WITH_PREFIX BINARY WORKING_DIRECTORY)
...
+    set(Rust_CARGO_TARGET wasm32-unknown-emscripten)
+
+    # QMAKE environment variable is needed by qt-build-utils to ensure that Cargo
+    # uses the same installation of Qt as CMake does.
+    if(NOT USE_QT5)
+        find_package(Qt6 COMPONENTS Core Gui Test)
+    endif()
```
ahayzen-kdab, line 183: Why do you need this code? This already exists at the top for all options

CMakeLists.txt
```diff
@@ -249,4 +261,7 @@ set_tests_properties(cargo_build_rerun PROPERTIES RUN_SERIAL TRUE)
 
 add_subdirectory(book)
 add_subdirectory(examples)
-add_subdirectory(tests)
+
+if(NOT BUILD_WASM)
```
ahayzen-kdab, line 265: i wonder if a `BUILD_TESTS` is better, as then people can skip building tests when building non-wasm builds?

CMakeLists.txt
```diff
@@ -9,6 +9,8 @@ cmake_minimum_required(VERSION 3.24)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
+option(BUILD_WASM "Build CXX-Qt for WASM" OFF)
```
ahayzen-kdab, line 12: probably change to `BUILD_TESTS` and `ON` by default

examples/CMakeLists.txt
```diff
@@ -4,9 +4,9 @@
...
-# This causes duplicate builds, as this crate is already entirely built
-# When using `cargo test`
-add_subdirectory(qml_features)
 add_subdirectory(qml_minimal)
-add_subdirectory(demo_threading)
+
+if(NOT BUILD_WASM)
+    add_subdirectory(qml_features)
+    add_subdirectory(demo_threading)
+endif()
```
ahayzen-kdab, line 12: ideally we should be building all examples in wasm to ensure that all our features build :-)

examples/qml_minimal/CMakeLists.txt
```diff
@@ -22,125 +22,208 @@ set(CMAKE_AUTOMOC ON)
...
-
-# ANCHOR: book_cmake_find_corrosion
-find_package(Corrosion QUIET)
-if(NOT Corrosion_FOUND)
-    include(FetchContent)
-    FetchContent_Declare(
-        Corrosion
-        GIT_REPOSITORY https://github.com/corrosion-rs/corrosion.git
-        GIT_TAG v0.4.5
+if(BUILD_WASM)
```
ahayzen-kdab, line 25: instead of putting this all behind an option, maybe instead try doing the following to make QmlImportScanner optional

```cmake
if(NOT USE_QT5)
    find_package(Qt6 COMPONENTS Core Gui Qml QuickControls2 OPTIONAL_COMPONENTS QmlImportScanner)
endif()
if(NOT Qt6_FOUND)
    find_package(Qt5 5.15 REQUIRED COMPONENTS Core Gui Qml QuickControls2 OPTIONAL_COMPONENTS QmlImportScanner)
endif()
```

CMakeLists.txt
```diff
@@ -171,39 +173,49 @@ function(add_valgrind_test NAME_WITH_PREFIX BINARY WORKING_DIRECTORY)
...
-        APPEND
-        RUNTIME_ENV
-        "PATH=path_list_append:${QT_INSTALL_BINS}"
-        "QT_PLUGIN_PATH=path_list_append:${QT_INSTALL_PLUGINS}"
-        "QML_IMPORT_PATH=path_list_append:${QT_INSTALL_QML}"
-        "QML2_IMPORT_PATH=path_list_append:${QT_INSTALL_QML}"
-    )
-    list(APPEND CARGO_ENV ${RUNTIME_ENV})
+if(BUILD_WASM)
+    set(Rust_CARGO_TARGET wasm32-unknown-emscripten)
```
ahayzen-kdab, line 177: can we try setting this on the CLI, as note some people try to build the examples standalone, eg they run CMake from the examples folder. So we either need to have this in all the examples or instead maybe it's better in the page in the book about WASM to state they need to run with the following or similar?

```
cmake -DRust_CARGO_TARGET=wasm32-unknown-emscripten
```

crates/cxx-qt-lib/src/assertion_utils.h
```diff
@@ -7,7 +7,11 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 #pragma once
 
+#ifdef CXX_QT_WASM
+#define assert_alignment_and_size(TYPE, ALIGNMENT, SIZE)
```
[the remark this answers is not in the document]
ahayzen-kdab: also until we figure out what is going on, can you add a code comment stating the craziness :-)

examples/qml_minimal/CMakeLists.txt
```diff
@@ -22,125 +22,208 @@ set(CMAKE_AUTOMOC ON)
...
+        Qt::Gui
+        Qt::Qml
+        Qt::QuickControls2
+    )
+    # ANCHOR_END: book_cmake_use_corrosion
 
-#
-# Unit test
-#
+    target_link_options(${APP_NAME}_lib INTERFACE "SHELL:-Wl,--whole-archive $<TARGET_FILE:${CRATE}-static> -Wl,--no-whole-archive")
```
ahayzen-kdab, line 94: generally instead of mass if blocks, we should try not to change this code too much as it's used in the books etc as well.

Instead can we have a minimal diff and have tight if blocks eg

```cmake
# WHOLE_ARCHIVE is needed for the generated QML plugin to register on startup,
# otherwise the linker will discard the static variables that initialize it.
if(EMSCRIPTEN)
    target_link_libraries(${APP_NAME}_lib INTERFACE
        Qt::Core
        Qt::Gui
        Qt::Qml
        Qt::QuickControls2
    )
    target_link_options(${APP_NAME}_lib INTERFACE "SHELL:-Wl,--whole-archive $<TARGET_FILE:${CRATE}-static> -Wl,--no-whole-archive")
else()
    target_link_libraries(${APP_NAME}_lib INTERFACE
       "$<LINK_LIBRARY:WHOLE_ARCHIVE,${CRATE}-static>"
       Qt::Core
       Qt::Gui
       Qt::Qml
       Qt::QuickControls2
    )
endif()
```

Then it's much easier to see where the differences are for wasm.

examples/qml_minimal/CMakeLists.txt
```diff
@@ -22,125 +22,208 @@ set(CMAKE_AUTOMOC ON)
...
+        if (COMMAND add_valgrind_test)
+            add_valgrind_test(
+                ${APP_TEST_NAME} "${TEST_CMD}" ${CMAKE_CURRENT_BINARY_DIR}
+            )
+        else()
+            MESSAGE(STATUS "add_valgrind_test is defined in the top level of CXX-Qt. It will not executed")
+        endif()
+    endfunction()
+
+    add_qml_test(myobject)
```
ahayzen-kdab, line 228: then all of the test stuff could go behind a new `BUILD_TESTS` option, although wonder what happens if you try to build the example standalone and it has a `if(BUILD_TESTS)` but that option is defined in the root CMake :thinking:

CMakeLists.txt
mattkdab: you get that if you use `emcmake cmake` or `emconfigure cmake`

CMakeLists.txt
mattkdab: oops

CMakeLists.txt
mattkdab: we'll probably still need BUILD_WASM to specify wasm-only stuff, right?

maybe have

```cmake
if(NOT BUILD_WASM)
    set(BUILD_TESTS OFF)
endif()
```

?

ahayzen-kdab: I guess unless there is another way to tell we are under wasm/emscripten from CMake ?

examples/CMakeLists.txt
mattkdab: agree, wanted to start with getting at least one working for this patch, but maybe i should get all?

CMakeLists.txt
```diff
@@ -121,12 +122,25 @@ add_compile_definitions(
...
+endif()
+
 # QMAKE environment variable is needed by qt-build-utils to ensure that Cargo
 # uses the same installation of Qt as CMake does.
 if(NOT USE_QT5)
     find_package(Qt6 COMPONENTS Core Gui Test)
 endif()
 if(NOT Qt6_FOUND)
+    if(BUILD_WASM)
+        message("
```
ahayzen-kdab, line 137: can we instead just fail the build ? eg have a `message(FATAL_ERROR "...")` ?

examples/CMakeLists.txt
```diff
@@ -4,9 +4,9 @@
 #
 # SPDX-License-Identifier: MIT OR Apache-2.0
 
-# No longer add cargo-without-cmake here.
-# This causes duplicate builds, as this crate is already entirely built
-# When using `cargo test`
```
ahayzen-kdab, line 9: can you leave this comment here :-)

examples/qml_minimal/CMakeLists.txt
```diff
@@ -78,20 +82,37 @@ add_library(${APP_NAME}_lib INTERFACE)
...
+        Qt::QuickControls2
+    )
+endif()
 # ANCHOR_END: book_cmake_use_corrosion
 
 # ANCHOR: book_cmake_executable
 # Define the executable with the C++ source
-add_executable(${APP_NAME} cpp/main.cpp)
+if(BUILD_WASM)
+    qt_add_executable(${APP_NAME} cpp/main.cpp)
```
ahayzen-kdab, line 112: assume we have to use `qt_add_executable` ? shame it was added in Qt 6 as otherwise we could just use it even for non-wasm builds. Hmmm

ahayzen-kdab: Appears if you build without the `qt_add_executable` you get the `.js` and `.wasm` files but are missing the `.html` file ?

Maybe we can look into how Qt 5.x WASM was working without the newer CMake commands

examples/qml_minimal/CMakeLists.txt
```diff
@@ -78,20 +82,37 @@ add_library(${APP_NAME}_lib INTERFACE)
...
-)
+if(BUILD_WASM)
+    target_link_libraries(${APP_NAME}_lib INTERFACE
+        Qt::Core
+        Qt::Gui
+        Qt::Qml
+        Qt::QuickControls2
+    )
+    
+    target_link_options(${APP_NAME}_lib INTERFACE
```
ahayzen-kdab, line 93: vscode seems to think there is an extra space at the end of this line and potentially in the line before too

examples/qml_minimal/CMakeLists.txt
```diff
@@ -78,20 +82,37 @@ add_library(${APP_NAME}_lib INTERFACE)
...
+    target_link_libraries(${APP_NAME}_lib INTERFACE
+        Qt::Core
+        Qt::Gui
+        Qt::Qml
+        Qt::QuickControls2
+    )
+    
+    target_link_options(${APP_NAME}_lib INTERFACE 
+        "SHELL:-Wl,--whole-archive $<TARGET_FILE:${CRATE}-static> -Wl,--no-whole-archive"
+    )
```
ahayzen-kdab, line 95: So because the library here is not in the `target_link_libraries` it isn't marked as a dependency of the target, so it isn't always built before the target reaches linking. There are a few options here

# Add dependency

Simplest is add a dependency until we figure out how to make `LINK_LIBRARY:WHOLE_ARCHIVE` work

```cmake
    # Ensure that the crate has been built to link against
    add_dependencies(${APP_NAME}_lib ${CRATE}-static)

    # FIXME: "$<LINK_LIBRARY:WHOLE_ARCHIVE,${CRATE}-static>" fails with the following
    #
    # Feature 'WHOLE_ARCHIVE', specified through generator-expression
    # '$<LINK_LIBRARY>' to link target 'example_qml_minimal', is not supported
    # for the 'CXX' link language.
    #
    # Does a newer version of Qt/emscripten solve this?
    target_link_options(${APP_NAME}_lib INTERFACE
        "SHELL:-Wl,--whole-archive $<TARGET_FILE:${CRATE}-static> -Wl,--no-whole-archive"
    )
```

# Add to target_link_libraries

We could potentially have the whole-archive args in the target_link_libraries.

```cmake
    target_link_libraries(${APP_NAME}_lib INTERFACE
        Qt::Core
        Qt::Gui
        Qt::Qml
        Qt::QuickControls2
        # WHOLE_ARCHIVE is needed for the generated QML plugin to register on startup,
        # otherwise the linker will discard the static variables that initialize it.
        -Wl,--whole-archive
        $<TARGET_PROPERTY:${CRATE}-static,INTERFACE_LINK_LIBRARIES>
        -Wl,--no-whole-archive
    )
```

Note that this would likely also work for non-WASM builds but is not preferred and `"$<LINK_LIBRARY:WHOLE_ARCHIVE,${CRATE}-static>"` should be used as that uses generators etc, we should try to figure out why this doesn't work with wasm first :thinking: 

# Mixture of link library and options

It seems possible also to specify the library and a link library then override the link options after.

Again seems better to figure out why the cmake generator route doesn't work.

```cmake
    target_link_libraries(${APP_NAME}_lib INTERFACE
        Qt::Core
        Qt::Gui
        Qt::Qml
        Qt::QuickControls2
        ${CRATE}-static
    )

    # WHOLE_ARCHIVE is needed for the generated QML plugin to register on startup,
    # otherwise the linker will discard the static variables that initialize it.
    target_link_options(${APP_NAME}_lib INTERFACE
        "SHELL:-Wl,--whole-archive $<TARGET_FILE:${CRATE}-static> -Wl,--no-whole-archive"
    )
```

examples/qml_minimal/CMakeLists.txt
```diff
@@ -78,20 +82,37 @@ add_library(${APP_NAME}_lib INTERFACE)
...
+    target_link_libraries(${APP_NAME}_lib INTERFACE
+        # WHOLE_ARCHIVE is needed for the generated QML plugin to register on startup,
+        # otherwise the linker will discard the static variables that initialize it.
+        "$<LINK_LIBRARY:WHOLE_ARCHIVE,${CRATE}-static>"
+        Qt::Core
+        Qt::Gui
+        Qt::Qml
+        Qt::QuickControls2
+    )
+endif()
```
ahayzen-kdab, line 106: So found out how to tell CMake how to link when it is in an unfamiliar build env (originally from https://discourse.cmake.org/t/error-when-crosscompiling-with-whole-archive-target-link/9394/3 ).

This results in a much simpler code as we can keep the original `target_link_libraries` and have a short difference for `BUILD_WASM`.

Note that you could probably put this if block higher up in the example cmake files so that it isn't shown in the general book pages and then have a separate book anchor for the wasm pages to show how you need to specify this when using wasm ?

Or maybe it even just goes globally in the top CMake file then you document in the book that you need to add those two lines for wasm builds by using a book anchor ?

Note this also solves the double build issue as the dependency is correctly done by `target_link_libraries` now

```cmake
# When building under emscripten tell CMake how to link using whole archive
if(BUILD_WASM)
    set(CMAKE_LINK_LIBRARY_USING_WHOLE_ARCHIVE "-Wl,--whole-archive" "<LINK_ITEM>" "-Wl,--no-whole-archive")
    set(CMAKE_LINK_LIBRARY_USING_WHOLE_ARCHIVE_SUPPORTED True)
endif()

target_link_libraries(${APP_NAME}_lib INTERFACE
    # WHOLE_ARCHIVE is needed for the generated QML plugin to register on startup,
    # otherwise the linker will discard the static variables that initialize it.
    "$<LINK_LIBRARY:WHOLE_ARCHIVE,${CRATE}-static>"
    Qt::Core
    Qt::Gui
    Qt::Qml
    Qt::QuickControls2
)
```

examples/qml_minimal/CMakeLists.txt
ahayzen-kdab: See other comment below about using `CMAKE_LINK_LIBRARY_USING_WHOLE_ARCHIVE` instead, this is now not relevant.

CMakeLists.txt
```diff
@@ -121,12 +122,25 @@ add_compile_definitions(
     QT_USE_QSTRINGBUILDER
 )
 
+if(BUILD_WASM)
+    set(BUILD_TESTING OFF)
+    set(Rust_CARGO_TARGET wasm32-unknown-emscripten)
```
ahayzen-kdab, line 127: If we could figure out the `qt_add_executable` different then maybe we just add the `CMAKE_LINK_LIBRARY_USING_WHOLE_ARCHIVE` here as then there would be no need for any `BUILD_WASM` in the examples?

Depends if we want to allow folks to build the examples standalone for wasm too i guess ? Or whether we just document all these difference in a book page?

CMakeLists.txt
mattkdab: sure

examples/CMakeLists.txt
mattkdab: huh i didnt even notice this got deleted

examples/qml_minimal/CMakeLists.txt
mattkdab: There is no Qt 5 wasm iiuc
and yea u dont get the html file, maybe there is a way around this idk

examples/qml_minimal/CMakeLists.txt
```diff
@@ -78,6 +88,7 @@ add_library(${APP_NAME}_lib INTERFACE)
 # overwriting another's files.
 target_include_directories(${APP_NAME}_lib INTERFACE "${CXXQT_EXPORT_DIR}/${CRATE}")
 
+
```
ahayzen-kdab, line 91: random new line, and another further down

examples/qml_minimal/CMakeLists.txt
```diff
@@ -36,6 +42,10 @@ endif()
 get_target_property(QMAKE Qt::qmake IMPORTED_LOCATION)
 # ANCHOR_END: book_cmake_find_qmake
 
+if(BUILD_WASM)
+    set(Rust_CARGO_TARGET wasm32-unknown-emscripten)
```
ahayzen-kdab, line 46: guess this could be combined with the other BUILD_WASM block further up ?

eg have

```cmake
if(BUILD_WASM)
    # When building under emscripten tell CMake how to link using whole archive
    set(CMAKE_LINK_LIBRARY_USING_WHOLE_ARCHIVE "-Wl,--whole-archive" "<LINK_ITEM>" "-Wl,--no-whole-archive")
    set(CMAKE_LINK_LIBRARY_USING_WHOLE_ARCHIVE_SUPPORTED True)
    # Ensure Rust build for the correct target
    set(Rust_CARGO_TARGET wasm32-unknown-emscripten)
endif()
```

there is also the question over whether this should be global or in examples, for now lets also put it in the examples. So you'll need this same snippet in qml_features and demo_threading

examples/qml_minimal/CMakeLists.txt
```diff
@@ -91,7 +99,11 @@ target_link_libraries(${APP_NAME}_lib INTERFACE
 
 # ANCHOR: book_cmake_executable
 # Define the executable with the C++ source
-add_executable(${APP_NAME} cpp/main.cpp)
+if(BUILD_WASM)
+    qt_add_executable(${APP_NAME} cpp/main.cpp)
```
ahayzen-kdab, line 103: would be good to have a comment as to why we need this (otherwise the html file is not generated). And would still be good if we could figure out how to make this work with `add_executable` so that it's an additional line for WASM instead of having an if block :thinking:

CMakeLists.txt
```diff
@@ -121,12 +122,25 @@ add_compile_definitions(
...
 # uses the same installation of Qt as CMake does.
 if(NOT USE_QT5)
     find_package(Qt6 COMPONENTS Core Gui Test)
 endif()
 if(NOT Qt6_FOUND)
+    if(BUILD_WASM)
+        message(FATAL_ERROR 
+                "CXX-Qt for WebAssembly only currently supports Qt 6 builds."
+        )
+        set(BUILD_WASM OFF)
```
ahayzen-kdab, line 142: probably don't need to set BUILD_WASM OFF if the previous line was fatal?

examples/demo_threading/CMakeLists.txt
```diff
@@ -16,13 +16,21 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
...
+    set(CMAKE_LINK_LIBRARY_USING_WHOLE_ARCHIVE "-Wl,--whole-archive" "<LINK_ITEM>" "-Wl,--no-whole-archive")
+    set(CMAKE_LINK_LIBRARY_USING_WHOLE_ARCHIVE_SUPPORTED True)
+    # Ensure Rust build for the correct target
+    set(Rust_CARGO_TARGET wasm32-unknown-emscripten)
+endif()
+
 set(CMAKE_AUTOMOC ON)
 set(CMAKE_AUTORCC ON)
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
ahayzen-kdab, line 31: you need to do the add_executable change to this CMake too

mattkdab: done, but this example is still not working. socket2 is failing to build

CMakeLists.txt
```diff
@@ -121,13 +121,27 @@ add_compile_definitions(
     QT_USE_QSTRINGBUILDER
 )
 
+if(BUILD_WASM)
```
ahayzen-kdab, line 124: could add something to the CHANGELOG for this :-)

mattkdab: done :D

book/src/getting-started/6-wasm-builds.md
```diff
@@ -0,0 +1,154 @@
...
+Example|Working
+-|-
+`qml-minimal-no-cmake`|❌ broken
+`demo_threading`|❌ broken
+`qml_features`|✅ working
+`qml_minimal`|✅ working
+
+For more information, see the [Known Issues](#known-issues) section at the bottom of this page.
+
+## Compiling CXX-Qt Projects for WebAssembly
```
ahayzen-kdab, line 103: as discussed maybe this comes higher up and the bits about building the `cxx-qt` repository are later. Note that `-DCMAKE_TOOLCHAIN_FILE=/path/to/qt.toolchain.cmake` info should probably be added to this part so folks know how to build it in their project

ahayzen-kdab: (or the `qt-cmake ` alternative route)

ahayzen-kdab: Maybe that whole bit explaining the toolchain situation can be it's own section ?

So have these as separate sections?
- CMake toolchain fun
- Building your own project
- Building CXX-Qt repository / examples

PR #948, opened 2024-04-30
CMake: option to build tests

mattkdab:


CMakeLists.txt
```diff
@@ -219,34 +221,39 @@ endif()
...
+    add_test(NAME cxx_qt_gen_test_inputs_gen COMMAND rustfmt --check ${CXX_QT_GEN_TEST_INPUTS})
+    add_test(NAME cxx_qt_gen_test_outputs_gen COMMAND rustfmt --check ${CXX_QT_GEN_TEST_OUTPUTS})
+    
+    # Add test which checks that a build rerun doesn't recompile and uses caches instead
+    add_test(NAME cargo_build_rerun COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/scripts/check_cargo_build_rerun.sh" "${CMAKE_CURRENT_SOURCE_DIR}")
+    
+    # Ensure that cargo_build_rerun doesn't run while we are already building
+    set_tests_properties(cargo_build_rerun PROPERTIES RUN_SERIAL TRUE)
+    
+    add_subdirectory(tests)
```
ahayzen-kdab, line 255: think this add_subdirectory might need to be after the other add subdirs for it to work. As appears maybe the tests aren't importing corrosion but getting it from the other subdirs automatically?

CMakeLists.txt
```diff
@@ -13,6 +13,8 @@ option(USE_QT5 "Use Qt5 even if Qt6 found" OFF)
 option(VCPKG "Use vcpkg for release mode dependencies" OFF)
 option(VCPKG_BUILDTREES_ROOT "Use custom vcpkg buildtrees" "")
 
+option(CXX_QT_BUILD_TESTS "Build the tests for CXX-Qt" ON)
```
ahayzen-kdab, line 16: do we need the `CXX_QT` prefix ? Could it just be `BUILD_TESTS` as it's kinda common to have `BUILD_TESTS` and `BUILD_EXAMPLES` i thought?

mattkdab: i felt it was better to be more descriptive, but maybe it doesnt matter

ahayzen-kdab: As we currently have `USE_QT5` and `VCPKG` we probably don't need the prefix, as either they should all have it or they should all not it to be consistent

CMakeLists.txt
mattkdab: aha!

CMakeLists.txt
redstrate: If we go with a generic name, make sure it's `BUILD_TESTING` as seems to be more of the standard across CMake projects

ahayzen-kdab: Aha thanks, i knew i had seen something like that before :sweat_smile:

CMakeLists.txt
```diff
@@ -219,34 +221,39 @@ endif()
...
+    file(GLOB CXX_QT_GEN_TEST_INPUTS ${CMAKE_CURRENT_SOURCE_DIR}/crates/cxx-qt-gen/test_inputs/*.rs)
+    file(GLOB CXX_QT_GEN_TEST_OUTPUTS ${CMAKE_CURRENT_SOURCE_DIR}/crates/cxx-qt-gen/test_outputs/*.rs)
+    add_test(NAME cxx_qt_gen_test_inputs_gen COMMAND rustfmt --check ${CXX_QT_GEN_TEST_INPUTS})
+    add_test(NAME cxx_qt_gen_test_outputs_gen COMMAND rustfmt --check ${CXX_QT_GEN_TEST_OUTPUTS})
+    
+    # Add test which checks that a build rerun doesn't recompile and uses caches instead
+    add_test(NAME cargo_build_rerun COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/scripts/check_cargo_build_rerun.sh" "${CMAKE_CURRENT_SOURCE_DIR}")
+    
+    # Ensure that cargo_build_rerun doesn't run while we are already building
+    set_tests_properties(cargo_build_rerun PROPERTIES RUN_SERIAL TRUE)
```
ahayzen-kdab, line 256: can we add these to the preview if block as they are linked to the ` add_test(NAME cargo_clippy` etc some are modifying the env for those and the rest are adding more cargo related tests

Then just leave the add_subdir of tests for this block

examples/qml_minimal/CMakeLists.txt
```diff
@@ -100,47 +100,49 @@ target_link_libraries(${APP_NAME} PRIVATE ${APP_NAME}_lib)
...
+            ${APP_NAME}_${TEST_NAME}_test PROPERTIES ENVIRONMENT_MODIFICATION
+                                                    "${RUNTIME_ENV}"
+        )
+        if (COMMAND add_valgrind_test)
+        add_valgrind_test(
+            ${APP_TEST_NAME} "${TEST_CMD}" ${CMAKE_CURRENT_BINARY_DIR}
+        )
+        else()
+        MESSAGE(STATUS "add_valgrind_test is defined in the top level of CXX-Qt. It will not executed")
+        endif()
```
ahayzen-kdab, line 144: looks like the indenting was lost here for the contents of the if branches

mattkdab: oop

CMakeLists.txt
ahayzen-kdab: i meant the preview `if(BUILD_TESTING)` block :-)  so around line 228

mattkdab: oh my

ahayzen-kdab (approved):
LGTM :-)

PR #950, opened 2024-05-08
build: for Rust 1.80 we need to specify rustc-check-cfg

ahayzen-kdab:
Otherwise you get build warnings when using rustc-cfg attributes.

https://blog.rust-lang.org/2024/05/06/check-cfg.html

### DIFF
--- a/crates/cxx-qt-build/src/lib.rs
+++ b/crates/cxx-qt-build/src/lib.rs
@@ -482,6 +482,20 @@ impl CxxQtBuilder {
         env::set_var(variable_cargo, value.unwrap_or("true"));
     }
 
+    fn define_cfg_check_variable(key: String, values: Option<Vec<&str>>) {
+        if let Some(values) = values {
+            let values = values
+                .iter()
+                // Escape and add quotes
+                .map(|value| format!("\"{}\"", value.escape_default()))
+                .collect::<Vec<_>>()
+                .join(", ");
+            println!("cargo:rustc-check-cfg=cfg({key}, values({values}))");
+        } else {
+            println!("cargo:rustc-check-cfg=cfg({key})");
+        }
+    }
+
     /// Generate and compile cxx-qt C++ code, as well as compile any additional files from
     /// [CxxQtBuilder::qobject_header] and [CxxQtBuilder::cc_builder].
     pub fn build(mut self) {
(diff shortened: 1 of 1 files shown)


ahayzen-kdab:
This could be backported to the 0.6.x series too.

crates/cxx-qt-build/src/lib.rs
```diff
@@ -476,6 +476,7 @@ impl CxxQtBuilder {
         if let Some(value) = value {
             println!("cargo:rustc-cfg={key}=\"{value}\"");
         } else {
+            println!("cargo:rustc-check-cfg=cfg({key})");
```
LeonMatthesKDAB, line 479: I don't think this is quite the right spot?
This will only be emitted when the config variable is actually defined.
But we need to emit check-cfg commands for every variable we **may** define.

This is quite annoying to implement though, so I'm not sure where the best place to put it would be...

crates/cxx-qt-build/src/lib.rs
```diff
@@ -495,13 +496,24 @@ impl CxxQtBuilder {
...
         // Find the Qt version and tell the Rust compiler
         // this allows us to have conditional Rust code
         CxxQtBuilder::define_cfg_variable(
             "cxxqt_qt_version_major".to_string(),
             Some(qtbuild.version().major.to_string().as_str()),
         );
 
+        // Seed all values from Qt 5.0 through to Qt 7.99
+        for major in 5..=7 {
+            println!("cargo:rustc-check-cfg=cfg(cxxqt_qt_version_at_least_{major})");
```
LeonMatthesKDAB, line 510: Maybe put all the rustc-check-cfg calls into a single function, so that we have a central place that defines all possible cfg flags?

crates/cxx-qt-build/src/lib.rs
LeonMatthesKDAB: With your current implementation I don't think we need this here at all, right?
You're already declaring all possible versions of qt_version_at_least and _major.

LeonMatthesKDAB (changes requested):

crates/cxx-qt-build/src/lib.rs
ahayzen-kdab: Oops that was an old version, that line should be removed

ahayzen-kdab: Fixed :-)

crates/cxx-qt-build/src/lib.rs
ahayzen-kdab: Attempted to make a method which escapes values etc

LeonMatthesKDAB (approved):

PR #952, opened 2024-05-13
refactor: Use Name for C++ QObject generation

LeonMatthesKDAB:
Part of #939

### DIFF
--- a/crates/cxx-qt-gen/src/naming/type_names.rs
+++ b/crates/cxx-qt-gen/src/naming/type_names.rs
@@ -11,9 +11,12 @@ use syn::{
     Path, Result,
 };
 
-use crate::syntax::{
-    attribute::attribute_find_path, expr::expr_to_string,
-    foreignmod::foreign_mod_to_foreign_item_types,
+use crate::{
+    parser::qobject::ParsedQObject,
+    syntax::{
+        attribute::attribute_find_path, expr::expr_to_string,
+        foreignmod::foreign_mod_to_foreign_item_types,
+    },
 };
 
 use super::Name;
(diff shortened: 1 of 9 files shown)


crates/cxx-qt-gen/src/naming/type_names.rs
```diff
@@ -139,14 +142,25 @@ impl TypeNames {
         Ok(())
     }
 
+    fn populate_qobject(&mut self, qobject: &ParsedQObject) -> Result<()> {
```
ahayzen-kdab, line 145: this makes the naming dependent on the parser which is slightly unfortunate for making it reusable later. Maybe later we could have an intermediate struct this is populated from ?

crates/cxx-qt-gen/src/naming/type_names.rs
```diff
@@ -268,7 +282,7 @@ impl TypeNames {
         Error::new_spanned(ident, format!("Undeclared type: `{ident}`!"))
     }
 
-    fn lookup(&self, ident: &Ident) -> Result<&Name> {
+    pub fn lookup(&self, ident: &Ident) -> Result<&Name> {
```
ahayzen-kdab, line 285: add docs if it's pub

crates/cxx-qt-gen/src/writer/cpp/mod.rs
```diff
@@ -44,24 +44,36 @@ mod tests {
...
-                    namespace: "cxx_qt::my_object".to_owned(),
-                    namespace_internals: "cxx_qt::my_object::cxx_qt_my_object".to_owned(),
+                    name: if let Some(namespace) = namespace {
+                        Name::mock_namespaced("MyObject", namespace)
+                    } else {
+                        Name::mock("MyObject")
+                    },
+                    rust_struct: Name::mock("MyObjectRust"),
+                    namespace_internals: if let Some(namespace) = namespace {
+                        namespace.to_owned() + "::cxx_qt_my_object"
```
ahayzen-kdab, line 73: is `format!("{namespace}::cxx_qt_my_object")` slightly more readable?

crates/cxx-qt-gen/src/naming/type_names.rs
LeonMatthesKDAB: done

ahayzen-kdab (approved):
LGTM i think :-)

crates/cxx-qt-gen/src/naming/type_names.rs
LeonMatthesKDAB: Right, that's true, but also already the case with the previous setup.

When we move this module out into its own crate, these functions will need to remain inside cxx-qt-gen, potentially as a free function.
They are the "naming" phase, so need to know how to extract the type names out of the parser data.

PR #957, opened 2024-05-21
fix: CI: Force use of Rust 1.77

LeonMatthesKDAB:
This may fix our CI issues on Windows, as they seem to have appeared with the 1.78 update


ahayzen-kdab (approved):
For now we go with this, we could limit it to just Windows using 1.77. But lets see if we can figure out the issue first...

PR #963, opened 2024-05-23
Only use whole-archive on the `-initalizers` library that contains QML plugins/resources

### LINKED ISSUE #958: MSVC CI doesn't compile with Rust 1.78.0
With Rust 1.78.0, we seem to run into an issue with duplicate symbol definitions of `__NULL_IMPORT_DESCRIPTOR`.

See:
https://github.com/KDAB/cxx-qt/actions/runs/9167401332/job/25204499375

This can be worked around by reverting to 1.77 for our CI (see #957 ), but we still need to fix this so that we remain compatible with upstream Rust.

LeonMatthesKDAB:
Closes #958


ahayzen-kdab:
Separately i'll explore having an init() method

LeonMatthesKDAB:
Closing in favor of #978

PR #964, opened 2024-05-23
WIP: cxx-qt-build: use init method rather than WHOLE_ARCHIVE for CMake

ahayzen-kdab:


crates/cxx-qt-build/src/lib.rs
```diff
@@ -496,6 +497,112 @@ impl CxxQtBuilder {
...
+                void
+                importPlugins()
+                {{
+                {import_plugins}
+                }}
+                "#
+            };
+        }
+
+        // TODO: other init, eg cxx-qt-lib
```
ahayzen-kdab, line 569: need to pass through cxx-qt-lib

crates/cxx-qt-build/src/lib.rs
```diff
@@ -496,6 +497,112 @@ impl CxxQtBuilder {
...
+        // Build the QML files
+        let mut import_plugins = self
+            .qml_modules
+            .iter()
+            .map(|qml_module| {
+                // The name is the URI with dots replaced by underscores
+                // unless the CLASS_NAME is specified (but we do not support this yet)
+                // https://doc.qt.io/qt-6/qqmlengineextensionplugin.html#Q_IMPORT_QML_PLUGIN
+                let plugin_name = qml_module.uri.replace(".", "_");
+                format!("  Q_IMPORT_PLUGIN({plugin_name});")
```
ahayzen-kdab, line 550: fails with this at the moment for qml_minimal

```
[build] /usr/bin/ld: examples/qml_minimal/libqml_minimal.a(a1053b249b42b775-init.o): in function `importPlugins()::Staticcom_kdab_cxx_qt_demoPluginInstance::Staticcom_kdab_cxx_qt_demoPluginInstance()':
[build] /var/home/andrew/Projects/cxx-qt/build/examples/qml_minimal/cxxqt/qml_minimal/cxx-qt/init.cpp:7: undefined reference to `qt_static_plugin_com_kdab_cxx_qt_demo()'
[build] collect2: error: ld returned 1 exit status
```

examples/qml_minimal/cpp/main.cpp
```diff
@@ -10,9 +10,13 @@
 #include <QtGui/QGuiApplication>
 #include <QtQml/QQmlApplicationEngine>
 
+#include <cxx-qt/init.h>
+
 int
 main(int argc, char* argv[])
 {
+  cxx_qt::init();
```
ahayzen-kdab, line 18: we need to document this in the book and changelog

crates/cxx-qt-build/src/lib.rs
```diff
@@ -496,6 +505,141 @@ impl CxxQtBuilder {
...
+            .qrc_files
+            .iter()
+            .filter_map(|qrc| qrc.file_stem().and_then(|file_stem| file_stem.to_str()))
+            .map(|file_stem| {
+                // The name is the base name of the file with characters that cannot
+                // be part of a C++ function replaced by underscores
+                // https://doc.qt.io/qt-6/qtresource-qtcore-proxy.html#Q_INIT_RESOURCE
+                let qrc_name = file_stem.replace('-', "_");
+                // Note this cannot be inside a namespace
+                format!("  Q_INIT_RESOURCE({qrc_name};")
```
ahayzen-kdab, line 520: need to check this works with qrc files as we don't have any in our tests?

crates/cxx-qt-build/src/lib.rs
```diff
@@ -623,6 +767,20 @@ impl CxxQtBuilder {
...
+        if qtbuild.version().major < 6 {
+            // Inject Qt 5 compat type registers
+            self.inits.push(CxxQtBuilderInit::new(
+                "cxx_qt_qt5_compat",
+                include_str!("std_types_qt5.cpp"),
+            ));
+        }
+
+        // Generate init method
+        cc_builder_whole_archive.file(self.write_init_method());
```
ahayzen-kdab, line 782: appears we still need the `+whole-archive` internally though? Unless is there a way of removing this too :thinking:

ahayzen-kdab: it might only be Cargo builds that now need it? As seems they can't tell that the extern's from Rust are linking through to the init method.

crates/cxx-qt-build/src/lib.rs
```diff
@@ -496,6 +505,139 @@ impl CxxQtBuilder {
...
+            init();
+            }}
+            "#
+        };
+        let source = formatdoc! {
+            r#"
+            #include <cxx-qt/init.h>
+
+            void
+            __cxx_qt_init()
```
ahayzen-kdab, line 612: When building with `cargo test` we get multiple definitions of some things.

Note we do not see the QML plugins though

```
  = note: /usr/bin/ld.gold: error: /tmp/rustchKpRbk/libqt-static-initializers.a(3e4a4eaacdc769a0-init.o): multiple definition of 'cxx_qt_qt5_compat()'
          /usr/bin/ld.gold: /home/runner/cxx-qt/build/./cargo/build/release/build/qml_minimal-e896cbf14dffc047/out/libqt-static-initializers.a(4c23f01c502e2d66-init.o): previous definition here
          /usr/bin/ld.gold: error: /tmp/rustchKpRbk/libqt-static-initializers.a(3e4a4eaacdc769a0-init.o): multiple definition of 'cxx_qt_lib_core_init()'
          /usr/bin/ld.gold: /home/runner/cxx-qt/build/./cargo/build/release/build/qml_minimal-e896cbf14dffc047/out/libqt-static-initializers.a(4c23f01c502e2d66-init.o): previous definition here
          /usr/bin/ld.gold: error: /tmp/rustchKpRbk/libqt-static-initializers.a(3e4a4eaacdc769a0-init.o): multiple definition of 'cxx_qt_lib_gui_init()'
          /usr/bin/ld.gold: /home/runner/cxx-qt/build/./cargo/build/release/build/qml_minimal-e896cbf14dffc047/out/libqt-static-initializers.a(4c23f01c502e2d66-init.o): previous definition here
          /usr/bin/ld.gold: error: /tmp/rustchKpRbk/libqt-static-initializers.a(3e4a4eaacdc769a0-init.o): multiple definition of 'cxx_qt::init()'
          /usr/bin/ld.gold: /home/runner/cxx-qt/build/./cargo/build/release/build/qml_minimal-e896cbf14dffc047/out/libqt-static-initializers.a(4c23f01c502e2d66-init.o): previous definition here
          /usr/bin/ld.gold: error: /tmp/rustchKpRbk/libqt-static-initializers.a(3e4a4eaacdc769a0-init.o): multiple definition of '__cxx_qt_init'
          /usr/bin/ld.gold: /home/runner/cxx-qt/build/./cargo/build/release/build/qml_minimal-e896cbf14dffc047/out/libqt-static-initializers.a(4c23f01c502e2d66-init.o): previous definition here
          collect2: error: ld returned 1 exit status
```

ahayzen-kdab: This could be due to them being in both cxx-qt-lib and the target like qml-minimal as they both build using CxxQtBuilder?

crates/cxx-qt-build/src/lib.rs
LeonMatthesKDAB: I think so, but this also means we don't need the `init` method at all in our cargo-only builds, right?
Initialization can just work "as-is" in that case, as we don't have a problem there iirc.

ahayzen-kdab: Yes it means we don't need the `init` currently in Cargo, although it'd be nice to remove this whole-archive and have the linker figure things out.

However i wonder if we still document that you should use it, as later it gives us a place to register/setup anything globally. Eg what if we wanted to setup Qt logging or something?

ahayzen-kdab:
The build system has changed instead as described in #983

PR #968, opened 2024-06-04
PropertNames: Use Name instead of CombinedIdent

LeonMatthesKDAB:


crates/cxx-qt-gen/src/generator/cpp/property/signal.rs
```diff
@@ -3,27 +3,31 @@
...
-    let notify_cpp = &idents.notify.cpp;
-    let notify_rust_str = idents.notify.rust.to_string();
+    let notify_cpp = &idents.notify.cxx_unqualified();
+    let notify_rust = idents.notify.rust_unqualified();
     let method: ForeignItemFn = syn::parse_quote! {
         #[doc = "Notify for the Q_PROPERTY"]
-        #[rust_name = #notify_rust_str]
-        fn #notify_cpp(self: Pin<&mut #cpp_class_rust>);
+        #[cxx_name = #notify_cpp]
+        fn #notify_rust(self: Pin<&mut #cpp_class_rust>);
```
ahayzen-kdab, line 23: do we have a common way of using cxx_name in extern C++ blocks or cxx_name in extern Rust blocks and rust_name in the opposite ? eg which way round are we trying to write new code?

crates/cxx-qt-gen/src/naming/name.rs
```diff
@@ -39,6 +39,47 @@ pub struct Name {
...
+
+    /// Create a new name from this name, with the given C++ namespace
+    pub fn with_namespace(self, namespace: String) -> Self {
+        Self {
+            namespace: Some(namespace),
+            ..self
+        }
+    }
+
+    pub fn with_rust_name(self, rust_name: Ident) -> Self {
```
ahayzen-kdab, line 76: missing doc line

crates/cxx-qt-gen/src/generator/cpp/property/signal.rs
LeonMatthesKDAB: With the new `Name` API, it makes more sense to do it this way around.

`rust_unqualified` returns an `Ident`, whilst `cxx_unqualified` returns a String.
This encourages the use of `cxx_name` with the `rust_name` as the main identifier.
As otherwise `parse_quote!` complains.

In most cases, we should just always emit both names, even if they're the same, it's easiest and most explicit.
With the new API the typing and naming all works out as long as you use the `cxx_name` attribute.

ahayzen-kdab: cool, just wanted to check it's been thought about

ahayzen-kdab (approved):
LGTM :-)

PR #969, opened 2024-06-04
Use Name for QSignal parsing & generation

LeonMatthesKDAB:


crates/cxx-qt-gen/test_outputs/passthrough_and_naming.h
```diff
@@ -70,7 +70,7 @@ QPushButton_clickedConnect(
 
 namespace mynamespace::rust::cxxqtgen1 {
 ::QMetaObject::Connection
-ExternObject_dataReadyConnect(
+ExternObjectCpp_dataReadyConnect(
```
ahayzen-kdab, line 73: assume the name change was intentional as looks like it was using the wrong name before? (eg it was using the rust name and not the possible cxx name)

ahayzen-kdab (approved):
LGTM :-)

PR #973, opened 2024-06-12
Test if windows-2019 passes CI

LeonMatthesKDAB:


LeonMatthesKDAB:
This may be related to: https://github.com/actions/runner-images/issues/10019
Apparently it will be fixed in the coming days :crossed_fingers:

LeonMatthesKDAB:
Github Runners are back to normal, so no need for this

PR #978, opened 2024-06-17
CMake: Add custom CMake code to simplify build

LeonMatthesKDAB:


crates/cxx-qt-build/src/lib.rs
```diff
@@ -255,13 +256,35 @@ fn generate_cxxqt_cpp_files(
     generated_file_paths
 }
 
+/// The export directory, if one was specified through the environment.
+/// Note that this is not namspaced by crate.
+fn export_dir() -> Option<PathBuf> {
```
ahayzen-kdab, line 261: Can be done in the next pull, but splitting things into different modules would be useful so we don't have a massive lib.rs. Eg these are all dirs or info that could go in their own mod.rs

crates/cxx-qt-build/src/lib.rs
```diff
@@ -536,110 +555,129 @@ impl CxxQtBuilder {
...
-                // with CMake builds, otherwise the statement below occurs where it's missing
-                // from the final binary.
-                //
-                // When building a staticlib -bundle means that the native static library
-                // is simply not included into the archive and some higher level build
-                // system will need to add it later during linking of the final binary.
-                // https://doc.rust-lang.org/rustc/command-line-arguments.html#option-l-link-lib
-                panic!("rustc 1.69.x is not supported with CXX-Qt due to a compiler bug.\nSee: https://github.com/rust-lang/rust/pull/110917\nPlease update your compiler using 'rustup update' or use an older compiler.");
-            }
+    fn setup_cc_builder<'a>(
```
ahayzen-kdab, line 585: this might be useful to have public and hidden, as note that cxx-qt needs to setup things in a similar way

crates/cxx-qt-build/src/lib.rs
```diff
@@ -536,110 +555,129 @@ impl CxxQtBuilder {
...
+    fn build_object_file(
+        builder: &cc::Build,
+        file_path: impl AsRef<Path>,
+        export_path: Option<(&str, &str)>,
+    ) {
+        let mut obj_builder = builder.clone();
+        obj_builder.file(file_path);
+        let obj_files = obj_builder.compile_intermediates();
+
+        if let [obj_file] = &obj_files[..] {
```
ahayzen-kdab, line 640: is this taking the first element? if so why not

```rust
if let Some(obj_file) = obj_files.first() {
...
}
```

crates/cxx-qt-build/src/lib.rs
```diff
@@ -536,110 +555,129 @@ impl CxxQtBuilder {
...
+            if let Some(export_dir) = export_dir() {
+                if let Some((out_directory, out_file_name)) = export_path {
+                    let obj_dir = export_dir.join(out_directory);
+                    std::fs::create_dir_all(&obj_dir).unwrap_or_else(|_| {
```
ahayzen-kdab, line 644: what is the difference between `unwrap_or_else(|_| { panic!(...) })` vs using expect?

```rust
std::fs::create_dir_all(&obj_dir).expect(
   &format("Could not create directory for object file: {}", obj_dir.to_string_lossy())
);
```

same in other places?

crates/cxx-qt-build/src/lib.rs
```diff
@@ -536,110 +555,129 @@ impl CxxQtBuilder {
...
+            } else {
+                println!("cargo::rustc-link-arg={}", obj_file.to_string_lossy());
+                // The linker argument order matters!
+                // We need to link the object file first, then link the static library.
+                // Otherwise, the linker will be unable to find the symbols in the static library file.
+                // See also: https://stackoverflow.com/questions/45135/why-does-the-order-in-which-libraries-are-linked-sometimes-cause-errors-in-gcc
+                println!("cargo::rustc-link-arg=-l{}", static_lib_name());
+            }
+        } else {
+            panic!(
```
ahayzen-kdab, line 667: ah this is why you didn't use `first()`, a comment higher up saying we want to hit the else case if there is not one would be good.

crates/cxx-qt-build/src/lib.rs
```diff
@@ -715,26 +767,104 @@ impl CxxQtBuilder {
...
+    }
+
+    fn build_qrc_files(&mut self, init_builder: &cc::Build, qtbuild: &mut qt_build_utils::QtBuild) {
+        for qrc_file in &self.qrc_files {
+            let obj_file_dir = format!("qrc/{}", crate_name());
+            let obj_file_name = format!(
+                "{}.o",
+                qrc_file.file_name().unwrap_or_default().to_string_lossy()
+            );
+            // TODO: Is it correct to use an obj file here, or should this just link normally?
```
ahayzen-kdab, line 793: looking around qrc files are normally converted to a `.cpp` file and then compiled to a `.cpp.o` file in Qt CMake, then i assume that is linked in.

ahayzen-kdab: then there is also a `_init` cpp and cpp.o as well for the initialiser

crates/cxx-qt-build/src/lib.rs
```diff
@@ -715,26 +767,104 @@ impl CxxQtBuilder {
...
+            Some(("", "initializers.o")),
+        );
+    }
+
+    fn build_qrc_files(&mut self, init_builder: &cc::Build, qtbuild: &mut qt_build_utils::QtBuild) {
+        for qrc_file in &self.qrc_files {
+            let obj_file_dir = format!("qrc/{}", crate_name());
+            let obj_file_name = format!(
+                "{}.o",
+                qrc_file.file_name().unwrap_or_default().to_string_lossy()
```
ahayzen-kdab, line 791: or_default seems fun here? could it try and create a `qrc/.o` file?

examples/demo_threading/CMakeLists.txt
```diff
@@ -27,38 +27,12 @@ endif()
...
-    "$<LINK_LIBRARY:WHOLE_ARCHIVE,${CRATE}-static>"
-    Qt::Core
-    Qt::Gui
-    Qt::Qml
-    Qt::QuickControls2
-)
+cxxqt_import_crate(MANIFEST_PATH rust/Cargo.toml CRATES ${CRATE})
+cxxqt_import_qml_module(${CRATE}qml
+    URI "com.kdab.energy"
+    SOURCE_CRATE ${CRATE})
```
ahayzen-kdab, line 35: nice :-)

examples/qml_minimal/CMakeLists.txt
```diff
@@ -33,68 +33,30 @@ endif()
...
-    Qt::Core
-    Qt::Gui
-    Qt::Qml
-    Qt::QuickControls2
-)
+# Corrosion (through CXX-Qt) creates a CMake target with the same name as the crate.
+cxxqt_import_crate(MANIFEST_PATH rust/Cargo.toml CRATES ${CRATE})
+target_link_libraries(${CRATE} INTERFACE Qt::Core Qt::Gui Qt::Qml Qt::QuickControls2)
+
+cxxqt_import_qml_module(${CRATE}_qml
```
ahayzen-kdab, line 49: some of the examples you use `_qml` and some `qml` would be good to be consistent. Also don't know if we put part of the uri in the target name so that it's obvious this is related to the plugin ?

examples/qml_minimal/CMakeLists.txt
```diff
@@ -33,68 +33,30 @@ endif()
...
+    SOURCE_CRATE ${CRATE})
 # ANCHOR_END: book_cmake_use_corrosion
 
 # ANCHOR: book_cmake_executable
 # Define the executable with the C++ source
 add_executable(${APP_NAME} cpp/main.cpp)
 
 # Link to the Rust library
-target_link_libraries(${APP_NAME} PRIVATE ${APP_NAME}_lib)
+target_link_libraries(${APP_NAME} PRIVATE ${CRATE}_qml)
```
ahayzen-kdab, line 59: maybe update to that we link to the QML module ?

I also note that here we don't need to link to the Rust library as the plugin does this for us? I assume this all works well if you have two QML modules.

tests/basic_cxx_qt/CMakeLists.txt
```diff
@@ -29,26 +29,15 @@ endif()
...
-    Qt::Core
-    Qt::Gui
-    Qt::Qml
-    Qt::QuickControls2
-)
+cxxqt_import_crate(MANIFEST_PATH rust/Cargo.toml CRATES ${CRATE})
 
 add_executable(${APP_NAME} cpp/main.cpp)
 target_link_libraries(${APP_NAME} PRIVATE
     ${CRATE}
```
ahayzen-kdab, line 38: nice so these C++ only tests link to just the Rust crate and have very minimal CMake from a typical project :-)

crates/cxx-qt-build/src/lib.rs
LeonMatthesKDAB: Right, this could maybe go into a `cxx-qt-build-shared` crate or similar.
Currently cxx-qt-build depends on cxx-qt, so we can't yet do this unfortunately.

crates/cxx-qt-build/src/lib.rs
LeonMatthesKDAB: Because this way we get to error out if there are no, or even more than one obj files as the output.
It should never be that there's more than one obj file produced by the builder, so I want to error out in that case.

I guess I could also check the `len`, but I think this is quite elegant, because we're confirming that the vec is structured as expected and accessing the right element at the same time.

crates/cxx-qt-build/src/lib.rs
LeonMatthesKDAB: clippy would be complaining, that the `to_string_lossy()` is a function call that always happens, even if we're not actually panicking.
This way we get to wrap it in a lambda.

So this is kind of like `expect_or_else()`, which unfortunately doesn't exist.

crates/cxx-qt-build/src/lib.rs
LeonMatthesKDAB: Done

crates/cxx-qt-build/src/lib.rs
LeonMatthesKDAB: This `unwrap_or_else` I actually got by accepting clippys suggestion on how to fix this.

I'm not a big fan, but apparently this is best practice...

crates/cxx-qt-build/src/lib.rs
LeonMatthesKDAB: Hm, good question, I guess you could...
I'll change it to unwrap_or_else and panic! instead

LeonMatthesKDAB: On the other hand, we're currently saying that `.qrc` won't work with CMake anyway, so we also don't really need to export this as an object file. From CMake it's far simpler to just re-add the qrc file as a source.

So I think I'll just not export the object file in that case.

LeonMatthesKDAB: I'll also add a warning. to the documentation of `.qrc`

ahayzen-kdab: For now, but once all the changes are in `qrc` should work via a Rust build script with a CMake build, so we are working towards making it work nicely

examples/qml_minimal/CMakeLists.txt
LeonMatthesKDAB: Ah, right that was feedback I got from Milian.
Especially in the examples that we show in the book, we'll probably not want to use this added ${CRATE} variable, or at least use another separator like `_` when appending something to that, as it's just hard to read otherwise.
I'll add that everywhere.

crates/cxx-qt-build/src/lib.rs
LeonMatthesKDAB: Yes, absolutely :+1:

examples/qml_minimal/CMakeLists.txt
LeonMatthesKDAB: Also: Updated the comment.

examples/qml_minimal/CMakeLists.txt
LeonMatthesKDAB: Done

crates/cxx-qt-build/src/lib.rs
LeonMatthesKDAB: Hm, qt-build-utils currently only dumps out the one rcc file, but splitting this would be a way to go in the future I guess :thinking:

crates/cxx-qt-build/src/lib.rs
```diff
@@ -360,6 +385,10 @@ impl CxxQtBuilder {
     ///     .qrc("src/my_resources.qrc")
     ///     .build();
     /// ```
+    ///
+    /// ⚠️  In CMake projects, the .qrc file is typically added to the `SOURCES` of the target.
+    /// Prefer this to adding the qrc file through cxx-qt-build.
+    /// When using CMake, the qrc file will **not** be built by cxx-qt-build!
```
ahayzen-kdab, line 391: we'll need to update this later

ahayzen-kdab (approved):
Lets go with this for now and hope that the manifest folders work as described in #983

PR #979, opened 2024-06-18
Make rust_mut() API convenient by implicit as_mut

LeonMatthesKDAB:
I just talked with Andy about this and he inspired me to take another crack at removing all the `as_mut()` calls.

So this is an idea of how to make this work.
However, it's very much non-standard and may be confusing, as it only drops the as_mut() requirement for `rust_mut()` and nothing else.

Maybe this is not worth it, but an idea we can play with.


LeonMatthesKDAB:
Not really standard or expected, so let's leave this out for now.

PR #981, opened 2024-06-20
examples: add a unit test to ensure that qrc works

ahayzen-kdab:


ahayzen-kdab:
I think it'd be tricky to explicitly disable for CMake builds and in theory shouldn't really be any different between CMake or Cargo builds as seen in the diff here.

For the object files the initialiser can just go into the crate's general initialiser so that we don't need to know the qrc name at the figure stage?

LeonMatthesKDAB:
That would work, as long as the qrc call happens in the build script of the final crate. 
However, it wouldn't work for libraries, unless we somehow add this to the opts as well.
Which would definitely require a different API, as you couldn't just call `qrc` on the builder :thinking:

ahayzen-kdab:
Hmm fun, but that is also true for cargo builds too? Any qrc's added in cxx-qt-lib would need an initialiser to be passed through via opts to the final build script?

LeonMatthesKDAB:
Cargo builds should work, even with dependencies as I'm adding the object file generated by each qrc file to the linker if it's a cargo-only build.

But I also haven't tested that yet.

LeonMatthesKDAB (approved):
more test, more better, and this should be doable to support in our upcoming changes in #986

PR #983, opened 2024-06-21
Add an "internals" section to the book

LeonMatthesKDAB:
This section should explain some of our internal architecture decisions so that we can remember them and contributors can understand them more easily


book/src/internals/build-system.md
```diff
@@ -0,0 +1,93 @@
...
+## Contents of `cxxqtbridge`
+
+### `crates` directory
+
+Inside cxxqtbridge, there should be a `crates` folder with one subfolder per crate.
+Each crates subfolder should contain the following:
+
+- `include/`
+  - `crate-name` - A folder for all headers that are exported by this crate
+  - `cxx-qt-lib -> ../../cxx-qt-lib/include` - Symbolic links for every dependency
```
ahayzen-kdab, line 64: ah yes good point then we can avoid having multiple copies like at the moment, also note that we have the `rust/cxx.h` that needs to be injected somewhere?

book/src/internals/build-system.md
```diff
@@ -0,0 +1,93 @@
...
+- `manifest.json` - This file describes which cxx-qt-build dependencies this crate has, if it needs any Qt modules, etc.
+- `initializers.cpp` - Any code required for initialization of this crate.
+- `initializers.o` - The initializers of this crate + all it's dependencies to be linked in by CMake
```
ahayzen-kdab, line 67: will qrc initialisers end up in here? as they are not part of the qml_modules but need to be somewhere?

book/src/internals/build-system.md
LeonMatthesKDAB: Rust, for the cxx.h we'll have to see where to put that, maybe we do duplicate that one...

book/src/internals/build-system.md
LeonMatthesKDAB: That will go in the qml_modules directories (see the `qml_modules directory` section)

LeonMatthesKDAB: We could theoretically add this to the initializers.o here as well, but we likely want to keep the `cxxqt_import_qml_module` call so we'll keep this separate

LeonMatthesKDAB:
I just had another idea on how to organize the build.

We had the additional problem that if you depend on a crate that is build with cxx-qt-build (e..g cxx-qt-lib), you need to make sure that crate is built before your own one.

To ensure this is actually the case, it needs to be added to the `[build-dependencies]`.
We were planning to make sure though that our users don't forget this step by adding some kind of token that they need to provide to cxx-qt-build.

Now, what if that token was just the manifest?
That way we would avoid having to write a manifest.json, which would pull in serde and serde_json as a dependency.
cxx-qt-build could simply generate a Rust function that can be included by the library which returns the appropriate struct.

So it would work like this:

1. The build script of e.g. cxx-qt-lib would automatically dump out a function into the `OUT_DIR` with the signature:
```
fn cxxqt_build_manifest() -> cxx_qt::BuildManifest
```

2. The user then needs to export this from their library (we will want to simplify this with a macro, e.g.:)
```
// In cxx-qt-lib lib.rs
cxx_qt::include_build_manifest! {};
```

3. When depending on cxx-qt-lib, you add it to both the dependencies and build-dependencies and then add the opts from cxx-qt-build:
```
// build.rs in your crate
    CxxQtBuilder::new()
        // ...
        .with_dependency(cxx_qt_lib::cxx_qt_build_manifest())
        .build();
```

This way we would completely avoid the need to serialize and deserialize the manifest and force our users to include everything in the right order.

@ahayzen-kdab what do you think?

I already discussed this with @vimpostor this morning, and he pointed out that the downside would be more code added to the library, but that should hopefully be optimized out, as it is only used in the build script, not in the final binary.

ahayzen-kdab:
@LeonMatthesKDAB Sounds good, how do you make the options set in the build.rs exposed to the function that is returning the struct? Is this all done automatically by cxx-qt-build ?

Eg if i add extra defines in the build.rs to the cxx-qt-build, does that write them somewhere in the out_dir ? Then somehow include_build_manifest includes those during the build and exposes the defines back in the method ? As this is before where we needed both a lib and headers crate.

LeonMatthesKDAB:
Yes, all that data can go into the generated function. The manifest can automatically include certain options (like the modules).

Out of that data, we generate a manifest.rs file and dump it in the OUT_DIR.

include_build_manifest can then expand to:
```rust
include!(concat!(env!("OUT_DIR"), "/cxx_qt_build/manifest.rs"));
```
which would include the Rust code inside the library itself.

book/src/internals/build-system.md
```diff
@@ -0,0 +1,100 @@
...
+
+### `qml_modules` directory
+
+Next to the crates directory, there should be a `qml_modules` directory, which contains one directory per declared QML module.
+
+Each module should include a `plugin_init.o`, `.qmltypes`, `qmldir`, and any other necessary files.
+
+## Integration with CMake
+
+Via the `CXXQT_EXPORT_DIR` environment variable CMake should be able to change the location of the `cxxqtbridge` directory.
```
ahayzen-kdab, line 88: We should probably have a separate issue to make sure our defines/env vars are consistent?

As i think we have `CXXQT_*` and `CXX_QT_*` ?

LeonMatthesKDAB:
```shell
cxx-qt on  985-build-system [$] is 📦 v0.6.1 via △ v3.27.7 via 🦀 v1.79.0
> rg CXXQT_ --vimgrep | wc -l
20

cxx-qt on  985-build-system [$] is 📦 v0.6.1 via △ v3.27.7 via 🦀 v1.79.0
> rg CXX_QT_ --vimgrep | wc -l
165
```
So it seems CXX_QT_ is the more dominant prefix...
We should probably track this in a separate issue though.

book/src/SUMMARY.md
```diff
@@ -26,3 +26,5 @@ SPDX-License-Identifier: MIT OR Apache-2.0
   - [Shared types](./bridge/shared_types.md)
   - [Attributes](./bridge/attributes.md)
   - [Traits](./bridge/traits.md)
+- [For Contributors: CXX-Qt Internals](./internals/index.md)
+  - [Build System](./internals/build-system.md)
```
ahayzen-kdab, line 30: could you add an entry into the changelog to detail that you have changed the build system as we forgot on the other changes

LeonMatthesKDAB: Ah, good point!

LeonMatthesKDAB: Done

book/src/internals/build-system.md
ahayzen-kdab: Right could be a good first issue task ;-)

book/src/internals/build-system.md
```diff
@@ -0,0 +1,100 @@
...
+SPDX-FileContributor: Leon Matthes <leon.matthes@kdab.com>
+
+SPDX-License-Identifier: MIT OR Apache-2.0
+-->
+
+# The CXX-Qt Build System
+
+Building with CXX-Qt is somewhat more complicated than it may sound at first.
+
+**STATUS:** In-progress - this is a draft, not entirely implemented yet!
```
ahayzen-kdab, line 12: is it still draft? most of it has landed now?

LeonMatthesKDAB: Good point, I removed the note.

book/src/internals/build-system.md
ahayzen-kdab: This is done now ? as we now know where it ends up?

book/src/internals/build-system.md
LeonMatthesKDAB: Moved into #996

book/src/internals/build-system.md
LeonMatthesKDAB: It is still duplicated into the include directory of every crate, but shouldn't bee too much of an issue.

ahayzen-kdab (approved):
LGTM :-) @LeonMatthesKDAB did you want to detail those CMake situations in this change or another?

LeonMatthesKDAB:
I'll work on a fix first then detail the CMake integration after that.
Especially as that's actually user-facing now.

PR #984, opened 2024-06-23
Add is_alive method to CxxQtThread

akiselev:
When combining async Rust channels and QT objects, it's really easy to end up in a situation where a message is received on the channel after the `QObject` has already been destroyed, causing `CxxQtThread::queue` to throw an exception. This PR introduces an `is_alive()` function to check the handle if it's been invalidated.

- Implement is_alive method in C++ CxxQtThread class
- Add is_alive to Rust Threading trait
- Generate FFI bindings for is_alive in cxx-qt-gen
- Expose is_alive as a public method on CxxQtThread in Rust
- Update documentation for Threading trait and CxxQtThread

### DIFF
--- a/crates/cxx-qt-gen/src/generator/rust/threading.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/threading.rs
@@ -263,6 +263,11 @@ mod tests {
                     }
 
                     #[doc(hidden)]
+                    fn is_destroyed(cxx_qt_thread: &qobject::MyObjectCxxQtThread) -> bool {
+                        qobject::cxx_qt_ffi_my_object_cxx_qt_thread_is_destroyed(cxx_qt_thread)
+                    }
+
+                    #[doc(hidden)]
                     fn queue<F>(cxx_qt_thread: &qobject::MyObjectCxxQtThread, f: F) -> std::result::Result<(), cxx::Exception>
                     where
                         F: FnOnce(core::pin::Pin<&mut qobject::MyObject>),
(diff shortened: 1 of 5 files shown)


LeonMatthesKDAB:
Hi @akiselev , thank you for taking the time to create this PR!

overall I think the `is_alive` method should be useful, so it's good to have it.
However, depending on what you want to use it for, this may not do what you intend!

If you want to check `is_alive` before calling `queue`, **this will not work**!

e.g. this:
```rust
if thread.is_alive() {
    thread.queue(...).unwrap()
}
```
May panic and crash your program. The reason for this is that the CxxQtThread does not have ownership of the QObject, it is only a handle to the thread the QObject lives in.
That other thread may delete the QObject at any time, including between the calls to `is_alive()` and `queue`.

So ideally you should really prefer just using.
```
thread.queue(...).ok();
```
if you want to avoid crashing your application in that case.

If however, you want to use `is_alive()` to e.g. exit a loop:
```
while qobject_thread.is_alive() {
    // do some rust work
    // maybe message the qobject somewhere
    qobject_thread.queue(...).ok();
    // do some more rust work
}
```
Then it's very much a good and convenient API.

So really you can only know for certain whether a QObject has already been destroyed, not whether it's still alive, as it may be destroyed at any point.
Therefore I'd suggest renaming this to `is_destructed`/`is_destroyed`/`is_dead` or something like this and prominently documenting this caveat.

akiselev:
I renamed the method to `is_destroyed` and added a better doc comment explaining the race condition

LeonMatthesKDAB:
Hi @akiselev , Thank you for the update to the PR.
I'm happy to approve the new API, but suspect CI will fail because the tests haven't been updated to the new generated output.
If your system supports it, can you run the `./update_expected.sh` shell script in the `cxx-qt-gen` directory?
That should automatically update the generation tests.

In case you have issues with the script, I can run it as well, just let me know.

akiselev:
Updated

ahayzen-kdab:
Looks there there is another case of a test needing updating :-)

Note you can run `cargo test` in the `crates/cxx-qt-gen` to see the ones that are failing.

LeonMatthesKDAB:
I updated the tests, so they hopefully pass CI now.

ahayzen-kdab (approved):
LGTM :-)

LeonMatthesKDAB:
Great to see this finally merged.

Thank you @akiselev for your contribution!

PR #986, opened 2024-06-27
Cxx-Qt-build: Refactor export_dir structure and introduce library Interface

LeonMatthesKDAB:
This should now be close to what we're aiming for as described in #983.


crates/cxx-qt-build/src/lib.rs
```diff
@@ -586,6 +582,23 @@ impl CxxQtBuilder {
         }
     }
 
+    fn setup_dependencies(&self) {
```
ahayzen-kdab, line 585: add doc lines of what this is doing? what are these dependencies ? it's not obvious here

crates/cxx-qt-build/src/lib.rs
```diff
@@ -806,21 +816,39 @@ impl CxxQtBuilder {
         }
     }
 
+    fn clean_directory(path: impl AsRef<Path>) {
+        let result = std::fs::remove_dir_all(&path);
+        if let Err(err) = result {
+            // If the path doesn't exist that's fine, otherwise we want to panic
+            if err.kind() != ErrorKind::NotFound {
+                panic!("Could not delete export directory: {}", err);
```
ahayzen-kdab, line 824: export dir? isn't this a generic path, and same on the expect below

ahayzen-kdab: As in would it not be better to just say `Could not delete {path}: {err}`

crates/cxx-qt-gen/src/generator/rust/mod.rs
```diff
@@ -79,7 +79,11 @@ impl GeneratedRustBlocks {
 
 /// Generate the include line for this parsed block
 fn generate_include(parser: &Parser) -> Result<Item> {
-    let import_path = format!("cxx-qt-gen/{}.cxxqt.h", parser.cxx_file_stem);
+    let import_path = format!(
+        "{crate_name}/{}.cxxqt.h",
+        parser.cxx_file_stem,
+        crate_name = std::env::var("CARGO_PKG_NAME").unwrap_or_else(|_| "cxx-qt-gen".to_owned())
```
ahayzen-kdab, line 85: is there a situation where `CARGO_PKG_NAME` is actually empty? wondering if we should error out instead of silently falling back to `cxx-qt-gen` ? As what happens if two crates try to write to that same path ?

crates/cxx-qt-lib/build.rs
```diff
@@ -4,11 +4,64 @@
...
+    println!("cargo::rerun-if-changed=include/{subfolder}");
+
+    for entry in
+        std::fs::read_dir(format!("include/{subfolder}")).expect("Failed to read include directory")
+    {
+        let entry = entry.expect("Failed to read header file!");
+        let header_name = entry.file_name();
+        println!("cargo::rerun-if-changed=include/{subfolder}/{header_name:?}",);
+
+        // TODO: Do we want to add the headers into a subdirectory?
```
ahayzen-kdab, line 35: btw i was thinking before 1.0 of putting things into `cxx-qt-lib/core/` / ``cxx-qt-lib/gui/` or something like Qt does but unsure.

There is also the debate of whether they should need to include every header of if there could be a single include for all of the headers or per module?

crates/cxx-qt/src/build.rs
```diff
@@ -0,0 +1,16 @@
...
+// SPDX-FileContributor: Leon Matthes <leon.matthes@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! This modules contains utilities for building crates that use cxx-qt-build.
+
+#[doc(hidden)]
+// This struct should not be used directly.
+// Only use the Manifests generated by cxx-qt-build.
+pub struct Manifest {
```
ahayzen-kdab, line 11: any reason for this not to be in `cxx-qt-build` ? Is this so that `cxx-qt` itself could create a manifest later as it just uses cxx_build directly ? and `cxx-qt-build` could then read the `cxx-qt` build manifest too ?

crates/cxx-qt-lib/build.rs
LeonMatthesKDAB: That probably makes sense to put it into directories.
We can always add forwarding headers later if we want to add convenience includes.

And we can always add headers that just include all headers of one module, e.g. `core/prelude.h`.

crates/cxx-qt-build/src/lib.rs
LeonMatthesKDAB: Ah, right, maybe I should just bubble up the error.

crates/cxx-qt-build/src/lib.rs
LeonMatthesKDAB: Done

crates/cxx-qt-build/src/lib.rs
LeonMatthesKDAB: Done, and also moved into the `dir` module.

crates/cxx-qt-lib/build.rs
LeonMatthesKDAB: I'd say let's do this in a separate PR :)

crates/cxx-qt/src/build.rs
LeonMatthesKDAB: With the `links` key, we no longer need this in cxx-qt, it's now in cxx-qt-build :)

crates/cxx-qt-gen/src/generator/rust/mod.rs
LeonMatthesKDAB: Refactored so that cxx-qt-build now writes the header prefix to disk and we can then read it back when building.

It's then still empty when testing so we still fall back to cxx-qt-gen and don't need to adjust our tests :)

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -313,6 +313,12 @@ jobs:
         QT_QPA_PLATFORM: ${{ matrix.qt_qpa_platform }}
         QT_SELECT: qt${{ matrix.qt_version }}
 
+    - name: "tmate session for debugging"
```
ahayzen-kdab, line 316: note this should be removed before merging :-)

crates/cxx-qt-build/Cargo.toml
```diff
@@ -22,6 +22,9 @@ quote.workspace = true
 qt-build-utils.workspace = true
 codespan-reporting = "0.11"
 version_check = "0.9"
+scratch = "1.0"
+serde = {version = "1.0", features = ["default", "derive"]}
```
ahayzen-kdab, line 26: elsewhere we have a space around the braces, so `serde = { version = "1.0", features = ["default", "derive"] }`

crates/cxx-qt-build/src/dependencies.rs
```diff
@@ -0,0 +1,301 @@
...
+//! This modules contains utilities for specifying dependencies with cxx-qt-build.
+
+use serde::{Deserialize, Serialize};
+
+use std::collections::{HashMap, HashSet};
+use std::path::{Path, PathBuf};
+
+/// When generating a library with cxx-qt-build, the library may need to export certain flags or headers.
+/// These are all specified by this Interface struct, which should be passed to the [crate::CxxQtBuilder::library] function.
+pub struct Interface {
```
ahayzen-kdab, line 15: does this need to be non exhaustive to avoid ABI breaks ? (eg if we are adding more fields) or does it work as the have to go through methods like default and cannot access the internals?

crates/cxx-qt-build/src/dependencies.rs
```diff
@@ -0,0 +1,301 @@
...
+    ///
+    /// Note that the link name may differ from the crate name.
+    /// Check your dependencies manifest file for the correct link name.
+    pub fn reexport_dependency(mut self, link_name: &str) -> Self {
+        self.reexport_links.insert(link_name.to_owned());
+        self
+    }
+}
+
+// cxx-qt-lib example
```
ahayzen-kdab, line 148: either remove these or make them a doc comment / test ?

Also should `python3` be in there?

crates/cxx-qt-build/src/dependencies.rs
```diff
@@ -0,0 +1,301 @@
...
+
+// cxx-qt-lib-headers example
+// Interface {
+//     export_crate_headers: true,
+//     exported_header_prefixes: Some(vec!["cxx-qt-lib-extras/..."]),
+//     reexport_links: vec!["cxx-qt-lib"],
+// }
+
+#[derive(Clone, Serialize, Deserialize)]
+// This struct is used by cxx-qt-build internally to propagate data through to downstream
```
ahayzen-kdab, line 163: could be a proper doc comment? then it would appear when you document locally with the private things too?

crates/cxx-qt-build/src/dependencies.rs
```diff
@@ -0,0 +1,301 @@
...
+    pub(crate) name: String,
+    pub(crate) link_name: String,
+    pub(crate) qt_modules: Vec<String>,
+    pub(crate) defines: Vec<(String, Option<String>)>,
+    pub(crate) initializers: Vec<PathBuf>,
+    pub(crate) exported_include_prefixes: Vec<String>,
+}
+
+#[derive(Clone)]
+pub(crate) struct Dependency {
```
ahayzen-kdab, line 175: doc line for what this is would be useful

crates/cxx-qt-build/src/dependencies.rs
```diff
@@ -0,0 +1,301 @@
...
+    pub(crate) path: PathBuf,
+    pub(crate) manifest: Manifest,
+}
+
+impl Dependency {
+    pub(crate) fn find_all() -> Vec<Dependency> {
```
ahayzen-kdab, line 181: doc line and/or comments as to what is happening here would be good

crates/cxx-qt-build/src/dependencies.rs
```diff
@@ -0,0 +1,301 @@
...
+                    entry.insert((value.to_owned(), dependency.manifest.name.clone()));
+                }
+                Occupied(entry) => {
+                    let existing_value = &entry.get().0;
+                    if existing_value != value {
+                        panic!("Conflicting compiler definitions requested!\nCrate {existing} exports {variable}={existing_value:?}, and crate {conflicting} exports {variable}={value:?}",
+                            existing=entry.get().1,
+                            conflicting = dependency.manifest.name);
+                    }
+                    // else: Ignore Duplicate definition with the same value
```
ahayzen-kdab, line 291: could have a reverse comment on top of the if instead

crates/cxx-qt-build/src/dependencies.rs
LeonMatthesKDAB: The internals are `pub(crate)`, so you can't access them from outside :)

crates/cxx-qt-build/src/dependencies.rs
LeonMatthesKDAB: Ah, right, those can go now.

crates/cxx-qt-build/src/dependencies.rs
LeonMatthesKDAB: done

crates/cxx-qt-build/src/dependencies.rs
LeonMatthesKDAB: Done

crates/cxx-qt-build/src/lib.rs
```diff
@@ -806,22 +865,102 @@ impl CxxQtBuilder {
...
+                        )
+                    });
+            }
+        }
+    }
+
     /// Generate and compile cxx-qt C++ code, as well as compile any additional files from
     /// [CxxQtBuilder::qobject_header] and [CxxQtBuilder::cc_builder].
     pub fn build(mut self) {
+        // TODO: Clean the directory before we start building
```
ahayzen-kdab, line 937: the code below runs clean, is this referring to something else ? it's not quite clear if this has been fixed or if there is a further clean ?

crates/cxx-qt-build/src/dependencies.rs
LeonMatthesKDAB: Done

crates/cxx-qt-build/src/lib.rs
LeonMatthesKDAB: Ah, right, I had to disable it for a while, it works now, so we can remove the comment :)

crates/cxx-qt-build/src/dependencies.rs
LeonMatthesKDAB: Done, with reference to the internals section of the book.

crates/cxx-qt-build/src/dependencies.rs
LeonMatthesKDAB: Done

.github/workflows/github-cxx-qt-tests.yml
LeonMatthesKDAB: Done

crates/cxx-qt-build/Cargo.toml
LeonMatthesKDAB: Done

ahayzen-kdab (approved):
LGTM :-)

PR #989, opened 2024-07-08
fix: Qrc test data

LeonMatthesKDAB:
This was accidentally only running the data from the "invalid" test case, as JS was merging everything into a single object, instead of two objects in an array.


ahayzen-kdab (approved):
LGTM :-)

ahayzen-kdab:
So this will be superseded by #986 ?

LeonMatthesKDAB:
Yes, it's part of #986 now.
Apparently this didn't even work before :sweat_smile:

PR #992, opened 2024-07-09
fix: Use QT_QPA_PLATFORM=offscreen in tests

LeonMatthesKDAB:
This avoids windows popping up when running the tests for qml_features and qml_minimal.

### DIFF
--- a/examples/qml_features/CMakeLists.txt
+++ b/examples/qml_features/CMakeLists.txt
@@ -77,6 +77,15 @@ if(BUILD_TESTING)
         NAME ${APP_TEST_NAME}
         COMMAND ${TEST_CMD}
     )
+    # Unfortunately due to the static linking in our CI on macOS we can't load the
+    # offscreen plugin, so just leave it at the default.
+    if (NOT APPLE)
+        set_tests_properties(${APP_TEST_NAME} PROPERTIES ENVIRONMENT "QT_QPA_PLATFORM=offscreen")
+    endif()
+    # RUNTIME_ENV comes from the CMakeLists.txt at the root of this repository.
+    set_tests_properties(
+        ${APP_TEST_NAME} PROPERTIES ENVIRONMENT_MODIFICATION "${RUNTIME_ENV}"
+    )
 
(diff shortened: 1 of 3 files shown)


examples/qml_features/CMakeLists.txt
```diff
@@ -77,6 +77,7 @@ if(BUILD_TESTING)
         NAME ${APP_TEST_NAME}
         COMMAND ${TEST_CMD}
     )
+    set_property(TEST ${APP_TEST_NAME} PROPERTY ENVIRONMENT "QT_QPA_PLATFORM=offscreen")
```
ahayzen-kdab, line 80: note you can also set this in the setup of the QUICK_TEST_MAIN_WITH_SETUP with `qputenv("QT_QPA_PLATFORM", "offscreen");` at the start of the constructor, unsure if that is better or worse here?

And unsure why we didn't enable this before?

examples/qml_minimal/CMakeLists.txt
```diff
@@ -88,6 +88,7 @@ if(BUILD_TESTING)
             NAME ${APP_TEST_NAME}
             COMMAND ${TEST_CMD}
         )
+        set_property(TEST ${APP_TEST_NAME} PROPERTY ENVIRONMENT "QT_QPA_PLATFORM=offscreen")
```
ahayzen-kdab, line 92: you could also add it to the qt_types_standalone test incase that starts testing widgets / qml objects enough that they try to show anything. And that could go in the main.cpp function i guess

examples/qml_features/CMakeLists.txt
LeonMatthesKDAB: Keeping it in CMake is a bit more flexible, as you can then still choose the platform after building by running the test executable manually.
So let's go with that for now.

examples/qml_minimal/CMakeLists.txt
LeonMatthesKDAB: Done.

ahayzen-kdab (approved):
LGTM :-)

PR #993, opened 2024-07-09
fix(build): Try to fix windows file sync issues

LeonMatthesKDAB:
See CI failures like this one: https://github.com/KDAB/cxx-qt/actions/runs/9855407973/job/27210248279?pr=990

### DIFF
--- a/crates/cxx-qt-build/src/out.rs
+++ b/crates/cxx-qt-build/src/out.rs
@@ -1,0 +1,28 @@
+// SPDX-FileCopyrightText: 2024 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Leon Matthes <leon.matthes@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! A module for handling anything output related
+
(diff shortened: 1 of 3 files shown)


crates/cxx-qt-build/src/out.rs
```diff
@@ -0,0 +1,25 @@
...
+
+#[cfg(windows)]
+pub(crate) fn write_file(
+    path: impl AsRef<Path>,
+    contents: impl AsRef<[u8]>,
+) -> std::io::Result<()> {
+    // On Windows we're experiencing some issues with file locking and possibly
+    // synchronization, especially when combined with cc::Build
+    // So force a synchronization here with sync_all()
+    let file = File::create(path)?;
```
ahayzen-kdab, line 19: need to add a `use` for this :-)

LeonMatthesKDAB: Done, fixed some other issues as well by disabling the `#[cfg(windows)]` temporarily.
It's amazing how many mistakes you can make when coding blind :sweat_smile:

LeonMatthesKDAB:
This should no longer be necessary as the issues were actually related to duplicate build scripts running as explained in #997

PR #994, opened 2024-07-09
Add HashSet of flags to ParsedQPoperty for optional getters and setters

BenFordTytherington:
Add option to specify whether QProperties generate getters or setters


crates/cxx-qt-gen/src/generator/cpp/property/meta.rs
```diff
@@ -7,12 +7,22 @@ use crate::generator::naming::property::QPropertyNames;
...
+        output += format!("WRITE {} ", name.cxx_unqualified()).as_str();
+    }
+    // Notify
+    if let Some(name) = &idents.notify {
+        output += format!("NOTIFY {}", name.cxx_unqualified()).as_str();
+    }
+
+    output += ")";
+
+    output
```
ahayzen-kdab, line 27: wonder if it is better or worse to build up a vec of the inner parts then write at the end joining them, rather than modifying a string?

eg (could have better naming but you get the idea)

```rust
let mut parts = vec![format!("READ {ident_getter}")];

// if write is enabled
parts.push(format!("WRITE {}", name.cxx_unqualified());

format!("Q_PROPERTY({ty} {ident} {meta_parts})", meta_parts = parts.join(" "))
```

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -3,14 +3,114 @@
...
+}
+
+impl QPropertyFlags {
+    pub fn new() -> Self {
+        Self {
+            read: None,
+            write: None,
+            notify: None,
+        }
+    }
```
ahayzen-kdab, line 36: could instead just have `#[derive(Default)]` on the `QPropertyFlags` then you can use `QPropertyFlags::default()`, but then i guess maybe this is not the "default" this is none of them

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -3,14 +3,114 @@
...
+
+    pub fn all_flags() -> Self {
```
ahayzen-kdab, line 38: note we are going to add more flags (like reset, constant, final etc) and the "default" (heh) would be effectively `read, write, notify` so this probably needs a better name?

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -3,14 +3,114 @@
...
+    }
+
+    fn set_field_by_string(&mut self, key: String, value: IdentFlag) -> Result<String> {
+        match key.as_str() {
+            "read" => self.read = value,
+            "write" => self.write = Some(value),
+            "notify" => self.notify = Some(value),
+            _ => {
+                return Err(Error::new(
+                    format_ident!("{}", key).span(), // TODO: check if this is an acceptable form of erroring for non Syn functions
```
ahayzen-kdab, line 53: i don't think this will point to the original span as the information has been lost, instead consider passing in the `Ident` type as the `key` not the `String` then you can use `key.span()`

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -3,14 +3,114 @@
...
     pub ident: Ident,
     /// The [syn::Type] of the property
     pub ty: Type,
+    /// Property flag collection
+    pub flags: QPropertyFlags,
+}
+
+fn parse_path_to_ident(path: &Path) -> Ident {
+    path.segments[0].ident.clone()
+}
```
ahayzen-kdab, line 96: hmm this assumes that there is only one part of the path

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -3,14 +3,114 @@
...
+
+    pub fn all_flags() -> Self {
+        Self {
+            read: None,
+            write: Some(None),
+            notify: Some(None),
+        }
+    }
+
+    fn set_field_by_string(&mut self, key: String, value: IdentFlag) -> Result<String> {
```
ahayzen-kdab, line 46: so you could instead change this to accept the `Path` as the `key`.

Then instead use our `syntax::path::path_compare_str` like this, unfortunately it means we can't use a `match` but it allows you to compare Path's as at the moment if you have `#[qproperty(T, name, read::thing)` you would consider that as `read` when it's actually an invalid path?

```write
if path_compare_str(&key /* which is of type Path */, &["read"]) {
  // we have a read
} else if path_compare_str(&key, &["write"]) {
  // we have a write
...
} else {
  // unknown we can error with the original Path span
  return Err(Error::new(key.span(), ...);
}

File: crates/cxx-qt-gen/src/parser/property.rs (hunk 5 above)
Comment (ahayzen-kdab) [reply crates/cxx-qt-gen/src/parser/property.rs:96]: see longer comment above about using `path_compare_str`

Comment (BenFordTytherington) [reply crates/cxx-qt-gen/src/parser/property.rs:96]: I think with the way it's being parsed in, there can only be one path segment, because they have to comma separated for the syn::Punctuated parsing, and if it was a List or NameValue it gets parsed properly?

File: crates/cxx-qt-gen/src/parser/property.rs (hunk 3 above)
Comment (BenFordTytherington) [reply crates/cxx-qt-gen/src/parser/property.rs:38]: Will change it to implementing Default

File: crates/cxx-qt-gen/src/parser/property.rs (hunk 6 above)
Comment (ahayzen-kdab) [reply crates/cxx-qt-gen/src/parser/property.rs:46]: Aha but as we are comparing only one level here you could use the built in method in `syn`.

There is `Path::is_ident` which you could then use like this...
https://docs.rs/syn/latest/syn/struct.Path.html#method.is_ident

```rust
if key.is_ident("read") {
  // we have a read
} else if key.is_ident("write") {
  // we have a write
...
} else {
  // unknown we can error with the original Path span
  return Err(Error::new(key.span(), ...);
}
```

ahayzen-kdab: As discussed there is also `get_ident` and `require_ident` which could be useful too

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -20,10 +120,43 @@ impl ParsedQProperty {
...
+                    let flag_string: String = passed_flags.add_from_meta(flag)?;
+                    flag_strings.push(flag_string);
+                }
+
+                if !flag_strings.contains(&String::from("read")) {
+                    return Err(Error::new(
+                        input.span(),
+                        "If flags are passed, read must be explicitly specified",
+                    ));
+                }
```
ahayzen-kdab, line 152: you don't actually need the vector of strings in the end other than for checking is there read in it ? So instead change to a bool and check in the loop.

eg

```rust
let mut found_read = false;

for flag in flags {
    let flag_string: String = passed_flags.add_from_meta(flag)?;
    found_read |= flag_string == "read";
}

if !found_read {
...
```

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -79,7 +287,7 @@ mod tests {
     #[test]
     fn test_parse_property_no_type() {
         let mut input: ItemStruct = parse_quote! {
-            #[qproperty(T)]
+            #[qproperty(num)]
```
ahayzen-kdab, line 290: any reason for this change? it doesn't seem to be relevant?

ahayzen-kdab: or should the name of the test be `_no_name` rather than `_no_type` :-)

BenFordTytherington: because the test is for testing no type, not no identifier, it was a duplicate of the test above it and existed already

crates/cxx-qt-gen/src/syntax/qtfile.rs
```diff
@@ -45,10 +45,21 @@ impl ToTokens for CxxQtFile {
 
 /// Parse the given [std::path::Path] into a [CxxQtFile]
 pub fn parse_qt_file(path: impl AsRef<std::path::Path>) -> Result<CxxQtFile> {
```
ahayzen-kdab, line 47: these changes could be in a separate pull request?

crates/cxx-qt-gen/src/parser/property.rs
ahayzen-kdab: Maybe the test should be removed as i don't think it can tell the difference between a name and a type, what we are really testing here is if there is 1 argument does it fail as we require 2.

Maybe instead this test could check that 0 arguments fails instead? So we have the test above for checking 1 argument fails and change this test for 0 arguments.

crates/cxx-qt-gen/src/generator/cpp/property/meta.rs
ahayzen-kdab: done

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
```diff
@@ -27,24 +27,49 @@ pub fn generate_cpp_properties(
...
+            if let Some(setter_wrapper) = setter::generate_wrapper(&idents, &cxx_ty) {
+                generated.private_methods.push(setter_wrapper)
+            }
+        }
+
+        // Checking custom notifier wasn't provided
+        if flags
+            .notify
+            .clone()
+            .is_some_and(|ident_option| ident_option.is_none())
```
ahayzen-kdab, line 67: guess all this fun is where the tri-state enum thing instead of Option<Name> in QPropertyNames could help ? did you have any further ideas on that?

eg
```rust
enum QPropertyFlagState {
   None,
   Auto(Name), // with auto generated name
   Custom(Name), // with custom name
}
```

Then here you would remove the flag check and just do this?

```rust
            if let Some(notify) = signal::generate(&idents, qobject_idents) {
                signals.push(notify)
            }
```

Then have the `generate` function return Some only when it's `Auto(...)` and not None or Custom ?

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
```diff
@@ -27,24 +27,49 @@ pub fn generate_cpp_properties(
...
+                .push(getter::generate(&idents, &qobject_ident, &cxx_ty));
+            generated
+                .private_methods
+                .push(getter::generate_wrapper(&idents, &cxx_ty));
+        }
+
+        // Checking custom setter wasn't provided
+        if flags
+            .write
+            .clone()
```
ahayzen-kdab, line 51: i think instead of `.clone()` you can use `.as_ref()` here, but this would also go away if you go the enum idea route

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
```diff
@@ -60,25 +85,49 @@ pub fn generate_cpp_properties(
...
+            struct MyStruct;
+        };
+        let property = ParsedQProperty::parse(input.attrs.remove(0)).unwrap();
+
+        let properties = vec![property];
+
+        let qobject_idents = create_qobjectname();
+
+        let type_names = TypeNames::mock();
+        let _generated =
```
ahayzen-kdab, line 110: ideally you want to be doing the assert_eq and assert_str_eq on the results here to check the right number and content of the things. You can likely copy and maybe even share some from below depending on the property names etc.

crates/cxx-qt-gen/src/generator/cpp/property/setter.rs
```diff
@@ -6,32 +6,39 @@
...
-                const ::rust::cxxqt1::MaybeLockGuard<{qobject_ident}> guard(*this);
-                {ident_setter_wrapper}(value);
-            }}
-            "#,
-            ident_setter = idents.setter.cxx_unqualified(),
-            ident_setter_wrapper = idents.setter_wrapper.cxx_unqualified(),
-        },
+pub fn generate(idents: &QPropertyNames, qobject_ident: &str, cxx_ty: &str) -> Option<CppFragment> {
+    // Checking whether a setter should be generated based off if it has a name
+    if let (Some(setter), Some(setter_wrapper)) = (&idents.setter, &idents.setter_wrapper) {
```
ahayzen-kdab, line 11: as we were discussing maybe the enum route is better?

This would then become
`if let (QPropertyFlagState::Auto(setter), QPropertyFlagState::Auto(setter_wrapper) = (&idents.setter, &idents.setter_wrapper) {`

Which would then mean that this generate function could be called even when a custom setter is set, as that is a different enum, which then removes that if condition on the flag in the other file

crates/cxx-qt-gen/src/generator/naming/property.rs
```diff
@@ -12,22 +12,49 @@ pub struct QPropertyNames {
     pub name: Name,
     pub getter: Name,
     pub getter_wrapper: Name,
-    pub setter: Name,
-    pub setter_wrapper: Name,
-    pub notify: Name,
+    pub setter: Option<Name>,
```
ahayzen-kdab, line 15: this would then become the enum QPropertyFlagState.

There would be the question over whether the getter ones should have an enum as well to do the same but without None as an option?

crates/cxx-qt-gen/src/generator/naming/property.rs
```diff
@@ -97,14 +126,20 @@ pub mod tests {
             names.getter.rust_unqualified(),
             &format_ident!("my_property")
         );
-        assert_eq!(names.setter.cxx_unqualified(), "setMyProperty");
         assert_eq!(
-            names.setter.rust_unqualified(),
+            names.setter.clone().unwrap().cxx_unqualified(),
```
ahayzen-kdab, line 130: as_ref can be used instead of clone i think?

crates/cxx-qt-gen/src/generator/rust/property/mod.rs
```diff
@@ -31,26 +31,47 @@ pub fn generate_rust_properties(
...
+            generated
+                .cxx_qt_mod_contents
+                .append(&mut getter.implementation_as_items()?);
+        }
+
+        // Checking that write flag was provided but no custom identifier
+        if flags
+            .write
+            .clone()
+            .is_some_and(|ident_option| ident_option.is_none())
```
ahayzen-kdab, line 51: same story on the rust side where the enum might simplify things

crates/cxx-qt-gen/src/generator/rust/property/setter.rs
```diff
@@ -19,46 +19,63 @@ pub fn generate(
...
+                    pub fn #setter_rust(mut self: core::pin::Pin<&mut Self>, value: #qualified_ty) {
+                        use cxx_qt::CxxQtType;
+                        //TODO: could be optimised to remove this condition?
+                        if self.#ident == value {
+                            // don't want to set the value again and reemit the signal,
+                            // as this can cause binding loops
+                            return;
+                        }
+                        self.as_mut().rust_mut().#ident = value;
+                        // self.as_mut().#notify_ident();
```
ahayzen-kdab, line 72: remove this line as notify_binding replaces this

crates/cxx-qt-gen/src/generator/rust/property/setter.rs
```diff
@@ -19,46 +19,63 @@ pub fn generate(
...
+                    #has_unsafe fn #setter_rust(self: Pin<&mut #cpp_class_name_rust>, value: #cxx_ty);
+                }
+            }],
+            implementation: vec![quote! {
+                impl #qualified_impl {
+                    #[doc = "Setter for the Q_PROPERTY "]
+                    #[doc = #ident_str]
+                    pub fn #setter_rust(mut self: core::pin::Pin<&mut Self>, value: #qualified_ty) {
+                        use cxx_qt::CxxQtType;
+                        //TODO: could be optimised to remove this condition?
```
ahayzen-kdab, line 65: right, might be worth doing not sure, lets leave a github comment here to discuss...

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
BenFordTytherington: Was thinking it would be Option<State> as then encapsulates that read can be Auto or Custom, but not None, and write and notify would be optional, I think I've left a comment somewhere about places where this refactor is useful

crates/cxx-qt-gen/src/generator/cpp/property/setter.rs
BenFordTytherington: Yeah exactly, or still using Some() if going for Option<State> as mentioned above

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -3,14 +3,112 @@
...
+
+/// An optional identifier of the functions passed with flags
+/// e.g. read = my_getter, IdentFlag would be used to store my_getter
+type IdentFlag = Option<Ident>;
+
+/// Struct for storing the flags provided for a QProperty and their optional identifiers ([IdentFlag])
+#[derive(Debug)]
+pub struct QPropertyFlags {
+    // TODO: Refactor to use an enum type for representing all 3 states of write and notify
+    // Or use an enum for {Auto, Custom(Ident)}, this is optional for write and notify, but not read
```
ahayzen-kdab, line 21: yes, as discussed do this as the enum refactor :-)

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -3,14 +3,112 @@
...
+    pub(crate) read: IdentFlag,
+    pub(crate) write: Option<IdentFlag>,
+    pub(crate) notify: Option<IdentFlag>,
+}
+
+impl Default for QPropertyFlags {
+    // Default represents the flags of the desugared version of #[qproperty(T, ident)]
```
ahayzen-kdab, line 28: maybe put three `///` so it's a doc line?

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -3,14 +3,112 @@
...
+            read: None,
+            write: Some(None),
+            notify: Some(None),
+        }
+    }
+}
+
+impl QPropertyFlags {
+    // Doesn't really represent a realistic state, just used for collecting in the parser
+    pub fn new() -> Self {
```
ahayzen-kdab, line 40: as this is only (?) used in this file and only once (?) maybe instead of making it a `pub` method either keep it private or at least `pub (crate)` or just use `QPropertyFlags { read: None, write: None, notify: None }` where the new() is called?

Or instead build them up before creating it so that you don't need read to be None

crates/cxx-qt-gen/src/parser/property.rs
ahayzen-kdab: done

crates/cxx-qt-gen/src/parser/property.rs
ahayzen-kdab: So now that you have the right `Path` coming in, instead use `Path::require_ident() -> Result<Ident>`, you could probably just remove this function and instead change it's usages?

eg instead of
```
Meta::Path(path) => Ok(self.set_field_by_ident(parse_path_to_ident(&path), None)?),
```
it could be
```
Meta::Path(path) => Ok(self.set_field_by_ident(path.require_ident()?, None)?),
```

You just might need to change some places to accept `&Ident` instead of `Ident` where possible, like `set_field_by_ident` can ? And other places that cannot can use `to_owned()` or `clone()`

https://docs.rs/syn/latest/syn/struct.Path.html#method.require_ident

crates/cxx-qt-gen/src/parser/property.rs
ahayzen-kdab: done

crates/cxx-qt-gen/src/parser/property.rs
ahayzen-kdab: changed

crates/cxx-qt-gen/src/parser/property.rs
ahayzen-kdab: See other comment about using `require_ident`

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -20,10 +118,46 @@ impl ParsedQProperty {
...
+            } else {
+                let _comma = input.parse::<Token![,]>()?; // Start of final identifiers
+
+                let punctuated_flags: Punctuated<Meta, Token![,]> =
+                    Punctuated::parse_terminated(input)?;
+
+                // Remove the commas and collect the individual meta items
+                let flags: Vec<Meta> = punctuated_flags.clone().into_iter().collect();
+
+                let mut passed_flags = QPropertyFlags::new();
```
ahayzen-kdab, line 137: parsed or passed? :sweat_smile:

crates/cxx-qt-gen/src/parser/property.rs
ahayzen-kdab: This is still valid as to remove the list and use a bool if QPropertyFlags changes so that read is not an Option so you cannot check if it's been set, see other comment.

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -46,6 +180,81 @@ mod tests {
...
+        let mut input: ItemStruct = parse_quote! {
+            #[qproperty(T, name, read = blah, a, notify = blahblah)]
+            struct MyStruct;
+        };
+        let property = ParsedQProperty::parse(input.attrs.remove(0));
+        assert!(property.is_err())
+    }
+
+    #[test]
+    fn test_parse_missing_flags() {
```
ahayzen-kdab, line 239: maybe change the test name to test_parse_missing_read_flag just to enforce that we are expecting it to fail as `read` is missing

crates/cxx-qt-gen/test_inputs/properties.rs
```diff
@@ -11,6 +11,7 @@ mod ffi {
         #[derive(Default)]
         #[qproperty(i32, primitive)]
         #[qproperty(QPoint, trivial)]
+        #[qproperty(i32, test_name, read = my_getter, write = my_setter, notify)]
         type MyObject = super::MyObjectRust;
     }
```
ahayzen-kdab, line 16: You also have some mixed cases that could go in here.

So you could have
```
#[qproperty(i32, custom_property, read = my_getter, write = my_setter, notify = my_notify)]
#[qproperty(i32, custom_partial_property, read = my_getter, write, notify = my_notify)]
```

Which would check that full custom generation works and also checks that eg the `my_notify` name is correctly chosen in the generated setter.

Unsure if you should add the methods/signals for the actual custom functions or not as they are in theory tested in other inputs and could make the output of this one even more verbose...

crates/cxx-qt-gen/src/syntax/qtfile.rs
ahayzen-kdab: done

crates/cxx-qt-gen/src/parser/property.rs
BenFordTytherington: Read was only one Option, so None means it exists, just has no Identifier. This should also be fixed with the move to a State enum inside QPropertyFlags

crates/cxx-qt-gen/src/generator/naming/property.rs
```diff
@@ -2,32 +2,107 @@
...
+        };
+
         Self {
             getter_wrapper: wrapper_name_from_function_name(&getter),
             getter,
-            setter_wrapper: wrapper_name_from_function_name(&setter),
+            setter_wrapper,
             setter,
-            notify: notify_name_from_property(&property_name),
+            notify,
```
ahayzen-kdab, line 105: something we need to figure out is how to read the cxx_name/rust_name correctly of the custom functions, eg...

i believe that this is will pick up the wrong notify name in the following case

```rust
extern "RustQt" {
   // here we should always use the rust name for custom attribute values
   #[qproperty(T, name, read = prop_getter, write = prop_setter, notify = prop_changed]
   #[qobject]

   #[cxx_name = "getProp"]
   fn prop_getter(self: &MyObject) -> T;
   #[cxx_name = "setProp"]
   fn prop_setter(self: Pin<&mut MyObject>, value: T);   

   #[qsignal]
   #[cxx_name = "propChanged"]
   fn prop_changed(self: Pin<&mut MyObject);
}
```

In these cases our Rust side would be correct but our C++ Q_PROPERTY would be incorrect as it would the custom name from the qproperty instead of the cxx_name. (there is also the reverse fun for the rust_name).

So when the flag is Custom we need to somehow lookup the given name in i think TypeNames, eg with `TypeNames::lookup` which takes an Ident and gives you a Name (which would have the cxx and rust name from the other blocks.

`generate_cpp_properties` already has `type_names: &TypeNames,` so i think it's just a matter of passing that into QPropertyNames::from ?

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
```diff
@@ -32,19 +32,27 @@ pub fn generate_cpp_properties(
...
-            .private_methods
-            .push(setter::generate_wrapper(&idents, &cxx_ty));
-        signals.push(signal::generate(&idents, qobject_idents));
+
+        if let Some(getter) = getter::generate(&idents, &qobject_ident, &cxx_ty) {
+            generated.methods.push(getter);
+            generated
+                .private_methods
+                .push(getter::generate_wrapper(&idents, &cxx_ty));
+        }
```
ahayzen-kdab, line 41: in theory you could add a else case here that is unreachable to ensure it doesn't happen?
```rust
} else {
    unreachable!("QProperty should always have a getter");
}
```

crates/cxx-qt-gen/src/generator/cpp/property/getter.rs
```diff
@@ -3,27 +3,38 @@
...
+                    }}
+                    "#,
+                ident_getter = name.cxx_unqualified(),
+                ident_getter_wrapper = idents.getter_wrapper.cxx_unqualified(),
+            ),
+        })
+    } else {
+        None
     }
 }
```
ahayzen-kdab, line 40: shouldn't the `generate_wrapper` below also use `if let NameState::Auto(name) = &idents.getter_wrapper {` for completeness ?

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
```diff
@@ -32,19 +32,27 @@ pub fn generate_cpp_properties(
...
-        generated
-            .private_methods
-            .push(setter::generate_wrapper(&idents, &cxx_ty));
-        signals.push(signal::generate(&idents, qobject_idents));
+
+        if let Some(getter) = getter::generate(&idents, &qobject_ident, &cxx_ty) {
+            generated.methods.push(getter);
+            generated
+                .private_methods
+                .push(getter::generate_wrapper(&idents, &cxx_ty));
```
ahayzen-kdab, line 40: if the `generate_wrapper` also returns Option then this should be put into another if block like the setter and then have an else block with the unreachable too etc

crates/cxx-qt-gen/src/parser/property.rs
ahayzen-kdab: done

crates/cxx-qt-gen/src/generator/cpp/property/getter.rs
ahayzen-kdab: ah i see `getter_wrapper` is a `Name` not `NameState` or `Option<NameState>`, so probably ignore this then

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
ahayzen-kdab: also i guess ignore this as `getter_wrapper` is `Name`

crates/cxx-qt-gen/src/generator/naming/property.rs
```diff
@@ -2,32 +2,83 @@
...
+        });
+
+        let setter_wrapper = if let Some(NameState::Auto(ref setter)) = setter {
+            Some(wrapper_name_from_function_name(setter))
+        } else {
+            None
+        };
+
         Self {
             getter_wrapper: wrapper_name_from_function_name(&getter),
```
ahayzen-kdab, line 77: if the `getter` is Custom here, shouldn't this then be None?

If so then getter_wrapper needs to become `Option<Name>` ? so then some of my previous comments are still valid ?

crates/cxx-qt-gen/src/generator/naming/property.rs
ahayzen-kdab: We'll do this in another branch, and comment has been added, so done.

crates/cxx-qt-gen/src/generator/rust/property/setter.rs
```diff
@@ -19,46 +22,64 @@ pub fn generate(
...
+                    #has_unsafe fn #setter_rust(self: Pin<&mut #cpp_class_name_rust>, value: #cxx_ty);
+                }
+            }],
+            implementation: vec![quote! {
+                impl #qualified_impl {
+                    #[doc = "Setter for the Q_PROPERTY "]
+                    #[doc = #ident_str]
+                    pub fn #setter_rust(mut self: core::pin::Pin<&mut Self>, value: #qualified_ty) {
+                        use cxx_qt::CxxQtType;
+
```
ahayzen-kdab, line 70: remove extra empty line that wasn't there before

crates/cxx-qt-gen/src/generator/cpp/property/signal.rs
```diff
@@ -6,24 +6,32 @@
...
+        qobject::QObjectNames,
+    },
     parser::signals::ParsedSignal,
 };
 
-pub fn generate(idents: &QPropertyNames, qobject_idents: &QObjectNames) -> ParsedSignal {
+pub fn generate(idents: &QPropertyNames, qobject_idents: &QObjectNames) -> Option<ParsedSignal> {
     // We build our signal in the generation phase as we need to use the naming
     // structs to build the signal name
     let cpp_class_rust = &qobject_idents.name.rust_unqualified();
```
ahayzen-kdab, line 19: why is this not inside the if statement?

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -3,14 +3,81 @@
...
+};
+
+#[derive(Debug, Clone)]
+pub enum FlagState {
+    Auto, // Might need to refactor to also store the generated ident here
+    Custom(Ident),
+}
+
+impl FlagState {
+    pub fn is_auto(&self) -> bool {
```
ahayzen-kdab, line 20: add `#[cfg(test)]` to this as it's only used by tests

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -3,14 +3,81 @@
...
 pub struct ParsedQProperty {
     /// The [syn::Ident] of the property
     pub ident: Ident,
     /// The [syn::Type] of the property
     pub ty: Type,
+    /// Property flag collection
+    pub flags: QPropertyFlags,
+}
+
+// TODO: Returning struct instead of tuple might be more descriptive
```
ahayzen-kdab, line 65: do you still think changing this or is a tuple OK?

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -20,10 +87,84 @@ impl ParsedQProperty {
...
+                for flag in flags {
+                    // Could maybe refactor parse_meta_name_value to parse_meta,
+                    // which would return an (Ident, Option<Ident>) and extract the logic below to a new fn
+                    match flag {
+                        Meta::Path(path) => {
+                            // Set flag as auto
+                            update_fields(path.require_ident()?, None)?;
+                        },
+                        Meta::NameValue(name_value) => {
+                            let kv_pair = parse_meta_name_value(&name_value)?;
```
ahayzen-kdab, line 137: could unpack here with like `let (field, value) = parse_meta_name_value(...)` then you don't need to use `.1` and `.0` etc

crates/cxx-qt-gen/src/parser/property.rs
ahayzen-kdab: Code has changed.

crates/cxx-qt-gen/src/parser/property.rs
ahayzen-kdab: done

crates/cxx-qt-gen/test_inputs/properties.rs
ahayzen-kdab: done

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
ahayzen-kdab: Changed

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
ahayzen-kdab: done

crates/cxx-qt-gen/src/generator/cpp/property/setter.rs
ahayzen-kdab: done

crates/cxx-qt-gen/src/generator/naming/property.rs
ahayzen-kdab: done

crates/cxx-qt-gen/src/generator/naming/property.rs
ahayzen-kdab: changed

crates/cxx-qt-gen/src/generator/rust/property/mod.rs
ahayzen-kdab: changed

crates/cxx-qt-gen/src/parser/property.rs
ahayzen-kdab: done

crates/cxx-qt-gen/src/parser/property.rs
ahayzen-kdab: done

crates/cxx-qt-gen/src/parser/property.rs
ahayzen-kdab: done

crates/cxx-qt-gen/src/parser/property.rs
ahayzen-kdab: changed

crates/cxx-qt-gen/src/parser/property.rs
ahayzen-kdab: this is unused now? think clippy was picking that up

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -20,10 +118,46 @@ impl ParsedQProperty {
...
+                // Can be removed in lieu of checking fields if the enum approach is used
+                let mut flag_strings: Vec<String> = Vec::new();
 
-            Ok(Self { ident, ty })
+                for flag in flags {
+                    let flag_string: String = passed_flags.add_from_meta(flag)?;
+                    flag_strings.push(flag_string);
+                }
+
+                if !flag_strings.contains(&String::from("read")) {
```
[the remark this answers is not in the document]
ahayzen-kdab: changed

crates/cxx-qt-gen/src/parser/property.rs
ahayzen-kdab: done

LeonMatthesKDAB (commented):
Overall looking good!
Mostly nitpicks left I think.

We can then follow up on this with the proper naming lookup in a second PR.

book/src/bridge/extern_rustqt.md
```diff
@@ -131,7 +131,19 @@ Where `<Property>` is the name of the property.
 
 These setters and getters assure that the changed signal is emitted every time the property is edited.
 
-> Note that in the future it will be possible to specify custom getters and setters
+It is also possible to specify custom getters, setters and on-changed functions, using flags passed like so:
```
LeonMatthesKDAB, line 134:
```suggestion
It is also possible to specify custom getters, setters and notify signals, using flags passed like so:
```

It's not really custom on_ functions, but a custom signal that you can specify.

book/src/bridge/extern_rustqt.md
```diff
@@ -131,7 +131,19 @@ Where `<Property>` is the name of the property.
...
+`#[qproperty(TYPE, NAME, read = myGetter, write = mySetter, notify = myOnChanged)]`
+
+> Note: currently TypeName lookups are not currently supported, so either the name must be camel case or specified like `#[cxx_name = "my_getter"]`
```
LeonMatthesKDAB, line 137: This is written from the perspective of a developer who knows what the TypeName struct is (so pretty much you, me and Andrew :wink: )
A user who reads this documentation won't know what this means.

book/src/bridge/extern_rustqt.md
```diff
@@ -131,7 +131,19 @@ Where `<Property>` is the name of the property.
...
+
+It is also possible to use any combination of custom functions or omitting entirely, but if flags are specified, read must be one of them as all properties need to be able to be read.
```
LeonMatthesKDAB, line 139: Please also describe what exactly `read` vs `read=my_getter` does.
I.e. if you use `read`, it will auto-generate an appropriate getter with the name of the property and assumes that a member of the same name existst in the Rust struct.
Whereas with `read=my_getter` nothing will be autogenerated and we expect a function with this exact name and a compatible signature to be

1. Implemented on the qobject::T
2. Exported to C++ within the bridge.

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -228,22 +213,29 @@ mod tests {
         assert_eq!(property.ty, parse_quote! { T });
 
         // Can use  assert_matches! when https://github.com/rust-lang/rust/issues/82775 gets stabilised
-        let expected_read = format_ident!("my_getter");
-        assert!(matches!(
+        // let expected_read = format_ident!("my_getter");
+        // assert!(matches!(
+        //     property.flags.read,
+        //     FlagState::Custom(ident) if ident == expected_read
+        // ));
```
LeonMatthesKDAB, line 220: Remove commented code

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -228,22 +213,29 @@ mod tests {
...
 
         assert!(property.flags.notify.is_some());
 
-        let expected_notify = format_ident!("my_notifier");
-        assert!(matches!(
+        // let expected_notify = format_ident!("my_notifier");
+        // assert!(matches!(
+        //     property.flags.notify,
+        //     Some(FlagState::Custom(ident)) if ident == expected_notify
+        // ));
```
LeonMatthesKDAB, line 234: Remove commented code

LeonMatthesKDAB (commented):
A decent start on the documentation, but it should to be phrased a bit more user-friendly.
The people reading this documentation are doing so because they don't know how certain things work or should be used.
So we need to try and describe features in a way that requires as little prerequisite knowledge as possible.

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/KDAB/cxx-qt/compare/v0.6.1...HEAD)
 
+- Add support for specifying read write and notify in qproperty macro, including support for custom user defined functions
```
ahayzen-kdab, line 20: this should be under the `### Added` section :-)

book/src/bridge/extern_rustqt.md
```diff
@@ -131,7 +131,23 @@ Where `<Property>` is the name of the property.
...
 
-> Note that in the future it will be possible to specify custom getters and setters
+It is also possible to specify custom getters, setters and notify signals, using flags passed like so:
+`#[qproperty(TYPE, NAME, read = myGetter, write = mySetter, notify = myOnChanged)]`
+
+> Note: currently the rust name must be in camel case or specified like `#[cxx_name = "my_getter"]` if not
+
+It is also possible to use any combination of custom functions or omit them entirely, but if flags are specified, read must be included as all properties need to be able to be read.
+
+Using the read flag will cause cxx_qt to generate a getter function with an automatic name based off the property. e.g. `#[qproperty(i32, num, read)]` will have a getter function generated called get_num in Rust, and getNum in C++.
```
ahayzen-kdab, line 141: use CXX-Qt when referring to the project in the book :-)

ahayzen-kdab: done

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
ahayzen-kdab: done, this was changed

crates/cxx-qt-gen/src/parser/property.rs
ahayzen-kdab: done

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
ahayzen-kdab: changed

crates/cxx-qt-gen/src/generator/naming/property.rs
ahayzen-kdab: done

crates/cxx-qt-gen/src/generator/cpp/property/getter.rs
ahayzen-kdab: changed

crates/cxx-qt-gen/src/generator/rust/property/setter.rs
ahayzen-kdab: done

crates/cxx-qt-gen/src/generator/cpp/property/signal.rs
ahayzen-kdab: done

crates/cxx-qt-gen/src/parser/property.rs
ahayzen-kdab: done

crates/cxx-qt-gen/src/parser/property.rs
ahayzen-kdab: done

crates/cxx-qt-gen/src/parser/property.rs
ahayzen-kdab: done

CHANGELOG.md
ahayzen-kdab: done

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
BenFordTytherington: The if let block is for generating the method, so the else case would basically be catching the getter not being generated (custom getter provided) so it's not an error to return None from generate here

ahayzen-kdab: cool, done

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -20,10 +87,84 @@ impl ParsedQProperty {
...
+                        }
+                    };
+                    *variable = Some(value.map_or(FlagState::Auto, FlagState::Custom));
+
+                    Ok(())
+                };
+
+                for flag in flags {
+                    // Could maybe refactor parse_meta_name_value to parse_meta,
+                    // which would return an (Ident, Option<Ident>) and extract the logic below to a new fn
```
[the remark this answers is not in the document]
BenFordTytherington: Done

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -38,14 +179,103 @@ mod tests {
...
+            #[qproperty(T, name, read, write, notify)]
+            struct MyStruct;
+        };
+        let property = ParsedQProperty::parse(input.attrs.remove(0)).unwrap();
+        assert_eq!(property.ident, format_ident!("name"));
+        assert_eq!(property.ty, parse_quote! { T });
+
+        assert!(property.flags.read.is_auto());
+
+        assert!(property.flags.write.is_some());
```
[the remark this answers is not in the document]
BenFordTytherington: Done

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -38,14 +179,103 @@ mod tests {
...
+        assert!(property.flags.write.unwrap().is_auto());
+
+        assert!(property.flags.notify.is_some());
+
+        let expected_notify = format_ident!("my_notifier");
+        assert!(matches!(
+            property.flags.notify,
+            Some(FlagState::Custom(ident)) if ident == expected_notify
+        ));
```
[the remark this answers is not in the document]
BenFordTytherington: Done

book/src/bridge/extern_rustqt.md
BenFordTytherington: Done

book/src/bridge/extern_rustqt.md
BenFordTytherington: Done

book/src/bridge/extern_rustqt.md
BenFordTytherington: Done

crates/cxx-qt-gen/src/parser/property.rs
BenFordTytherington: Done

crates/cxx-qt-gen/src/parser/property.rs
BenFordTytherington: Done

crates/cxx-qt-gen/src/parser/property.rs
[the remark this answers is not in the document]
LeonMatthesKDAB: This hasn't been incorporated in the assert! statements yet, right?

crates/cxx-qt-gen/src/generator/rust/property/setter.rs
BenFordTytherington: left an issue to remember this: https://github.com/KDAB/cxx-qt/issues/1000

ahayzen-kdab: Cool :-)

crates/cxx-qt-gen/src/parser/property.rs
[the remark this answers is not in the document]
BenFordTytherington: should be fixed now

PR #995, opened 2024-07-09
Implement QAnyStringView

### LINKED ISSUE #629: cxx-qt-lib bindings for QAnyStringView
[QAnyStringView](https://doc.qt.io/qt-6/qanystringview.html) is a new class in Qt6 that is being used in new Qt APIs that take string arguments, for example [QQmlApplicationEngine::loadFromModule](https://doc-snapshots.qt.io/qt6-6.5/qqmlapplicationengine.html#loadFromModule). It supports UTF-8 strings, so it may be possible to implement cheap conversion from Rust &str to QAnyStringView without needing to copy data or convert to UTF-16 for a QString first.

redstrate:
This is a read-only string type that's used in several new Qt6 APIs such as QQmlApplicationEngine::loadFromModule.

Fixes #629

### DIFF
--- a/crates/cxx-qt-lib/build.rs
+++ b/crates/cxx-qt-lib/build.rs
@@ -69,6 +69,9 @@ fn write_headers() {
 }
 
 fn main() {
+    let qtbuild = qt_build_utils::QtBuild::new(vec!["Core".to_owned()])
+        .expect("Could not find Qt installation");
+
     write_headers();
 
     let emscripten_targeted = match std::env::var("CARGO_CFG_TARGET_OS") {
@@ -201,6 +204,10 @@ fn main() {
         "core/qvector/qvector_u64",
     ];
 
+    if qtbuild.version().major > 5 {
+        rust_bridges.extend(["core/qanystringview"]);
+    }
+
     if qt_gui_enabled() {
         rust_bridges.extend([
             "core/qlist/qlist_qcolor",
@@ -268,6 +275,10 @@ fn main() {
         "core/qvariant/qvariant",
         "core/qvector/qvector",
     ];
+
+    if qtbuild.version().major > 5 {
+        cpp_files.extend(["core/qanystringview"]);
+    }
 
     if qt_gui_enabled() {
         cpp_files.extend([
--- a/crates/cxx-qt-lib/src/core/qanystringview.cpp
+++ b/crates/cxx-qt-lib/src/core/qanystringview.cpp
@@ -1,0 +1,48 @@
+// clang-format off
+// SPDX-FileCopyrightText: 2024 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// clang-format on
+// SPDX-FileContributor: Joshua Goins <joshua.goins@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+#include "cxx-qt-lib/qanystringview.h"
+
+#include <cxx-qt-lib/assertion_utils.h>
+
+// QAnyStringView has two members.
+// A union of (void*, char*, char_16*) and a size_t.
+// https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/text/qanystringview.h
+assert_alignment_and_size(QAnyStringView, {
+  ::std::size_t a0;
+  void* a1;
+});
+
+static_assert(::std::is_trivially_copy_assignable<QAnyStringView>::value);
+static_assert(::std::is_trivially_copy_constructible<QAnyStringView>::value);
+
+static_assert(::std::is_trivially_destructible<QAnyStringView>::value);
+
+static_assert(QTypeInfo<QAnyStringView>::isRelocatable);
+
+namespace rust {
+namespace cxxqtlib1 {
+
+QAnyStringView
+qanystringviewInitFromRustString(::rust::Str string)
+{
+  return QAnyStringView(string.data(), string.size());
+}
+
+QAnyStringView
+qanystringviewInitFromQString(const QString& string)
+{
+  return QAnyStringView(string);
+}
+
+::rust::isize
+qanystringviewLen(const QAnyStringView& string)
+{
+  return static_cast<::rust::isize>(string.size());
+}
+
+}
+}
--- a/crates/cxx-qt-lib/src/core/qanystringview.rs
+++ b/crates/cxx-qt-lib/src/core/qanystringview.rs
@@ -1,0 +1,118 @@
+// SPDX-FileCopyrightText: 2024 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Joshua Goins <josh@redstrate.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+use crate::QString;
+use core::ffi::c_void;
+use core::marker::PhantomData;
+use core::mem::MaybeUninit;
+use cxx::{type_id, ExternType};
+
+#[cxx::bridge]
+mod ffi {
+    unsafe extern "C++" {
+        include!("cxx-qt-lib/qanystringview.h");
+        type QAnyStringView<'a> = super::QAnyStringView<'a>;
+
+        include!("cxx-qt-lib/qstring.h");
+        type QString = crate::QString;
+
+        /// Returns true if the string has no characters; otherwise returns false.
+        #[rust_name = "is_empty"]
+        fn isEmpty(self: &QAnyStringView) -> bool;
+
+        /// Returns true if this string is null; otherwise returns false.
+        #[rust_name = "is_null"]
+        fn isNull(self: &QAnyStringView) -> bool;
+    }
+
+    #[namespace = "rust::cxxqtlib1"]
+    unsafe extern "C++" {
+        include!("cxx-qt-lib/common.h");
+
+        #[doc(hidden)]
+        #[rust_name = "QAnyStringView_init_default"]
+        fn construct() -> QAnyStringView<'static>;
+        #[doc(hidden)]
+        #[rust_name = "QAnyStringView_init_from_rust_string"]
+        fn qanystringviewInitFromRustString<'a>(string: &str) -> QAnyStringView<'a>;
+        #[doc(hidden)]
+        #[rust_name = "QAnyStringView_init_from_qstring"]
+        fn qanystringviewInitFromQString<'a>(string: &QString) -> QAnyStringView<'a>;
+        #[doc(hidden)]
+        #[rust_name = "QAnyStringView_init_from_QAnyStringView"]
+        fn construct<'a>(string: &QAnyStringView) -> QAnyStringView<'a>;
+
+        #[doc(hidden)]
+        #[rust_name = "QAnyStringView_eq"]
+        fn operatorEq(a: &QAnyStringView, b: &QAnyStringView) -> bool;
+
+        #[doc(hidden)]
+        #[rust_name = "QAnyStringView_len"]
+        fn qanystringviewLen(string: &QAnyStringView) -> isize;
+    }
+}
+
+/// The QAnyStringView class provides a unified view of a Latin-1, UTF-8, or UTF-16 string.
+#[repr(C)]
+pub struct QAnyStringView<'a> {
+    /// QAnyStringView has two members, a pointer and a size_t
+    _space: MaybeUninit<[usize; 1]>,
+    _space2: MaybeUninit<[c_void; 1]>,
+
+    /// Needed to keep the lifetime in check
+    _phantom: PhantomData<&'a usize>,
+}
+
+impl<'a> Clone for QAnyStringView<'a> {
+    /// Constructs a copy of other.
+    ///
+    /// This operation takes constant time, because QAnyStringView is a view-only string.
+    fn clone(&self) -> QAnyStringView<'a> {
+        ffi::QAnyStringView_init_from_QAnyStringView(self)
+    }
+}
+
+impl Default for QAnyStringView<'_> {
+    /// Constructs a null string. Null strings are also empty.
+    fn default() -> Self {
+        ffi::QAnyStringView_init_default()
+    }
+}
+
+impl PartialEq for QAnyStringView<'_> {
+    fn eq(&self, other: &Self) -> bool {
+        ffi::QAnyStringView_eq(self, other)
+    }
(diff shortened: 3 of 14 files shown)


crates/cxx-qt-lib/build.rs
```diff
@@ -71,6 +71,7 @@ fn main() {
     };
 
     let mut rust_bridges = vec![
+        "core/qanystringview",
```
redstrate, line 74: Note to me tomorrow: lock this behind Qt6

crates/cxx-qt-lib/src/core/qanystringview.cpp
```diff
@@ -0,0 +1,46 @@
...
+{
+  // Note that rust::Str here is borrowed
+  // and we convert back from UTF-8 to UTF-16
+  return QAnyStringView(string.data(), string.size());
+}
+
+::rust::isize
+qanystringviewLen(const QAnyStringView& string)
+{
+  // In Qt 5 the type was int now it is qsizetype, so we need to ensure the type
```
ahayzen-kdab, line 40: guess this comment isn't required if it's only Qt 6?

crates/cxx-qt-lib/src/core/qanystringview.cpp
```diff
@@ -0,0 +1,46 @@
+// clang-format off
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// clang-format on
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+// SPDX-FileContributor: Leon Matthes <leon.matthes@kdab.com>
```
ahayzen-kdab, line 5: change to your name and email :-)

crates/cxx-qt-lib/src/core/qanystringview.cpp
```diff
@@ -0,0 +1,46 @@
+// clang-format off
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
```
ahayzen-kdab, line 2: 2034 :-)

crates/cxx-qt-lib/src/core/qanystringview.rs
```diff
@@ -0,0 +1,142 @@
...
+use cxx::{type_id, ExternType};
+use std::ffi::c_void;
+use std::mem::MaybeUninit;
+
+#[cxx::bridge]
+mod ffi {
+    #[namespace = "Qt"]
+    unsafe extern "C++" {
+        include!("cxx-qt-lib/qt.h");
+    }
```
ahayzen-kdab, line 14: this block looks unused? i think you can remove as you aren't using any enums from the Qt namespace?

crates/cxx-qt-lib/src/core/qanystringview.rs
```diff
@@ -0,0 +1,142 @@
...
+        #[rust_name = "QAnyStringView_init_from_rust_string"]
+        fn qanystringviewInitFromRustString(string: &str) -> QAnyStringView;
+        #[doc(hidden)]
+        #[rust_name = "QAnyStringView_init_from_QAnyStringView"]
+        fn construct(string: &QAnyStringView) -> QAnyStringView;
+
+        #[doc(hidden)]
+        #[rust_name = "QAnyStringView_eq"]
+        fn operatorEq(a: &QAnyStringView, b: &QAnyStringView) -> bool;
+        #[doc(hidden)]
```
ahayzen-kdab, line 50: remove extra attribute by itself

crates/cxx-qt-lib/src/core/qanystringview.rs
```diff
@@ -0,0 +1,142 @@
...
+        #[rust_name = "QAnyStringView_len"]
+        fn qanystringviewLen(string: &QAnyStringView) -> isize;
+    }
+}
+
+/// The QAnyStringView class provides a Unicode character string.
+///
+/// Note that QAnyStringView is a UTF-16 whereas Rust strings are a UTF-8
+#[repr(C)]
+pub struct QAnyStringView {
```
ahayzen-kdab, line 62: would be good to add a test to `qt_types_standalone` which constructs, clones, and then drops the type (see the other ones in there). As then the type is used as part of a test which means it is run through valgrind :-)

ahayzen-kdab, line 62: add an entry to the changelog to mention that you have added a new type :-)

crates/cxx-qt-lib/build.rs
ahayzen-kdab, line 74: guess there is also the question of whether it'll be used by other types in cxx-qt-lib or whether it should be in cxx-qt-lib-extras? It sounds like it is used by types already in cxx-qt-lib in Qt 6 though so would be useful ?

LeonMatthesKDAB (changes requested):
Hi @redstrate , thanks for working on this!
It can hopefully really improve some of the Rust-C++ string interop :crossed_fingers: 

For now, some initial comments, that you may already be aware of.
If you think this PR isn't ready for review yet, also consider converting it to a draft under the "reviewers" section.

crates/cxx-qt-lib/src/core/qanystringview.cpp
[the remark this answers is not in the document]
redstrate: Correct, copy & paste error. Fixed

crates/cxx-qt-lib/src/core/qanystringview.rs
```diff
@@ -0,0 +1,142 @@
...
+
+impl From<&str> for QAnyStringView {
+    /// Constructs a QAnyStringView from a Rust string
+    ///
+    /// Note that this converts from UTF-8 to UTF-16
+    fn from(str: &str) -> Self {
+        ffi::QAnyStringView_init_from_rust_string(str)
+    }
+}
+
```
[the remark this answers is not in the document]
redstrate: Done and added to the tests

crates/cxx-qt-lib/src/core/qanystringview.rs
```diff
@@ -0,0 +1,142 @@
...
+impl From<&String> for QAnyStringView {
```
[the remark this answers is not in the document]
redstrate: Agree, removed

crates/cxx-qt-lib/src/core/qanystringview.rs
```diff
@@ -0,0 +1,142 @@
...
+    /// Constructs a QAnyStringView from a Rust string
+    ///
+    /// Note that this converts from UTF-8 to UTF-16
+    fn from(str: &String) -> Self {
+        ffi::QAnyStringView_init_from_rust_string(str)
+    }
+}
+
+impl From<String> for QAnyStringView {
```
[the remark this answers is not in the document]
redstrate: Correct, removed!

crates/cxx-qt-lib/src/core/qanystringview.rs
```diff
@@ -0,0 +1,142 @@
...
+///
+/// Note that QAnyStringView is a UTF-16 whereas Rust strings are a UTF-8
+#[repr(C)]
+pub struct QAnyStringView {
+    /// QAnyStringView has two members, a pointer and a size_t
+    _space: MaybeUninit<[usize; 1]>,
+    _space2: MaybeUninit<[c_void; 1]>,
+}
+
+impl Clone for QAnyStringView {
```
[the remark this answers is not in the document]
redstrate: Hmm, I would agree but it turns out `MaybeUninit` is not happy:

```
error[E0204]: the trait `std::marker::Copy` cannot be implemented for this type
  --> crates/cxx-qt-lib/src/core/qanystringview.rs:76:19
   |
61 |     _space2: MaybeUninit<[c_void; 1]>,
   |     --------------------------------- this field does not implement `std::marker::Copy`
...
76 | impl<'a> Copy for QAnyStringView<'a> {}
   |                   ^^^^^^^^^^^^^^^^^^
   |
note: the `std::marker::Copy` impl for `MaybeUninit<[std::ffi::c_void; 1]>` requires that `std::ffi::c_void: std::marker::Copy`
  --> crates/cxx-qt-lib/src/core/qanystringview.rs:61:14
   |
61 |     _space2: MaybeUninit<[c_void; 1]>,
   |              ^^^^^^^^^^^^^^^^^^^^^^^^
```

Maybe should skip for now?

crates/cxx-qt-lib/src/core/qanystringview.rs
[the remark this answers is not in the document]
redstrate: Done, the lifetimes are fixed now

crates/cxx-qt-lib/src/core/qanystringview.rs
```diff
@@ -0,0 +1,142 @@
...
+
+impl PartialEq for QAnyStringView {
+    fn eq(&self, other: &Self) -> bool {
+        ffi::QAnyStringView_eq(self, other)
+    }
+}
+
+impl Eq for QAnyStringView {}
+
+impl Drop for QAnyStringView {
```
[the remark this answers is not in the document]
redstrate: Copy & paste error for QString :see_no_evil: removed

crates/cxx-qt-lib/build.rs
redstrate: I plan on implementing loadFromModule, which is part of QQmlApplication and thus should stay in cxx-qt-lib. That's under the assumption that  -lib can't depend on -extras, I haven't checked

redstrate: Done

crates/cxx-qt-lib/src/core/qanystringview.cpp
redstrate: Done

crates/cxx-qt-lib/src/core/qanystringview.cpp
redstrate: Done

crates/cxx-qt-lib/src/core/qanystringview.rs
redstrate: Oops good catch, removed

crates/cxx-qt-lib/src/core/qanystringview.rs
redstrate: Done

crates/cxx-qt-lib/src/core/qanystringview.rs
redstrate: Done

crates/cxx-qt-lib/src/core/qanystringview.cpp
redstrate: Correct, it was a copy & paste error from QString. Removed

crates/cxx-qt-lib/src/core/qanystringview.rs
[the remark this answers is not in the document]
LeonMatthesKDAB: Ah, well then let's skip it for now, we can readd it later if needed.

crates/cxx-qt-lib/src/core/qanystringview.rs
redstrate: Done

crates/cxx-qt-lib/build.rs
```diff
@@ -191,6 +194,10 @@ fn main() {
         "core/qvector/qvector_u64",
     ];
 
+    if qtbuild.version().major > 5 {
```
ahayzen-kdab, line 197: wonder if it'd be useful to have this `version` exposed onto `CxxQtBuilder` so that you don't need to depend on `qt-build-utils` directly.

But then `CxxQtBuilder` doesn't actually create it's `QtBuild` until the `build()` method.

Something to think about separately though cc @LeonMatthesKDAB

crates/cxx-qt-lib/src/core/qanystringview.cpp
```diff
@@ -0,0 +1,47 @@
...
+qanystringviewInitFromQString(const QString& string)
+{
+  return QAnyStringView(string);
+}
+
+::rust::isize
+qanystringviewLen(const QAnyStringView& string)
+{
+  return static_cast<::rust::isize>(string.size());
+}
```
ahayzen-kdab, line 44: another reason to add `qsizetype` support as it removes this :-) #882

crates/cxx-qt-lib/src/core/qanystringview.cpp
```diff
@@ -0,0 +1,47 @@
...
+// A union of (void*, char*, char_16*) and a size_t.
+// https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/text/qanystringview.h
+assert_alignment_and_size(QAnyStringView,
+                          alignof(::std::size_t),
+                          sizeof(::std::size_t) + sizeof(void*));
+
+static_assert(::std::is_trivially_copy_assignable<QAnyStringView>::value);
+static_assert(::std::is_trivially_copy_constructible<QAnyStringView>::value);
+
+static_assert(::std::is_trivially_destructible<QAnyStringView>::value);
```
ahayzen-kdab, line 21: i assume it is not `is_trivially_copyable` ? As if it was then you can derive Clone etc. But i assume not as there are pointers here internally.

Then it is trivially copy constructable/assignable and destructible so it needs a Clone implementation but does it also need a Drop implementation ? We don't seem to have an existing type like this :-) cc @LeonMatthesKDAB

ahayzen-kdab: Aha i see there is no deconstructor :-D

ahayzen-kdab (dismissed):
LGTM, lets see what Leon thinks when he is back

crates/cxx-qt-lib/build.rs
redstrate: (note to future self: If we decide to do it later, might be worth adding TODOs so the implementer doesn't forget :-)

LeonMatthesKDAB (approved):
LGTM :)

LeonMatthesKDAB:
@redstrate The changes to the assert_alignment_and_size macro have broken the rebase. Can you please update the check :)

LeonMatthesKDAB:
I've updated this PR, hopefully it passes CI now :crossed_fingers:

ahayzen-kdab (approved):
LGTM :-)

PR #997, opened 2024-07-12
fix: Remove dev-dependencies on cxx-qt-lib

LeonMatthesKDAB:
This was causing a duplicate build of cxx-qt-lib, which caused the file locking issues we observed in CI.

This should also speed up the CI run for cargo_test, as it's not building -lib twice now.

However, this doesn't yet solve the inherent issue that Cargo may run build scripts of different configurations in parallel. So this remains a TODO!


ahayzen-kdab (approved):
LGTM :-)

PR #999, opened 2024-07-12
Allow cxx-qt-build to deal with build scripts running in parallel

LeonMatthesKDAB:
As we discovered after merging our latest build system refactorings, Cargo may actually run build scripts in parallel when multiple configurations are used.

We want to deal with this by using the OUT_DIR for cargo builds, as is encouraged by Cargo.

For CMake integration we want to only export the crate that is imported by CMake and require the user to manually provide different export directories when building multiple configurations.


LeonMatthesKDAB:
Oh, Github item #999, nice. One more and we made it to 1000 :partying_face:

book/src/getting-started/5-cmake-integration.md
```diff
@@ -130,27 +132,34 @@ For this example, we are [supporting both Qt5 and Qt6 with CMake](https://doc.qt
 {{#include ../../../examples/qml_minimal/CMakeLists.txt:book_cmake_setup}}
 ```
 
-Locate [Corrosion](https://github.com/corrosion-rs/corrosion), a tool for integrating Rust libraries into CMake.
-If Corrosion is not installed, automatically download it:
+Download CXX-Qts CMake code with FetchContent:
 
 ```cmake,ignore
 {{#include ../../../examples/qml_minimal/CMakeLists.txt:book_cmake_find_corrosion}}
```
ahayzen-kdab, line 138: `cmake_find_corrosion` should probably be changed ? `cmake_find_cxx_qt` or something?

book/src/getting-started/5-cmake-integration.md
```diff
@@ -130,27 +132,34 @@ For this example, we are [supporting both Qt5 and Qt6 with CMake](https://doc.qt
...
+    - `CXX_QT_EXPORT_DIR` - Manually specify the path where CXX-Qt artifacts will be exported to.
+        - This is usually not necessary. However, if you're importing the same crate with different feature sets in the same CMake build configuration, you will need to specify seperate `CXX_QT_EXPORT_DIR`s to avoid multiple versions of the crate exporting to the same directory.
+    - `QMAKE` - Override the path to the QMAKE executable
+2. `cxx_qt_import_qml_module` - This function imports a QML modules as a new target. It requires the following arguments:
+    - TARGET_NAME - Specify the name of the CMake target that this function will create
+    - `URI` - The URI of the qml module to import - this needs to exactly match the URI in the `CxxQtBuilder::qml_module` call in your build script.
+    - `SOURCE_CRATE` The crate that exports the QML module (this crate must have been imported with `cxx_qt_import_crate`).
 
 ```cmake,ignore
 {{#include ../../../examples/qml_minimal/CMakeLists.txt:book_cmake_use_corrosion}}
```
ahayzen-kdab, line 153: `use_corrosion` should this be changed too?

book/src/getting-started/4-cargo-executable.md
```diff
@@ -27,6 +27,8 @@ Add the dependencies to the `Cargo.toml` file.
 We'll need `cxx`, `cxx-qt`, `cxx-qt-lib` and `cxx-qt-build`:
 
 ```toml,ignore
+[package]
+name = "qml_minimal"
```
ahayzen-kdab, line 31: could use the package info from the examples/qml_minimal toml ? to try and keep as much as possible from real code?

book/src/getting-started/5-cmake-integration.md
```diff
@@ -130,27 +132,34 @@ For this example, we are [supporting both Qt5 and Qt6 with CMake](https://doc.qt
 {{#include ../../../examples/qml_minimal/CMakeLists.txt:book_cmake_setup}}
 ```
 
-Locate [Corrosion](https://github.com/corrosion-rs/corrosion), a tool for integrating Rust libraries into CMake.
-If Corrosion is not installed, automatically download it:
+Download CXX-Qts CMake code with FetchContent:
```
ahayzen-kdab, line 135: we don't actually use FetchContent yet? can we add an example of how to do that yet and how are we doing to do this? Can we point fetchcontent at a subsection of the git repo or do we need another git repo that stays in lockstep ? etc

book/src/getting-started/5-cmake-integration.md
LeonMatthesKDAB: Done

book/src/getting-started/4-cargo-executable.md
LeonMatthesKDAB: Done

book/src/getting-started/5-cmake-integration.md
LeonMatthesKDAB: If you render the book, you'll actually see that the code example says:
```
# TODO: Replace with fetch-content co cxx-qt-cmake
```

So that's at least a reminder for us that we still need to set this up.

I had really hoped that FetchContent has the ability to only download a certain directory out of a given Github repository, but apparently that's not possible.
which to me sounds like it would require us to have a second repository, unless we always want to download the entire repo, which is rather large.

But as that's a larger refactoring step, we'll want to do that as a separate PR I think.

LeonMatthesKDAB: Hmmm, I guess we could also have another `branch`, instead of another repository actually. :thinking: 
Similar to how our Github-pages branch has completely different content, it should be possible to have another branch and fetch that with the `GIT_TAG`. If used together with `GIT_SHALLOW`, it should also just download the necessary content :thinking: 

So it would look something like this:
```cmake
FetchContent_Declare(CxxQt
    GIT_REPOSITORY https://github.com/kdab/cxx-qt
    GIT_TAG cmake-v0.6 # or just `cmake` to get the latest from the branch
    GIT_SHALLOW ON
)
```

LeonMatthesKDAB: Unsure whether that's a bit too hacky though...

ahayzen-kdab: Hmm what do other kdab projects do that require cmake modules like this? Or do they always require you to build the C++ code anyway as you don't have cargo bringing in the libs :thinking:

ahayzen-kdab: Guess if they were in their own repo we could use a submodule to bring them into the cxx-qt repository at least so that we don't duplicate the code?

eg
```
cxx-qt/
  - cmake/cxx-qt-cmake/ -> submodule to the other repo

cxx-qt-cmake/
  - CxxQt.cmake
```

ahayzen-kdab: Or would we use fetch content in all of our examples so that would be irrelevant actually :thinking:

LeonMatthesKDAB: So previously we used this to find Corrosion:
```cmake
find_package(Corrosion QUIET)
if(NOT Corrosion_FOUND)
    include(FetchContent)
    FetchContent_Declare(
        Corrosion
        GIT_REPOSITORY https://github.com/corrosion-rs/corrosion.git
        GIT_TAG v0.5.0
    )

    FetchContent_MakeAvailable(Corrosion)
endif()
```

So we would change this to:
```cmake
find_package(CxxQt QUIET)
if(NOT CxxQt_FOUND)
    include(FetchContent)
    FetchContent_Declare(
        CxxQt
        GIT_REPOSITORY https://github.com/KDAB/cxx-qt-cmake.git
        GIT_TAG v0.7.0
    )
    FetchContent_MakeAvailable(CxxQt)
endif()
```
https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure

So we could include our CMake code as a submodule and set up `find_package` to search inside the submodule, which would mean our local builds wouldn't have to use FetchContent and would always have the right version.
However, then we'd need to sync the submodule with the FetchContent_Declare call as well :thinking: 

In any case, I think this is something for a follow-up PR to figure out.

ahayzen-kdab: Yup, figure out in another PR...

ahayzen-kdab (approved):
LGTM :-)

Issue #1002, opened 2024-07-23
"final" is a reserved keyword but also the name of an attribute in QProperty


BenFordTytherington:
since final is a reserved word in rust, it cannot be used in struct fields, which is easy enough to avoid, but also currently not supported in the macro e.g. `#[qproperty(T, NAME, read, final)]` will error in the parser, as it does not allow keywords in the declaration. Potential fix by changing macro to use all caps names like `#[qproperty(T, NAME, READ, WRITE = my_setter, FINAL)]`. This also matches Qt but looks a bit odd for Rust code.

ahayzen-kdab:
I wonder if we should copy what Qt is doing, as `final` is also a keyword in C++, and as we have found before copying Qt avoids issues later on. So maybe we use capitals which then for a Qt developer at least feels familiar :-) @LeonMatthesKDAB any thoughts?

LeonMatthesKDAB:
Hm, not a big fan of having to use all-caps, but until we have a better idea, let's go for it.
We can always support lower-case as well if we figure out a way around it.

Good catch @BenFordTytherington

ahayzen-kdab:
I think the way around it is probably `r#final` like #211 but tbh `#[qproperty(T, name, ..., r#final)]` seems much worse than `#[qproperty(T, name, ..., FINAL)]`. And can't remember what happens with syn with these r-values, think they might have been a slightly odd type.

So for me capitals is best for now as well.

LeonMatthesKDAB:
As we have a solution for this in #1003 I think we can close this.

PR #1003, opened 2024-07-23
Add support for new QProperty flags (Reset, Constant, Required, Final)

BenFordTytherington:
Constant, Reset and Required are now supported, with final as a WIP (see [issue #1002](https://github.com/KDAB/cxx-qt/issues/1002))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Serde support for `QString` (requires "serde" feature on cxx-qt-lib)
 - A new QuickControls module, which exposes `QQuickStyle`. This module is enabled by default and is behind the `qt_quickcontrols` feature.
 - Add support for specifying read write and notify in qproperty macro, including support for custom user defined functions
-- Add support for the constant, required, reset and final flags
+- Add support for the constant, required, reset and final flags in the qproperty macro
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Serde support for `QString` (requires "serde" feature on cxx-qt-lib)
 - A new QuickControls module, which exposes `QQuickStyle`. This module is enabled by default and is behind the `qt_quickcontrols` feature.
 - Add support for specifying read write and notify in qproperty macro, including support for custom user defined functions
+- Add support for the constant, required, reset and final flags
 
 ### Changed
 
--- a/book/src/bridge/extern_rustqt.md
+++ b/book/src/bridge/extern_rustqt.md
@@ -133,7 +133,7 @@ These setters and getters assure that the changed signal is emitted every time t
 
 It is also possible to specify custom getters, setters and notify signals, using flags passed like so:
 `#[qproperty(TYPE, NAME, READ = myGetter, WRITE = mySetter, NOTIFY = myOnChanged)]`
-Which must be in all capitals.
+> Note: the key for the flags use all capitals like in the Qt version of qproperty
 
 > Note: currently the rust name must be in camel case or specified like `#[cxx_name = "my_getter"]` if not
 
@@ -159,8 +159,8 @@ If a custom function is specified, an implementation both in qobject::MyObject a
 - `NOTIFY` or `NOTIFY = my_on_changed`
   - Specifies that the property should emit a notify signal on change, with optional user defined signal name
 - `CONSTANT`
-  - Specifies that the property should be constant (implication is that the getter returns the same value every time for that particular instance, but this is not yet enforced)
-  - __`CONSTANT` is not available for QProperties which use `WRITE` or `NOTIFY` and will not compile__
+  - Specifies that the property should be constant (implication is that the getter returns the same value every time for that particular instance)
+  - __`CONSTANT` is not available for properties which use `WRITE` or `NOTIFY` and will not compile__
 - `REQUIRED`
   - Specifies that the property must be set by a user of the class, useful in QML as the class cannot be instantiated unless the property has been set
 - `FINAL`
--- a/book/src/bridge/extern_rustqt.md
+++ b/book/src/bridge/extern_rustqt.md
@@ -132,22 +132,41 @@ Where `<Property>` is the name of the property.
 These setters and getters assure that the changed signal is emitted every time the property is edited.
 
 It is also possible to specify custom getters, setters and notify signals, using flags passed like so:
-`#[qproperty(TYPE, NAME, read = myGetter, write = mySetter, notify = myOnChanged)]`
+`#[qproperty(TYPE, NAME, READ = myGetter, WRITE = mySetter, NOTIFY = myOnChanged)]`
+Which must be in all capitals.
 
 > Note: currently the rust name must be in camel case or specified like `#[cxx_name = "my_getter"]` if not
 
 It is also possible to use any combination of custom functions or omit them entirely, but if flags are specified, read must be included as all properties need to be able to be read.
 
-Using the read flag will cause CXX-Qt to generate a getter function with an automatic name based off the property. e.g. `#[qproperty(i32, num, read)]` will have a getter function generated called get_num in Rust, and getNum in C++.
+Using the read flag will cause CXX-Qt to generate a getter function with an automatic name based off the property. e.g. `#[qproperty(i32, num, READ)]` will have a getter function generated called get_num in Rust, and getNum in C++.
 
 If a custom function is specified, an implementation both in qobject::MyObject and and export in the bridge is expected.
 
 ### Examples
 
-- `#[qproperty(TYPE, NAME, read)]` A read only prop
-- `#[qproperty(TYPE, NAME, read = myGetter, write, notify)]` custom getter provided but will generate setter and on-changed
-- `#[qproperty(TYPE, NAME)]` is syntactic sugar for `#[qproperty(TYPE, NAME, read, write, notify)]`
-- `#[qproperty(TYPE, NAME, write)]` is an error as read was not explicitly passed
+- `#[qproperty(TYPE, NAME, READ)]` A read only property
+- `#[qproperty(TYPE, NAME, READ = myGetter, WRITE, NOTIFY)]` custom getter provided, but will generate setter and on-changed
+- `#[qproperty(TYPE, NAME)]` is shorthand for `#[qproperty(TYPE, NAME, READ, WRITE, NOTIFY)]`
+- `#[qproperty(TYPE, NAME, WRITE)]` is an error as read was not explicitly passed
+
+### Available Flags
+
+- `READ` or `READ = my_getter`
+  - Specifies that the property should be readable (*always required if flags are passed*), with optional user defined getter
+- `WRITE` or `WRITE = my_setter`
+  - Specifies that the property should be writeable, with optional user defined setter
+- `NOTIFY` or `NOTIFY = my_on_changed`
+  - Specifies that the property should emit a notify signal on change, with optional user defined signal name
+- `CONSTANT`
+  - Specifies that the property should be constant (implication is that the getter returns the same value every time for that particular instance, but this is not yet enforced)
+  - __`CONSTANT` is not available for QProperties which use `WRITE` or `NOTIFY` and will not compile__
+- `REQUIRED`
+  - Specifies that the property must be set by a user of the class, useful in QML as the class cannot be instantiated unless the property has been set
+- `FINAL`
+  - Specifies that the property will not be overriden by a derived class
+- `RESET = my_reset`
+  - Specifies a function to reset the property to a default value, user function __must__ be provided or it will not compile
 
 ## Methods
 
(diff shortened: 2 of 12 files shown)


crates/cxx-qt-gen/src/generator/cpp/property/meta.rs
```diff
@@ -20,6 +21,18 @@ pub fn generate(idents: &QPropertyNames, cxx_ty: &str) -> String {
...
+        parts.push(format!("RESET {}", reset.cxx_unqualified()));
+    }
+
+    if flags.constant {
+        parts.push(String::from("CONSTANT"))
+    }
+
+    if flags.required {
+        parts.push(String::from("REQUIRED"))
+    }
```
ahayzen-kdab, line 34: you could add the `is_final` implementation here as that part is known at least

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -110,20 +121,36 @@ impl ParsedQProperty {
...
                 // Create mutable closure to capture the variables for setting with the Meta values
                 let mut update_fields = |ident: &Ident, value: Option<Ident>| -> Result<()> {
-                    let variable = match ident.to_string().as_str() {
-                        "read" => &mut read,
-                        "write" => &mut write,
-                        "notify" => &mut notify,
+                    match ident.to_string().as_str() {
+                        "read" => read = Some(value.map_or(FlagState::Auto, FlagState::Custom)), // Might be able to use DRY here
+                        "write" => write = Some(value.map_or(FlagState::Auto, FlagState::Custom)),
+                        "notify" => notify = Some(value.map_or(FlagState::Auto, FlagState::Custom)),
```
ahayzen-kdab, line 133: did the closure route not work? eg

```
"read" => map_auto_or_custom(&mut read),
```

could even pass in `&value` if capturing is an issue. But then at least you have a closure that does the
`*mutable = Some(value.map_or(FlagState::Auto, FlagState::Custom));`

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -29,6 +32,10 @@ pub struct QPropertyFlags {
     pub(crate) read: FlagState,
     pub(crate) write: Option<FlagState>,
     pub(crate) notify: Option<FlagState>,
+    pub(crate) reset: Option<Ident>, // TODO: in future might be able to generate the function if T has a default
+    pub(crate) is_final: bool, // TODO: `final` is a keyword, see https://github.com/KDAB/cxx-qt/issues/1002
```
ahayzen-kdab, line 36: fun, lets discuss in the issue

ahayzen-kdab: Consensus seems to be to use capitals for all of them for now to match Qt, so `READ`, `WRITE`, `FINAL` etc

crates/cxx-qt-gen/src/parser/property.rs
BenFordTytherington: I just forgot to clean that bit up actually, will do

crates/cxx-qt-gen/src/parser/property.rs
ahayzen-kdab: the TODO comment for is_final could be removed here

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -110,20 +114,42 @@ impl ParsedQProperty {
...
                 let mut notify = None;
 
+                let mut constant = false;
+                let mut required = false;
+                let mut is_final = false;
+                let mut reset = None;
+
+                let map_auto_or_custom = |variable: &mut Option<FlagState>, value: &Option<Ident>| {
+                    *variable = Some(value.as_ref().map_or(FlagState::Auto, |ident| FlagState::Custom(ident.clone())));
+                };
```
ahayzen-kdab, line 124: if this closure was inside the other closure, then it could probably capture the `value`, but also fine to be outside if you think it's cleaner to have no capturing

BenFordTytherington: it did capture it, but it moves it, so then can't be used later in that closure in the reset block

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -110,20 +114,42 @@ impl ParsedQProperty {
...
+                        "RESET" => {
+                            // Might be refactorable to be simpler
+                            if let Some(ident) = value {
+                                reset = Some(ident);
+                            } else {
+                                return Err(Error::new(
+                                    ident.span(),
+                                    "The identifier of a reset function must be provided explicitly",
+                                ));
+                            }
```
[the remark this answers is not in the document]
BenFordTytherington: that would create a Some(Ok(Ident)) so not that I don't think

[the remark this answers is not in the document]
ahayzen-kdab: yes it would create `Some(OK(Ident))` initially but there is then the `?` at the end of the `ok_or_else` which unwraps it to either return an error or become `Some(Ident)`

crates/cxx-qt-gen/src/parser/property.rs
ahayzen-kdab: Ah makes sense

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Serde support for `QString` (requires "serde" feature on cxx-qt-lib)
 - A new QuickControls module, which exposes `QQuickStyle`. This module is enabled by default and is behind the `qt_quickcontrols` feature.
 - Add support for specifying read write and notify in qproperty macro, including support for custom user defined functions
+- Add support for the constant, required, reset and final flags
```
ahayzen-kdab, line 30: mention that this is for the qproperty macro

book/src/bridge/extern_rustqt.md
```diff
@@ -132,22 +132,41 @@ Where `<Property>` is the name of the property.
 These setters and getters assure that the changed signal is emitted every time the property is edited.
 
 It is also possible to specify custom getters, setters and notify signals, using flags passed like so:
-`#[qproperty(TYPE, NAME, read = myGetter, write = mySetter, notify = myOnChanged)]`
+`#[qproperty(TYPE, NAME, READ = myGetter, WRITE = mySetter, NOTIFY = myOnChanged)]`
+Which must be in all capitals.
```
ahayzen-kdab, line 136: Maybe instead write

```
> Note the key for the flags are defined using all capitals like in Qt
```

book/src/bridge/extern_rustqt.md
```diff
@@ -132,22 +132,41 @@ Where `<Property>` is the name of the property.
...
+
+- `READ` or `READ = my_getter`
+  - Specifies that the property should be readable (*always required if flags are passed*), with optional user defined getter
+- `WRITE` or `WRITE = my_setter`
+  - Specifies that the property should be writeable, with optional user defined setter
+- `NOTIFY` or `NOTIFY = my_on_changed`
+  - Specifies that the property should emit a notify signal on change, with optional user defined signal name
+- `CONSTANT`
+  - Specifies that the property should be constant (implication is that the getter returns the same value every time for that particular instance, but this is not yet enforced)
+  - __`CONSTANT` is not available for QProperties which use `WRITE` or `NOTIFY` and will not compile__
```
ahayzen-kdab, line 163: swap "QProperties" for "properties" as elsewhere we are calling them property

CHANGELOG.md
BenFordTytherington: done

crates/cxx-qt-gen/src/parser/property.rs
BenFordTytherington: done

crates/cxx-qt-gen/src/parser/property.rs
BenFordTytherington: done

crates/cxx-qt-gen/src/parser/property.rs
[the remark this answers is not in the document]
BenFordTytherington: done

book/src/bridge/extern_rustqt.md
```diff
@@ -132,22 +132,41 @@ Where `<Property>` is the name of the property.
...
+### Available Flags
+
+- `READ` or `READ = my_getter`
+  - Specifies that the property should be readable (*always required if flags are passed*), with optional user defined getter
+- `WRITE` or `WRITE = my_setter`
+  - Specifies that the property should be writeable, with optional user defined setter
+- `NOTIFY` or `NOTIFY = my_on_changed`
+  - Specifies that the property should emit a notify signal on change, with optional user defined signal name
+- `CONSTANT`
+  - Specifies that the property should be constant (implication is that the getter returns the same value every time for that particular instance, but this is not yet enforced)
```
ahayzen-kdab, line 162: i think it's unlikely that we can enforce this, so maybe instead change to say that it is implied that the read/getter should return the same value every time for that instance.

BenFordTytherington: done

book/src/bridge/extern_rustqt.md
BenFordTytherington: done

book/src/bridge/extern_rustqt.md
BenFordTytherington: done

book/src/getting-started/2-our-first-cxx-qt-module.md
```diff
@@ -176,7 +176,7 @@ Also, we can't move the `QObject`, which is why it is behind a Rust [`Pin`](http
 
 CXX-Qt will generate getters and setters for all properties of our `struct`.
 That's where the `number()` and `set_number()` functions used by `increment_number()` come from.
-For more details on what you can do with the `QObject` from Rust and what functions CXX-Qt will generate for you, take a look at the [`QObject` page](../concepts/generated_qobject.md).
+For more details on what you can do with the `QObject` from Rust and what functions CXX-Qt will generate for you, take a look at the [`QObject` page](../concepts/generated_qobject.md). For details on how to provide user defined methods for the `QObject`, take a look at the [`QProperty` section](../bridge/extern_rustqt.md#properties) and the [Available Flags](../bridge/extern_rustqt.md#available-flags)
```
ahayzen-kdab, line 179: there are many things other than properties that can be done with qobjects so specifically mentioning them here probably isn't useful as then why not mention all the other things? Instead either link generally to the `extern_rustqt.md` or leave just the link to the QObject page as that then links to the extern RustQt page anyway along with other details for QObject that are outside the extern RustQt block

BenFordTytherington: I just put it there as it mentions that cxx always generates the getters and setters which is no longer true, so thought it best it to mention there that it's optional now. will delete

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -110,20 +114,37 @@ impl ParsedQProperty {
...
+                                    ident.span(),
+                                    "Invalid flag passed, must be one of READ, WRITE, NOTIFY, RESET, CONSTANT, REQUIRED, FINAL",
+                                ))?
+                            ),
                         _ => {
                             return Err(Error::new(
                                 ident.span(),
-                                "Invalid flag passed, must be one of read, write, notify",
+                                "Invalid flag passed, must be one of READ, WRITE, NOTIFY, RESET, CONSTANT, REQUIRED, FINAL",
                             ));
```
ahayzen-kdab, line 145: there is the same error here and instead of `ok_or` we should try to use `ok_or_else` so that it doesn't construct the error when not required.

Instead can we have a closure for the error?

```rust
let unknown_flag_error = || {
   Error::new(
        ident.span(),
       "Invalid flag passed, must be one of READ, WRITE, NOTIFY, RESET, CONSTANT, REQUIRED, FINAL",
  )
};

...

reset = Some(value.ok_or_else(unknown_flag_error));

...

_ => return Err(unknown_flag_error())
```

book/src/getting-started/2-our-first-cxx-qt-module.md
ahayzen-kdab: Maybe to the end of the previous paragraph add .... 
```
See the [`QProperty` section](../bridge/extern_rustqt.md#properties) for details on user defined properties.
```

crates/cxx-qt-gen/src/parser/property.rs
BenFordTytherington: done

ahayzen-kdab: however i guess for the RESET one you could instead be saying that RESET requires a value and cannot have RESET rather than `RESET = myFunction` as that is what the actual error is in that case?

So in that case the closure would go and you would return like before, but use `ok_or_else(|| Error...)`, and update the message, what do you think?

book/src/getting-started/2-our-first-cxx-qt-module.md
BenFordTytherington: done

ahayzen-kdab (approved):
LGTM :-)

PR #1005, opened 2024-07-25
Refactor ParsedMethod and ParsedSignal to use Name

BenFordTytherington:
- [ ] Remove use of CombinedIdent (mainly in QMethodName)
- [ ] Refactor ParsedMethod to have a `name: Name` field

CombinedIdent is deprecated and now replaced with Name which is better implemented with more utility methods.


crates/cxx-qt-gen/src/generator/naming/mod.rs
```diff
@@ -2,20 +2,10 @@
...
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
-pub mod functions;
 pub mod method;
 pub mod namespace;
 pub mod property;
 pub mod qobject;
 pub mod signals;
 
 use syn::Ident;
```
ahayzen-kdab, line 11: this seems unused now?

BenFordTytherington: I thought I had removed that and the functions file in the first commit?

crates/cxx-qt-gen/src/naming/name.rs
```diff
@@ -84,6 +84,28 @@ impl Name {
...
 
+    /// For a given Name generate the Rust and C++ wrapper names
+    pub fn wrapper_from(self) -> Self {
+        const CXX_WRAPPER_SUFFIX: &str = "Wrapper";
+        const RUST_WRAPPER_SUFFIX: &str = "_wrapper";
+
+        let rust_name = self.rust.clone();
+        let cxx = self.cxx.clone();
+
+        let cxx_name = if let Some(name) = cxx {
```
ahayzen-kdab, line 95: Could add the same comment on this as in `from_rust_ident_and_attrs`.

But I would also note that does this get used as if `from_rust_ident_and_attrs` is used to create the Name then it ensures that `.cxx` is not None and where you are using it is using that route.

So is the else case actually required?

Eg if cxx is Some then append the suffix to it, but if cxx is None then only do it on the rust one as the cxx side is not used/set anyway in the original Name so it doesn't have a difference?

crates/cxx-qt-gen/src/naming/name.rs
```diff
@@ -84,6 +84,28 @@ impl Name {
         }
     }
 
+    /// For a given Name generate the Rust and C++ wrapper names
+    pub fn wrapper_from(self) -> Self {
```
ahayzen-kdab, line 88: as you are cloning things internally anyway you could change this to be `(&self)` so then you don't need another clone at the call site.

ahayzen-kdab:
@BenFordTytherington Overall looks good, appears that clippy has found some unused imports :-)

crates/cxx-qt-gen/src/naming/name.rs
BenFordTytherington: if cxx is None then it needs to generate a cxx_name anyway it seems from the tests, so I did what the old implementation used, which is that it just converts the rust name to camelcase and uses that to create a wrapper name

BenFordTytherington: in other words, if cxx is set to None you still need to generate a cxx wrapper name, just assuming it is the camel-case version of the rust name, not any custom name they provided with cxx_name

ahayzen-kdab: Right, so guess that works

crates/cxx-qt-gen/src/naming/name.rs
BenFordTytherington: done

crates/cxx-qt-gen/src/generator/naming/mod.rs
BenFordTytherington: pretty sure this is already sorted

PR #1006, opened 2024-07-26
Move ParsedMethods into structuring phase

BenFordTytherington:
All methods are now collected in structuring phase, as `ParsedMethod`, and then associated with `StructuredQObjects`, which have a `HashMap<Ident, &ParsedMethod>` which stores the `rust_unqualified` name as a key, and the method as the value, so that they can be looked up.


crates/cxx-qt-gen/src/parser/cxxqtdata.rs
```diff
@@ -21,6 +21,9 @@ use syn::{
 };
 use syn::{ItemMacro, Meta};
 
+// PROTOTYPING ONLY
+use crate::naming::Name;
```
ahayzen-kdab, line 25: think clippy is raging about this :-)

crates/cxx-qt-gen/src/generator/cpp/qobject.rs
```diff
@@ -175,7 +177,7 @@ impl GeneratedCppQObject {
             let (initializer, mut blocks) = threading::generate(&qobject_idents)?;
             generated.blocks.append(&mut blocks);
             class_initializers.push(initializer);
-        // If this type has locking enabled then add generation
+            // If this type has locking enabled then add generation
```
ahayzen-kdab, line 180: that looks like a fun formatting change, maybe move this comment down into the else if block below as it's supposed to be for that block

```
} else if qobject.locking {
    // If this type has locking enabled then add generation
```

crates/cxx-qt-gen/src/generator/structuring/mod.rs
```diff
@@ -29,15 +31,30 @@ impl<'a> Structures<'a> {
     /// Create a new `Structures` object from the given `ParsedCxxQtData`
     /// Returns an error, if any references could not be resolved.
     pub fn new(cxxqtdata: &'a ParsedCxxQtData) -> Result<Self> {
+        let methods_found: Vec<_> = cxxqtdata
+            .methods
+            .iter()
+            .map(|method| (method.name.clone(), method.qobject_ident.clone()))
+            .collect();
+        for method_pair in methods_found {
```
ahayzen-kdab, line 39: i know it's just tracing but remember you can destructure :-)

`for (key, value) in methods_found {`

crates/cxx-qt-gen/src/generator/structuring/mod.rs
```diff
@@ -54,6 +71,30 @@ impl<'a> Structures<'a> {
...
         }
 
+        for method in &cxxqtdata.methods {
+            println!(
+                "Looking for QObject match for method: {:?}",
+                method.name.clone()
+            );
+            if let Some(qobject_ident) = qobjects.iter_mut().find(|qobject| {
+                qobject.declaration.name.rust_unqualified() == &method.qobject_ident
+            }) {
```
ahayzen-kdab, line 81: This looks quite intense to be doing for every method we iterate over all qobjects?

Does this point that we should store the qobjects above as a hash instead by their rust declaration name?

```
let mut qobjects: HashMap<Ident, StructuredQObject>
```

Or should there be a cache of the index or a mut ref (might be very tricky with lifetimes). Eg `HashMap<Ident, usize>` or `HashMap<Ident, &mut StructuredQobject>` so we can get from the qobject ident -> structured qobject easily?

crates/cxx-qt-gen/src/parser/cxxqtdata.rs
```diff
@@ -204,31 +213,38 @@ impl ParsedCxxQtData {
...
+                    let parsed_signal_method_self = ParsedSignal::parse(foreign_fn, safe_call)?;
+                    self.signals.push(parsed_signal_method_self);
 
                     self.with_qobject(&parsed_signal_method.qobject_ident)?
                         .signals
                         .push(parsed_signal_method);
-                // Test if the function is an inheritance method
-                //
-                // Note that we need to test for qsignal first as qsignals have their own inherit meaning
+                    // Test if the function is an inheritance method
```
ahayzen-kdab, line 224: seems to be in lots of places :-) Is this jetbrains formatting subtly differently ? or has Rust 1.80 changed things :thinking: 

As it's creating noise unrelated to your change, either should be in it's own change if that is a cargo fmt change or we need to figure out why your IDE is subtly different :sweat_smile:

crates/cxx-qt-gen/src/parser/cxxqtdata.rs
BenFordTytherington: Done

crates/cxx-qt-gen/src/generator/cpp/signal.rs
```diff
@@ -521,7 +521,7 @@ mod tests {
         let mut type_names = TypeNames::default();
         type_names.mock_insert("ObjRust", None, None, None);
         let qobject_name = type_names.lookup(&signal.qobject_ident).unwrap();
-        let generated = generate_cpp_signal(&signal, &qobject_name, &type_names).unwrap();
+        let generated = generate_cpp_signal(&signal, qobject_name, &type_names).unwrap();
```
ahayzen-kdab, line 524: where is this change coming from as the signature hasn't changed ? Seems like a clippy style fix ?

crates/cxx-qt-gen/src/generator/rust/qobject.rs
```diff
@@ -23,6 +23,7 @@ use quote::quote;
 use syn::{Ident, Result};
 
 impl GeneratedRustFragment {
+    // Might need to be refactored to use a StructuredQObject instead (confirm with Leon)
```
ahayzen-kdab, line 26: seems likely :-)

crates/cxx-qt-gen/src/generator/cpp/signal.rs
BenFordTytherington: yeah, was "immediately dereferencing" warning, so assuming qobject_name was already a reference, so function got &&T and just instantly coerced to &T, so it wasn't needed

crates/cxx-qt-gen/src/generator/structuring/mod.rs
```diff
@@ -54,6 +51,113 @@ impl<'a> Structures<'a> {
...
+            if let Some(qobject) = qobjects
+                .iter_mut()
+                .find(|qobject| qobject.has_qobject_name(&method.qobject_ident))
+            {
+                qobject.methods.push(method);
+            } else {
+                return Err(Error::new_spanned(
+                    &method.qobject_ident,
+                    format!("Unknown QObject: {:?}", &method.qobject_ident),
+                ));
```
ahayzen-kdab, line 65: Wonder if this is more readable then these big if let Some blocks ?

```rust
            let mut qobject = qobjects
                .iter_mut()
                .find(|qobject| qobject.has_qobject_name(qobject_ident))
                .ok_or_else(|| {
                    Error::new_spanned(qobject_ident, format!("Unknown QObject: {qobject_ident}"))
                })?;
            qobject.methods.push(method);
```

BenFordTytherington: Done

crates/cxx-qt-gen/src/parser/cxxqtdata.rs
BenFordTytherington: Done

crates/cxx-qt-gen/src/generator/rust/qobject.rs
ahayzen-kdab: Yup, assuming the next one is working on the generator phase

crates/cxx-qt-gen/src/generator/cpp/qobject.rs
BenFordTytherington: Removed

crates/cxx-qt-gen/src/parser/cxxqtdata.rs
BenFordTytherington: Removed

crates/cxx-qt-gen/src/generator/cpp/signal.rs
ahayzen-kdab: Even running with `--all-targets` doesn't help fun :-)

ahayzen-kdab (approved):
LGTM, lets go with this as a first step of adding structuring support for methods then generation phase refactor is another PR.

PR #1007, opened 2024-07-29
cmake: Use cxx-qt-cmake repo

LeonMatthesKDAB:


examples/qml_minimal/CMakeLists.txt
```diff
@@ -29,14 +29,18 @@ if(NOT Qt6_FOUND)
...
-
 # ANCHOR: book_cmake_find_cxx_qt
-# TODO: Replace with fetch-content co cxx-qt-cmake
+find_package(CxxQt QUIET)
+if(NOT CxxQt_FOUND)
+    include(FetchContent)
+    FetchContent_Declare(
+        CxxQt
+        GIT_REPOSITORY https://github.com/kdab/cxx-qt-cmake.git
+        GIT_TAG main
```
ahayzen-kdab, line 39: how do we make it clear in the book that they should use `va.b.c` ? or do we do that already ?

LeonMatthesKDAB: One way I could see us going about this is to create different branches for each minor version (so e.g. `0.7`), then create tags for each patch version.
So e.g. we could use `GIT_TAG 0.7` which would keep us up-to-date and would be an okay default for end-users as well. As long as e.g. `0.7` isn't released yet, we can just keep pushing changes to that branch so we can change whatever we want. After release, we move on to `0.8` and only publish compatible patches to the `0.7` branch.
Which ensures we can always work with the "bleeding-edge" version of the cmake code, and our users get the latest patches by default.

And if you want to tag to a specific version, you can use `GIT_TAG 0.7.0`.

LeonMatthesKDAB: @ahayzen-kdab Let me know what you think

ahayzen-kdab: I was more meaning the usual problem of that we want our example to build but we also want the book to use stable tags. So we need the book to show `GIT_TAG v0.7.0` but we want the example in our repo to use `GIT_TAG main`.

Regarding the repo management for now i would just keep adding things to `main` branch then we can tag each release with `v0.7.0` or `v0.7.1`, if we feel it's useful we could create a `0.7.x` branch like we in this repo too later but probably not required as it won't be changing much?

LeonMatthesKDAB: But that's the thing, we could already create a 0.7 branch in cxx-qt-cmake, and simply tag to that.

When we build the book during the 0.7 release, it would correctly say to just use `GIT_TAG 0.7`, which would be fine. It's not entirely stable, but a good default, as it includes the latest patches.
And we can then move the CMakeLists.txt on to `0.8` so we have it always point to a sensible value and don't need to point it to `main`.

ahayzen-kdab: But normally your branch names and tag names are different as they collide otherwise. So you tend to do branches like `0.7.x` and tags as `v0.7.0` (prefixed with a v).

We could yes make the examples use `0.7.x` but then our development branch in the cmake repo is odd as it's not `main`.

So can we not just develop on `main` and in the book use `v0.7.1` as the tag for the tutorial etc? As we are already IIRC injecting hardcoded values in certain places.

LeonMatthesKDAB: Right, I guess we can do that.
Let me try.

LeonMatthesKDAB: Done

ahayzen-kdab (approved):
LGTM :-)

PR #1009, opened 2024-07-30
Refactor to use Structuring for methods and signals

BenFordTytherington:
The generators and tests should now be using StructuredQObject and its fields, methods and signals, in place of ParsedQobject and it's fields of the same name, this includes:

- Rewriting some of the generate functions in order to accept &Vec<&T> where they used to accept &Vec<T> and all the use of these to follow that
- Rewriting the rust generators to accept StructuredQObject instead of ParsedQObject
- Rewriting the generator functions to push methods from StructuredQObject

### DIFF
--- a/crates/cxx-qt-gen/src/generator/cpp/method.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/method.rs
@@ -21,13 +21,13 @@ use indoc::formatdoc;
 use syn::{spanned::Spanned, Error, FnArg, Pat, PatIdent, PatType, Result};
 
 pub fn generate_cpp_methods(
-    invokables: &Vec<ParsedMethod>,
+    invokables: &Vec<&ParsedMethod>,
     qobject_idents: &QObjectNames,
     type_names: &TypeNames,
 ) -> Result<GeneratedCppQObjectBlocks> {
     let mut generated = GeneratedCppQObjectBlocks::default();
     let qobject_ident = qobject_idents.name.cxx_unqualified();
-    for invokable in invokables {
+    for &invokable in invokables {
         let idents = QMethodName::try_from(invokable)?;
         let return_cxx_ty = syn_type_to_cpp_return_type(&invokable.method.sig.output, type_names)?;
 
@@ -274,7 +274,12 @@ mod tests {
         let mut type_names = TypeNames::mock();
         type_names.mock_insert("QColor", None, None, None);
 
-        let generated = generate_cpp_methods(&invokables, &qobject_idents, &type_names).unwrap();
+        let generated = generate_cpp_methods(
+            &invokables.iter().map(|method| method).collect(),
+            &qobject_idents,
+            &type_names,
+        )
+        .unwrap();
 
         // methods
         assert_eq!(generated.methods.len(), 5);
@@ -450,7 +455,12 @@ mod tests {
         type_names.mock_insert("A", None, Some("A1"), None);
         type_names.mock_insert("B", None, Some("B2"), None);
 
-        let generated = generate_cpp_methods(&invokables, &qobject_idents, &type_names).unwrap();
+        let generated = generate_cpp_methods(
+            &invokables.iter().map(|method| method).collect(),
+            &qobject_idents,
+            &type_names,
+        )
+        .unwrap();
 
         // methods
         assert_eq!(generated.methods.len(), 1);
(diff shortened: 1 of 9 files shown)


crates/cxx-qt-gen/src/generator/cpp/method.rs
```diff
@@ -274,7 +274,12 @@ mod tests {
         let mut type_names = TypeNames::mock();
         type_names.mock_insert("QColor", None, None, None);
 
-        let generated = generate_cpp_methods(&invokables, &qobject_idents, &type_names).unwrap();
+        let generated = generate_cpp_methods(
+            &invokables.iter().map(|method| method).collect(),
```
ahayzen-kdab, line 278: `&invokables.as_ref()` ?

crates/cxx-qt-gen/src/generator/cpp/method.rs
```diff
@@ -21,13 +21,13 @@ use indoc::formatdoc;
...
 pub fn generate_cpp_methods(
-    invokables: &Vec<ParsedMethod>,
+    invokables: &Vec<&ParsedMethod>,
     qobject_idents: &QObjectNames,
     type_names: &TypeNames,
 ) -> Result<GeneratedCppQObjectBlocks> {
     let mut generated = GeneratedCppQObjectBlocks::default();
     let qobject_ident = qobject_idents.name.cxx_unqualified();
-    for invokable in invokables {
+    for &invokable in invokables {
```
ahayzen-kdab, line 30: do you need the `&` here?

crates/cxx-qt-gen/src/generator/cpp/method.rs
```diff
@@ -450,7 +455,12 @@ mod tests {
         type_names.mock_insert("A", None, Some("A1"), None);
         type_names.mock_insert("B", None, Some("B2"), None);
 
-        let generated = generate_cpp_methods(&invokables, &qobject_idents, &type_names).unwrap();
+        let generated = generate_cpp_methods(
+            &invokables.iter().map(|method| method).collect(),
```
ahayzen-kdab, line 459: `&invokables.as_ref()`

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
```diff
@@ -57,7 +57,7 @@ pub fn generate_cpp_properties(
     }
 
     generated.append(&mut generate_cpp_signals(
-        &signals,
+        &signals.iter().map(|signal| signal).collect(), // Converting Vec<ParsedSignal> to Vec<&ParsedSignal>
```
ahayzen-kdab, line 60: `&signals.as_ref()`

crates/cxx-qt-gen/src/generator/cpp/method.rs
BenFordTytherington: Don't think that worked, I can just use iter().collect() though which is slightly cleaner, and the places that only had 1 or 2 methods in tests I have refactored to just declare the methods outside the vec

crates/cxx-qt-gen/src/generator/cpp/method.rs
BenFordTytherington: Yeah it's used in the QMethod::try_from, was erroring since the trait isn't implemented for &&ParsedMethod, might be able to put the ref inside that fn call, but I think it makes more sense since invokables is a Vec of references

crates/cxx-qt-gen/src/generator/cpp/method.rs
ahayzen-kdab: Yeah seems it doesn't and you'd need to implement a load of other traits

crates/cxx-qt-gen/src/generator/rust/qobject.rs
```diff
@@ -195,9 +197,10 @@ mod tests {
             }
         };
         let parser = Parser::from(module).unwrap();
+        let structures = Structures::new(&parser.cxx_qt_data).unwrap();
 
         let rust = GeneratedRustFragment::from_qobject(
-            parser.cxx_qt_data.qobjects.values().next().unwrap(),
+            &structures.qobjects[0],
```
ahayzen-kdab, line 203: would probably use `&structures.qobjects.get(0).unwrap()` here so you don't get a random index failure thing when it goes wrong

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
```diff
@@ -57,7 +57,7 @@ pub fn generate_cpp_properties(
     }
 
     generated.append(&mut generate_cpp_signals(
-        &signals,
+        &signals.iter().collect(), // Converting Vec<ParsedSignal> to Vec<&ParsedSignal>
```
ahayzen-kdab, line 60: do we need the comment, as the others don't have the comment so maybe remove :-)

ahayzen-kdab (approved):
LGTM :-)

PR #1010, opened 2024-07-30
Refactor to remove method fields in ParsedQObject

BenFordTytherington:
Remove the now unused method and signals fields from ParsedQObject since they are now created inside StructuredQObject instead, and update a few tests


ahayzen-kdab (approved):
LGTM :-)

PR #1012, opened 2024-07-30
Refactor ParsedSignal to remove old code

BenFordTytherington:
Removes some old code in ParsedSignal creation for adding cxx_name, as this is now handled by Name. Refactored from just using the ForeignItemFn that was passed, to custom building the token stream like is done for ParsedMethod, including docs passthrough and cxx_names on all signals

- Update tests with new order for doc and cxx_name
- Update test_outputs for new attribute parsing
- Add doc field to ParsedSignal for passthrough
- Update tests to ignore unknown attributes like #[my_attribute]
- Update from_property_method of ParsedSignal to accept docs

### DIFF
--- a/crates/cxx-qt-gen/src/generator/cpp/property/signal.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/property/signal.rs
@@ -5,6 +5,7 @@
 
 use syn::ForeignItemFn;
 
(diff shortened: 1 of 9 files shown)


ahayzen-kdab (approved):
LGTM :-)

PR #1013, opened 2024-07-31
Clean up some long functions and tests

BenFordTytherington:
Created some helper functions for ParsedMethod and ParsedSignal to improve readability of tests and generator code, some of which were quite long and verbose

### DIFF
--- a/crates/cxx-qt-gen/src/parser/method.rs
+++ b/crates/cxx-qt-gen/src/parser/method.rs
@@ -3,13 +3,18 @@
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
+use crate::generator::cpp::fragment::CppNamedType;
+use crate::naming::cpp::syn_type_to_cpp_type;
+use crate::naming::TypeNames;
 use crate::{
     naming::Name,
     parser::parameter::ParsedFunctionParameter,
     syntax::{attribute::attribute_take_path, foreignmod, safety::Safety, types},
 };
+use proc_macro2::TokenStream;
+use quote::quote;
 use std::collections::HashSet;
-use syn::{spanned::Spanned, Error, ForeignItemFn, Ident, Result};
+use syn::{spanned::Spanned, Error, FnArg, ForeignItemFn, Ident, Pat, PatIdent, PatType, Result};
 
 /// Describes a C++ specifier for the Q_INVOKABLE
 #[derive(Eq, Hash, PartialEq)]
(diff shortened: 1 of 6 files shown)


crates/cxx-qt-gen/src/parser/method.rs
```diff
@@ -102,4 +110,90 @@ impl ParsedMethod {
             name,
         })
     }
+
+    /// Returns a [TokenStream] of the methods parameters, which can be used in generation stage
+    /// class_name is the rust class name which this method is defined for
+    pub fn get_params_tokens(&self, class_name: &Ident) -> TokenStream {
```
ahayzen-kdab, line 116: wonder if this method should be in the parser or generator :') Maybe it's fine here though :thinking:

crates/cxx-qt-gen/src/parser/signals.rs
```diff
@@ -110,6 +132,30 @@ impl ParsedSignal {
             docs,
         })
     }
+
+    /// Returns a [TokenStream] of the signals parameters, which can be used in generation stage
+    /// class_name is the rust class name which this method is defined for
+    pub fn get_params_tokens(&self, class_name: &Ident) -> TokenStream {
```
ahayzen-kdab, line 138: assume we can't sanely shared code with the one in parsed method?

Guess this is what you were saying about traits? If ParsedMethod and ParsedSignal implement a ParamTokens trait or something that had methods to get the mutable / parameters etc. Then the generation phase could have one method of `get_param_token` for a type that `impl ParamTokens` or something?

ahayzen-kdab: @LeonMatthes any thoughts here if it's worth making a trait to have a common way for ParsedSignal/ParsedMethod to expose their info so that the generator phase could have a single method which figures out the tokens (and then moves this logic into the generator phase)

Other route would be to return a common struct or just have enough args on the function.

As note sure at the moment i like how this is effectively "generating" something in the parser phase :thinking:

ahayzen-kdab: @LeonMatthesKDAB oops wrong account ^^ :sweat_smile:

LeonMatthesKDAB: Right, this probably should go in the generator :thinking: 

And I guess it could just take all the data it needs directly
```rust
fn generate_member_parameter_list(mutability: bool /*or some kind of mutability enum*/, class_name: &Ident, parameters: &[ParsedFunctionParameter])
```
Then calling it for the signal would just be:
```rust
generate_member_parameter_list(signal.mutable, format_ident!("..."), &signal.parameters);
// and basically the same for methods
generate_member_parameter_list(method.mutable, format_ident!("..."), &method.parameters);
```
Which would be simple enough I think.

BenFordTytherington: So go with just using a common function then? I think a trait could potentially be useful elsewhere since method and signal have so much in common but might be a bit much work here just to simplify generation

crates/cxx-qt-gen/src/parser/method.rs
BenFordTytherington: in generator/mod.rs now

ahayzen-kdab (approved):
Looks good, there can be further improvements but lets keep moving forward.

PR #1016, opened 2024-08-01
Move generator helpers to their language specific directories

BenFordTytherington:
get_params_tokens is now in generator/rust/mod.rs, and get_cpp_params is now in generator/cpp/mod.rs.


ahayzen-kdab (approved):
Nice! LGTM :-)

PR #1017, opened 2024-08-01
Introduce method lookup into NameState so that cxx_names can be used

BenFordTytherington:
StructuredQObject is passed through various functions so that in NameState, a method lookup can be used for the user defined functions, and this also requires modifying some test data to create Structures

### DIFF
--- a/crates/cxx-qt-gen/src/generator/cpp/mod.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/mod.rs
@@ -17,6 +17,8 @@ pub mod qobject;
 pub mod signal;
 pub mod threading;
 
+mod utils;
+
 use std::collections::BTreeSet;
 
 use crate::generator::cpp::fragment::CppNamedType;
@@ -79,7 +81,38 @@ impl GeneratedCppBlocks {
     }
 }
 
-mod utils;
+/// Returns a vector of the names and types ([CppNamedType] of the parameters of this method, used in cpp generation step
+pub fn get_cpp_params(method: &ForeignItemFn, type_names: &TypeNames) -> Result<Vec<CppNamedType>> {
+    method
+        .sig
+        .inputs
+        .iter()
+        .map(|input| {
+            // Match parameters to extract their idents
+            if let FnArg::Typed(PatType { pat, ty, .. }) = input {
+                let ident = if let Pat::Ident(PatIdent { ident, .. }) = &**pat {
+                    ident
+                } else {
+                    return Err(Error::new(input.span(), "Unknown pattern for type"));
+                };
+
+                // If the name of the argument is self then ignore,
+                // as this is likely the self: Pin<T>
+                if ident == "self" {
+                    Ok(None)
+                } else {
+                    Ok(Some(CppNamedType {
+                        ident: ident.to_string(),
+                        ty: syn_type_to_cpp_type(ty, type_names)?,
+                    }))
+                }
+            } else {
+                Ok(None)
+            }
+        })
+        .filter_map(|result| result.map_or_else(|e| Some(Err(e)), |v| v.map(Ok)))
+        .collect()
+}
 
 #[cfg(test)]
 mod tests {
@@ -142,37 +175,4 @@ mod tests {
         let cpp = GeneratedCppBlocks::from(&parser).unwrap();
         assert_eq!(cpp.qobjects[0].name.namespace(), Some("cxx_qt"));
     }
-}
-
-/// Returns a vector of the names and types ([CppNamedType] of the parameters of this method, used in cpp generation step
-pub fn get_cpp_params(method: &ForeignItemFn, type_names: &TypeNames) -> Result<Vec<CppNamedType>> {
-    method
-        .sig
-        .inputs
-        .iter()
-        .map(|input| {
-            // Match parameters to extract their idents
-            if let FnArg::Typed(PatType { pat, ty, .. }) = input {
-                let ident = if let Pat::Ident(PatIdent { ident, .. }) = &**pat {
-                    ident
-                } else {
-                    return Err(Error::new(input.span(), "Unknown pattern for type"));
-                };
-
-                // If the name of the argument is self then ignore,
-                // as this is likely the self: Pin<T>
-                if ident == "self" {
-                    Ok(None)
-                } else {
-                    Ok(Some(CppNamedType {
-                        ident: ident.to_string(),
-                        ty: syn_type_to_cpp_type(ty, type_names)?,
-                    }))
-                }
-            } else {
-                Ok(None)
-            }
-        })
-        .filter_map(|result| result.map_or_else(|e| Some(Err(e)), |v| v.map(Ok)))
-        .collect()
 }
--- a/crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
@@ -89,9 +89,9 @@ mod tests {
 
         let qobject_idents = create_qobjectname();
 
-        let obj = ParsedQObject::mock_parsed_qobject();
+        let obj = ParsedQObject::mock();
 
-        let structured_qobject = StructuredQObject::mock_structured_qobject(&obj);
+        let structured_qobject = StructuredQObject::mock(&obj);
 
         let type_names = TypeNames::mock();
         generate_cpp_properties(
@@ -145,8 +145,6 @@ mod tests {
             struct MyStruct;
         };
 
-        // let generated = setup_generated(&mut input).unwrap();
-
         let property = ParsedQProperty::parse(input.attrs.remove(0)).unwrap();
 
         let properties = vec![property];
@@ -157,14 +155,14 @@ mod tests {
         let parser = Parser::from(module).unwrap();
         let structures = Structures::new(&parser.cxx_qt_data).unwrap();
 
-        let structured_qobject = structures.qobjects.get(0).unwrap();
+        let structured_qobject = structures.qobjects.first().unwrap();
 
         let type_names = TypeNames::mock();
         let generated = generate_cpp_properties(
             &properties,
             &qobject_idents,
             &type_names,
-            &structured_qobject,
+            structured_qobject,
         )
         .unwrap();
 
@@ -203,8 +201,6 @@ mod tests {
             struct MyStruct;
         };
 
-        // let generated = setup_generated(&mut input).unwrap();
-
         let property = ParsedQProperty::parse(input.attrs.remove(0)).unwrap();
 
         let properties = vec![property];
@@ -216,14 +212,14 @@ mod tests {
         let parser = Parser::from(module).unwrap();
         let structures = Structures::new(&parser.cxx_qt_data).unwrap();
 
-        let structured_qobject = structures.qobjects.get(0).unwrap();
+        let structured_qobject = structures.qobjects.first().unwrap();
 
         let type_names = TypeNames::mock();
         let generated = generate_cpp_properties(
             &properties,
             &qobject_idents,
             &type_names,
-            &structured_qobject,
+            structured_qobject,
         )
         .unwrap();
 
@@ -266,9 +262,9 @@ mod tests {
 
         let qobject_idents = create_qobjectname();
 
-        let obj = ParsedQObject::mock_parsed_qobject();
+        let obj = ParsedQObject::mock();
 
-        let structured_qobject = StructuredQObject::mock_structured_qobject(&obj);
+        let structured_qobject = StructuredQObject::mock(&obj);
 
         let mut type_names = TypeNames::mock();
         type_names.mock_insert("QColor", None, None, None);
@@ -559,9 +555,9 @@ mod tests {
         }];
         let qobject_idents = create_qobjectname();
 
-        let obj = ParsedQObject::mock_parsed_qobject();
+        let obj = ParsedQObject::mock();
 
-        let structured_qobject = StructuredQObject::mock_structured_qobject(&obj);
+        let structured_qobject = StructuredQObject::mock(&obj);
 
         let mut type_names = TypeNames::mock();
         type_names.mock_insert("A", None, Some("A1"), None);
--- a/crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
@@ -102,6 +102,42 @@ mod tests {
         )
     }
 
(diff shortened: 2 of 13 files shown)


crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
```diff
@@ -96,7 +129,41 @@ mod tests {
...
+        let structured_qobject = structures.qobjects.get(0).unwrap();
+
+        let type_names = TypeNames::mock();
+        let generated = generate_cpp_properties(
+            &properties,
+            &qobject_idents,
+            &type_names,
+            &structured_qobject,
+        )
+        .unwrap();
```
ahayzen-kdab, line 166: @LeonMatthesKDAB any suggestions on how we can nicely mock the `Structures` from our other discussions ? Eg should we be building the whole parser and using parse_quote etc here. Or should we manually have ways to build up the Structure and ParsedMethod etc.

LeonMatthesKDAB: Hm, we should be mocking it, yes. I'll need to think a bit about this after the workshop.

BenFordTytherington: Have introduced functions just to remove that block from tests, but still could be improved I think

BenFordTytherington:
Might've messed up integrating latest changes in main there, oops

crates/cxx-qt-gen/src/generator/cpp/mod.rs
```diff
@@ -78,6 +78,38 @@ impl GeneratedCppBlocks {
...
+                }
+            } else {
+                Ok(None)
+            }
+        })
+        .filter_map(|result| result.map_or_else(|e| Some(Err(e)), |v| v.map(Ok)))
+        .collect()
+}
 
 mod utils;
```
ahayzen-kdab, line 114: this mod should probably be near the top with the others ?

crates/cxx-qt-gen/src/generator/cpp/mod.rs
```diff
@@ -78,6 +78,38 @@ impl GeneratedCppBlocks {
         })
     }
 }
+/// Returns a vector of the names and types ([CppNamedType] of the parameters of this method, used in cpp generation step
```
ahayzen-kdab, line 81: should be a new line between functions, so add a new line above this comment

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
```diff
@@ -96,7 +145,28 @@ mod tests {
             struct MyStruct;
         };
 
-        let generated = setup_generated(&mut input).unwrap();
+        // let generated = setup_generated(&mut input).unwrap();
```
ahayzen-kdab, line 148: remove commented line

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
```diff
@@ -133,7 +203,29 @@ mod tests {
...
```
ahayzen-kdab, line 206: same here

crates/cxx-qt-gen/src/generator/naming/property.rs
```diff
@@ -49,24 +54,41 @@ pub struct QPropertyNames {
...
-        let setter = flags.write.as_ref().map(|setter| {
-            NameState::from_flag_with_auto_fn(setter, || setter_name_from_property(&property_name))
-        });
+        let getter = NameState::from_flag_with_auto_fn(
+            &flags.read,
+            || getter_name_from_property(&property_name),
+            structured_qobject,
+        )?;
+
+        let setter = if let Some(setter) = &flags.write {
```
ahayzen-kdab, line 73: Feels like it should be possible to simplify this but can't think of how right now :sweat_smile:

crates/cxx-qt-gen/src/generator/structuring/qobject.rs
```diff
@@ -31,4 +33,25 @@ impl<'a> StructuredQObject<'a> {
...
+            .find(|method| method.name.rust_unqualified() == id);
+
+        if let Some(method) = method {
+            Ok(method.name.clone())
+        } else {
+            Err(Error::new_spanned(
+                id,
+                format!("Method with name '{id}' not found!"),
+            ))
+        }
```
ahayzen-kdab, line 50: This looks like it could be simplified to something like this?

```rust
method
  .map(|method| method.name.clone())
  .ok_or_else(|| Error:new_spanned(...));
```

crates/cxx-qt-gen/src/generator/naming/property.rs
BenFordTytherington: My old way of writing those was to use if let and if it could be done with map, clippy usually tells you, I think its not possible here maybe because of the Result? I'm sure there is a way to improve it though

crates/cxx-qt-gen/src/generator/structuring/qobject.rs
BenFordTytherington: Done

crates/cxx-qt-gen/src/generator/naming/property.rs
```diff
@@ -30,10 +31,14 @@ impl Deref for NameState {
 }
 
 impl NameState {
-    pub fn from_flag_with_auto_fn(state: &FlagState, auto_fn: impl Fn() -> Name) -> Self {
+    pub fn from_flag_with_auto_fn(
+        state: &FlagState,
+        auto_fn: impl Fn() -> Name,
+        structured_qobject: &StructuredQObject,
+    ) -> Result<Self> {
         match state {
```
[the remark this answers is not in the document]
BenFordTytherington: DOne

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
BenFordTytherington: Done

crates/cxx-qt-gen/src/generator/cpp/mod.rs
BenFordTytherington: Done

crates/cxx-qt-gen/src/generator/cpp/mod.rs
BenFordTytherington: Done

crates/cxx-qt-gen/src/generator/structuring/qobject.rs
```diff
@@ -31,4 +33,20 @@ impl<'a> StructuredQObject<'a> {
...
+    pub fn method_lookup(&self, id: &Ident) -> Result<Name> {
+        let method = self
+            .methods
+            .iter()
+            .find(|method| method.name.rust_unqualified() == id);
+
+        method
+            .map(|method| method.name.clone())
+            .ok_or_else(|| Error::new_spanned(id, format!("Method with name '{id}' not found!")))
+    }
```
ahayzen-kdab, line 46: this could all be simplified into one glorious chain of items with no `let` as well ? :tada: 

```rust
   pub fn method_lookup(&self, id: &Ident) -> Result<Name> {
        self
            .methods
            .iter()
            .find(|method| method.name.rust_unqualified() == id)
            .map(|method| method.name.clone())
            .ok_or_else(|| Error::new_spanned(id, format!("Method with name '{id}' not found!")))
    }
```

LeonMatthesKDAB: or even:
```rust
   pub fn method_lookup(&self, id: &Ident) -> Result<Name> {
        self
            .methods
            .iter()
            .map(|method| &method.name)
            .find(|name| name.rust_unqualified() == id)
            .cloned()
            .ok_or_else(|| Error::new_spanned(id, format!("Method with name '{id}' not found!")))
    }
```

crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -54,6 +56,27 @@ pub struct ParsedQObject {
 }
 
 impl ParsedQObject {
+    #[cfg(test)]
+    pub fn mock_parsed_qobject() -> Self {
+        ParsedQObject {
+            base_class: None,
+            name: Name::new(format_ident!("my_property")),
+            rust_type: format_ident!("i32"),
```
LeonMatthesKDAB, line 64: This doesn't really make sense, you're mocking a parsed QObject, not a property.
So that should be `MyObject` and `MyObjectRust`.

crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -54,6 +56,27 @@ pub struct ParsedQObject {
 }
 
 impl ParsedQObject {
+    #[cfg(test)]
+    pub fn mock_parsed_qobject() -> Self {
```
LeonMatthesKDAB, line 60: Why not just `mock()`? The function is already on ParsedQObject.

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
LeonMatthesKDAB: Hm, the new mocking functions look alright, so maybe that's just the way to go :)

crates/cxx-qt-gen/src/generator/naming/property.rs
LeonMatthesKDAB: You should be able to use `map` with [`transpose`](https://doc.rust-lang.org/std/option/enum.Option.html#method.transpose).
So something like:
```rust
let setter = flags.write.map(|setter| NameState::from_flag_with_auto_fn(
                setter,
                || setter_name_from_property(&property_name),
                structured_qobject,
            )).transpose()?;
```

LeonMatthesKDAB: Which allows you to go from `Result<Option<>>` to `Option<Result<>>` and vice-versa

crates/cxx-qt-gen/test_inputs/properties.rs
```diff
@@ -20,4 +20,14 @@ mod ffi {
         #[qproperty(i32, final_prop, READ, WRITE, FINAL)]
         type MyObject = super::MyObjectRust;
     }
+
+    unsafe extern "RustQt" {
+        fn my_getter(self: &MyObject) -> i32;
+
+        fn my_setter(self: Pin<&mut MyObject>, value: i32);
+
+        fn myOnChanged(self: Pin<&mut MyObject>);
```
LeonMatthesKDAB, line 29: This should have `#[qsignal]` on it.

crates/cxx-qt-gen/test_inputs/properties.rs
```diff
@@ -20,4 +20,14 @@ mod ffi {
         #[qproperty(i32, final_prop, READ, WRITE, FINAL)]
         type MyObject = super::MyObjectRust;
     }
+
+    unsafe extern "RustQt" {
+        fn my_getter(self: &MyObject) -> i32;
+
+        fn my_setter(self: Pin<&mut MyObject>, value: i32);
```
LeonMatthesKDAB, line 27: Maybe add some `#[cxx_name]` or `#[rust_name]` to some of these to test the renaming actually works with custom renaming.

crates/cxx-qt-gen/test_inputs/properties.rs
BenFordTytherington: This I think is acc a bug I need to fix, as using qsignal means the code won't compile as the lookup searches for a method not a signal, and all the flags are using the same logic for custom functions

ahayzen-kdab: Right a notify should be a signal, others like read, write, reset should be methods.

crates/cxx-qt-gen/src/generator/naming/property.rs
ahayzen-kdab: Aha nice :-)

crates/cxx-qt-gen/test_inputs/properties.rs
ahayzen-kdab: Guess this should ideally be the snake case variant, which then means it'd check if it picks up the automatic camelCase conversion?

eg use

```
#[qproperty(i32, custom_on_changed_prop, READ, WRITE, NOTIFY = my_on_changed)]

...

       #[qsignal]
        fn my_on_changed(self: Pin<&mut MyObject>);
```

Then we should see `my_on_changed` on the Rust side and `myOnChanged` in the C++ generation.

ahayzen-kdab: as then we have changes for everything

- rust_name (getter one)
- cxx_name (setter one)
- automatic camel conversion (notify one)
- already camel in Rust no conversion (reset one)

ahayzen-kdab: Nice looks good :-)

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
BenFordTytherington: so sorted then?

crates/cxx-qt-gen/src/generator/naming/property.rs
BenFordTytherington: used it, did need a clone of the flag though I think

crates/cxx-qt-gen/src/generator/structuring/qobject.rs
BenFordTytherington: Done

crates/cxx-qt-gen/src/parser/qobject.rs
BenFordTytherington: Done

crates/cxx-qt-gen/src/parser/qobject.rs
BenFordTytherington: Done, same for the structured obj

crates/cxx-qt-gen/test_inputs/properties.rs
BenFordTytherington: sorted

crates/cxx-qt-gen/test_inputs/properties.rs
BenFordTytherington: Done

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
LeonMatthesKDAB: I think so.

LeonMatthesKDAB (approved):
Nice work :+1:

PR #1018, opened 2024-08-01
Add support for inherited method to structuring phase

BenFordTytherington:
Start moving ParsedInheritedMethod into structuring phase by adding fields to StructuredQObject and ParsedCxxQtData, and associate inherited methods with their StructuredQObject in structuring phase.

### DIFF
--- a/crates/cxx-qt-gen/src/parser/cxxqtdata.rs
+++ b/crates/cxx-qt-gen/src/parser/cxxqtdata.rs
@@ -35,6 +35,8 @@ pub struct ParsedCxxQtData {
     pub methods: Vec<ParsedMethod>,
     /// List of the Q_SIGNALS found
     pub signals: Vec<ParsedSignal>,
+    /// List of the inherited methods found
+    pub inherited_methods: Vec<ParsedInheritedMethod>,
     /// List of QNamespace declarations
     pub qnamespaces: Vec<ParsedQNamespace>,
     /// Blocks of extern "C++Qt"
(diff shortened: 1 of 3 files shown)


BenFordTytherington:
Weird error about the examples having a struct which is not constructed?

ahayzen-kdab:
> Weird error about the examples having a struct which is not constructed?

Looks like `QQuickItem` isn't required anymore in custom_parent_class, just remove it's type and include

crates/cxx-qt-gen/src/parser/cxxqtdata.rs
```diff
@@ -217,9 +220,14 @@ impl ParsedCxxQtData {
...
                     // Note that we need to test for qsignal first as qsignals have their own inherit meaning
                 } else if attribute_take_path(&mut foreign_fn.attrs, &["inherit"]).is_some() {
+                    let parsed_inherited_method_self =
+                        ParsedInheritedMethod::parse(foreign_fn.clone(), safe_call)?;
+
                     let parsed_inherited_method =
                         ParsedInheritedMethod::parse(foreign_fn, safe_call)?;
 
+                    self.inherited_methods.push(parsed_inherited_method_self);
+
```
ahayzen-kdab, line 230: why do we have the same thing twice?  why not this

```rust
let parsed_inherited_method = ParsedInheritedMethod::parse(foreign_fn, safe_call)?;
self.inherited_methods.push(parsed_inherited_method.clone());

self.with_qobject(&parsed_inherited_method.qobject_ident)?...
```

BenFordTytherington: so that it adds it to the methods contained in cxxqtdata but doesn't break the rest of the code, which is all still using the old method (same way we did things for methods and signals, where I implemented it all before removing the old code)

ahayzen-kdab: OK, it gets removed in a later branch anyway, so lets go with it.

ahayzen-kdab (approved):
LGTM :-)

PR #1019, opened 2024-08-01
Refactor use inherited methods from structuring

BenFordTytherington:
Use structured_qobject.inherited_methods in place of qobject.inherited_methods


ahayzen-kdab (approved):
LGTM :-)

PR #1020, opened 2024-08-01
Refactor remove dead inherited method

BenFordTytherington:
ParsedQObject doesn't have an inherited_methods field anymore, it's been fully moved into StructuredQObject


ahayzen-kdab (approved):
LGTM :-)

PR #1021, opened 2024-08-01
Hotfix update clippy example

BenFordTytherington:
Clippy started complaining about an unused struct which it previously didn't mind, so just removed those lines


ahayzen-kdab (approved):
The constructor implementation changed 1yr ago, now the latest version of clippy is able to find unused types :tada: So we need this to pass CI.

PR #1022, opened 2024-08-02
Create script for generating data and reports for code coverage

BenFordTytherington:


scripts/grcov_cxx_qt.sh
```diff
@@ -0,0 +1,5 @@
+export RUSTFLAGS="-Cinstrument-coverage"
```
ahayzen-kdab, line 1: can you add the bash shebang and license (SPDX info copy from other files and use your name) to the top

```bash
#!/usr/bin/env bash

# SPDX-FileCopyrightText: ...
# SPDX-FileContributor: ...
#
```

ahayzen-kdab, line 1: can you add  this near the top so if fails when the first command fails

```
set -e
```

ahayzen-kdab, line 1: Can you add this snippet so we ensure we are running in the right directory
```bash
# Ensure we are in the right directory
SCRIPT=$(realpath "$0")
cd "$SCRIPTPATH/../"
```

scripts/grcov_cxx_qt.sh
```diff
@@ -1,5 +1,22 @@
...
+
 export RUSTFLAGS="-Cinstrument-coverage"
+export LLVM_PROFILE_FILE="$SCRIPTDIR/coverage/coverage_data-%p-%m.profraw"
 cargo build --package cxx-qt-gen
-export LLVM_PROFILE_FILE="coverage/coverage_data-%p-%m.profraw"
 cargo test --package cxx-qt-gen
-grcov . -s . --binary-path ./target/debug/ -t html --branch --ignore-not-existing -o ./target/debug/
\ No newline at end of file
+grcov . -s . --binary-path ./target/debug/ -t html --branch --ignore-not-existing -o ./target/debug/
+echo "Coverage html report generated in $SCRIPTDIR/target/debug/html"
```
ahayzen-kdab, line 22: Oops that should be `$SCRIPTPATH` not `$SCRIPTDIR` and same in the LLVM_PROFILE_FILE above

scripts/grcov_cxx_qt.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+# SPDX-FileCopyrightText: 2021 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+# SPDX-FileContributor: Ben Ford <ben.ford@kdab.com>
+#
+# SPDX-License-Identifier: MIT OR Apache-2.0
+# Assumes you have grcov and llvm-tools
```
ahayzen-kdab, line 7: lets add a new line between the SPDX info and the comment

ahayzen-kdab (approved):
LGTM :-)

PR #1026, opened 2024-08-06
Add exclude flags to grcov in CI

BenFordTytherington:
Also adds markers in src/lib.rs to exclude lines used in the update_expected script
- CODECOV_EXCLUDE_START in a comment before the block to exclude
- CODECOV_EXCLUDE_STOP in a comment after the block

### DIFF
--- a/.github/workflows/github-cxx-qt-tests.yml
+++ b/.github/workflows/github-cxx-qt-tests.yml
@@ -78,7 +78,7 @@ jobs:
           LLVM_PROFILE_FILE: coverage/coverage_data-%p-%m.profraw
         run: cargo test --lib --package cxx-qt-gen
       - name: generate-report
-        run: grcov . -s . --binary-path ./target/debug/ -t lcov --branch --ignore-not-existing -o ./target/debug/lcov.info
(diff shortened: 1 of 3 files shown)


.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -78,7 +78,7 @@ jobs:
           LLVM_PROFILE_FILE: coverage/coverage_data-%p-%m.profraw
         run: cargo test --lib --package cxx-qt-gen
       - name: generate-report
-        run: grcov . -s . --binary-path ./target/debug/ -t lcov --branch --ignore-not-existing -o ./target/debug/lcov.info
+        run: grcov . -s . --binary-path ./target/debug/ -t lcov --branch --ignore-not-existing -o ./target/debug/lcov.info --excl-start CODECOV_EXCLUDE_START --excl-stop CODECOV_EXCLUDE_STOP
```
ahayzen-kdab, line 81: should probably add this to the script as well so that the local script matches the same output?

BenFordTytherington: yep will do

ahayzen-kdab (approved):
LGTM :-)

PR #1027, opened 2024-08-06
Improve code coverage in syntax directory

BenFordTytherington:
Includes adding a new test input file for reading shebang, and also for parsing of cxx and regular rust items by CxxQtFile

### DIFF
--- a/crates/cxx-qt-gen/src/syntax/foreignmod.rs
+++ b/crates/cxx-qt-gen/src/syntax/foreignmod.rs
@@ -232,6 +232,17 @@ mod tests {
     }
 
     #[test]
+    fn test_foreign_mod_to_foreign_item_types_invalid() {
+        let item: ItemForeignMod = parse_quote! {
+            extern "RustQt" {
+                fn my_function() {}
+            }
+        };
+        let result = foreign_mod_to_foreign_item_types(&item);
+        assert!(result.is_err())
+    }
+
+    #[test]
     fn test_foreign_fn_self() {
         let foreign_fn: ForeignItemFn = parse_quote! {
             fn foo(self: &T, a: A) -> B;
--- a/crates/cxx-qt-gen/src/syntax/foreignmod.rs
+++ b/crates/cxx-qt-gen/src/syntax/foreignmod.rs
@@ -37,7 +37,7 @@ pub(crate) fn foreign_mod_to_foreign_item_types(
         .filter_map(|item| match foreign_item_to_type(item) {
             Ok(Some(value)) => Some(Ok(value)),
             Ok(None) => None,
-            Err(err) => Some(Err(err)), // Hard to cover since foreign_item_to_type never returns error
+            Err(err) => Some(Err(err)),
         })
         .collect::<Result<Vec<ForeignItemType>>>()
 }
--- a/crates/cxx-qt-gen/src/syntax/foreignmod.rs
+++ b/crates/cxx-qt-gen/src/syntax/foreignmod.rs
@@ -37,7 +37,7 @@ pub(crate) fn foreign_mod_to_foreign_item_types(
         .filter_map(|item| match foreign_item_to_type(item) {
             Ok(Some(value)) => Some(Ok(value)),
             Ok(None) => None,
-            Err(err) => Some(Err(err)),
+            Err(err) => Some(Err(err)), // Hard to cover since foreign_item_to_type never returns error
         })
         .collect::<Result<Vec<ForeignItemType>>>()
 }
--- a/crates/cxx-qt-gen/src/syntax/foreignmod.rs
+++ b/crates/cxx-qt-gen/src/syntax/foreignmod.rs
@@ -3,7 +3,7 @@
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
-use proc_macro2::{TokenStream, TokenTree};
+use proc_macro2::TokenStream;
 use quote::{quote, ToTokens};
 use syn::{
     parse::{Parse, ParseStream, Parser},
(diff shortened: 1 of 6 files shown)


crates/cxx-qt-gen/src/syntax/foreignmod.rs
```diff
@@ -37,7 +37,7 @@ pub(crate) fn foreign_mod_to_foreign_item_types(
         .filter_map(|item| match foreign_item_to_type(item) {
             Ok(Some(value)) => Some(Ok(value)),
             Ok(None) => None,
-            Err(err) => Some(Err(err)),
+            Err(err) => Some(Err(err)), // Hard to cover since foreign_item_to_type never returns error
```
ahayzen-kdab, line 40: You would need `verbatim_to_foreign_type` to have failed i believe?

So eg if you had a `type T = abc` with not `;` at the end this would Err ?

So if you had an input of
```rust
extern "RustQt" {
   type A = B // no semi-colon should fail
}
```

does this work?

crates/cxx-qt-gen/src/syntax/lifetimes.rs
```diff
@@ -148,5 +150,7 @@ mod tests {
         assert_unsupported_type! { &dyn Foo };
         assert_unsupported_type! { fn(A) };
         assert_unsupported_type! { fn(i64) -> i32 };
+        // assert_unsupported_type! { Fn(i64, i64) -> i32 }; // This currently just panics so test fails
```
ahayzen-kdab, line 153: we're trying to hit `PathArguments::Parenthesized` so maybe needs to be like  `assert_unsupported_type! { Vec<Fn(i64, i64) -> i32> }` as it should be in the arguments of a path?

ahayzen-kdab: Hmm no i don't think that is quite right either, seems it the path part would be the `Fn`

ahayzen-kdab: Lets dig into it later :sweat_smile:

crates/cxx-qt-gen/src/syntax/foreignmod.rs
ahayzen-kdab: As discussed this that semi-colon check is never hit, as this happens in syn

- When a `type` is found https://github.com/dtolnay/syn/blob/94d74d52ca8ec4fb46317f261edd678f9231e178/src/item.rs#L1864 it uses `parse_foreign_item_type`
- `parse_foreign_item_type` build a `FlexibleItemType` and returns Verbatim if the type has a colon or type
- When `FlexibleItemType` is run through Parse it requires valid syntax with a `;` at the end otherwise it fails https://github.com/dtolnay/syn/blob/94d74d52ca8ec4fb46317f261edd678f9231e178/src/item.rs#L1149

So if we ever get a Verbatim block that starts with a `type` it must also have a matching `;` at the end, and we only care about the type and the ident in this function, so that whole `input.step` can be removed

ahayzen-kdab, line 40: to reach the `Unsupported verbatim input in ForeignItem` error we just need to construct a Verbatim input that doesn't start with `type`.

A fn one is quite simple to do https://github.com/dtolnay/syn/blob/94d74d52ca8ec4fb46317f261edd678f9231e178/src/item.rs#L1825

So basically this should then hit that error case, as it has a Brace
```rust
extern "RustQt" {
   fn my_function() {}
}
```

ahayzen-kdab, line 40: can remove this comment as we have an error case now?

crates/cxx-qt-gen/src/syntax/qtfile.rs
```diff
@@ -58,3 +58,91 @@ pub fn parse_qt_file(path: impl AsRef<std::path::Path>) -> Result<CxxQtFile> {
...
+    fn test_parse_qt_file_shebang_strip() {
+        let manifest_dir = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
+
+        let qt_file = parse_qt_file(manifest_dir.join("test_inputs/shebang.rs")).unwrap();
+
+        assert_eq!(qt_file.items.len(), 3);
+        // Testing the debug formatter
+        for item in qt_file.items {
+            println!("[Debug] item: {:?}", item);
+        }
```
ahayzen-kdab, line 93: this should be a test in `qtitem.rs` i think? And there you could do things much tighter so you could have

```rust
let cxx: CxxQtItem = parse_quote!{
  #[cxx::bridge]
  mod ffi {}
};
let debug_formatted = format!("{:?}", cxx);
assert_eq!(debug_formatted, ...);
```
and you can then do the same for Item / Cxx / CxxQt

crates/cxx-qt-gen/src/syntax/foreignmod.rs
BenFordTytherington: done

crates/cxx-qt-gen/src/syntax/qtfile.rs
```diff
@@ -58,3 +58,91 @@ pub fn parse_qt_file(path: impl AsRef<std::path::Path>) -> Result<CxxQtFile> {
...
+    }
+
+    #[test]
+    #[should_panic]
+    fn test_parse_invalid_qt_file() {
+        let manifest_dir = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
+
+        let _qt_file = parse_qt_file(manifest_dir.join("not/real"));
```
ahayzen-kdab, line 101: could just be `let _ =`

crates/cxx-qt-gen/test_inputs/shebang.rs
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/
+// Test for qtfile
```
ahayzen-kdab, line 2: i would still have one item in here and then in your `test_parse_qt_file_shebang_strip` have those checks that are in the other one

```
        assert!(qt_file.attrs.is_empty());
        assert_eq!(qt_file.items.len(), 1);
        assert!(matches!(qt_file.items.first(), Some(CxxQt(_))))
````

Just so that you ensure that when there is a shebang, yes we strip it but do we also still parse the remainder of the file OK?

**Review by ahayzen-kdab — APPROVED** [review]
LGTM, lets merge this then do more iterations to get all the way to 100%

File: crates/cxx-qt-gen/src/syntax/qtitem.rs [hunk 6]
```diff
@@ -82,3 +82,42 @@ impl ToTokens for CxxQtItem {
...
+
+    #[test]
+    fn test_format_rust_item() {
+        let rust: CxxQtItem = parse_quote! {
+          struct MyStruct {
+                name: &str
+          }
+        };
+        let debug_formatted = format!("{:?}", rust);
+        print!("rust: {:?}", debug_formatted);
```
ahayzen-kdab, line 120: don't need the print lines here (and in the others)

ahayzen-kdab (approved):
LGTM :-)

PR #1028, opened 2024-08-07
Improve coverage in naming

BenFordTytherington:


crates/cxx-qt-gen/src/naming/cpp.rs
```diff
@@ -332,6 +332,64 @@ mod tests {
...
+        );
+
+        let ty = parse_quote! { CxxVector<A> };
+        let mut type_names = TypeNames::default();
+        type_names.mock_insert("A", None, Some("A1"), None);
+        assert_eq!(
+            syn_type_to_cpp_type(&ty, &type_names).unwrap(),
+            "::std::vector<A1>"
+        );
+    }
```
ahayzen-kdab, line 383: i think these ones can go into the other test `test_syn_type_to_cpp_type` that has a helper so these would be
```
{ SharedPtr<T } => "::std::shared_ptr<T>",
```

Or did you specifically need the A -> A1 conversion also to happen? If so you could add
`type_names.mock_insert("A", None, Some("A1"), None);`  to the `test_syn_types_to_cpp_types` then use `SharedPtr<A>` etc

BenFordTytherington: Used them in the macro helper instead now

ahayzen-kdab (approved):
LGTM :-)

PR #1030, opened 2024-08-07
Improve coverage in generator

BenFordTytherington:


crates/cxx-qt-gen/src/generator/structuring/mod.rs
```diff
@@ -124,6 +124,83 @@ mod tests {
         assert!(structures.is_err());
     }
 
+    #[test]
+    fn test_module_invalid_qobject() {
```
ahayzen-kdab, line 128: prefer lots of small tests, so can you split these into individual ones as when they fail it's much more obvious what is causing it then
```
fn test_module_invalid_qobject_enum()
fn test_module_invalid_qobject_invokable()
fn test_module_invalid_qobject_signal()
fn test_module_invalid_qobject_inherit()
```

crates/cxx-qt-gen/src/generator/cpp/fragment.rs
```diff
@@ -32,3 +32,21 @@ pub struct CppNamedType {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_default_creation() {
+        let default_fragment = CppFragment::default();
+        // This feels like there should be a better way to test the default
```
ahayzen-kdab, line 43: maybe something like
```rust
if let CppFragment::Pair { header, source } = default_fragment {
   assert!(header.is_empty());
   assert!(source.is_empty());
} else {
   panic!("Expected a CppFragment::Pair");
}

ahayzen-kdab: Seems codecov thinks that's worse, lets go back to what you had before then :-)

ahayzen-kdab (approved):
LGTM :-)

PR #1040, opened 2024-08-19
Improve coverage in generator

BenFordTytherington:


crates/cxx-qt-gen/src/generator/rust/signals.rs
```diff
@@ -65,14 +65,18 @@ pub fn generate_rust_signal(
...
+                // CODECOV_EXCLUDE_STOP
+            }
         })
         .collect::<Result<_>>()?;
 
     let self_type_cxx = if signal.mutable {
         parse_quote! { Pin<&mut #qobject_name_rust> }
     } else {
-        parse_quote! { &#qobject_name_rust }
+        parse_quote! { &#qobject_name_rust } // Signals cannot be mutable
```
ahayzen-kdab, line 79: Think that comment should be `Signals cannot be immutable` ? As they have to be mutable :-) (and this is another unreachable!() place i assume?)

crates/cxx-qt-gen/src/generator/cpp/fragment.rs
```diff
@@ -19,6 +19,19 @@ pub enum CppFragment {
...
+    fn into(self) -> String {
+        match self {
+            CppFragment::Pair { header, source } => {
+                format!("Header:\n  {header}\nSource:\n  {source}")
+            }
+            CppFragment::Header(header) => header,
+            CppFragment::Source(source) => source,
+        }
+    }
+}
```
ahayzen-kdab, line 33: this is not used i don't think? was it a previous attempt before the `require_header` / `require_pair` things?

crates/cxx-qt-gen/src/generator/cpp/fragment.rs
```diff
@@ -47,4 +60,11 @@ mod tests {
             }
         )
     }
+
+    #[test]
+    fn test_into_string() {
+        let source = CppFragment::Source(String::from("SOURCE"));
+        let str_repr: String = source.into();
+        assert_eq!(str_repr.as_str(), "SOURCE")
+    }
```
ahayzen-kdab, line 69: this could be removed too?

crates/cxx-qt-gen/src/generator/cpp/mod.rs
```diff
@@ -92,20 +91,17 @@ pub fn get_cpp_params(method: &ForeignItemFn, type_names: &TypeNames) -> Result<
...
                 let ident = if let Pat::Ident(PatIdent { ident, .. }) = &**pat {
                     ident
-                } else {
-                    return Err(Error::new(input.span(), "Unknown pattern for type"));
+                }
+                // CODECOV_EXCLUDE_START
+                else {
+                    unreachable!("Unknown pattern for type, FnArg can only have Pat::Ident")
                 };
+                // CODECOV_EXCLUDE_STOP
```
ahayzen-kdab, line 99: in other places you have done

```
                    } else {
                        // CODECOV_EXCLUDE_START
                        unreachable!(...);
                        // CODECOV_EXCLUDE_STOP
                    }
```

which then doesn't break the coding style of having `} else {` instead of `else` on it's own line. Can the same be done here?

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
```diff
@@ -80,8 +81,37 @@ mod tests {
...
     use quote::format_ident;
-    use syn::{parse_quote, ItemMod, ItemStruct};
+    use syn::{parse_quote, Error, ItemMod, ItemStruct};
+
+    pub fn require_pair(fragment: &CppFragment) -> Result<(String, String)> {
+        match fragment {
+            CppFragment::Pair { header, source } => Ok((header.clone(), source.clone())),
+            _ => {
+                let fragment_string: String = fragment.clone().into();
+                Err(Error::new_spanned(fragment_string, "Expected Pair!"))
```
ahayzen-kdab, line 91: ah i see the string thing is coming from here

crates/cxx-qt-gen/src/generator/cpp/fragment.rs
```diff
@@ -19,6 +19,19 @@ pub enum CppFragment {
     Source(String),
 }
 
+#[cfg(test)]
+impl Into<String> for CppFragment {
```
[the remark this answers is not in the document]
ahayzen-kdab: (Or you could use `Debug` too https://doc.rust-lang.org/stable/std/fmt/trait.Debug.html )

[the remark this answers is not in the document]
ahayzen-kdab: Which may be best as this is developer facing info?

crates/cxx-qt-gen/src/generator/cpp/fragment.rs
```diff
@@ -3,7 +3,7 @@
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
-#[derive(PartialEq, Eq, Debug)]
+#[derive(PartialEq, Eq, Debug, Clone)]
```
ahayzen-kdab, line 6: remove the derive of `Clone` once you change to `impl Debug`

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
ahayzen-kdab, line 91: Have you tested what happens when the error occurs? Eg do the tokens even create a nice error message?

As I would note that here we are in C++ land so you don't have to return a `syn::Error` you could return a normal error type or panic / unwrap in the test :thinking:

crates/cxx-qt-gen/src/generator/rust/signals.rs
BenFordTytherington: yeah it is for now, I was wondering if that indicates a bigger refactor though, that we have a signal mutability field and options handling it, but they can never be immutable

ahayzen-kdab: For now we are leaving the door open to having them be immutable too so just erroring in the parser for now, but i'm unsure if that works on the Qt side as you typically make them mutable in C++. This is something we need to check and decide before 1.0.

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
ahayzen-kdab: But maybe a `syn::Error` is able to create a nice message here?

crates/cxx-qt-gen/src/generator/cpp/fragment.rs
BenFordTytherington: It's used in the require functions so I can format an error in the catchall block, e.g. if I'm matching:
```rust
CppFragment::Pair => { ... },
_ => Err(Error::new_spanned( ... )
```
I call fragment.clone.into() and it'll format that into the error. It lets me avoid matching Source, so requires less testing

crates/cxx-qt-gen/src/generator/cpp/fragment.rs
BenFordTytherington: Needed for coverage of the into() trait I believe

crates/cxx-qt-gen/src/generator/cpp/mod.rs
BenFordTytherington: yep, fixed

crates/cxx-qt-gen/src/generator/cpp/fragment.rs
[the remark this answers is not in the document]
BenFordTytherington: Debug is derived already so I think using that does the trick

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
BenFordTytherington: panic!() isn't a good option because it will be an uncovered line, but yeah I was only using the tokens to create a syn error so can maybe switch to a regular result and avoid needing it tbh

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
```diff
@@ -71,18 +72,39 @@ pub fn generate_cpp_properties(
 mod tests {
     use super::*;
 
-    use crate::parser::property::QPropertyFlags;
+    use core::result::Result;
```
ahayzen-kdab, line 75: instead of an import that shadows you could use the fully qualified name for the returns...

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
```diff
@@ -71,18 +72,39 @@ pub fn generate_cpp_properties(
...
+    use crate::parser::property::QPropertyFlags;
     use crate::parser::qobject::ParsedQObject;
     use crate::{CppFragment, Parser};
     use indoc::indoc;
     use pretty_assertions::assert_str_eq;
     use quote::format_ident;
     use syn::{parse_quote, ItemMod, ItemStruct};
 
-    fn setup_generated(input: &mut ItemStruct) -> Result<GeneratedCppQObjectBlocks> {
+    pub fn require_pair(fragment: &CppFragment) -> Result<(String, String), String> {
```
ahayzen-kdab, line 87: so here use `core::result::Result<...>`

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
```diff
@@ -71,18 +72,39 @@ pub fn generate_cpp_properties(
...
+        }
+    }
+
+    pub fn require_source(fragment: &CppFragment) -> Result<String, String> {
+        match fragment {
+            CppFragment::Source(source) => Ok(source.clone()),
+            _ => Err(format!("Expected just a source, got {fragment:?} instead")),
+        }
+    }
+    fn setup_generated(input: &mut ItemStruct) -> syn::Result<GeneratedCppQObjectBlocks> {
```
ahayzen-kdab, line 107: then add a new line above this function and put it back to just `Result<...>`

crates/cxx-qt-gen/src/generator/rust/signals.rs
```diff
@@ -65,13 +65,18 @@ pub fn generate_rust_signal(
...
+                unreachable!("should only have typed no receiver")
+                // CODECOV_EXCLUDE_STOP
+            }
         })
         .collect::<Result<_>>()?;
 
     let self_type_cxx = if signal.mutable {
         parse_quote! { Pin<&mut #qobject_name_rust> }
     } else {
+        // Signals cannot be immutable
```
ahayzen-kdab, line 79: Change to at least `// Note: Signals cannot be immutable so this could isn't reached`

crates/cxx-qt-gen/src/generator/cpp/fragment.rs
```diff
@@ -47,4 +47,11 @@ mod tests {
             }
         )
     }
+
+    #[test]
+    fn test_into_string() {
+        let source = CppFragment::Source(String::from("SOURCE"));
+        let str_repr: String = source.into();
+        assert_eq!(str_repr.as_str(), "SOURCE")
+    }
```
ahayzen-kdab, line 56: remove this change?

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
```diff
@@ -71,17 +72,37 @@ pub fn generate_cpp_properties(
 mod tests {
```
ahayzen-kdab, line 72: appears this needs to be at least `pub(crate)` otherwise other places cannot reach it

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
```diff
@@ -11,6 +11,7 @@ use crate::generator::{
 use crate::{
     naming::cpp::syn_type_to_cpp_type, naming::TypeNames, parser::property::ParsedQProperty,
 };
+
```
ahayzen-kdab, line 14: can remove this empty line change

ahayzen-kdab (approved):
LGTM :-)

PR #1041, opened 2024-08-19
Fix alignment & size assertions

mattkdab:
Padding is currently manually added to size.
Some integer sizes are also assumed based on 64bit arch. This causes issues when compiling for 32-bit arch's and hardcoding padding is generally not the cleanest.

Resolve by calculating sizes and padding via traversing data members in reverse. Fully constexpr.

### DIFF
--- a/crates/cxx-qt-lib-extras/src/assertion_utils.h
+++ b/crates/cxx-qt-lib-extras/src/assertion_utils.h
@@ -11,7 +11,7 @@
 #include <cassert>
 #include <cstdint>
 
-namespace assertion_util {
+namespace rust::cxxqtlib1 {
 template<typename iter>
 constexpr static ::std::size_t
 calc_align_size(const iter rbegin,
@@ -44,7 +44,7 @@ calc_align_size(const iter rbegin,
 
   return rows * actual_alignment;
 }
-} // namespace assertion_util
+} // namespace rust::cxxqtlib1
 
 #define assert_alignment_and_size(TYPE, EXP_ALIGN, ARR, ARR_SZ)                \
   static_assert(EXP_ALIGN == alignof(TYPE));                                   \
--- a/crates/cxx-qt-lib-extras/src/assertion_utils.h
+++ b/crates/cxx-qt-lib-extras/src/assertion_utils.h
@@ -8,8 +8,8 @@
 #pragma once
 
 #include <array>
-#include <cstdint>
 #include <cassert>
+#include <cstdint>
 
 namespace assertion_util {
 constexpr static ::std::size_t
--- a/crates/cxx-qt-lib-extras/src/assertion_utils.h
+++ b/crates/cxx-qt-lib-extras/src/assertion_utils.h
@@ -12,13 +12,13 @@
 #include <cstdint>
 
 namespace assertion_util {
+template<typename iter>
 constexpr static ::std::size_t
-calc_align_size(const ::std::reverse_iterator<const ::std::size_t*> rbegin,
-                const ::std::reverse_iterator<const ::std::size_t*> rend,
+calc_align_size(const iter rbegin,
+                const iter rend,
                 const ::std::size_t actual_alignment)
 {
   ::std::size_t rows = 0;
-
   ::std::size_t accum = 0;
 
   for (auto it = rbegin; it != rend; ++it) {
@@ -46,8 +46,9 @@ calc_align_size(const ::std::reverse_iterator<const ::std::size_t*> rbegin,
 }
 } // namespace assertion_util
 
-#define assert_alignment_and_size(TYPE, ALIGNMENT, ARR)                        \
-  static_assert(ALIGNMENT == alignof(TYPE));                                   \
-  static_assert(assertion_util::calc_align_size(                               \
-                  ::std::rbegin(ARR), ::std::rend(ARR), alignof(TYPE)) ==      \
-                sizeof(TYPE));
+#define assert_alignment_and_size(TYPE, EXP_ALIGN, ARR, ARR_SZ)                \
+  static_assert(EXP_ALIGN == alignof(TYPE));                                   \
+  static_assert(                                                               \
+    assertion_util::calc_align_size<                                           \
+      ::std::array<::std::size_t, ARR_SZ>::const_reverse_iterator>(            \
+      ::std::rbegin(ARR), ::std::rend(ARR), alignof(TYPE)) == sizeof(TYPE));
--- a/crates/cxx-qt-lib-extras/src/assertion_utils.h
+++ b/crates/cxx-qt-lib-extras/src/assertion_utils.h
@@ -7,7 +7,47 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 #pragma once
 
-#define assert_alignment_and_size(TYPE, ALIGNMENT, SIZE)                       \
-  static_assert(alignof(TYPE) <= (ALIGNMENT),                                  \
-                "unexpectedly large " #TYPE " alignment!");                    \
-  static_assert(sizeof(TYPE) == (SIZE), "unexpected " #TYPE " size!");
+#include <array>
+#include <cstdint>
+#include <cassert>
+
+namespace assertion_util {
+constexpr static ::std::size_t
+calc_align_size(const ::std::reverse_iterator<const ::std::size_t*> rbegin,
+                const ::std::reverse_iterator<const ::std::size_t*> rend,
+                const ::std::size_t actual_alignment)
+{
+  ::std::size_t rows = 0;
+
+  ::std::size_t accum = 0;
+
+  for (auto it = rbegin; it != rend; ++it) {
+    assert(*it <= actual_alignment);
+
+    if (it + 1 != rend) {
+      if (accum + *it == actual_alignment) {
+        accum += *it;
+      } else {
+        if (accum + *it + *(it + 1) <= actual_alignment)
+          accum += *it;
+        else
+          accum = actual_alignment;
+      }
+      if (accum == actual_alignment) {
+        accum = 0;
+        ++rows;
+      }
+    } else {
+      ++rows;
+    }
+  }
+
+  return rows * actual_alignment;
+}
+} // namespace assertion_util
+
+#define assert_alignment_and_size(TYPE, ALIGNMENT, ARR)                        \
+  static_assert(ALIGNMENT == alignof(TYPE));                                   \
+  static_assert(assertion_util::calc_align_size(                               \
+                  ::std::rbegin(ARR), ::std::rend(ARR), alignof(TYPE)) ==      \
+                sizeof(TYPE));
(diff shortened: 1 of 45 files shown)


crates/cxx-qt-lib/src/core/qvariant/qvariant.cpp
```diff
@@ -24,34 +24,17 @@
...
-// but due to compiler optimisation it ends up as
-// 3 * ushort (6) + union with double (8) + 2 bytes padding
-// alignment is 8 byte on 32bit systems as well due to the double
-assert_alignment_and_size(
-  QVariant,
-  alignof(double),
-  (sizeof(::std::uint16_t /* compiler optimised from ::std::uint32_t */) * 3) +
-    sizeof(double) + 2 /* compiler padding */);
-
+constexpr static ::std::array<::std::size_t, 4> arr{ sizeof(uint),
```
ahayzen-kdab, line 33: as discussed in places like this, i think we should still use `::std::uint16_t` as we are using `u16` in Rust. As we want to check that the Rust size ends up the same as the Qt size. We know that the Qt type uses `uint` so it's not actually checking the right thing if we have `uint` here.

crates/cxx-qt-lib/src/gui/qcolor.cpp
```diff
@@ -16,11 +16,11 @@
 // compiler padding this results in a sizeof 16.
 // https://code.qt.io/cgit/qt/qtbase.git/tree/src/gui/painting/qcolor.h?h=v5.15.6-lts-lgpl#n262
 // https://code.qt.io/cgit/qt/qtbase.git/tree/src/gui/painting/qcolor.h?h=v6.2.4#n237
-assert_alignment_and_size(QColor,
-                          alignof(::std::size_t),
-                          sizeof(::std::int32_t) +
-                            (sizeof(::std::uint16_t) * 5) +
-                            2 /* compiler padding */);
```
ahayzen-kdab, line 23: note that on the Rust side we have
```rust
#[repr(C)]                                                                                                                                                                                                                             
pub struct QColor {                                                                                                                                                                                                                    
    _cspec: MaybeUninit<i32>,                                                                                                                                                                                                          
    _ct: MaybeUninit<[u16; 5]>,                                                                                                                                                                                                        
    _padding: MaybeUninit<u16>,                                                                                                                                                                                                        
} 
```

Wonder if we should not have the padding here on the Rust side and leave the compiler to fill any padding automatically?

crates/cxx-qt-lib/src/core/qvariant/qvariant.cpp
mattkdab: as we discussed before, the packed uints should probably be represented as a single u32 rather than three u16's. maybe a task for another PR? or should i go ahead and do it here?

mattkdab: nevermind, done.

crates/cxx-qt-lib/src/gui/qimage.cpp
```diff
@@ -20,15 +20,18 @@
...
 #else
 // In Qt6 the QPaintDevice doesn't contain the `reserved` pointer, making it 1
 // pointer smaller
-assert_alignment_and_size(QImage,
-                          alignof(::std::size_t),
-                          sizeof(::std::size_t) * 3);
+constexpr static ::std::array<::std::size_t, 3> arr{ sizeof(ushort),
+                                                     sizeof(::std::size_t),
+                                                     sizeof(::std::size_t) };
+assert_alignment_and_size(QImage, alignof(::std::size_t), arr, arr.size());
```
ahayzen-kdab, line 34: Rust side uses either 3x or 4x of `usize` this should be changed to `u16` + 2/3x `usize`?

And then the C++ side should use `::std::uint16_t` instead of ushort as we want to use the modern numerics that enforce the type size to be the same as Rust?

crates/cxx-qt-lib/src/gui/qcolor.cpp
```diff
@@ -16,11 +16,11 @@
...
+constexpr static ::std::array<::std::size_t, 6> arr{
+  sizeof(::std::int32_t),  sizeof(::std::uint16_t), sizeof(::std::uint16_t),
+  sizeof(::std::uint16_t), sizeof(::std::uint16_t), sizeof(::std::uint16_t)
+};
+assert_alignment_and_size(QColor, alignof(::std::int32_t), arr, arr.size());
```
ahayzen-kdab, line 23: the rust side of `QColor` has a field called `_padding` seems like this should be removed as `#[repr(C)]` should ensure the right layout and alignment should occur? Can you try removing that on the Rust side?

Then the Rust side has i32 + 5x u16 which matches the size array here

ahayzen-kdab: ah wait you did that below, i missed that :-) ...done!

crates/cxx-qt-lib/src/gui/qfont.cpp
```diff
@@ -11,9 +11,9 @@
 
 // https://code.qt.io/cgit/qt/qtbase.git/tree/src/gui/text/qfont.h?h=v5.15.6-lts-lgpl#n344
 // https://code.qt.io/cgit/qt/qtbase.git/tree/src/gui/text/qfont.h?h=v6.2.4#n323
-assert_alignment_and_size(QFont,
-                          alignof(::std::size_t),
-                          sizeof(::std::size_t) + sizeof(::std::int64_t));
+constexpr static ::std::array<::std::size_t, 2> arr{ sizeof(::std::size_t),
+                                                     sizeof(uint) };
+assert_alignment_and_size(QFont, alignof(::std::size_t), arr, arr.size());
```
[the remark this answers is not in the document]
mattkdab: it should probably be changed to `usize` + `u32` as the members are pointer and uint
it will align to the pointer so regardless of whether the uint is 16 or 32 it should be the same size i think

crates/cxx-qt-lib/src/gui/qimage.cpp
mattkdab: yes

crates/cxx-qt-lib-extras/src/assertion_utils.h
```diff
@@ -7,7 +7,48 @@
...
 
-#define assert_alignment_and_size(TYPE, ALIGNMENT, SIZE)                       \
-  static_assert(alignof(TYPE) <= (ALIGNMENT),                                  \
-                "unexpectedly large " #TYPE " alignment!");                    \
-  static_assert(sizeof(TYPE) == (SIZE), "unexpected " #TYPE " size!");
+#include <array>
+#include <cassert>
+#include <cstdint>
+
+namespace assertion_util {
```
ahayzen-kdab, line 14: should this be under `rust::cxxqtlib1` like all of our other C++ types?

crates/cxx-qt-lib/src/gui/qimage.rs
```diff
@@ -274,9 +274,13 @@ pub struct QImage {
...
     #[cfg(cxxqt_qt_version_major = "5")]
-    _data: MaybeUninit<[usize; 4]>,
+    _painters: MaybeUninit<u16>,
+    #[cfg(cxxqt_qt_version_major = "5")]
+    _pointers: MaybeUninit<[usize; 3]>,
+    #[cfg(cxxqt_qt_version_major = "6")]
+    _painters: MaybeUninit<u16>,
     #[cfg(cxxqt_qt_version_major = "6")]
-    _data: MaybeUninit<[usize; 3]>,
+    _pointers: MaybeUninit<[usize; 2]>,
```
ahayzen-kdab, line 283: Can be simplified to this as the `_painters` part is the same for Qt 5 or 6 ?
```rust
    _painters: MaybeUninit<u16>,

    #[cfg(cxxqt_qt_version_major = "5")]
    _pointers: MaybeUninit<[usize; 3]>,
    #[cfg(cxxqt_qt_version_major = "6")]
    _pointers: MaybeUninit<[usize; 2]>,
```

crates/cxx-qt-lib/src/assertion_utils.h
```diff
@@ -7,7 +7,48 @@
...
 
-#define assert_alignment_and_size(TYPE, ALIGNMENT, SIZE)                       \
-  static_assert(alignof(TYPE) <= (ALIGNMENT),                                  \
-                "unexpectedly large " #TYPE " alignment!");                    \
-  static_assert(sizeof(TYPE) == (SIZE), "unexpected " #TYPE " size!");
+#include <array>
+#include <cassert>
+#include <cstdint>
+
+namespace assertion_util {
```
ahayzen-kdab, line 14: same namespace thing here this one would be within `rust::cxxqtlib1`

crates/cxx-qt-lib/src/gui/qfont.rs
```diff
@@ -368,8 +368,8 @@ pub use ffi::{
 
 #[repr(C)]
 pub struct QFont {
-    _cspec: MaybeUninit<i32>,
-    _resolve_mask: MaybeUninit<u16>,
+    _cspec: MaybeUninit<usize>,
+    _resolve_mask: MaybeUninit<i32>,
```
ahayzen-kdab, line 372: small thing but shouldn't this be `u32` not `i32` as it's std::uint32 / uint on the C++ side ?

crates/cxx-qt-lib/src/assertion_utils.h
```diff
@@ -7,7 +7,48 @@
...
+  }
+
+  return rows * actual_alignment;
+}
+} // namespace rust::cxxqtlib1::assertion_utils
+
+#define assert_alignment_and_size(TYPE, EXP_ALIGN, ARR, ARR_SZ)                \
+  static_assert(EXP_ALIGN == alignof(TYPE));                                   \
+  static_assert(                                                               \
+    rust::cxxqtlib1::assertion_utils::calc_align_size<                         \
```
ahayzen-kdab, line 52: while making changes use `::rust::...` instead of `rust::...` to be fully qualified

crates/cxx-qt-lib-extras/src/assertion_utils.h
```diff
@@ -7,7 +7,48 @@
...
+  }
+
+  return rows * actual_alignment;
+}
+} // namespace rust::cxxqtlib1::assertion_utils
+
+#define assert_alignment_and_size(TYPE, EXP_ALIGN, ARR, ARR_SZ)                \
+  static_assert(EXP_ALIGN == alignof(TYPE));                                   \
+  static_assert(                                                               \
+    rust::cxxqtlib1::assertion_utils::calc_align_size<                         \
```
ahayzen-kdab, line 52: while making changes use `::rust::...` instead of `rust::...` to be fully qualified

crates/cxx-qt-lib/src/gui/qfont.rs
mattkdab: good catch

ahayzen-kdab (approved):
LGTM :-)

PR #1042, opened 2024-08-20
Improve coverage in naming

BenFordTytherington:


crates/cxx-qt-gen/src/naming/name.rs
```diff
@@ -146,6 +146,7 @@ impl Name {
         let mut namespace = if let Some(index) = attribute_find_path(attrs, &["namespace"]) {
             Some(expr_to_string(
                 &attrs[index].meta.require_name_value()?.value,
+                // requires non string literal so might be uncatchable
```
ahayzen-kdab, line 149: unreachable?

crates/cxx-qt-gen/src/naming/cpp.rs
```diff
@@ -118,12 +128,8 @@ pub(crate) fn syn_type_to_cpp_type(ty: &Type, type_names: &TypeNames) -> Result<
...
                 .map(|generic| path_segment_to_string(generic, type_names))
                 .collect::<Result<Vec<String>>>()?;
-            if ty_strings.len() == 1 {
-                let first = ty_strings.first().unwrap();
-                Ok(first.to_owned())
-            } else {
-                Ok(ty_strings.join("::"))
-            }
+            let first = ty_strings.first().unwrap();
+            Ok(first.to_owned())
```
ahayzen-kdab, line 132: wonder if we should put unreachable or an error here? As we think that it is wrong if we have a path that has multiple segments somehow.

I think that one that would have hit the old could would be if it had been like `T::T`  where `T` is a type known in `type_names`? 

So maybe change to 
```rust
            if ty_strings.len() == 1 {
                let first = ty_strings.first().unwrap();
                Ok(first.to_owned())
            } else {
                // TODO: Implement a way of resolving multiple segments in TypeNames
                // this would need changes also to the map(...) above where each segment
                // is changed to the C++ type, instead the whole path needs to be considered
                Err(Error::new(ty.span(), "Paths with multiple segments are not supported");
            }
```

Then write a test that has a type in TypeNames and use it like above twice in a path. As the map(...) will resolve each segment individually to it's C++ matching name, but then the old could would have combined them wrongly together so best to error for now

crates/cxx-qt-gen/src/naming/name.rs
BenFordTytherington: Not sure but I think so?

ahayzen-kdab: What if you do like `#[namespace = 1]` then it's not a string literal?

ahayzen-kdab: (as it should be `#[namespace = "1"]`

crates/cxx-qt-gen/src/naming/cpp.rs
ahayzen-kdab: Maybe `path_segment_to_string` was outputting `None` for each of them? As `std` -> None, `collections` -> None, `HashMap` -> `::std::map`.

If you eg did `HashMap::HashMap` as the input i would expect then to see a vector with `::std::map`, `::std::map` as the output which was before being combined to `::std::map::::stdmap` :-D

BenFordTytherington: I'll give that a go now

crates/cxx-qt-gen/src/naming/name.rs
BenFordTytherington: Added a test for this

ahayzen-kdab (approved):
LGTM :-)

PR #1044, opened 2024-08-21
Final coverage fixes

BenFordTytherington:

PR #1046, opened 2024-08-22
Update errors to use consistent formatting where applicable

BenFordTytherington:
Checked all use of Error for well written messages and changed the format / typos in some of them.


BenFordTytherington:
I think the coverage checks here will pass once #1044 is merged and rebased here

crates/cxx-qt-gen/src/generator/structuring/qobject.rs
```diff
@@ -44,7 +44,7 @@ impl<'a> StructuredQObject<'a> {
             .map(|method| &method.name)
             .find(|name| name.rust_unqualified() == id)
             .cloned()
-            .ok_or_else(|| Error::new_spanned(id, format!("Method with name '{id}' not found!")))
+            .ok_or_else(|| Error::new_spanned(id, format!("Method with name `{id}` not found!")))
```
LeonMatthesKDAB, line 47: Hm, this is pretty much a duplicate with line 56 apart from `Method`->`Signal`.

Maybe put this in a shared function:
```rust
fn err_member_not_found(kind: &str, id: &str) -> Error
```

crates/cxx-qt-gen/src/naming/cpp.rs
```diff
@@ -87,14 +91,14 @@ pub(crate) fn syn_type_to_cpp_type(ty: &Type, type_names: &TypeNames) -> Result<
                 if let Lit::Int(len) = &len.lit {
                     len.base10_parse::<usize>()?
                 } else {
-                    return Err(Error::new(ty.span(), "Array length must be a integer"));
+                    return Err(Error::new(ty.span(), "Array length must be an integer!"));
```
LeonMatthesKDAB, line 94: Hm, code duplication with line 97. :thinking: 

Can we maybe merge the two consecutive `if let` expressions into one somehow?

Something like:
```rust
if let Expr::Lit(ExprLit { lit: Lit::Int(len), ...}) = &len
```

crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -152,30 +152,30 @@ impl ParsedQObject {
...
-                    "cxx_qt::Locking must be an unsafe impl",
+                    "cxx_qt::Locking must be an `unsafe impl`!",
                 ));
             }
 
             if not.is_none() {
                 return Err(Error::new_spanned(
                     trait_path,
-                    "cxx_qt::Locking is enabled by default, it can only be negated.",
+                    "cxx_qt::Locking is enabled by default, it can only be negated! \ne.g. `impl !cxx_qt::Locking for T`",
```
LeonMatthesKDAB, line 170: Trailing space before `\n`

crates/cxx-qt-gen/src/syntax/foreignmod.rs
```diff
@@ -163,36 +163,36 @@ pub fn self_type_from_foreign_fn(signature: &Signature) -> Result<Receiver> {
...
-                "Attributes on the `self:` receiver are not supported",
+                "Attributes on the `self:` receiver are not supported!",
             ));
         }
 
         if receiver.mutability.is_some() {
             return Err(Error::new(
                 receiver.span(),
-                "mut on self (i.e. `&mut self`) are not supported, use `self: Pin<&mut T>` instead",
+                "mut on self (i.e. `&mut self`) are not supported! use `self: Pin<&mut T>` instead",
```
LeonMatthesKDAB, line 173: `use` -> `Use` due to new sentence now.

Same for the other errors in this file.

crates/cxx-qt-gen/src/syntax/types.rs
```diff
@@ -118,7 +118,7 @@ pub fn extract_qobject_ident(ty: &Type) -> Result<(Ident, Option<Mut>)> {
             if mutability.is_some() {
                 return Err(Error::new_spanned(
                     type_ref,
-                    "Cannot take T by mutable reference, use either `self: &T`, or `Pin<&mut T>`",
+                    "Cannot take T by mutable reference! use either `self: &T`, or `Pin<&mut T>`",
```
LeonMatthesKDAB, line 121: `use` -> `Use`

LeonMatthesKDAB (changes requested):
Hi @BenFordTytherington, thanks a lot for going through these and cleaning them up!

Whilst you're at it, we should try to remove some of the code duplication.
Otherwise, great change!

crates/cxx-qt-gen/src/naming/cpp.rs
LeonMatthesKDAB: Hm, you should be able to get rid of the `attrs: _empty` part with `..` (two dots only, not three, sorry)
See also: https://doc.rust-lang.org/rust-by-example/flow_control/match/destructuring/destructure_structures.html

Overall I think that's cleaner, as you're destructuring in one go.

crates/cxx-qt-gen/src/naming/cpp.rs
```diff
@@ -83,18 +90,29 @@ pub(crate) fn syn_type_to_cpp_return_type(
...
-                }
+            // let len = if let Expr::Lit(len) = &len {
+            //     if let Lit::Int(len) = &len.lit {
+            //         len.base10_parse::<usize>()?
+            //     } else {
+            //         return Err(Error::new(ty.span(), "Array length must be an integer!"));
+            //     }
+            // } else {
+            //     return Err(Error::new(ty.span(), "Array length must be an integer!"));
+            // };
```
LeonMatthesKDAB, line 101: commented code, please remove

crates/cxx-qt-gen/src/naming/cpp.rs
```diff
@@ -170,20 +188,17 @@ pub(crate) fn syn_type_to_cpp_type(ty: &Type, type_names: &TypeNames) -> Result<
...
+        _others => Err(err_unsupported_type(ty)),
     }
 }
 
 /// Convert any generic arguments to C++, eg A and B in Ty<A, B>
 fn generic_argument_to_string(generic: &GenericArgument, type_names: &TypeNames) -> Result<String> {
     match generic {
         GenericArgument::Type(ty) => syn_type_to_cpp_type(ty, type_names),
-        _others => Err(Error::new(
+        other => Err(Error::new(
```
LeonMatthesKDAB, line 199:
```suggestion
        _other => Err(Error::new(
```
This seems to be the reason why CI is failing.
I think you used this in an intermediate commit, but don't anymore, so it needs the `_` back.

crates/cxx-qt-gen/src/syntax/types.rs
```diff
@@ -85,26 +85,26 @@ fn extract_qobject_from_mut_pin(ty: &TypePath) -> Result<(Ident, Mut)> {
...
                 return Err(Error::new_spanned(
                     ty,
-                    "Non reference args in Pin are not allowed",
+                    "Non reference args in Pin are not allowed!",
                 ));
             }
         } else {
             // CODECOV_EXCLUDE_START
-            unreachable!("Pin must use angle brackets for generic args like Pin<&mut T>");
+            unreachable!("Pin must use angle brackets for generic args like Pin<&mut T>!");
```
LeonMatthesKDAB, line 100: This isn't actually `unreachable!`
If I write this:
```rust
    unsafe extern "RustQt" {
        #[qinvokable]
        fn increment_number(self: Pin);
    }
```
The bridge macro will panic!

So this should return a spanned error instead.
Wonder how we missed this for so long :thinking: 
You can also add a corresponding test and remove the CODECOVE_EXCLUDE region.

LeonMatthesKDAB (commented):
Just some small things to go, then we can merge.

crates/cxx-qt-gen/src/naming/cpp.rs
BenFordTytherington: Done

crates/cxx-qt-gen/src/syntax/types.rs
BenFordTytherington: hmm I guess we didn't think of that case when we marked it unreachable, I think maybe it was erroring in a use of parse_quote somewhere so we assumed it would be caught before that block. I've fixed this with a new test

LeonMatthesKDAB (approved):

PR #1048, opened 2024-08-23
Fix calc_align_size correctness

mattkdab:
This function does not currently break anything but it is possible to calculate padding incorrectly in very specific situations.


mattkdab:
Consider the following:

```cpp
class SomeClass {
    char s;
    int i;
    char c;
    double f;
};

int main() {
    static constexpr ::std::array<::std::size_t, 4> arr{
        sizeof(char), sizeof(int), sizeof(char), sizeof(double)};
    
    assert_alignment_and_size(SomeClass, alignof(double), arr, arr.size())
}
```

This will fail currently because the current `calc_align_size` function will calculate this memory layout:

|||||||||
|-|-|-|-|-|-|-|-|
| s | i | i | i | i | c |   |   |
| f | f | f | f | f | f | f | f |

for a total of 16 bytes.

The true layout looks something like this:

|||||||||
|-|-|-|-|-|-|-|-|
| s |   |   |   | i | i | i | i |
| c |   |   |   |   |   |   |   |
| f | f | f | f | f | f | f | f |

which totals 24 bytes.

This is because the `int i` should be at offsets divisible by its size.

Consider this case:

```cpp
class E {
    char c1, c2;
    int i;
    double f;
};
```
Here, the true layout is this:

|||||||||
|-|-|-|-|-|-|-|-|
| c1 | c2 |   |   | i | i | i | i |
| f | f | f | f | f | f | f | f |

which is only 16 bytes, as the `int i` can be offset by 4 bytes.

Another example:

```cpp
class F {
    char c1;
    char c2;
    short s1;
    char c3;
    short s2;
    double f;
};
```

Here, even though we have chars on either side of the first short, the memory layout will look like this:

|||||||||
|-|-|-|-|-|-|-|-|
| c1 | c2 | s1 | s1 | c3 |   | s2 | s2 |
| f | f | f | f | f | f | f | f |

which is only 16 bytes.

The two shorts can both be placed  at offsets divisible by 2 bytes, with the chars fitting in between.

This fix is currently incomplete, but the code needs to be changed to ensure it is calculating correctly.

LeonMatthesKDAB:
@mattkdab This PR is still marked as draft, is it ready for review? If so, please convert it to a normal PR :)

crates/cxx-qt-lib-extras/src/assertion_utils.h
```diff
@@ -14,27 +15,39 @@
...
-      if (accum + *it == actual_alignment) {
-        accum += *it;
+  for (auto it = begin; it != end; ++it) {
+    auto val = *it;
+    assert(val <= actual_alignment);
+
+    if (val > max_size_in_row)
+      max_size_in_row = val;
+
+    if ((row_accum + val) % max_size_in_row == 0) {
```
LeonMatthesKDAB, line 34: I believe this is incorrect.
Currently, it checks whether the alignment of the new value is correctly aligned with whatever the largest value in this row was.
But it should really check whether the alignment of the new value is correctly aligned with its own alignment rules.

Changing this to: `(row_accum + val) % max_size_in_row == 0` should fix it for all trivial types, as those always have `sizeof() == alignof()`.
Though in reality we would really like to check whether `val` is `% alignof(the type val represents)`.

crates/cxx-qt-lib-extras/src/assertion_utils.h
```diff
@@ -14,27 +15,39 @@
...
+
+      if (row_accum + pad + val <= actual_alignment) {
+        row_accum += (pad + val);
       } else {
-        if (accum + *it + *(it + 1) <= actual_alignment)
-          accum += *it;
-        else
-          accum = actual_alignment;
+        ++rows;
+        row_accum = actual_alignment;
```
LeonMatthesKDAB, line 43: shouldn't this be `row_accum = val`? As otherwise we're immediately filling the next row entirely, no matter how large `val` is?

crates/cxx-qt-lib/src/assertion_utils.h
```diff
@@ -14,27 +15,39 @@
 namespace rust::cxxqtlib1::assertion_utils {
 template<typename iter>
 constexpr static ::std::size_t
-calc_align_size(const iter rbegin,
-                const iter rend,
+calc_align_size(const iter begin,
```
LeonMatthesKDAB, line 18: Side-note: we should be able to avoid duplicating this in cxx-qt-lib-extras by simply doing an `#include <cxx-qt-lib/assertion_utils.h>` or something like this.

cxx-qt-lib-extras depends on cxx-qt-lib and reexports the headers.

LeonMatthesKDAB (changes requested):
Unfortunately this version is still incorrect.

If I have a struct like this:
```cpp
struct X
{
  int i;
  char c;
  char c2;
  double d;
};
static_assert(sizeof(X) == 16);
```
It should be layed out as:

| | | | | | | | |
|-|-|-|-|-|-|-|-|
| i | i | i | i | c |c2 | | |
| d | d |d |d | d |d |d |d |

Which gives a size of `2 * 8 = 16`, as validated by the static_assert.

However, the current implementation returns 24.
This can be tested with this failing static_assert:
```
constexpr static std::array<std::size_t, 4> test1{ sizeof(int),
                                                   sizeof(char),
                                                   sizeof(char),
                                                   sizeof(double) };
static_assert(
  rust::cxxqtlib1::assertion_utils::calc_align_size(std::begin(test1),
                                                    std::end(test1),
                                                    alignof(X)) == 16);
```

This can be fixed with the suggestions in my other comments, but I'm still not 100% certain that this is correct either.


Overall I wonder whether the approach calculating this at compile-time is the right way forward, or whether we should try to revert back to the simpler route we had before, but try adjusting it to work with 32-bit architectures as well.
At this point this would be the 3rd iteration of this function and we're still not certain whether we've found a correct implementation...

What do you think @ahayzen-kdab , @mattkdab ?

ahayzen-kdab:
As discussed this will change to the struct route, where the macro builds a struct with the fields matching the Rust fields, then static assert that that struct is the same size/alignment as the Qt type. Therefore using the compiler to sort out any alignment/padding fun between 32bit vs 64bit vs 128bit etc

mattkdab:
> As discussed this will change to the struct route, where the macro builds a struct with the fields matching the Rust fields, then static assert that that struct is the same size/alignment as the Qt type. Therefore using the compiler to sort out any alignment/padding fun between 32bit vs 64bit vs 128bit etc

refactored to do this

crates/cxx-qt-lib-extras/src/assertion_utils.h
```diff
@@ -3,52 +3,9 @@
...
-}
-} // namespace rust::cxxqtlib1::assertion_utils
-
-#define assert_alignment_and_size(TYPE, EXP_ALIGN, ARR, ARR_SZ)                \
-  static_assert(EXP_ALIGN == alignof(TYPE));                                   \
-  static_assert(                                                               \
-    ::rust::cxxqtlib1::assertion_utils::calc_align_size<                       \
-      ::std::array<::std::size_t, ARR_SZ>::const_reverse_iterator>(            \
-      ::std::rbegin(ARR), ::std::rend(ARR), alignof(TYPE)) == sizeof(TYPE));
+#include "cxx-qt-lib/src/assertion_utils.h"
```
ahayzen-kdab, line 11: seems this fails? Wonder if we aren't bundling the assertion_utils.h header into the cxx-qt-lib export?

ahayzen-kdab: @mattkdab i think you can put it back to `#include <cxx-qt-lib/assertion_utils.h>` then what you need to do is

- in `crates/cxx-qt-lib/` move the assertion_utils.h from the src/ folder to the include/ folder as we export the include folder here https://github.com/KDAB/cxx-qt/blob/86d00e56a431192410a0ba19936c745f8040962a/crates/cxx-qt-lib/build.rs#L295
- add `assertion_utils.h` to the write_headers like we do for common here https://github.com/KDAB/cxx-qt/blob/86d00e56a431192410a0ba19936c745f8040962a/crates/cxx-qt-lib/build.rs#L50
- remove `assertion_utils` from the manual rerun-if at the end https://github.com/KDAB/cxx-qt/blob/86d00e56a431192410a0ba19936c745f8040962a/crates/cxx-qt-lib/build.rs#L337
- update any includes of it in cxx-qt-lib .cpp files to use the new `cxx-qt-lib/assertion_utils.h` path rather than the relative `../assertion_utils.h`

crates/cxx-qt-lib-extras/src/assertion_utils.h
```diff
@@ -3,52 +3,9 @@
 // clang-format on
```
ahayzen-kdab, line 3: can this file just be removed? as now `#include <cxx-qt-lib/assertion_utils.h>` is used?

crates/cxx-qt-lib/build.rs
```diff
@@ -49,6 +49,11 @@ fn write_headers() {
     println!("cargo::rerun-if-changed=include/common.h");
     std::fs::copy("include/common.h", header_dir().join("common.h"))
         .expect("Failed to copy header file!");
+    std::fs::copy(
```
ahayzen-kdab, line 52: you need this line here too, like the above common.h has. Otherwise the cargo won't know to rebuild if the assertion_utils changes.
```
println!("cargo::rerun-if-changed=include/assertion_utils.h");
```

crates/cxx-qt-lib/include/assertion_utils.h
```diff
@@ -0,0 +1,16 @@
...
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// clang-format on
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+// SPDX-FileContributor: Leon Matthes <leon.matthes@kdab.com>
+// SPDX-FileContributor: Matt Aber <matt.aber@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+#pragma once
+
+#define assert_alignment_and_size(TYPE, EXP_ALIGN, MEMBERS)                    \
```
ahayzen-kdab, line 11: guess there is also the question, do we need to pass in the expected alignment? What are we trying to compare here? Can we compare the struct alignment to the Qt type? Or do we always just need to check that it is `alignof(::std::size_t)` ?

mattkdab: true we can just do this now that we define a type
```cpp
static_assert(alignof(s##TYPE) == alignof(TYPE));
static_assert(sizeof(s##TYPE) == sizeof(TYPE))
```

LeonMatthesKDAB (approved):
Would be great if we didn't have to give the members names in the macro input, but I also don't know how to do that, so this is already a really great improvement.

LGTM :partying_face:

PR #1049, opened 2024-08-26
fix: Allow transitive subclassing

### LINKED ISSUE #1034: Deriving from a cxx-qt QObject child
### Discussed in https://github.com/KDAB/cxx-qt/discussions/1031

<div type='discussions-op-text'>

<sup>Originally posted by **SanderVocke** August 13, 2024</sup>
Hi there,

Another day, another question from my side.

I am trying to have inheritance between my cxx-qt objects. In my particular case, I derive a class from QQuickItem, then I want another class to derive from this class. The code is at the bottom of my message.

The problem I am running into is that it seems all cxx-qt generated items derive from `rust::cxxqtlib1::CxxQtLocking`, and trying to derive a new cxx-qt class from an existing one causes multiple inheritance errors:

```
C:\dev\shoopdaloop\build\cp312-cp312-win_amd64\cargo\build\x86_64-pc-windows-msvc\debug\build\shoop_rs_frontend-ea53a39f3219adf3\out\cxx-qt-gen/qobj_autoconnect.cxxqt.h(11): warning C4584: 'AutoConnect': base-class 'rust::cxxqtlib1::CxxQtLocking' is already a base-class of 'FindParentBackendWrapper'
```

I don't know if there is a way around this, and don't see any similar cases in the examples. I think not being able to do inheritance like this will put some serious constraints on code reusability.

I guess one way could be to make the base class not be a Qt-derived object at all, but implement the common functionality via traits or the like. But that would not allow for e.g. inheriting properties and invokeables.

Thanks in any case!

The relevant code:

Base class:

```rust
use crate::logging::macros::*;
shoop_log_unit!("Frontend.FindParentBackendWrapper");

use std::pin::Pin;

#[cxx_qt::bridge]
pub mod qobj_find_parent_backend_wrapper {
    unsafe extern "C++" {
        include!(<QtQuick/QQuickItem>);
        type QQuickItem;
    }

    unsafe extern "RustQt" {
        #[qobject]
        #[base = "QQuickItem"]
        #[qproperty(*mut QQuickItem, backend_wrapper)]
        #[qproperty(bool, backend_wrapper_initialized)]
        type FindParentBackendWrapper = super::FindParentBackendWrapperRust;
    }

    unsafe extern "C++" {
        include!("cxx-qt-shoop/qquickitem_parent.h");
        #[rust_name = "maybe_parent_qquickitem_qquickitem_findparentbackendwrapper"]
        fn maybe_parent_qquickitem(item : &QQuickItem) -> *mut QQuickItem;
        #[rust_name = "maybe_parent_qquickitem_findparentbackendwrapper"]
        fn maybe_parent_qquickitem(item : &FindParentBackendWrapper) -> *mut QQuickItem;

        include!("cxx-qt-shoop/qobject_classname.h");
        #[rust_name = "qobject_class_name_qquickitem_findparentbackendwrapper"]
        fn qobject_class_name(obj : &QQuickItem) -> &str;

        include!("cxx-qt-shoop/cast_ptr.h");
        #[rust_name = "cast_ptr_findparentbackendwrapper_qquickitem"]
        unsafe fn cast_ptr(item : *mut FindParentBackendWrapper) -> *mut QQuickItem;
    }

    unsafe extern "RustQt" {
        #[qinvokable]
        unsafe fn rescan_parents(self: Pin<&mut FindParentBackendWrapper>);
    }

    unsafe extern "C++" {
        include!("cxx-qt-shoop/make_unique.h");

        #[rust_name = "make_unique_findparentbackendwrapper"]
        fn make_unique() -> UniquePtr<FindParentBackendWrapper>;
    }

    impl cxx_qt::Constructor<(*mut QQuickItem,), NewArguments=(*mut QQuickItem,)> for FindParentBackendWrapper {}
    impl cxx_qt::Constructor<(), NewArguments=()> for FindParentBackendWrapper {}
}

use qobj_find_parent_backend_wrapper::*;

unsafe fn find_parent_backend(item : *mut QQuickItem) -> Result<*mut QQuickItem, String> {
    if item.is_null() {
        return Err(String::from("Could not find back-end item."));
    }
    let is_backend = qobject_class_name_qquickitem_findparentbackendwrapper(item.as_ref().unwrap()) == "Backend";
    if is_backend {
        return Ok(item);
    }
    let parent = maybe_parent_qquickitem_qquickitem_findparentbackendwrapper(item.as_ref().unwrap());
    return find_parent_backend (parent);
}
pub struct FindParentBackendWrapperRust {
    backend_wrapper : *mut QQuickItem,
    backend_wrapper_initialized : bool,
}

impl Default for FindParentBackendWrapperRust {
    fn default() -> FindParentBackendWrapperRust {
        FindParentBackendWrapperRust {
            backend_wrapper : std::ptr::null_mut(),
            backend_wrapper_initialized : false,
        }
    }
}

impl FindParentBackendWrapper {
    pub unsafe fn rescan_parents(mut self: Pin<&mut FindParentBackendWrapper>) {
        let self_ptr = self.as_mut().get_unchecked_mut() as *mut FindParentBackendWrapper;
        let self_qquick_ptr = cast_ptr_findparentbackendwrapper_qquickitem(self_ptr);
        let backend = find_parent_backend(self_qquick_ptr).unwrap();
        self.set_backend_wrapper(backend);
    }
}

impl cxx_qt::Constructor<(*mut QQuickItem,)> for FindParentBackendWrapper {
    type BaseArguments = (*mut QQuickItem,); // Will be passed to the base class constructor
    type InitializeArguments = (); // Will be passed to the "initialize" function
    type NewArguments = (*mut QQuickItem,); // Will be passed to the "new" function

    fn route_arguments(args: (*mut QQuickItem,)) -> (
        Self::NewArguments,
        Self::BaseArguments,
        Self::InitializeArguments
    ) {
        (args, args, ())
    }

    fn new(_parent : (*mut QQuickItem,)) -> FindParentBackendWrapperRust {
        FindParentBackendWrapperRust::default()
    }
}

impl cxx_qt::Constructor<()> for FindParentBackendWrapper {
    type BaseArguments = (); // Will be passed to the base class constructor
    type InitializeArguments = (); // Will be passed to the "initialize" function
    type NewArguments = (); // Will be passed to the "new" function

    fn route_arguments(_args: ()) -> (
        Self::NewArguments,
        Self::BaseArguments,
        Self::InitializeArguments
    ) {
        ((), (), ())
    }

    fn new(_args: ()) -> FindParentBackendWrapperRust {
        FindParentBackendWrapperRust::default()
    }
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test_class_name() {
        let obj = make_unique_findparentbackendwrapper();
        unsafe {
            let ptr = obj.into_raw();
            let quick = cast_ptr_findparentbackendwrapper_qquickitem(ptr);
            let classname = qobject_class_name_qquickitem_findparentbackendwrapper(quick.as_ref().unwrap());
            assert_eq!(classname, "FindParentBackendWrapper");
        }
    }
}
```

Child class:

```rust
use crate::logging::macros::*;
shoop_log_unit!("Frontend.AutoConnect");

// use crate::cxx_qt_shoop::qobj_find_parent_backend_wrapper;

#[cxx_qt::bridge]
pub mod qobj_autoconnect {
    unsafe extern "C++" {
        include!(<QtQuick/QQuickItem>);
        type QQuickItem;
    }

    unsafe extern "C++" {
        include!("cxx-qt-gen/qobj_find_parent_backend_wrapper.cxxqt.h");
    }

    unsafe extern "RustQt" {
        #[qobject]
        #[base = "FindParentBackendWrapper"]
        type AutoConnect = super::AutoConnectRust;
    }

    unsafe extern "C++" {
        include!("cxx-qt-shoop/make_unique.h");

        #[rust_name = "make_unique_autoconnect"]
        fn make_unique() -> UniquePtr<AutoConnect>;
    }

    unsafe extern "C++" {
        include!("cxx-qt-shoop/qobject_classname.h");
        #[rust_name = "qobject_class_name_autoconnect"]
        fn qobject_class_name(obj : &AutoConnect) -> &str;
    }

    impl cxx_qt::Constructor<(*mut QQuickItem,), NewArguments=(*mut QQuickItem,)> for AutoConnect {}
    impl cxx_qt::Constructor<(), NewArguments=()> for AutoConnect {}
}

pub struct AutoConnectRust {
}

impl Default for AutoConnectRust {
    fn default() -> AutoConnectRust {
        AutoConnectRust {}
    }
}

impl cxx_qt::Constructor<(*mut QQuickItem,)> for AutoConnect {
    type BaseArguments = (*mut QQuickItem,); // Will be passed to the base class constructor
    type InitializeArguments = (); // Will be passed to the "initialize" function
    type NewArguments = (*mut QQuickItem,); // Will be passed to the "new" function

    fn route_arguments(args: (*mut QQuickItem,)) -> (
        Self::NewArguments,
        Self::BaseArguments,
        Self::InitializeArguments
    ) {
        (args, args, ())
    }

    fn new(_parent : (*mut QQuickItem,)) -> AutoConnectRust {
        AutoConnectRust::default()
    }
}

impl cxx_qt::Constructor<()> for AutoConnect {
    type BaseArguments = (); // Will be passed to the base class constructor
    type InitializeArguments = (); // Will be passed to the "initialize" function
    type NewArguments = (); // Will be passed to the "new" function

    fn route_arguments(_args: ()) -> (
        Self::NewArguments,
        Self::BaseArguments,
        Self::InitializeArguments
    ) {
        ((), (), ())
    }

    fn new(_args: ()) -> AutoConnectRust {
        AutoConnectRust::default()
    }
}

#[cfg(test)]
mod tests {
    use super::qobj_autoconnect::*;

    #[test]
    fn test_class_name() {
        let obj = make_unique_autoconnect();
        let classname = qobject_class_name_autoconnect (obj.as_ref().unwrap());
        assert_eq!(classname, "AutoConnect");
    }
}
```</div>

### PR DESCRIPTION
**Pull request by LeonMatthesKDAB**
When subclassing from another CXX-Qt subclass, this created problems with duplicate subclassing from our own trait-like types (e.g. CxxQtLocking).

Some functions also need disambiguation, like the unsafeRust function.

We can fix this with the help of template-metaprogramming, by conditionally disabling duplicate inheritance and disambiguating the function calls.

Fixes #1034

### CODE REVIEW

**Comment by LeonMatthesKDAB** [comment]
TODO: Add an example to qml_features

File: crates/cxx-qt-gen/src/generator/cpp/externcxxqt.rs [hunk 1]
```diff
@@ -32,10 +32,6 @@ pub fn generate(
             let qobject_name = type_names.lookup(&signal.qobject_ident)?;
```
ahayzen-kdab, line 32: can we have a changelog entry as the API has changed a bit here?

Also does the book need updating it probably mentions the locking trait?

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
```diff
@@ -186,15 +177,14 @@ pub mod tests {
         )
         .unwrap();
 
-        // should be a pair
-        let result = require_header(&generated.methods[0]);
-        assert!(result.is_err());
-
-        let result = require_pair(&generated.private_methods[0]);
+        // should be a header
```
ahayzen-kdab, line 180: this comment is now incorrect?

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
```diff
@@ -186,15 +177,14 @@ pub mod tests {
...
+        let result = require_pair(&generated.methods[0]);
         assert!(result.is_err());
 
         let result = require_source(&generated.methods[0]);
         assert!(result.is_err());
```
ahayzen-kdab, line 185: the above checks for a pair so these lines can be removed?

crates/cxx-qt-gen/src/generator/rust/cxxqttype.rs
```diff
@@ -8,57 +8,62 @@ use crate::{
...
 use quote::quote;
-use syn::Result;
+use syn::{Ident, Result};
 
 use super::fragment::RustFragmentPair;
 
 pub fn generate(
     qobject_ident: &QObjectNames,
     type_names: &TypeNames,
+    module_ident: &Ident,
```
ahayzen-kdab, line 18: once #1060 lands or if this lands first, either branch will need updating to the ` qobject_names.name.module_ident()` route, cc @BenFordTytherington

crates/cxx-qt-gen/src/generator/rust/cxxqttype.rs
```diff
@@ -8,57 +8,62 @@ use crate::{
...
 ) -> Result<GeneratedRustFragment> {
     let mut blocks = GeneratedRustFragment::default();
 
     let cpp_struct_ident = &qobject_ident.name.rust_unqualified();
     let rust_struct_ident = &qobject_ident.rust_struct.rust_unqualified();
+    let rust_fn = qobject_ident.cxx_qt_ffi_method("rust");
```
ahayzen-kdab, line 24: If this returned a `Name` then couldn't the rust_qualified be used instead of doing `#module_ident::#rust_fn` further down?

crates/cxx-qt-gen/src/generator/rust/cxxqttype.rs
```diff
@@ -8,57 +8,62 @@ use crate::{
...
+    let rust_mut_fn = qobject_ident.cxx_qt_ffi_method("rust_mut");
```
ahayzen-kdab, line 25: should we have these as const's somewhere rather than magic values?

crates/cxx-qt/include/threading.h
```diff
@@ -32,11 +31,20 @@ class CxxQtThreading : public CxxQtLocking
 
   CxxQtThread<T> qtThread() const
   {
-    return CxxQtThread<T>(m_cxxQtThreadObj, m_rustObjMutex);
+    // Note that this assumes that T inherits from CxxQtLocking, which we assert
+    // during parsing.
```
ahayzen-kdab, line 35: do we add a `Q_ASSERT` to confirm this?

crates/cxx-qt/include/threading.h
```diff
@@ -32,11 +31,20 @@ class CxxQtThreading : public CxxQtLocking
...
 
 private:
   ::std::shared_ptr<CxxQtGuardedPointer<T>> m_cxxQtThreadObj;
 };
 
+template<typename T>
+CxxQtThread<T>
+qtThread(const T& qobject)
+{
+  return static_cast<const CxxQtThreading<T>&>(qobject).qtThread();
```
ahayzen-kdab, line 47: do we add a `Q_ASSERT` that the cast can work?

examples/qml_features/qml/pages/CustomBaseClassPage.qml
```diff
@@ -9,83 +9,127 @@ import QtQuick.Layouts 1.12
 import com.kdab.cxx_qt.demo 1.0
 
 Page {
+    property var activeModel: tabs.currentIndex == 0 ? customBaseClass : transitiveInheritance
```
ahayzen-kdab, line 12: triple equals, use a type, and readonly so

````
readonly property bool activeModel: tabs.currentIndex === 0 ? customBaseClass : transitiveInheritance
```

examples/qml_features/qml/pages/CustomBaseClassPage.qml
```diff
@@ -9,83 +9,127 @@ import QtQuick.Layouts 1.12
...
+
     header: ToolBar {
         RowLayout {
             anchors.fill: parent
 
             ToolButton {
                 text: qsTr("Add Row")
 
-                onClicked: customBaseClass.add()
+                onClicked: activeModel.add()
```
ahayzen-kdab, line 22: prefer fully qualified over using context, so add `id: root` then `root.activeModel.add()`

We probably don't do this in all places yet but it's the modern way and what things like qmllint will recommend

examples/qml_features/qml/pages/CustomBaseClassPage.qml
```diff
@@ -9,83 +9,127 @@ import QtQuick.Layouts 1.12
 import com.kdab.cxx_qt.demo 1.0
 
 Page {
+    property var activeModel: tabs.currentIndex == 0 ? customBaseClass : transitiveInheritance
+    property var activeListView: tabs.currentIndex == 0 ? customBaseList : transitiveList
```
ahayzen-kdab, line 13: same as above ^^

examples/qml_features/qml/pages/CustomBaseClassPage.qml
```diff
@@ -9,83 +9,127 @@ import QtQuick.Layouts 1.12
...
+
+                ListView {
+                    id: customBaseList
+                    currentIndex: -1
+                    model: CustomBaseClass {
+                        id: customBaseClass
+                    }
+                    delegate: ItemDelegate {
+                        highlighted: ListView.isCurrentItem
+                        text: model.id + ": " + model.value
```
ahayzen-kdab, line 95: can also use the required property route, but this is probably fine for now

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
LeonMatthesKDAB: No, it's correct. We're asserting that `require_pair` returns an error, because it's actually a header.

I think this just exists to cover all branches of the `require_pair` function...

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
LeonMatthesKDAB: Like the call above I think this just exists to get 100% coverage on require_source for the error case.

crates/cxx-qt-gen/src/generator/rust/cxxqttype.rs
LeonMatthesKDAB: Refactored to Name, which became a bit larger of a refactor, so did that in a separate commit.

crates/cxx-qt/include/threading.h
LeonMatthesKDAB: Ah, Locking is gone, so that comment needs to go as well.

crates/cxx-qt/include/threading.h
LeonMatthesKDAB: It's a static cast, so it should fail to compile if the qobject doesn't inherit from CxxQtThreading.

examples/qml_features/qml/pages/CustomBaseClassPage.qml
LeonMatthesKDAB: The type of this is not bool, it's either `TransitiveInheritance`, or `CustomBaseClass`, so I don't think this can have a type apart from var.

Added the `readonly` though.

examples/qml_features/qml/pages/CustomBaseClassPage.qml
LeonMatthesKDAB: See my other comment :)

examples/qml_features/qml/pages/CustomBaseClassPage.qml
LeonMatthesKDAB: Done

examples/qml_features/qml/pages/CustomBaseClassPage.qml
ahayzen-kdab: Ah yes the type would have to be one shared by them both. Var is fine then

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
ahayzen-kdab: Ah it's hitting the errors states i see.

crates/cxx-qt-gen/src/generator/rust/cxxqttype.rs
LeonMatthesKDAB: Hm, we don't do that anywhere else, so don't know why we should start here :thinking: 
After the refactor, these are the names of the functions in C++, so the compiler will check that they match.

ahayzen-kdab: Right can be done separately, just wondered if we have these magic values used in multiple parts of the generation if they should be a const static somewhere

ahayzen-kdab (approved):
LGTM :-)

PR #1053, opened 2024-08-28
Implement QListElement for QLine(F)

SanderVocke:
QListElement was not implemented for these types yet.

Let me know if I need to do anything in particular to be able to contribute a PR, license/copyright-wise. - right now I just copied the header from other similar files and put my name there.

This change builds for me and I am able to use the new list types.

Note that I also added two convenience methods to QPainter (which exist in Qt too), for drawing these lists of lines.


crates/cxx-qt-lib/src/gui/qpainter.rs
```diff
@@ -239,6 +244,18 @@ mod ffi {
...
+        /// Draws a line defined by line.
+        #[rust_name = "draw_linef"]
+        fn drawLine(self: Pin<&mut QPainter>, line: &QLineF);
+
+        /// Draws the set of lines defined by the list lines using the current pen and brush.
+        #[rust_name = "draw_lines"]
+        fn drawLines(self: Pin<&mut QPainter>, lines: &QList_QLine);
+
+        /// Draws the set of lines defined by the list lines using the current pen and brush.
+        #[rust_name = "draw_linefs"]
```
SanderVocke, line 256: I put `draw_linefs` as the rust name because I believe aliasing the names is not supported, right? In Qt they are both `drawLines()`.

LeonMatthesKDAB: Yes, they need to be different names, so `draw_linefs` is good :+1:

LeonMatthesKDAB (commented):
Hi @SanderVocke and thank you for the contribution!

The code it and of itself is good.
We do actually have a `generate.sh` script inside `cxx-qt-lib/src/core/qlist/` that we used to generate all of these files.
So instead of adding the files for QLine(F) manually, you should be able to simply add a line to that script and execute it.

Can you please change to using that?
As it would allow us to update the code far easier if we ever need to e.g. add a new function to the trait.

Regarding the licensing:
We don't have a CLA, so you shouldn't have to do anything license-wise.
As you're contributing to this repo you implicitly agree that your code will made open-source and dual-licensed as apache 2.0 and MIT.
If you edit the generate.sh script to generate these files, your name won't be in the SPDX-FileContributor list of the QLine/F files. But you could simply add yourself as a SPDX-FileContributor on the `generate.sh` script itself.
Having multiple contributors is not a problem.

SanderVocke:
OK, switched to the auto-generated version and included the change to generate.sh.

LeonMatthesKDAB (dismissed):

SanderVocke:
The formatting should be fixed now.

LeonMatthesKDAB (approved):

SanderVocke:
Ah, before you approve the CI run - I still need to run this locally against Qt5.

SanderVocke:
I tried this locally against Qt5 and it builds fine, but some tests fail. In most cases it seems to be tests that make a diff between two result files, and the differences it finds are purely whitespace.

Would you mind running the CI and having a look for yourself?

ahayzen-kdab:
Locally it could be a difference in clang-format versions or something, but CI seems happy :-)

LeonMatthesKDAB (approved):
@SanderVocke sorry I missed your latest update.
Looks great now and can be merged.

Thank you for the contribution!

ahayzen-kdab:
(Note CI is failing due to macOS changing clang-format versions, see discussion in #1083 )

SanderVocke:
Cheers, so if I understand correctly it is a matter of waiting for the clang-format to be fixed, then updating this branch?

ahayzen-kdab:
Yes, this is separate from your change, once it is fixed we'll rebase yours and it'll land :-)

PR #1054, opened 2024-08-29
Refactor ParsedMethod, ParsedSignal and ParsedInheritedMethod

BenFordTytherington:
- Refactor method and inherited method to have docs passthrough
- Group common functionality between these 3 similar structs

### DIFF
--- a/crates/cxx-qt-gen/src/generator/rust/method.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/method.rs
@@ -42,8 +42,6 @@ pub fn generate_rust_methods(
             std::mem::swap(&mut unsafe_call, &mut None);
         }
 
-        let doc_comments = &invokable.docs;
-
         let fragment = RustFragmentPair {
             cxx_bridge: vec![quote_spanned! {
                 invokable.method.span() =>
@@ -52,8 +50,8 @@ pub fn generate_rust_methods(
                     // Note that we are exposing a Rust method on the C++ type to C++
                     //
                     // CXX ends up generating the source, then we generate the matching header.
-                    #(#doc_comments)*
                     #[cxx_name = #wrapper_ident_cpp]
+                    #[doc(hidden)]
                     // TODO: Add #[namespace] of the QObject
                     #unsafe_call fn #invokable_ident_rust(#parameter_signatures) #return_type;
                 }
@@ -108,7 +106,6 @@ mod tests {
                     None,
                 )
                 .unwrap(),
-                docs: vec![],
             },
             ParsedMethod {
                 method: method2.clone(),
@@ -128,7 +125,6 @@ mod tests {
                     None,
                 )
                 .unwrap(),
-                docs: vec![],
             },
             ParsedMethod {
                 method: method3.clone(),
@@ -148,7 +144,6 @@ mod tests {
                     None,
                 )
                 .unwrap(),
-                docs: vec![],
             },
             ParsedMethod {
                 method: method4.clone(),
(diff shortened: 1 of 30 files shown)


BenFordTytherington:
Is it useful to combine all these invokable types in `ParsedCxxQtData` using trait objects and enum markers? or is it better to keep them separate for preserving order.
Something like `invokables: BTreeMap<InvokableKind, Box<Vec<dyn Invokable>>` might be too complex but could also combine lots of common logic?

crates/cxx-qt-gen/src/parser/signals.rs
```diff
@@ -34,10 +33,16 @@ pub struct ParsedSignal {
...
+    /// All the doc attributes (each line) of the signal
     pub docs: Vec<Attribute>,
 }
 
+impl Invokable for &ParsedSignal {
+    fn name(&self) -> &Name {
+        &self.name
+    }
+}
+
```
LeonMatthesKDAB, line 45: A whole trait seems a bit overkill. Maybe prefer using a lambda in the structuring phase instead that just accesses the right member.

crates/cxx-qt-gen/src/generator/rust/method.rs
```diff
@@ -179,7 +178,6 @@ mod tests {
             &generated.cxx_mod_contents[0],
             quote! {
                 extern "Rust" {
-                    #[doc(hidden)]
```
LeonMatthesKDAB, line 182: These tests, as well as the expected output test probably still need updating.
I assume that's also why CI fails.

crates/cxx-qt-gen/src/parser/mod.rs
```diff
@@ -25,9 +29,67 @@ use syn::{
...
+/// Struct with common fields between Invokable types.
+/// These types are ParsedSignal, ParsedMethod and ParsedInheritedMethod
+pub struct MethodFields {
+    qobject_ident: Ident,
+    mutable: bool,
+    parameters: Vec<ParsedFunctionParameter>,
+    safe: bool,
+    name: Name,
+    docs: Vec<Attribute>, // TODO: Remove this
+}
```
LeonMatthesKDAB, line 59: Quite nice to see this being deduplicated.
I would prefer to move this type and associated functions into its own file though. You could probably also just put it into method.rs. But it doesn't really belong into the parser module.

crates/cxx-qt-gen/src/parser/mod.rs
```diff
@@ -25,9 +29,67 @@ use syn::{
...
+
+/// Function for creating an [MethodFields] from a method and docs.
+/// These fields are shared by ParsedSignal, ParsedMethod and ParsedInheritedMethod so grouped into common logic.
+pub fn extract_common_fields(method: &ForeignItemFn, docs: Vec<Attribute>) -> Result<MethodFields> {
```
LeonMatthesKDAB, line 63: Why isn't this function `MethodFields::parse`?
That's the pattern we use for all other parsing.

LeonMatthesKDAB (changes requested):
The tests seem to not be up-to-date, but this is going in the right direction :)

crates/cxx-qt-gen/src/parser/mod.rs

crates/cxx-qt-gen/src/parser/mod.rs
BenFordTytherington: Moved into method.rs

crates/cxx-qt-gen/src/generator/rust/method.rs
BenFordTytherington: Done

crates/cxx-qt-gen/src/parser/signals.rs

LeonMatthesKDAB (approved):

PR #1055, opened 2024-08-29
Add namespace of qobject in place of TODO comments

BenFordTytherington:
Some TODO comments were left in the repo to add in #[namespace] of the QObject during generation, so that has been added

### DIFF
--- a/crates/cxx-qt-gen/test_outputs/invokables.rs
+++ b/crates/cxx-qt-gen/test_outputs/invokables.rs
@@ -43,56 +43,67 @@ mod ffi {
     extern "Rust" {
         #[doc(hidden)]
         #[cxx_name = "cppMethodWrapper"]
+        #[namespace = "cxx_qt::my_object"]
         fn cpp_method(self: &MyObject);
     }
     extern "Rust" {
         #[doc(hidden)]
         #[cxx_name = "invokableWrapper"]
+        #[namespace = "cxx_qt::my_object"]
         fn invokable(self: &MyObject);
     }
     extern "Rust" {
         #[doc(hidden)]
         #[cxx_name = "invokableMutableWrapper"]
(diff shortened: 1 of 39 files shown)


crates/cxx-qt-gen/test_outputs/invokables.rs
```diff
@@ -43,56 +43,67 @@ mod ffi {
     extern "Rust" {
         #[doc(hidden)]
         #[cxx_name = "cppMethodWrapper"]
+        #[namespace = "cxx_qt::my_object"]
```
LeonMatthesKDAB, line 46: Hm, does this actually have any effect?
The QObject itself is already associated with a namespace. As these are member functions, they can't have a namespace of their own, so does this change anything?

BenFordTytherington: Not actually sure, I just replaced all the TODOs, assuming whoever left them knew whether it was necessary, might need to check over some of them

LeonMatthesKDAB: @ahayzen-kdab Do you have any idea of why we had that in there?

ahayzen-kdab: Was this historical when we used to have free methods and before they were using `self` due to other reasons?

(eg we used to have a free method like `my_invokable(this, ...args)` instead of `T::my_invokable(..args)` )

As yes I think this isn't required anymore and instead the comments could be removed?

BenFordTytherington: Done, and replaced Todos with comments indicating namespace shouldn't be added

ahayzen-kdab (approved):
LGTM :-)

PR #1057, opened 2024-09-02
Refactor MethodFields

BenFordTytherington:
Refactor ForeignItemFn into MethodFields and simplify some other aspects surrounding methods and signals

### DIFF
--- a/crates/cxx-qt-gen/src/parser/inherit.rs
+++ b/crates/cxx-qt-gen/src/parser/inherit.rs
@@ -6,8 +6,8 @@
 use crate::parser::method::MethodFields;
 use crate::parser::{check_safety, separate_docs};
 use crate::syntax::{attribute::attribute_take_path, safety::Safety};
+use core::ops::Deref;
 use quote::format_ident;
-use std::ops::Deref;
 use syn::{Attribute, ForeignItemFn, Ident, Result};
 
 /// Describes a method found in an extern "RustQt" with #[inherit]
--- a/crates/cxx-qt-gen/src/parser/inherit.rs
+++ b/crates/cxx-qt-gen/src/parser/inherit.rs
@@ -5,28 +5,17 @@
 
 use crate::parser::method::MethodFields;
 use crate::parser::{check_safety, separate_docs};
-use crate::{
-    naming::Name,
-    parser::parameter::ParsedFunctionParameter,
-    syntax::{attribute::attribute_take_path, safety::Safety},
-};
+use crate::syntax::{attribute::attribute_take_path, safety::Safety};
 use quote::format_ident;
+use std::ops::Deref;
 use syn::{Attribute, ForeignItemFn, Ident, Result};
 
 /// Describes a method found in an extern "RustQt" with #[inherit]
 pub struct ParsedInheritedMethod {
     /// The original [syn::ForeignItemFn] of the inherited method declaration
     pub method: ForeignItemFn,
-    /// The type of the self argument
-    pub qobject_ident: Ident,
-    /// whether the inherited method is marked as mutable
-    pub mutable: bool,
-    /// Whether the method is safe to call.
-    pub safe: bool,
-    /// the parameters of the method, without the `self` argument
-    pub parameters: Vec<ParsedFunctionParameter>,
-    /// the name of the function in Rust, as well as C++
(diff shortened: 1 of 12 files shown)


crates/cxx-qt-gen/src/parser/method.rs
```diff
@@ -34,22 +35,12 @@ impl ParsedQInvokableSpecifiers {
...
-    pub parameters: Vec<ParsedFunctionParameter>,
+    /// The common fields which are available on all callable types
+    pub method_fields: MethodFields,
     /// Any specifiers that declared on the invokable
     pub specifiers: HashSet<ParsedQInvokableSpecifiers>,
     /// Whether the method is qinvokable
     pub is_qinvokable: bool,
-    /// The rust and cxx name of the function
-    pub name: Name,
 }
```
ahayzen-kdab, line 44: shouldn't there be `doc` here, as a separate item, like in ParsedSignal / ParsedInheritedMethod ?

crates/cxx-qt-gen/src/parser/inherit.rs
```diff
@@ -5,28 +5,15 @@
...
 use crate::parser::method::MethodFields;
 use crate::parser::{check_safety, separate_docs};
-use crate::{
-    naming::Name,
-    parser::parameter::ParsedFunctionParameter,
-    syntax::{attribute::attribute_take_path, safety::Safety},
-};
+use crate::syntax::{attribute::attribute_take_path, safety::Safety};
 use quote::format_ident;
+use std::ops::Deref;
```
ahayzen-kdab, line 10: prefer `core::ops::Deref` (eg prefer crates like alloc/core instead of std where possible). Then later then makes things easier to have no_std if ever possible in the future

crates/cxx-qt-gen/src/parser/method.rs
```diff
@@ -9,9 +9,10 @@ use crate::{
     syntax::{attribute::attribute_take_path, safety::Safety},
 };
 use std::collections::HashSet;
-use syn::{Attribute, Error, ForeignItemFn, Ident, Result};
+use std::ops::Deref;
```
ahayzen-kdab, line 12: same

crates/cxx-qt-gen/src/parser/signals.rs
```diff
@@ -3,31 +3,18 @@
...
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
-use crate::{
-    naming::Name,
-    parser::parameter::ParsedFunctionParameter,
-    syntax::{attribute::attribute_take_path, path::path_compare_str, safety::Safety},
-};
-use syn::{spanned::Spanned, Attribute, Error, ForeignItemFn, Ident, Result, Visibility};
+use crate::syntax::{attribute::attribute_take_path, path::path_compare_str, safety::Safety};
+use std::ops::Deref;
```
ahayzen-kdab, line 7: same

crates/cxx-qt-gen/src/parser/method.rs
BenFordTytherington: I think I discussed this with @LeonMatthesKDAB , and since you have to implement your methods outside the bridge, the docs used are the ones from the implementation (which are also required by clippy), so any docs given here are ignored. Might need to confirm though.

ahayzen-kdab: aha, makes sense, maybe add a comment here stating this as it's an oddity otherwise when you compare the structs :-)

crates/cxx-qt-gen/src/parser/inherit.rs
BenFordTytherington: Done

crates/cxx-qt-gen/src/parser/method.rs
BenFordTytherington: Done

crates/cxx-qt-gen/src/parser/signals.rs

crates/cxx-qt-gen/src/parser/method.rs
BenFordTytherington: yep will do

ahayzen-kdab (approved):
LGTM :-)

PR #1059, opened 2024-09-04
Refactor assert_parse_errors macro to be more generic

BenFordTytherington:
assert_parse_errors is now more generic thanks to some help from Leon, and I've used this in as many places as possible.

### DIFF
--- a/crates/cxx-qt-gen/src/generator/structuring/mod.rs
+++ b/crates/cxx-qt-gen/src/generator/structuring/mod.rs
@@ -119,6 +119,7 @@ impl<'a> Structures<'a> {
 mod tests {
     use super::*;
     // use crate::tests::assert_parse_errors;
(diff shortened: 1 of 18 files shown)


crates/cxx-qt-gen/src/generator/structuring/mod.rs
```diff
@@ -118,6 +118,8 @@ impl<'a> Structures<'a> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    // use crate::tests::assert_parse_errors;
```
ahayzen-kdab, line 121: remove comment

BenFordTytherington: done

ahayzen-kdab (approved):
LGTM :-)

PR #1061, opened 2024-09-04
Fix namespace bug

BenFordTytherington:
Putting Namespace attrs on a type in RustQt or the block itself doesn't compile, with a big string of cpp errors.

I've tried putting namespace back into the rust generate fn, along with a host of other options, such as putting the namespace on the method block as well.

However, putting namespace on the bridge works fine, and everything works properly.

### DIFF
--- a/crates/cxx-qt-gen/src/generator/naming/qobject.rs
+++ b/crates/cxx-qt-gen/src/generator/naming/qobject.rs
@@ -7,7 +7,8 @@ use crate::{
     parser::qobject::ParsedQObject,
 };
 use convert_case::{Case, Casing};
-use quote::format_ident;
+use proc_macro2::TokenStream;
+use quote::{format_ident, quote};
 use syn::{Ident, Result};
 
 /// Names for parts of a Q_OBJECT
@@ -61,6 +62,15 @@ impl QObjectNames {
             "cxx_qt_ffi_{ident}_{suffix}",
             ident = self.name.cxx_unqualified().to_case(Case::Snake)
         )
+    }
+
+    pub fn namespace_tokens(&self) -> TokenStream {
+        let namespace = self.name.namespace();
+        if namespace.is_none() {
+            quote! {}
+        } else {
+            quote! { #[namespace = #namespace ] }
+        }
     }
 }
 
--- a/crates/cxx-qt-gen/src/generator/naming/qobject.rs
+++ b/crates/cxx-qt-gen/src/generator/naming/qobject.rs
@@ -64,12 +64,13 @@ impl QObjectNames {
         )
     }
 
+    /// Returns the tokens of the namespace attribute to be added to a rust line, or no tokens if this instance has no namespace
+    /// attribute looks like `#[namespace = "namespace::here"]`
     pub fn namespace_tokens(&self) -> TokenStream {
-        let namespace = self.name.namespace();
-        if namespace.is_none() {
-            quote! {}
-        } else {
+        if let Some(namespace) = self.name.namespace() {
             quote! { #[namespace = #namespace ] }
+        } else {
+            quote! {}
         }
     }
 }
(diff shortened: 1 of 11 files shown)


crates/cxx-qt-gen/src/generator/rust/method.rs
```diff
@@ -52,6 +59,7 @@ pub fn generate_rust_methods(
                     // CXX ends up generating the source, then we generate the matching header.
                     #[cxx_name = #wrapper_ident_cpp]
                     // Namespace is not needed here
```
ahayzen-kdab, line 61: turns out it is when the namespace is on the `type MyObject` and not on the cxx_qt::bridge

crates/cxx-qt-gen/src/generator/naming/qobject.rs
```diff
@@ -62,6 +63,15 @@ impl QObjectNames {
...
         )
     }
+
+    pub fn namespace_tokens(&self) -> TokenStream {
+        let namespace = self.name.namespace();
+        if namespace.is_none() {
+            quote! {}
+        } else {
+            quote! { #[namespace = #namespace ] }
+        }
```
ahayzen-kdab, line 73: could be
```rust
if let Some(namespace) = self.name.namespace() {
    quote! { #[namespace = #namespace] }
else {
    quote! {}
}
```

crates/cxx-qt-gen/src/generator/rust/qobject.rs
```diff
@@ -160,6 +160,7 @@ fn generate_qobject_definitions(
             },
             quote! {
                 extern "Rust" {
+                    #namespace // Needed for namespacing QObjects
```
ahayzen-kdab, line 163: can the place where `let namespace` is done here use your new helper too ?

crates/cxx-qt-gen/src/generator/naming/qobject.rs
```diff
@@ -62,6 +63,15 @@ impl QObjectNames {
             ident = self.name.cxx_unqualified().to_case(Case::Snake)
         )
     }
+
+    pub fn namespace_tokens(&self) -> TokenStream {
```
ahayzen-kdab, line 67: doc line would be good

crates/cxx-qt-gen/src/generator/rust/property/getter.rs
```diff
@@ -34,11 +34,13 @@ pub fn generate(
...
         let qualified_impl = type_names.rust_qualified(cpp_class_name_rust)?;
 
+        let cxx_namespace = qobject_idents.namespace_tokens();
+
         Ok(Some(RustFragmentPair {
             cxx_bridge: vec![quote! {
                 extern "Rust" {
                     #[cxx_name = #getter_wrapper_cpp]
-                    // Namespace is not needed here
+                    #cxx_namespace // Needed so QObjects can have a namespace
```
ahayzen-kdab, line 43: maybe change the comment to something like the following
```
// Needed for QObjects that have a namespace on their type item or extern block
//
// Note that a namespace in the cxx_qt::bridge would be applied to all child items
// but for other cases there is no parent so we need to be explicit
#cxx_namespace
```

ahayzen-kdab: (same for the others)

crates/cxx-qt-gen/src/generator/rust/qobject.rs
```diff
@@ -160,6 +150,7 @@ fn generate_qobject_definitions(
             },
             quote! {
                 extern "Rust" {
+                    #namespace // Needed for namespacing QObjects individually, not just by namespacing entire bridge
```
ahayzen-kdab, line 153: update the longer comment

BenFordTytherington: Done

ahayzen-kdab (approved):
LGTM :-)

PR #1063, opened 2024-09-09
Improve attribute parsing

BenFordTytherington:
Attribute parsing is somewhat inconsistent, and invalid / unknown attributes are mostly ignored and or passed through.

This PR aims to improve consistency of attribute parsing and hopefully remove the use of attribute_take_path in order to preserve all original attributes

### DIFF
--- a/crates/cxx-qt-gen/src/generator/rust/externcxxqt.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/externcxxqt.rs
@@ -2,7 +2,6 @@
 // SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
-
 use crate::{
     generator::rust::{
         fragment::{GeneratedRustFragment, RustFragmentPair},
@@ -10,9 +9,10 @@ use crate::{
     },
     naming::TypeNames,
     parser::externcxxqt::ParsedExternCxxQt,
+    syntax::path::path_compare_str,
 };
 use quote::quote;
-use syn::Result;
+use syn::{Attribute, Result};
 
 impl GeneratedRustFragment {
     pub fn from_extern_cxx_qt(
@@ -51,9 +51,16 @@ impl GeneratedRustFragment {
                         #[cxx_name = #cxx_name]
                     }
                 };
+                let docs: Vec<&Attribute> = ty
+                    .declaration
+                    .attrs
+                    .iter()
+                    .filter(|attr| path_compare_str(attr.meta.path(), &["doc"]))
+                    .collect();
                 quote! {
                     #namespace
                     #cxx_name
+                    #(#docs)*
                     #vis type #ident;
                 }
             })
--- a/crates/cxx-qt-gen/src/generator/rust/externcxxqt.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/externcxxqt.rs
@@ -21,7 +21,7 @@ impl GeneratedRustFragment {
     ) -> Result<Self> {
         let mut generated = Self::default();
 
-        let namespace = if let Some(namespace) = &extern_cxxqt_block.namespace {
+        let extern_block_namespace = if let Some(namespace) = &extern_cxxqt_block.namespace {
             quote! { #[namespace = #namespace ] }
         } else {
             quote! {}
@@ -68,7 +68,7 @@ impl GeneratedRustFragment {
 
         let fragment = RustFragmentPair {
             cxx_bridge: vec![quote! {
-                #namespace
+                #extern_block_namespace
                 #unsafety extern "C++" {
                     #(#items)*
 
--- a/crates/cxx-qt-gen/src/generator/rust/externcxxqt.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/externcxxqt.rs
@@ -30,12 +30,42 @@ impl GeneratedRustFragment {
         // Add the pass through blocks
         let unsafety = &extern_cxxqt_block.unsafety;
         let items = &extern_cxxqt_block.passthrough_items;
+        let types = &extern_cxxqt_block
+            .qobjects
+            .iter()
+            .map(|ty| {
+                let namespace = if let Some(namespace) = &ty.name.namespace() {
+                    quote! { #[namespace = #namespace ] }
+                } else {
+                    quote! {}
+                };
+                let cpp_name = &ty.name.cxx_unqualified();
+                let rust_name = &ty.name.rust_unqualified();
+                let vis = &ty.declaration.vis;
+                let ident = &ty.name.rust_unqualified();
+                let cxx_name = if &rust_name.to_string() == cpp_name {
+                    quote! {}
+                } else {
+                    let cxx_name = cpp_name.to_string();
+                    quote! {
+                        #[cxx_name = #cxx_name]
+                    }
+                };
+                quote! {
+                    #namespace
+                    #cxx_name
+                    #vis type #ident;
+                }
+            })
+            .collect::<Vec<_>>();
 
         let fragment = RustFragmentPair {
             cxx_bridge: vec![quote! {
                 #namespace
                 #unsafety extern "C++" {
                     #(#items)*
+
+                    #(#types)*
                 }
             }],
             implementation: vec![],
--- a/crates/cxx-qt-gen/src/generator/rust/externcxxqt.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/externcxxqt.rs
@@ -21,13 +21,18 @@ impl GeneratedRustFragment {
     ) -> Result<Self> {
         let mut generated = Self::default();
 
+        let namespace = if let Some(namespace) = &extern_cxxqt_block.namespace {
+            quote! { #[namespace = #namespace ] }
+        } else {
+            quote! {}
+        };
+
         // Add the pass through blocks
-        let attrs = &extern_cxxqt_block.attrs;
         let unsafety = &extern_cxxqt_block.unsafety;
         let items = &extern_cxxqt_block.passthrough_items;
         let fragment = RustFragmentPair {
             cxx_bridge: vec![quote! {
-                #(#attrs)*
+                #namespace
                 #unsafety extern "C++" {
                     #(#items)*
                 }
--- a/crates/cxx-qt-gen/src/generator/rust/externcxxqt.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/externcxxqt.rs
@@ -30,6 +30,7 @@ impl GeneratedRustFragment {
         // Add the pass through blocks
         let unsafety = &extern_cxxqt_block.unsafety;
         let items = &extern_cxxqt_block.passthrough_items;
+
         let fragment = RustFragmentPair {
             cxx_bridge: vec![quote! {
                 #namespace
--- a/crates/cxx-qt-gen/src/generator/rust/inherit.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/inherit.rs
@@ -76,7 +76,7 @@ mod tests {
         method: ForeignItemFn,
         safety: Safety,
     ) -> Result<GeneratedRustFragment> {
-        let method = ParsedInheritedMethod::parse(method, safety).unwrap();
+        let method = ParsedInheritedMethod::parse(method, safety)?;
         let inherited_methods = vec![&method];
         generate(&create_qobjectname(), &inherited_methods)
     }
--- a/crates/cxx-qt-gen/src/generator/rust/inherit.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/inherit.rs
@@ -46,12 +46,13 @@ pub fn generate(
                 std::mem::swap(&mut unsafe_call, &mut unsafe_block);
             }
             let doc_comments = &method.docs;
+            let namespace = qobject_names.namespace_tokens();
 
             syn::parse2(quote_spanned! {
                 method.method.span() =>
                 #unsafe_block extern "C++" {
-                    // TODO: should namespace be added here?
                     #[cxx_name = #cxx_name_string]
+                    #namespace
                     #(#doc_comments)*
                     #unsafe_call fn #ident(#self_param, #(#parameters),*) #return_type;
                 }
--- a/crates/cxx-qt-gen/src/generator/rust/inherit.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/inherit.rs
@@ -45,14 +45,12 @@ pub fn generate(
             if method.safe {
                 std::mem::swap(&mut unsafe_call, &mut unsafe_block);
             }
-
-            let attrs = &method.method.attrs;
             let doc_comments = &method.docs;
 
             syn::parse2(quote_spanned! {
                 method.method.span() =>
                 #unsafe_block extern "C++" {
-                    #(#attrs)*
+                    // TODO: should namespace be added here?
                     #[cxx_name = #cxx_name_string]
                     #(#doc_comments)*
                     #unsafe_call fn #ident(#self_param, #(#parameters),*) #return_type;
--- a/crates/cxx-qt-gen/src/naming/type_names.rs
+++ b/crates/cxx-qt-gen/src/naming/type_names.rs
@@ -5,10 +5,9 @@
 
 use std::collections::{btree_map::Entry, BTreeMap, BTreeSet};
 
-use quote::format_ident;
+use quote::{format_ident, quote};
 use syn::{
-    parse_quote, token::Brace, Attribute, Error, Ident, Item, ItemEnum, ItemForeignMod, ItemStruct,
-    Path, Result,
+    parse_quote, Attribute, Error, Ident, Item, ItemEnum, ItemForeignMod, ItemStruct, Path, Result,
 };
 
 use crate::{
@@ -168,16 +167,18 @@ impl TypeNames {
         }
 
         for extern_cxxqt in &cxx_qt_data.extern_cxxqt_blocks {
-            // TODO: Refactor, this is a hack to reconstruct the original ItemForeignMod
-            let foreign_mod = ItemForeignMod {
-                attrs: extern_cxxqt.attrs.clone(),
-                unsafety: None,
-                brace_token: Brace::default(),
-                items: extern_cxxqt.passthrough_items.clone(),
-                abi: syn::Abi {
-                    extern_token: syn::parse_quote!(extern),
-                    name: None,
-                },
+            let namespace = if let Some(namespace) = &extern_cxxqt.namespace {
+                quote! { #[namespace = #namespace ] }
+            } else {
+                quote! {}
+            };
+
+            let items = extern_cxxqt.passthrough_items.clone();
+            let foreign_mod: ItemForeignMod = parse_quote! {
+                #namespace
+                extern "C++" {
+                    #(#items)*
+                }
             };
             self.populate_from_foreign_mod_item(&foreign_mod, bridge_namespace, module_ident)?;
         }
--- a/crates/cxx-qt-gen/src/naming/type_names.rs
+++ b/crates/cxx-qt-gen/src/naming/type_names.rs
@@ -337,7 +337,7 @@ impl TypeNames {
     fn populate(
         &mut self,
         ident: &Ident,
-        attrs: &[Attribute],
+        attrs: &[Attribute], // TODO: Should these be only valid attrs relevant to name?
         parent_namespace: Option<&str>,
         module_ident: &Ident,
     ) -> Result<()> {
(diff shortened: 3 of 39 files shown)


crates/cxx-qt-gen/src/generator/rust/externcxxqt.rs
```diff
@@ -21,15 +21,58 @@ impl GeneratedRustFragment {
     ) -> Result<Self> {
         let mut generated = Self::default();
 
+        let namespace = if let Some(namespace) = &extern_cxxqt_block.namespace {
```
ahayzen-kdab, line 24: could call this `extern_block_namespace` to avoid any confusion / shadowing in the `map` where that has a namespace too

crates/cxx-qt-gen/src/generator/rust/inherit.rs
```diff
@@ -45,14 +45,12 @@ pub fn generate(
...
             }
-
-            let attrs = &method.method.attrs;
             let doc_comments = &method.docs;
 
             syn::parse2(quote_spanned! {
                 method.method.span() =>
                 #unsafe_block extern "C++" {
-                    #(#attrs)*
+                    // TODO: should namespace be added here?
```
ahayzen-kdab, line 53: good question, it seems to work at the moment or maybe a test case doesn't try this out yet. Would be good to check that CXX generates the right thing when the `cxx::bridge` does not have a namespace, as this extern block doesn't have a namespace either, but the self QObject is in a namespace.

crates/cxx-qt-gen/src/naming/type_names.rs
```diff
@@ -168,18 +159,24 @@ impl TypeNames {
...
+            let items = extern_cxxqt.passthrough_items.clone();
+            let foreign_mod: ItemForeignMod = parse_quote! {
+                #namespace
+                extern "C++" {
+                    #(#items)*
+                }
             };
             self.populate_from_foreign_mod_item(&foreign_mod, bridge_namespace, module_ident)?;
+
+            for qobject in extern_cxxqt.qobjects.iter() {
```
ahayzen-kdab, line 177: do you also need to add any qsignals in a extern_cxxqt qobject to the type names ? As the `#[qsignal]` in a C++Qt block are not in the passthrough_items (?) and might have a cxx_name/rust_name?

crates/cxx-qt-gen/src/parser/cxxqtdata.rs
```diff
@@ -126,8 +127,11 @@ impl ParsedCxxQtData {
...
     fn parse_foreign_mod_rust_qt(&mut self, mut foreign_mod: ItemForeignMod) -> Result<()> {
-        let namespace = attribute_find_path(&foreign_mod.attrs, &["namespace"])
-            .map(|index| expr_to_string(&foreign_mod.attrs[index].meta.require_name_value()?.value))
+        let attrs = parse_attributes(&foreign_mod.attrs, &["namespace"])?;
+
+        let namespace = attrs
+            .get("namespace")
+            .map(|attr| expr_to_string(&attr.meta.require_name_value()?.value))
             .transpose()?
             .or_else(|| self.namespace.clone());
```
ahayzen-kdab, line 136: why does this not use attribute_get_path like elsewhere?

ahayzen-kdab: Same further down, what is the difference between parse_attributes and attribute_get_path?

I guess is parse_attributes only allows the given attributes and fails if they aren't there whereas attribute_get_path doesn't mind about unknown attributes?

Maybe instead `parse_attributes` should be called `require_attributes` or `parse_require_attributes` or something with `require` in the name so it's clear it fails if none of the attributes in the list are found?

Then places that use `attribute_get_path` should be checked if they should be using the require method?

crates/cxx-qt-gen/src/parser/mod.rs
```diff
@@ -45,11 +47,36 @@ fn check_safety(method: &ForeignItemFn, safety: &Safety) -> Result<()> {
...
-        docs.push(doc);
+    method
+        .attrs
+        .iter()
+        .filter(|attr| path_compare_str(attr.meta.path(), &["doc"]))
+        .cloned()
+        .collect()
+}
+
+pub fn parse_attributes<'a>(
```
ahayzen-kdab, line 58: doc line would be good, and as described in another comment would be good to get the word `require` into the name so that it matches the syn API. eg `parse_require_attributes`

crates/cxx-qt-gen/src/parser/mod.rs
```diff
@@ -100,6 +128,16 @@ impl Parser {
...
 
+        // Maybe not the best but removes use of take_path
+        let filtered: Vec<Attribute> = module
+            .attrs
+            .iter()
+            .filter(|attr| !path_compare_str(attr.meta.path(), &["cxx_qt", "bridge"]))
+            .cloned()
+            .collect();
+
+        parse_attributes(&filtered, &["doc"])?;
```
ahayzen-kdab, line 139: ah yes would be nice to make `parse_attributes` support longer paths like `cxx_qt::bridge`. Could you make `parse_attributes` check if the string contains a `::` if it does then split it and use path_compare_str on the list if it doesn't then just do `[string]` like it does currently?

crates/cxx-qt-gen/src/parser/mod.rs
```diff
@@ -280,7 +306,7 @@ mod tests {
 
         assert_ne!(parser.passthrough_module, module);
 
-        assert_eq!(parser.passthrough_module.attrs.len(), 0);
+        assert_eq!(parser.passthrough_module.attrs.len(), 1);
```
ahayzen-kdab, line 309: what is the attribute on the passthrough_module? As we should not be passing through `cxx_qt::bridge` i guess.

Instead we might need to change the passthrough_module to be a `passthrough_items: Vec<Item>` + `module_ident: Ident` + visibility and rebuild it in the generation with something like this?
```
#[cxx::bridge]
#module_visibility mod #module_ident {
  #(#items)*
}
```

crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -46,6 +44,19 @@ pub struct ParsedQObject {
 }
 
 impl ParsedQObject {
+    const ALLOWED_ATTRS: [&'static str; 11] = [
+        "cxx_name",
+        "rust_name",
+        "namespace",
+        "derive",
```
ahayzen-kdab, line 51: what is `derive` used for on the QObject? And would that even work? As wouldn't it be on the Rust struct side not the QObject type in the extern "RustQt" ?

crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -120,47 +132,35 @@ impl ParsedQObject {
...
                 name,
                 uncreatable,
                 singleton,
             }));
         }
-
         Ok(None)
     }
 
     fn parse_property_attributes(attrs: &mut Vec<Attribute>) -> Result<Vec<ParsedQProperty>> {
```
ahayzen-kdab, line 155: wonder if this should be `&Vec<Attribute>`

crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -120,47 +132,35 @@ impl ParsedQObject {
...
+        // Once extract_if is stable, this would allow comparing all the elements using
+        // path_compare_str and building ParsedQProperty from the extracted elements.
         // https://doc.rust-lang.org/nightly/std/vec/struct.Vec.html#method.extract_if
-        while let Some(attr) = attribute_take_path(attrs, &["qproperty"]) {
-            properties.push(ParsedQProperty::parse(attr)?);
-        }
-
-        Ok(properties)
+        attrs
+            .drain(..)
```
ahayzen-kdab, line 160: wonder if this should be `iter()` as this modifies the attribute list rather than parsing and rebuilding which is the new philosophy?

crates/cxx-qt-gen/src/syntax/foreignmod.rs
```diff
@@ -63,7 +63,7 @@ fn verbatim_to_foreign_type(tokens: &TokenStream) -> Result<Option<ForeignItemTy
             })?;
 
             Ok(Some(parse_quote! {
-                #(#attrs)*
+                #(#attrs)* // TODO: Should these be checked at this point? Passthrough of random attrs could cause problems
```
ahayzen-kdab, line 66: in theory this is in the `syntax` module which is used by the `parser` module. So these are just helpers for `syn` and are not tightly bound to the CXX-Qt syntax.

So here i think it would be fine to have all attributes go through unchecked, as whatever is using this method should be checking the resultant `ForiegnItemType` and parsing that elsewhere

crates/cxx-qt-gen/src/syntax/foreignmod.rs
```diff
@@ -95,6 +95,7 @@ pub struct ForeignTypeIdentAlias {
 impl Parse for ForeignTypeIdentAlias {
     fn parse(input: ParseStream) -> Result<Self> {
         let attrs = input.call(Attribute::parse_outer)?;
+        // TODO: Potentially check attr validity here?
```
ahayzen-kdab, line 98: same, these are just helpers

crates/cxx-qt-gen/src/syntax/qtfile.rs
```diff
@@ -12,7 +12,7 @@ use syn::{AttrStyle, Attribute, Result};
 /// Representation of a CxxQtFile as Syn items
 pub struct CxxQtFile {
     /// A vector of [syn::Attribute] in the file
-    pub attrs: Vec<Attribute>,
+    pub attrs: Vec<Attribute>, // TODO: Should these be checked at this point?
```
ahayzen-kdab, line 15: same i think, this is just a helper for the shape of the CXX-Qt syntax in a syn-like struct that then parser uses

crates/cxx-qt-gen/src/writer/rust/mod.rs
```diff
@@ -47,6 +48,15 @@ pub fn write_rust(generated: &GeneratedRustBlocks, include_path: Option<&str>) -
...
 
+    let docs: Vec<Attribute> = cxx_mod
+        .attrs
+        .iter()
+        .filter(|attr| path_compare_str(attr.meta.path(), &["doc"]))
+        .cloned()
+        .collect();
+
+    cxx_mod.attrs = vec![]; // Reset all attributes to stop cxx_qt::bridge being passed through
+
```
ahayzen-kdab, line 59: instead see the comment further back about how we maybe should store the module as items and then rebuild here :thinking:

crates/cxx-qt-gen/test_inputs/passthrough_and_naming.rs
```diff
@@ -37,7 +35,6 @@ pub mod ffi {
     extern "C" {}
 
     #[namespace = "namespace"]
-    #[custom_attr = "test"]
```
ahayzen-kdab, line 40: this one should still work as it's not a CXX-Qt related block

crates/cxx-qt-gen/test_inputs/passthrough_and_naming.rs
```diff
@@ -46,7 +43,6 @@ pub mod ffi {
     unsafe extern "C++" {}
 
     #[namespace = "namespace"]
-    #[custom_attr = "test"]
```
ahayzen-kdab, line 49: this one should still work as it's not a CXX-Qt related block

crates/cxx-qt-gen/test_inputs/inheritance.rs
```diff
@@ -9,6 +9,11 @@ mod inheritance {
         type QAbstractItemModel;
     }
 
+    extern "C++Qt" {
+        #[qobject]
+        type MyStruct;
+    }
+
```
ahayzen-kdab, line 16: what is this for?

crates/cxx-qt-gen/test_outputs/qenum.rs
```diff
@@ -18,6 +18,7 @@ mod ffi {
         include!("directory/file_ident.cxxqt.h");
     }
     #[repr(i32)]
+    #[namespace = "cxx_qt::my_object"]
```
ahayzen-kdab, line 21: so before these were getting the namespace from the bridge level one, guess we are now explicit.

But i note looking at some of the other outputs not everything has an explicit namespace, looks like signals are missing them? (if we are going down the everything we generate has an explicit namespace)

crates/cxx-qt-gen/src/generator/rust/externcxxqt.rs
BenFordTytherington: Done

crates/cxx-qt-gen/src/naming/type_names.rs
```diff
@@ -168,18 +159,24 @@ impl TypeNames {
...
+                self.insert(qobject.name.clone())?;
+            }
```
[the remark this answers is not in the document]
BenFordTytherington: I moved qobjects out of the passthrough_items vector since it impacted how I could rebuild them, so they then have to be added to type_names, else it will cause an unknown type error. So this is registering qobjects from extern C++Qt yes

crates/cxx-qt-gen/src/naming/type_names.rs
BenFordTytherington: I don't think those are added ever you're right, although I also think that's been the case prior to this change and it seems to be working?

ahayzen-kdab: I'm not sure a rename of a qsignal in a C++Qt (not a RustQt) block has ever been tried/tested maybe. So would be good for completeness i guess.

crates/cxx-qt-gen/src/parser/cxxqtdata.rs
BenFordTytherington: Yeah, I'll change it to require. It will error if an attribute is found that isn't in the allowed list and also creates the Map of all attributes that it does find, so then it does all the collecting in one place, which did clean things up in one or two places. Some places attribute_get_path are still used are because they would require an absurd list of allowed attrs, and I want to search for an attribute but not error.

crates/cxx-qt-gen/src/parser/mod.rs
```diff
@@ -6,6 +6,7 @@
 pub mod constructor;
 pub mod cxxqtdata;
 pub mod externcxxqt;
+mod externqobject;
```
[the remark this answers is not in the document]
BenFordTytherington: Done

crates/cxx-qt-gen/src/parser/mod.rs
BenFordTytherington: It's the #[cxx_qt::bridge] there, since I moved to not removing it, I think that could be one candidate where removing it is not a bad idea.

crates/cxx-qt-gen/src/parser/qobject.rs
BenFordTytherington: #[derive(Default)] for construction, It was in some tests or examples somewhere I believe

crates/cxx-qt-gen/src/parser/mod.rs
ahayzen-kdab: See other comment somewhere below where instead of storing the module as passthrough maybe we have

- vector of items
- doc attributes
- module ident
- module visibility

And then rebuild it

crates/cxx-qt-gen/src/parser/qobject.rs
BenFordTytherington: Drain needs a mutable reference which is used in parse_property_attributes

crates/cxx-qt-gen/src/parser/qobject.rs
ahayzen-kdab: Hmm, i think most of those would be on the Rust struct?

```rust
#[cxx_qt::bridge]
mod ffi {
   extern "RustQt" {
     #[qobject]  // there is no derive here
   }
}

#[derive(Default)]   // this struct must implement Default for the default constructor
struct MyObjectRust;
```

I think maybe those are the derive on the struct outside of the bridge? So i think it's likely we don't need derive here ? :thinking:

crates/cxx-qt-gen/src/parser/qobject.rs
ahayzen-kdab: See next comment, i don't think we should drain if we are not removing attributes

crates/cxx-qt-gen/src/parser/qobject.rs
BenFordTytherington: ParsedQProperty::parse needs to take an owned attr, so using iter creates a reference, and it cannot be moved. So either a mutable reference and drain needs to be used, or attr needs to be cloned I think

crates/cxx-qt-gen/src/syntax/foreignmod.rs
BenFordTytherington: Cool, removed comment

crates/cxx-qt-gen/src/parser/qobject.rs
ahayzen-kdab: Probably clone it into the ParsedQProperty::parse then, or change ParsedQProperty to take a reference (but what do other parse's do?)

crates/cxx-qt-gen/src/syntax/foreignmod.rs
BenFordTytherington: Same as above, removed comment

crates/cxx-qt-gen/src/syntax/qtfile.rs
BenFordTytherington: Removed comment

crates/cxx-qt-gen/test_inputs/passthrough_and_naming.rs
BenFordTytherington: Are there any invalid attributes for there? that one originates in TypeNames::populate_from_foreign_mod_item, so I replaced like everywhere else, but that might need to just be attribute_get_path

ahayzen-kdab: Ah right, so these are just CXX blocks so we should probably just pass through anything to CXX and let it deal with them. And in TypeNames yes use attribute_get_path to find any namespace attribute and just ignore the rest

crates/cxx-qt-gen/test_inputs/inheritance.rs
ahayzen-kdab: Would be good to have a tighter unit test instead if that is the case.

Note that in `test_intputs/signals.rs` it has this code

```rust
unsafe extern "C++Qt" {
...
   #[qobject]
   type QTimer;
}
```

Isn't this the same?

crates/cxx-qt-gen/src/naming/type_names.rs
BenFordTytherington: Last push added in that registration and it didn't even break any of the tests, potentially need to add some tests for collisions

crates/cxx-qt-gen/src/parser/mod.rs

crates/cxx-qt-gen/src/parser/qobject.rs
BenFordTytherington: ahh it was just in some of the test_inputs, and removing them was fine so I've done that

crates/cxx-qt-gen/test_inputs/inheritance.rs
BenFordTytherington: Hmm, I wonder if that gets parsed differently because it's unsafe? there was one line missing in coverage, I'll remove that block so next CI run we should be able to see if that did affect coverage or if it's now fixed by elsewhere

ahayzen-kdab: Right this could be the case, but then a unit test tighter could be created rather than another entry in test_inputs

crates/cxx-qt-gen/test_inputs/passthrough_and_naming.rs
BenFordTytherington: Cool, done

crates/cxx-qt-gen/src/parser/mod.rs
BenFordTytherington: Done, added a path split helper function which might be useful elsewhere too

crates/cxx-qt-gen/src/writer/rust/mod.rs
BenFordTytherington: Done with use of new PassthroughMod struct

crates/cxx-qt-gen/src/parser/mod.rs
BenFordTytherington: New commit with PassthroughMod struct should do this correctly

crates/cxx-qt-gen/src/parser/cxxqtdata.rs
BenFordTytherington: Done

ahayzen-kdab (approved):
LGTM :-)

PR #1066, opened 2024-09-10
WIP: cxx-qt-build: use isolated cc::Build for QmlModules

### LINKED ISSUE #1065: QmlModule in cxx-qt-build should use their own isolated cc::Build
As we move to building QmlModule with their include paths being added #856 we need to avoid different QmlModule causing collisions between each other.

At the moment we have one large cc::Build that is used for the main CxxQtBuilder and any QmlModules, any files or include paths that are added to the cc::Build are then common everywhere as the same cc::Build is shared.

Instead we need to adjust our API so that QmlModules can also specify things like C++ source files to compile, C++ header files to give to moc, cc_builder closures etc.

Consider if we need a whole refactor of the public build API so that QmlModules become a sub / nested build.

Eg

```rust
CxxQtBuilder::qml_module("com.kdab.uri")
   .file(...)
   .qml(...)
   .qrc(...)
   .qobject_header(...)
   .cc_builder(|cc| ... )
   .build();

CxxQtBuilder::library()
   .file(...)
   .qml(...)
   .qrc(...)
   .qobject_header(...)
   .cc_builder(|cc| ... )
   .build();

CxxQtBuilder::executable()
   .file(...)
   .qml(...)
   .qrc(...)
   .qobject_header(...)
   .cc_builder(|cc| ... )
   .build();
```

This could then mean that one crate could potentially be one Qml Module or a library or an executable etc.

ahayzen-kdab:
Closes #1065


LeonMatthesKDAB:
We ended up going the opposite route for 0.8, with one runner that just builds the QML module.

PR #1068, opened 2024-09-10
Add base upcast

BenFordTytherington:
Add support to impl Upcast under the hood to silence dead code warning of base types.
Also switch from `#[base = "BaseClass"]` to `#[base = BaseClass]`

### DIFF
--- a/crates/cxx-qt-gen/src/generator/cpp/qobject.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/qobject.rs
@@ -234,7 +234,7 @@ mod tests {
             mod ffi {
                 extern "RustQt" {
                     #[qobject]
-                    #[base = "QStringListModel"]
+                    #[base = QStringListModel]
                     type MyObject = super::MyObjectRust;
                 }
             }
--- a/crates/cxx-qt-gen/src/generator/cpp/qobject.rs
+++ b/crates/cxx-qt-gen/src/generator/cpp/qobject.rs
@@ -15,6 +15,7 @@ use crate::{
     naming::Name,
 };
 use crate::{naming::TypeNames, parser::qobject::ParsedQObject};
+use quote::format_ident;
 use std::collections::BTreeSet;
 use syn::Result;
 
@@ -117,7 +118,7 @@ impl GeneratedCppQObject {
         let base_class = qobject.base_class.clone().unwrap_or_else(|| {
             // If there is a QObject macro then assume the base class is QObject
             if qobject.has_qobject_macro {
-                "QObject".to_string()
+                Name::new(format_ident!("QObject"))
             } else {
                 // CODECOV_EXCLUDE_START
                 unreachable!(
@@ -126,7 +127,10 @@ impl GeneratedCppQObject {
                 // CODECOV_EXCLUDE_STOP
             }
         });
-        generated.blocks.base_classes.push(base_class.clone());
+        generated
+            .blocks
+            .base_classes
+            .push(base_class.cxx_unqualified());
 
         // Add the CxxQtType rust and rust_mut methods
         generated
@@ -152,7 +156,10 @@ impl GeneratedCppQObject {
 
         generated.blocks.append(&mut inherit::generate(
             &structured_qobject.inherited_methods,
-            &qobject.base_class,
+            &qobject
+                .base_class
+                .clone() // TODO: can this be done without clone?
+                .map(|name| name.cxx_unqualified()),
             type_names,
         )?);
         generated.blocks.append(&mut qenum::generate_on_qobject(
@@ -173,7 +180,7 @@ impl GeneratedCppQObject {
         generated.blocks.append(&mut constructor::generate(
             &generated,
             &structured_qobject.constructors,
-            base_class,
+            base_class.cxx_unqualified(),
             &class_initializers,
             type_names,
         )?);
--- a/crates/cxx-qt-gen/src/generator/rust/qobject.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/qobject.rs
@@ -34,7 +34,13 @@ impl GeneratedRustFragment {
         let namespace_idents = NamespaceName::from(qobject);
         let mut generated = Self::default();
 
-        generated.append(&mut generate_qobject_definitions(&qobject_names)?);
+        generated.append(&mut generate_qobject_definitions(
+            &qobject_names,
+            qobject
+                .base_class
+                .clone()
+                .map(|name| name.cxx_unqualified()),
+        )?);
 
         // Generate methods for the properties, invokables, signals
         generated.append(&mut generate_rust_properties(
@@ -97,7 +103,10 @@ impl GeneratedRustFragment {
 }
 
 /// Generate the C++ and Rust CXX definitions for the QObject
-fn generate_qobject_definitions(qobject_idents: &QObjectNames) -> Result<GeneratedRustFragment> {
+fn generate_qobject_definitions(
+    qobject_idents: &QObjectNames,
+    base: Option<String>,
+) -> Result<GeneratedRustFragment> {
     let mut generated = GeneratedRustFragment::default();
     let cpp_class_name_rust = &qobject_idents.name.rust_unqualified();
     let cpp_class_name_cpp = &qobject_idents.name.cxx_unqualified();
@@ -114,6 +123,12 @@ fn generate_qobject_definitions(qobject_idents: &QObjectNames) -> Result<Generat
             #[doc = #cpp_class_name_cpp]
             #[cxx_name = #cpp_class_name_cpp]
         }
+    };
+
+    let base_upcast = if let Some(base) = base {
+        quote! { impl cxx_qt::Upcast<#base> for #rust_struct_name_rust {} }
+    } else {
+        quote! {}
     };
 
     let fragment = RustFragmentPair {
@@ -142,7 +157,7 @@ fn generate_qobject_definitions(qobject_idents: &QObjectNames) -> Result<Generat
                 }
             },
         ],
-        implementation: vec![],
+        implementation: vec![base_upcast],
     };
 
     generated
--- a/crates/cxx-qt-gen/src/parser/qobject.rs
+++ b/crates/cxx-qt-gen/src/parser/qobject.rs
@@ -10,7 +10,6 @@ use crate::{
         attribute::attribute_take_path, expr::expr_to_string, foreignmod::ForeignTypeIdentAlias,
     },
 };
-#[cfg(test)]
 use quote::format_ident;
 
 use syn::{Attribute, Error, Ident, Meta, Result};
@@ -28,7 +27,7 @@ pub struct QmlElementMetadata {
 /// then mutate these [syn::Item]'s for generation purposes.
 pub struct ParsedQObject {
     /// The base class of the struct
-    pub base_class: Option<String>,
+    pub base_class: Option<Name>,
     /// The name of the QObject
     pub name: Name,
     /// The ident of the inner type of the QObject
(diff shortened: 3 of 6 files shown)


crates/cxx-qt-gen/test_inputs/passthrough_and_naming.rs
```diff
@@ -115,7 +115,7 @@ pub mod ffi {
 
     extern "RustQt" {
         #[qobject]
-        #[base = "QStringListModel"]
+        #[base = QStringListModel]
```
ahayzen-kdab, line 118: add a changelog entry under changed for this, as base is now not a string

crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -73,19 +73,33 @@ impl ParsedQObject {
...
-                    ));
+            .map(|attr| -> Result<Ident> {
+                let expr = &attr.meta.require_name_value()?.value;
+                if let Expr::Path(path_expr) = expr {
+                    Ok(path_expr.path.require_ident()?.clone())
+                } else {
+                    Err(Error::new_spanned(
+                        expr,
+                        "Base must be a identifier and cannot be empty!",
+                    ))
```
ahayzen-kdab, line 101: somewhere, unsure where? Might even have to be later in the generator phase where it does that fake trait. We need to ensure that whatever type is used in the `#[base_class = T]` has been defined as a type and known to the TypeNames.

eg
```rust
extern "RustQt" {
   #[qobject]
   #[base_class = T]
   type MyObject = ...;
}
```

should fail without having a block that also defines the T
```rust
extern "C++" {
   include!("t.h");
   type T;
}
```

crates/cxx-qt-gen/src/generator/rust/qobject.rs
```diff
@@ -116,6 +122,12 @@ fn generate_qobject_definitions(qobject_idents: &QObjectNames) -> Result<Generat
         }
     };
 
+    let base_upcast = if let Some(base) = base {
+        quote! { impl cxx_qt::Upcast<#base> for #rust_struct_name_rust {} }
```
ahayzen-kdab, line 126: base here is a string rather than ident? so it's probably do `impl cxx_qt::Upcast<"T">` ?

BenFordTytherington: ah yup, forgot i'd switched back to string there

crates/cxx-qt-gen/src/generator/rust/qobject.rs
```diff
@@ -116,6 +124,15 @@ fn generate_qobject_definitions(qobject_idents: &QObjectNames) -> Result<Generat
         }
     };
 
+    let base_upcast = if let Some(base) = base {
+        let base_name = type_names
+            .lookup(&format_ident!("{base}"))?
```
ahayzen-kdab, line 129: isn't base an Ident elsewhere why not pass the Option<Ident> through?

crates/cxx-qt-gen/src/generator/cpp/qobject.rs
```diff
@@ -126,7 +127,7 @@ impl GeneratedCppQObject {
                 // CODECOV_EXCLUDE_STOP
             }
         });
-        generated.blocks.base_classes.push(base_class.clone());
+        generated.blocks.base_classes.push(base_class.to_string());
```
ahayzen-kdab, line 130: i think when this is not using the fallback of a QObject and the user has specified something. We need to do the lookup a type_names.lookup with it and then find the `cxx_qualified()`. As we might need the string to be `namespace::Base` or it might have a cxx_name etc

crates/cxx-qt-gen/src/generator/rust/qobject.rs
```diff
@@ -116,6 +124,15 @@ fn generate_qobject_definitions(qobject_idents: &QObjectNames) -> Result<Generat
...
+            .rust_unqualified();
+        quote! { impl cxx_qt::Upcast<#base_name> for #rust_struct_name_rust {} }
```
ahayzen-kdab, line 131: does the compiler or clippy complain about the type not being used? i think Leon needed to add a allow unused to the trait impl or trait itself or something?

BenFordTytherington: Not sure, He just said that using that trait prevents dead code warnings for the base type declaration which is now mandatory

crates/cxx-qt-gen/src/generator/rust/qobject.rs
```diff
@@ -116,6 +124,15 @@ fn generate_qobject_definitions(qobject_idents: &QObjectNames) -> Result<Generat
         }
     };
 
+    let base_upcast = if let Some(base) = base {
+        let base_name = type_names
+            .lookup(&format_ident!("{base}"))?
+            .rust_unqualified();
```
ahayzen-kdab, line 130: think this should be qualified as it needs the module? as it needs to be `ffi::T` not just `T`

crates/cxx-qt-gen/src/generator/cpp/qobject.rs
```diff
@@ -113,19 +113,12 @@ impl GeneratedCppQObject {
...
-                unreachable!(
-                    "Cannot have an empty #[base] attribute  with no #[qobject] attribute"
-                );
-                // CODECOV_EXCLUDE_STOP
-            }
-        });
+        let base_class = if let Some(ident) = &qobject.base_class {
+            type_names.lookup(ident)?.cxx_qualified()
+        } else {
+            "QObject".to_string()
```
ahayzen-kdab, line 119: this part is losing the `if qobject.has_qobject_macro` and then the unreachable block? That should still be in this else block i think

BenFordTytherington: Didn't think it was necessary anymore as we check for qobject and base better, but will add it back

ahayzen-kdab: Think it's a sanity check that we don't end up in a situation where there was no #[qobject] macro on a type without a #[base]. 

There is the question over whether we should require #[qobject] on all types in RustQt, do we require that already i forget? As there is the question over supporting #[qgadget] later. But maybe for now we could require all types to have qobject macro and then optionally have a base class? Then that `has_qobject_macro` could go and so could this block? Unless there is another reason for it?

ahayzen-kdab: I remember now, the `has_qobject_macro` was so that we can support https://doc.qt.io/qt-6/qquickframebufferobject-renderer.html which you need to inherit from but has now QObject as a base class

crates/cxx-qt-gen/src/generator/rust/qobject.rs
LeonMatthesKDAB: I think if the trait is coming from another crate, it probably can't figure out that this is unused, so shouldn't complain. :crossed_fingers: 
Otherwise the `#[allow(dead_code)]` needs to be on the trait in the cxx_qt crate itself.

crates/cxx-qt-gen/src/parser/qobject.rs
```diff
@@ -73,19 +73,33 @@ impl ParsedQObject {
...
+        //             // - when there is a qobject macro it is not valid
+        //             // - when there is not a qobject macro it is not valid
+        //             return Err(Error::new_spanned(
+        //                 attr,
+        //                 "The #[base] attribute cannot be empty!",
+        //             ));
+        //         }
+        //         Ok(string)
+        //     })
+        //     .transpose()?;
```
ahayzen-kdab, line 90: remove commented code?

crates/cxx-qt-gen/src/generator/rust/qobject.rs
ahayzen-kdab: I think this is still valid, instead of `        let base_name = type_names.lookup(&base)?.rust_unqualified();` it should be `        let base_name = type_names.lookup(&base)?.rust_qualified();` or something? As it needs to be `ffi::T` not `T` ?

crates/cxx-qt-gen/test_outputs/inheritance.rs
```diff
@@ -5,6 +5,7 @@ mod inheritance {
         type QModelIndex = cxx_qt_lib::QModelIndex;
         include!("cxx-qt-lib/qvariant.h");
         type QVariant = cxx_qt_lib::QVariant;
+        type QAbstractItemModel;
```
ahayzen-kdab, line 8: why are we not seeing the upcast trait being impl'd in the output? Seems maybe that is going wrong or needs `update_expected` running again?

crates/cxx-qt-gen/src/generator/rust/qobject.rs
```diff
@@ -116,6 +124,13 @@ fn generate_qobject_definitions(qobject_idents: &QObjectNames) -> Result<Generat
         }
     };
 
+    let base_upcast = if let Some(base) = base {
+        let base_name = type_names.lookup(&base)?.rust_unqualified();
+        quote! { impl cxx_qt::Upcast<#base_name> for #rust_struct_name_rust {} }
```
ahayzen-kdab, line 129: this should be implemented on the Rust version of the C++ class not the Rust struct? And it needs to be qualified?

Eg this should be `impl cxx_qt::Upcast<qobject::Base> for qobject::MyObject {}` and not `impl cxx_qt::Upcast<Base> for MyObjectRust {}`

crates/cxx-qt-gen/test_outputs/passthrough_and_naming.cpp
```diff
@@ -279,7 +279,7 @@ MyObject_readyConnect(
 
 namespace cxx_qt::multi_object {
 MyObject::MyObject(QObject* parent)
-  : QStringListModel(parent)
+  : cxx_qt::multi_object::QStringListModel(parent)
```
ahayzen-kdab, line 282: this is not quite right, but i think it's because the input has a namespace on the bridge?

Can you add a `#[namespace = ""]` onto the `QStringListModel` or it's extern block so we have a sane value here?

crates/cxx-qt-gen/test_outputs/passthrough_and_naming.rs
```diff
@@ -85,6 +86,7 @@ pub mod ffi {
         #[namespace = "cxx_qt::multi_object"]
         type MyObjectRust;
     }
+    impl cxx_qt::Upcast<ffi::QStringListModel> for ffi::MyObject {}
```
ahayzen-kdab, line 89: this needs to end up outside of the bridge, so must be inserted into the wrong list...

crates/cxx-qt-gen/src/generator/rust/qobject.rs
```diff
@@ -142,12 +158,15 @@ fn generate_qobject_definitions(qobject_idents: &QObjectNames) -> Result<Generat
...
         ],
-        implementation: vec![],
+        implementation: base_upcast,
     };
 
     generated
         .cxx_mod_contents
         .append(&mut fragment.cxx_bridge_as_items()?);
+    generated
+        .cxx_mod_contents
```
ahayzen-kdab, line 168: this should be `.cxx_qt_mod_contents` not `.cxx_mod_contents` that should fix the build failures

crates/cxx-qt-gen/test_outputs/passthrough_and_naming.rs
ahayzen-kdab: (see other comment about `cxx_qt_mod_contents` vs `cxx_mod_contents`)

LeonMatthesKDAB (approved):

PR #1071, opened 2024-09-12
Fix: Silence warning about unused base class

LeonMatthesKDAB:
For some reason when using CXX-Qt as a dependency outside of this repository, the warning about the unused struct wasn't being silenced. This PR should fix that issue


ahayzen-kdab (approved):
LGTM :sweat_smile:

PR #1072, opened 2024-09-13
Feature naming on properties

BenFordTytherington:
Allow the use of cxx_name and rust_name flags in qproperty in order to disable use of automatic camelcase conversion. 

example use:
```rust
#[qproperty(i32, num, cxx_name = "numberProp")]
struct MyStruct;
```


crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -59,18 +64,29 @@ pub struct ParsedQProperty {
...
+
+    // Only cxx_name and rust_name can use string literals for values
+    if ident == "cxx_name" || ident == "rust_name" {
+        let string = expr_to_string(expr).map_err(|_| {
+            Error::new(
+                ident.span(),
+                "cxx_name and rust_name must use string values like `cxx_name = \"myName\"`!",
+            )
+        })?;
+        value = format_ident!("{}", string);
```
ahayzen-kdab, line 77: seems a bit unfortunate that it ends up back as a Ident when this is a string :thinking: but would require larger changes i guess....

BenFordTytherington: Yeah it's just a prototype atm to see if we can get the correct behaviour

BenFordTytherington: On 2nd thoughts this is probably fine, as it is still identifying something, and used to fill fields in Name, but we just want the user to specify it in quotes?

ahayzen-kdab: Right we still wants the quotes for the input. But yes on the Rust side this is actually turning into an Ident and on the C++ side a String as usual, so not terrible and probably fine to keep this. Was more an observation.

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -162,25 +189,44 @@ impl ParsedQProperty {
                     ))
                 }
 
-                if let Some(read) = read {
+                let cxx_name_str = if let Some(name) = cxx_name {
+                    name.to_string()
+                } else {
+                    ident.to_string().to_case(Case::Camel)
```
ahayzen-kdab, line 195: i think the idea was to camelCase only if there was no rust_name or cxx_name ? Rather than if there is no cxx_name. Or have i misremembered ?

(but then all of this will start changing soon anyway)

ahayzen-kdab: So it might become more like this maybe?
```rust
let name = if cxx_name.is_none() && rust_name.is_none() {
   Name::new(ident)
     .with_cxx_name(/* to camel */)
} else {
  Name::new(ident)
    /* if there is a cxx_name */
    .with_cxx_name(...)
   /* if there is a rust_name */
    .with_rust_name(...)

  // or can this just be the Name::from_ident_and_attrs(...) ?
};
```

crates/cxx-qt-gen/test_inputs/properties.rs
```diff
@@ -12,7 +12,13 @@ mod ffi {
         #[qproperty(i32, primitive)]
         #[qproperty(QPoint, trivial)]
         #[qproperty(i32, custom_function_prop, READ = my_getter, WRITE = my_setter, NOTIFY)]
-        #[qproperty(i32, readonly_prop, READ)]
+        #[qproperty(
+            i32,
+            readonly_prop,
+            READ,
+            cxx_name = "renamedProperty",
+            rust_name = "renamed_property"
```
ahayzen-kdab, line 20: maybe call these renamed_readonly_property and renamedReadonlyProperty ?

BenFordTytherington: Done, made them write too so we can see the name of the setter

ahayzen-kdab: Now it's not read-only though :-D maybe instead of editing the existing readonly_prop, revert that to how it was and add a new property with a custom read/write/notify called renamed_property / renamedProperty to check all cases?

BenFordTytherington: Fair point, don't know why I specifically picked that one to begin with

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -162,25 +189,46 @@ impl ParsedQProperty {
...
+                                required,
+                            },
+                        })
+                    } else {
+                        Err(Error::new(
+                            punctuated_flags.span(),
+                            "If any flags are passed, READ must be explicitly specified!",
+                        ))
+                    }
+                } else {
```
ahayzen-kdab, line 226: some code comments through this part might be useful to state what each if statement is doing

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
```diff
@@ -62,6 +62,7 @@ pub fn generate_cpp_properties(
 pub mod tests {
     use super::*;
```
ahayzen-kdab, line 63: a compiled example of this in qml-features `naming.rs` would be good to check that rust_name / cxx_name work correctly

crates/cxx-qt-gen/src/generator/cpp/property/mod.rs
```diff
@@ -62,6 +62,7 @@ pub fn generate_cpp_properties(
 pub mod tests {
```
ahayzen-kdab, line 62: entry in the CHANGELOG would be good to under Added for adding cxx_name / rust_name support to qproperty attribute

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -59,18 +64,26 @@ pub struct ParsedQProperty {
...
-        func_signature = path_expr.path.require_ident()?.clone();
+    // Only cxx_name and rust_name can use string literals for values
+    let value = if ident == "cxx_name" || ident == "rust_name" {
+        let string = expr_to_string(expr).map_err(|_| {
+            Error::new(
+                ident.span(),
+                "cxx_name and rust_name must use string values like `cxx_name = \"myName\"`!",
+            )
+        })?;
+        format_ident!("{}", string)
```
LeonMatthesKDAB, line 76: Do not use format_ident here, as it will panic if the string cannot be converted to an ident.
So if I did e.g. `rust_name = "hello world"`, cxx-qt-gen would panic, instead of handling this gracefully.

If you can get hold of a LitStr instance, you should be able to call `parse::<Ident>` on it:
https://docs.rs/syn/latest/syn/struct.LitStr.html#method.parse

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -162,25 +189,46 @@ impl ParsedQProperty {
...
+                } else if rust_name.is_none() {
+                    ident.to_string().to_case(Case::Camel)
+                } else {
+                    ident.to_string()
+                };
+                let mut name = Name::new(ident).with_cxx_name(cxx_name_str);
+
+                if let Some(ident) = rust_name {
+                    name = name.with_rust_name(ident);
+                }
```
LeonMatthesKDAB, line 203: This feels like it belongs on `Name` itself, and should ideally share the logic with `Name::from_ident_and_attrs`.
Can we achieve that somehow?
That was kind of the point of having the `Name` type, to not have to reimplement this logic for eveything we parse.

LeonMatthesKDAB (commented):
Only a first review, and it's looking okay thus far.
We can hopefully share the logic that is already in `Name`.
Otherwise, this should finally close #939

crates/cxx-qt-gen/src/parser/property.rs
BenFordTytherington: I think we did have a Name::property_name type method which did this but with the old auto camel case conversion, so I removed it. I can add a new method back which handles that

crates/cxx-qt-gen/src/parser/property.rs
BenFordTytherington: Done using syn::parse_string

crates/cxx-qt-gen/src/parser/property.rs
BenFordTytherington: Done, and that naming logic has been extracted, so should be more readable

crates/cxx-qt-gen/src/naming/name.rs
```diff
@@ -136,34 +136,56 @@ impl Name {
...
+            rust: ident.clone(),
+            cxx: None,
+            namespace,
+            module: module.cloned().map(Path::from),
+        }
+        .with_options(cxx_name, rust_name, false))
+    }
+
+    /// Applies naming options to an existing name, applying logic about what should cause renaming
+    pub fn with_options(self, cxx: Option<String>, rust: Option<Ident>, auto_camel: bool) -> Self {
```
ahayzen-kdab, line 166: generally i prefer to use enums rather than bools in method args, otherwise you can end up with `method(true, false, true)` which has no meaning compared to `method(AutoCamel::Enabled, ...)` etc.

I do however note that this auto_camel arg might disappear in the other change? But then it may also come back if we add `auto_camel` as an option to the bridge :-)

BenFordTytherington: Done

ahayzen-kdab (approved):
LGTM :-)

PR #1074, opened 2024-09-13
Add CI runner for WASM build

### LINKED ISSUE #414: WASM Example

mattkdab:
Closes #414 (note #947 for further wasm code)

### DIFF
--- a/.github/workflows/github-cxx-qt-tests.yml
+++ b/.github/workflows/github-cxx-qt-tests.yml
@@ -88,6 +88,135 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           verbose: true
 
+  build-wasm:
+    name: Ubuntu 22.04 (wasm_32) Qt 6
+    needs: [clang_format, license_check, rust_format_check, markdown_lint, shellcheck]
+    runs-on: ubuntu-22.04
+    env:
+      SCCACHE_CACHE_SIZE: 600M
+    steps:
+      - name: "Checkout repository"
+        uses: actions/checkout@v4
+      - name: Setup toolchain
+        run: |
+          rustup component add clippy rustfmt
+          rustup target add wasm32-unknown-emscripten
+      - name: "Rust tools cache"
+        uses: actions/cache@v4
+        id: rust-tools-cache
+        with:
+          path: |
+            ~/.cargo/bin/sccache
+            ~/.cargo/bin/mdbook
+            ~/.cargo/bin/mdbook-linkcheck
+          key: ubuntu-22.04_sccache-0.7.6_mdbook-0.4.36_mdbook-linkcheck-0.7.7
+      - name: "Build Rust tools"
+        if: steps.rust-tools-cache.outputs.cache-hit != 'true'
+        # Do not build with storage backends enabled, we only need local
+        run: |
+          cargo install --no-default-features sccache
+          cargo install mdbook mdbook-linkcheck
+      # We want our compiler cache to always update to the newest state.
+      # The best way for us to achieve this is to **always** update the cache after every landed commit.
+      # That way it will closely follow our development.
+      # And if a PR diverges a lot with its cache that's not a big deal, as it will be merged eventually.
+      #
+      # This is a workaround for the fact that GH doesn't support updating existing caches.
+      # See: https://github.com/azu/github-actions-overwrite-cache-example
+      #
+      # Ideally we'd like to use this:
+      # - name: "Compiler cache"
+      #   uses: actions/cache@v4
+      #   with:
+      #     update: true <------- THIS DOESN'T EXIST YET
+      #     path: /home/runner/.cache/sccache
+      #     key: "Ubuntu 22.04 (wasm_32) Qt6_compiler_cache"
+      - name: "Restore Compiler Cache"
+        id: compiler-cache-restore
+        uses: actions/cache/restore@v4
+        with:
+          path: /home/runner/.cache/sccache
+          key: "Ubuntu 22.04 (wasm_32) Qt6_compiler_cache"
+      - name: "emsdk cache"
+        uses: actions/cache@v4
+        id: emsdk-cache
+        with:
+          path: ./emsdk
+          key: emsdk_2.0.14
+      - name: "Setup emsdk"
+        if: steps.emsdk-cache.outputs.cache-hit != 'true'
+        run: |
+          git clone https://github.com/emscripten-core/emsdk
+          ./emsdk/emsdk install 2.0.14
+          ./emsdk/emsdk activate 2.0.14
+      - name: "Install Dependencies"
+        run: sudo apt-get update && sudo apt-get install -y ninja-build clang-format-12
+      - name: Install Qt GCC
+        uses: jurplel/install-qt-action@v3
+        with:
+          aqtversion: '==3.1.*'
+          version: '6.2.4'
+          host: 'linux'
+          target: 'desktop'
+          arch: 'gcc_64'
+          cache: true
+      - name: Install Qt WASM
+        uses: jurplel/install-qt-action@v3
+        with:
+          aqtversion: '==3.1.*'
+          version: '6.2.4'
+          host: 'linux'
+          target: 'desktop'
+          arch: 'wasm_32'
+          tools: 'tools_cmake'
+          cache: true
+      - name: "Configure"
+        # FIXME: ninja is not being found?
+        env:
+          RUSTC_WRAPPER: sccache
+        run: |
+          source ./emsdk/emsdk_env.sh
+          /home/runner/work/cxx-qt/Qt/6.2.4/wasm_32/bin/qt-cmake -DQT_HOST_PATH=/home/runner/work/cxx-qt/Qt/6.2.4/gcc_64 -DBUILD_WASM=ON -Bbuild .
+      - name: "Build"
+        env:
+          RUSTC_WRAPPER: sccache
+        run: cmake --build build --parallel 4
+      - name: Test output files exist
+        run: |
+          test -f ./build/examples/qml_minimal/example_qml_minimal.html
+          test -f ./build/examples/qml_minimal/example_qml_minimal.js
+          test -f ./build/examples/qml_minimal/libqml_minimal.a
+          test -x ./build/examples/qml_minimal/example_qml_minimal.wasm
+          test -f ./build/examples/qml_minimal/qtloader.js
+          test -f ./build/examples/qml_features/example_qml_features.html
+          test -f ./build/examples/qml_features/example_qml_features.js
+          test -f ./build/examples/qml_features/libqml_features.a
+          test -x ./build/examples/qml_features/example_qml_features.wasm
+          test -f ./build/examples/qml_features/qtloader.js
+      - name: "Print compiler cache statistics"
+        run: sccache --show-stats
+      # This is a workaround for the fact that GH doesn't support updating existing caches.
+      # See: https://github.com/azu/github-actions-overwrite-cache-example
+      - name: "Delete previous compiler cache"
+        # Updating th cache doesn't work from forks
+        # So update it once it's merged into the repo
+        if: ${{ steps.compiler-cache-restore.outputs.cache-hit &&  github.event_name == 'push' }}
+        working-directory: /home/runner/cxx-qt
+        continue-on-error: true
+        run: |
+          gh extension install actions/gh-actions-cache
+          gh actions-cache delete "Ubuntu 22.04 (wasm_32) Qt6_compiler_cache" --confirm
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: "Save Compiler Cache"
+        # Updating th cache doesn't work from forks
+        # So update it once it's merged into the repo
+        if: ${{ github.event_name == 'push' }}
+        uses: actions/cache/save@v4
+        with:
+          path: /home/runner/.cache/sccache
+          key: "Ubuntu 22.04 (wasm_32) Qt6_compiler_cache"
+
   build:
     # Run after pre checks
     needs: [clang_format, license_check, rust_format_check, markdown_lint, shellcheck]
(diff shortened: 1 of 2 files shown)


.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -88,6 +88,55 @@ jobs:
...
+          arch: 'gcc_64'
+          cache: true
+      - name: Install Qt WASM
+        uses: jurplel/install-qt-action@v3
+        with:
+          aqtversion: '==3.1.*'
+          version: '6.2.4'
+          host: 'linux'
+          target: 'desktop'
+          arch: 'wasm_32'
```
ahayzen-kdab, line 129: assume it cannot install two archs at the same time via some kind of comma separated thing and you have to have two blocks?

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -88,6 +88,55 @@ jobs:
...
+          tools: 'tools_cmake'
+          cache: true
+      - name: "Configure"
+        # ninja is not being found?
```
ahayzen-kdab, line 133: add `FIXME` to the start of the comment

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -88,6 +88,55 @@ jobs:
...
+          rustup component add clippy rustfmt
+          rustup target add wasm32-unknown-emscripten
+          git clone https://github.com/emscripten-core/emsdk
+          ./emsdk/emsdk install 2.0.14
+          ./emsdk/emsdk activate 2.0.14
+          source ./emsdk/emsdk_env.sh
+      - name: "Build Rust tools"
+        run: |
+          cargo install --no-default-features sccache
+          cargo install mdbook mdbook-linkcheck
```
ahayzen-kdab, line 110: change to the similar to the matrix so that it uses caching (as this step takes 5.5 mins) and we can just reuse the existing Ubuntu cache for the tools

```
    - name: "Rust tools cache"
      uses: actions/cache@v4
      id: rust-tools-cache
      with:
        path: |
          ~/.cargo/bin/sccache
          ~/.cargo/bin/mdbook
          ~/.cargo/bin/mdbook-linkcheck
        key: ubuntu-22.04_sccache-0.7.6_mdbook-0.4.36_mdbook-linkcheck-0.7.7
    - name: "Build Rust tools"
      # Do not build with storage backends enabled, we only need local
      run: |
        cargo install --no-default-features sccache
        cargo install mdbook mdbook-linkcheck
```

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -88,6 +88,55 @@ jobs:
...
+      - name: "Install Dependencies"
```
ahayzen-kdab, line 111: add in the compiler cache as we are setting up sccache

```
    # We want our compiler cache to always update to the newest state.
    # The best way for us to achieve this is to **always** update the cache after every landed commit.
    # That way it will closely follow our development.
    # And if a PR diverges a lot with its cache that's not a big deal, as it will be merged eventually.
    #
    # This is a workaround for the fact that GH doesn't support updating existing caches.
    # See: https://github.com/azu/github-actions-overwrite-cache-example
    #
    # Ideally we'd like to use this:
    # - name: "Compiler cache"
    #   uses: actions/cache@v4
    #   with:
    #     update: true <------- THIS DOESN'T EXIST YET
    #     path: /home/runner/.cache/sccache
    #     key: "Ubuntu 22.04 (wasm_32) Qt6_compiler_cache"
    - name: "Restore Compiler Cache"
      id: compiler-cache-restore
      uses: actions/cache/restore@v4
      with:
        path: /home/runner/.cache/sccache
        key: "Ubuntu 22.04 (wasm_32) Qt6_compiler_cache"
```

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -88,6 +88,55 @@ jobs:
...
+          tools: 'tools_cmake'
+          cache: true
+      - name: "Configure"
+        # ninja is not being found?
+        run: |
+          source ./emsdk/emsdk_env.sh
+          /home/runner/work/cxx-qt/Qt/6.2.4/wasm_32/bin/qt-cmake -DQT_HOST_PATH=/home/runner/work/cxx-qt/Qt/6.2.4/gcc_64 -DBUILD_WASM=ON -Bbuild .
+      - name: "Build"
+        run: cmake --build build --parallel 4
+
```
ahayzen-kdab, line 139: add a tidy up of the cache
```
    # This is a workaround for the fact that GH doesn't support updating existing caches.
    # See: https://github.com/azu/github-actions-overwrite-cache-example
    - name: "Delete previous compiler cache"
      # Updating th cache doesn't work from forks
      # So update it once it's merged into the repo
      if: ${{ steps.compiler-cache-restore.outputs.cache-hit &&  github.event_name == 'push' }}
      working-directory: ${{ matrix.workspace }}
      continue-on-error: true
      run: |
        gh actions-cache delete "Ubuntu 22.04 (wasm_32) Qt6_compiler_cache" --confirm
      env:
        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
    - name: "Save Compiler Cache"
      # Updating th cache doesn't work from forks
      # So update it once it's merged into the repo
      if: ${{ github.event_name == 'push' }}
      uses: actions/cache/save@v4
      with:
        path: /home/runner/.cache/sccache
        key: "Ubuntu 22.04 (wasm_32) Qt6_compiler_cache"
```

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -88,6 +88,55 @@ jobs:
...
+        uses: jurplel/install-qt-action@v3
+        with:
+          aqtversion: '==3.1.*'
+          version: '6.2.4'
+          host: 'linux'
+          target: 'desktop'
+          arch: 'wasm_32'
+          tools: 'tools_cmake'
+          cache: true
+      - name: "Configure"
```
ahayzen-kdab, line 132: enable `sccache` for configure step

```
      env:
        RUSTC_WRAPPER: sccache
```

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -88,6 +88,55 @@ jobs:
...
+        # ninja is not being found?
+        run: |
+          source ./emsdk/emsdk_env.sh
+          /home/runner/work/cxx-qt/Qt/6.2.4/wasm_32/bin/qt-cmake -DQT_HOST_PATH=/home/runner/work/cxx-qt/Qt/6.2.4/gcc_64 -DBUILD_WASM=ON -Bbuild .
+      - name: "Build"
```
ahayzen-kdab, line 137: enable `sccache` for build step

```
      env:
        RUSTC_WRAPPER: sccache
```

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -88,6 +88,55 @@ jobs:
...
+        run: cmake --build build --parallel 4
```
ahayzen-kdab, line 138: after the build add another step to print the caching stats (useful for debugging when things aren't working)

```
    - name: "Print compiler cache statistics"
      run: sccache --show-stats
```

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -88,6 +88,55 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           verbose: true
 
+  build-wasm:
+    needs: [clang_format, license_check, rust_format_check, markdown_lint, shellcheck]
+    runs-on: ubuntu-22.04
+    env:
+      SCCACHE_CACHE_SIZE: 600M
+    steps:
```
ahayzen-kdab, line 96: as a last step, could we check that the html file is created and optionally any other things that are stable paths like the .wasm bundle?

Use `test -f /path/to/file` to check if it's a file or `test -x /path/to/executable` to check if it's an executable file.

Assume as it's just a .html and .wasm you probably want like
```
     - name: Test output files exist
        run: |
          test -f ./build/../index.html
          test -f ./build/../bundle.wasm
```

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -88,6 +88,55 @@ jobs:
...
+      - name: "Checkout repository"
+        uses: actions/checkout@v4
+      - name: Setup toolchain
+        run: |
+          rustup component add clippy rustfmt
+          rustup target add wasm32-unknown-emscripten
+          git clone https://github.com/emscripten-core/emsdk
+          ./emsdk/emsdk install 2.0.14
+          ./emsdk/emsdk activate 2.0.14
+          source ./emsdk/emsdk_env.sh
```
ahayzen-kdab, line 106: This could be moved into it's own step of `Setup emsdk`.

Then see below about `Rust tool cache` + `Build rust tools` as you could copy this and use a cache with a key for `emsdk_2.0.14` and cache the sdk paths to save another 30s.

Also note that the `source` here is probably pointless as we found that each step has it's own env?

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -88,6 +88,55 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           verbose: true
 
+  build-wasm:
```
ahayzen-kdab, line 91: can you add a `name: Ubuntu 22.04 (wasm_32) Qt 6` as a key, then this will appear nicely in the CI workflow

.github/workflows/github-cxx-qt-tests.yml
mattkdab: yeah i dont think so: https://github.com/jurplel/install-qt-action/tree/v3/?tab=readme-ov-file#arch

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -88,6 +88,134 @@ jobs:
...
+      - name: "Setup emsdk"
+        run: |
+          git clone https://github.com/emscripten-core/emsdk
+          ./emsdk/emsdk install 2.0.14
+          ./emsdk/emsdk activate 2.0.14
+      - name: "emsdk cache"
+        uses: actions/cache@v4
+        id: emsdk-cache
+        with:
+          path: ./emsdk
```
ahayzen-kdab, line 149: You only need to setup if there is no cache, so swap them around and add a condition. Also we need the cache id to be versioned otherwise we cannot bump the emsdk version later to break out of the cache.

So it should look like this in the end

```
      - name: "emsdk cache"
        uses: actions/cache@v4
        id: emsdk-cache-2.0.14
        with:
          path: ./emsdk
      - name: "Setup emsdk"
        if: steps.emsdk-cache.outputs.cache-hit != 'true'
        run: |
          git clone https://github.com/emscripten-core/emsdk
          ./emsdk/emsdk install 2.0.14
          ./emsdk/emsdk activate 2.0.14
```

I assume that the `install` command only puts things inside the `emsdk` folder and not in some node_modules folder elsewhere?

.github/workflows/github-cxx-qt-tests.yml
ahayzen-kdab: Right

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -88,6 +88,134 @@ jobs:
...
+          test -f ./build/examples/qml_minimal/example_qml_minimal.html
+          test -f ./build/examples/qml_minimal/example_qml_minimal.js
+          test -f ./build/examples/qml_minimal/libqml_minimal.a
+          test -x ./build/examples/qml_minimal/example_qml_minimal.wasm
+          test -f ./build/examples/qml_minimal/qtloader.js
+          test -f ./build/examples/qml_features/example_qml_features.html
+          test -f ./build/examples/qml_features/example_qml_features.js
+          test -f ./build/examples/qml_features/libqml_features.a
+          test -x ./build/examples/qml_features/example_qml_features.wasm
+          test -f ./build/examples/qml_features/qtloader.js
```
ahayzen-kdab, line 217: while making changes, move this test step to be after the `cmake --build` and before the sccache stats / tidying up of cache. So it's in the same order as other tests

.github/workflows/github-cxx-qt-tests.yml
ahayzen-kdab: Ah wait i missed the `key` somehow in my copy, so the version is there already!

So just the reorder and if statement
```
      - name: "emsdk cache"
        uses: actions/cache@v4
        id: emsdk-cache
        with:
          path: ./emsdk
          key: emsdk_2.0.14
      - name: "Setup emsdk"
        if: steps.emsdk-cache.outputs.cache-hit != 'true'
        run: |
          git clone https://github.com/emscripten-core/emsdk
          ./emsdk/emsdk install 2.0.14
          ./emsdk/emsdk activate 2.0.14
```

mattkdab: i believe everything is within `emsdk`:

```console
╔ 📁 emsdk   main  🐍 v3.10.12 
╚══ ls -l --group-directories-first
total 256
drwxrwxr-x 6 sus sus   4096 Sep 12 10:54 bazel
drwxrwxr-x 2 sus sus   4096 Sep 12 10:54 docker
drwxrwxr-x 2 sus sus   4096 Sep 16 11:38 downloads
drwxrwxr-x 3 sus sus   4096 Sep 12 10:56 node
drwxrwxr-x 2 sus sus   4096 Sep 12 10:54 scripts
drwxrwxr-x 2 sus sus   4096 Sep 12 10:54 test
drwxrwxr-x 6 sus sus   4096 Sep 16 11:41 upstream
```

mattkdab: ^ there are other things in the dir too but those are the subfolders which contain for example a node distro etc

ahayzen-kdab: Right lets go with it for now :-)

ahayzen-kdab (approved):
LGTM :-)

PR #1076, opened 2024-09-17
Add full_qt feature and replace full with it, full now includes other deps

BenFordTytherington:
cxx-qt-lib-extras now uses full_qt feature, and examples now use that over full, since full has new added deps.


ahayzen-kdab (approved):
LGTM :-)

PR #1077, opened 2024-09-17
Fix error in book

BenFordTytherington:
One line fix in book, had deprecated qml_name attribute, which has been replaced by using `#[qml_element = "Name"]`


ahayzen-kdab (approved):
LGTM :-)

PR #1081, opened 2024-09-25
Hotfix: rust_name on #[inherit]

LeonMatthesKDAB:
This used ident directly, instead of rust_unqualified()

Also, add a corresponding regression test


ahayzen-kdab (approved):
LGTM :-)

PR #1082, opened 2024-09-30
Update README.md with Qt support

LeonMatthesKDAB:
We should explicitly state which versions of Qt we support, and which we don't.


ahayzen-kdab (approved):
LGTM :-)

PR #1083, opened 2024-09-30
ci: disable failing tests due to differing clang-format on macOS

ahayzen-kdab:
A better way would be to have a consistent clang-format version everywhere

### DIFF
--- a/.github/workflows/github-cxx-qt-tests.yml
+++ b/.github/workflows/github-cxx-qt-tests.yml
@@ -23,7 +23,7 @@ jobs:
       - uses: actions/checkout@v4
       # Note ensure that matrix runner is updated too
       - run: |
-          pip install --user clang-format==14.0.6
+          pip install --user clang-format==18.1.8
           test -x ~/.local/bin/clang-format
           test -x /home/runner/.local/bin/clang-format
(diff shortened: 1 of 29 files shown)


.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -282,7 +282,10 @@ jobs:
             # once Ubuntu 22.04 is we can move to clang-format-14 everywhere
             # for now we need at least clang-format-12 otherwise include reordering fails after clang-format off
             # https://github.com/KDAB/cxx-qt/issues/121
-            ctest_args: --exclude-regex '^(cargo_clippy|cargo_doc|example_qml_features_test|example_qml_minimal_myobject_test|cargo_build_rerun|.*valgrind)$'
+            #
+            # FIXME: differing clang-format version causes differences in generated C++ enum headers
```
ahayzen-kdab, line 286: i think we should use something like `pypi` to give us a consistently pinned version clang-format.

https://pypi.org/project/clang-format/

This could also then be added to something like pre-commit if we wanted that too later.

LeonMatthesKDAB (approved):

PR #1084, opened 2024-10-01
Fix: Require Send on closures connected to signals

### LINKED ISSUE #577: Add support for `extern "C++Qt"` with type, fn, and signals

LeonMatthesKDAB:
The closures may be executed on a different thread, depending on where
the emitting object lives.

Also closes #577. We have determined that is indeed safe to queue a
deferred closure, as the closure will not be called if the object is
deleted, as we use the object itself as the context.


ahayzen-kdab (approved):
LGTM :-)

PR #1086, opened 2024-10-01
WIP: Add QLocale

nicolasfella:
Not complete yet because:

- It's a lot
- I'm not sure how to wrap some API


crates/cxx-qt-lib-extras/src/core/qlocale.rs
```diff
@@ -0,0 +1,568 @@
...
+        /// Returns the exponent separator for this locale.
+        #[rust_name = "exponential"]
+        fn exponential(self: &QLocale) -> QString;
+
+        /// Returns the first day of the week according to the current locale.
+        // Qt::DayOfWeek QLocale::firstDayOfWeek() const
+        #[rust_name = "first_day_of_week"]
+        fn firstDayOfWeek(self: &QLocale) -> DayOfWeek;
+
+        // QString QLocale::formattedDataSize(qint64 bytes, int precision = 2, QLocale::DataSizeFormats format = DataSizeIecFormat) const
```
nicolasfella, line 479: DataSizeFormats is QFlags, do we have prior art on how to model that?

crates/cxx-qt-lib-extras/src/core/qlocale.cpp
```diff
@@ -0,0 +1,23 @@
...
+    return QLocale(name);
+}
+
+QLocale defaultLocale()
+{
+    return QLocale();
+}
+}
+}
+
```
ahayzen-kdab, line 22: the rust side has a `usize` as it's size, add a check here using `assert_alignment_and_size(QLocale, { ::std::size_t a0; });` to ensure that is true.

There needs to be a combination of static asserts on if it's trivially copyable/assignable/constructible/destructible etc.

If it is not one of these then you need to implement Drop/Clone manually etc (have a look around some of our other types to get ideas :-) )

crates/cxx-qt-lib-extras/include/core/qlocale.h
```diff
@@ -0,0 +1,31 @@
...
+
+using QLocaleTagSeparator = QLocale::TagSeparator;
+using QLocaleCurrencySymbolFormat = QLocale::CurrencySymbolFormat;
+using QLocaleFormatType = QLocale::FormatType;
+using QLocaleLanguage = QLocale::Language;
+using QLocaleMeasurementSystem = QLocale::MeasurementSystem;
+
+QLocale fromName(const QString& name);
+
+QLocale defaultLocale();
```
ahayzen-kdab, line 29: you likely don't need these and can use our constructor helper from cxx-qt-lib/common.h

eg

```rust
unsafe extern "C++ {
   include!("cxx-qt-lib/common.h");
  
   #[doc(hidden)]
   #[rust_name = "qlocale_init_default"]
   fn construct() -> QLocale;

   #[doc(hidden)]
   #[rust_name = "qlocale_init_from_name"]
   fn construct(name: &QString) -> QLocale;
}
```

This would likely work in your CXX bridge then you can do `ffi::qlocale_init_from_name(my_string)` and `ffi::qlocale_init_default()`

crates/cxx-qt-lib-extras/src/core/qlocale.rs
ahayzen-kdab: We unfortunately don't have anything for flags yet i don't think, we support enums but not flags.

@LeonMatthesKDAB something to figure out for 0.8 ?

nicolasfella:
Added a test

crates/cxx-qt-lib-extras/src/core/qlocale.rs
LeonMatthesKDAB: Right, sounds useful to find an ergonomic way to use Flags for 0.8. We don't have anything yet though.

crates/cxx-qt-lib-extras/src/core/qlocale.rs
```diff
@@ -0,0 +1,570 @@
...
+pub use ffi::QLocaleLanguage;
+pub use ffi::QLocaleMeasurementSystem;
+pub use ffi::QLocaleTagSeparator;
+
+#[repr(C)]
+pub struct QLocale {
+    _cspec: MaybeUninit<usize>,
+}
+
+impl Default for QLocale {
```
ahayzen-kdab, line 558: As QLocale is not trivially copyable or destructible can you add a `impl Clone for QLocale` and a `impl Drop for QLocale` implementation similar to other places (note we have templates for those too eg see how QString is done)

tests/qt_types_standalone/rust/src/qlocale.rs
```diff
@@ -0,0 +1,22 @@
...
+        include!("cxx-qt-lib-extras/qlocale.h");
+        type QLocale = cxx_qt_lib_extras::QLocale;
+    }
+
+    extern "Rust" {
+        fn construct_qlocale() -> QLocale;
+    }
+}
+
+fn construct_qlocale() -> QLocale {
```
ahayzen-kdab, line 20: once you have Clone implemented you could add a clone test

tests/qt_types_standalone/rust/src/qlocale.rs
```diff
@@ -0,0 +1,27 @@
+// SPDX-FileCopyrightText: 2024 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Nicolas Fella <nicolas.fella@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use cxx_qt_lib_extras::QLocale;
+
+#[cxx::bridge]
+mod qline_cxx {
```
nicolasfella, line 9: copypaste error

tests/qt_types_standalone/CMakeLists.txt
```diff
@@ -46,7 +46,7 @@ set(CRATE qt_types_standalone)
 cxx_qt_import_crate(
     MANIFEST_PATH rust/Cargo.toml
     CRATES ${CRATE}
-    QT_MODULES Qt::Core Qt::Gui Qt::Qml
+    QT_MODULES Qt::Core Qt::Gui Qt::Widgets Qt::Qml
```
ahayzen-kdab, line 49: Could try adding QuickControls2 here which might help with the order, as qt_quickcontrols is enabled in cxx-qt-lib. But this ordering issue is still painful when it happens :-/

LeonMatthesKDAB:
Closing due to inactivity, feel free to reopen when development restarts.

PR #1088, opened 2024-10-07
Disable workflow runs for pushes outside main

LeonMatthesKDAB:
This prevents dependabot/... branches from running their tests twice.
First as a push to the dependabot/... branch and then as the PR into
main.

### DIFF
--- a/.github/workflows/github-cxx-qt-tests.yml
+++ b/.github/workflows/github-cxx-qt-tests.yml
@@ -7,6 +7,9 @@
 name: cxx-qt tests
 on:
   push:
+    branches:
+      - 'main'
+      - '[0-9]+\.[0-9]+\.x'
   pull_request:
   schedule:
     # Run daily at 01:01


.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -7,6 +7,8 @@
 name: cxx-qt tests
 on:
   push:
+    branches:
+      - 'main'
```
ahayzen-kdab, line 11: do we also want it to run on stable branches? So `0.5.x` and `0.6.x` etc? 

We already run the book on a tag with `    tags: v[0-9]+\.[0-9]+\.[0-9]+` wonder if we can have like
```
branches:
   - 'main'
   - [0-9]+\.[0-9]+\.\x
```

LeonMatthesKDAB: Done

ahayzen-kdab: Assume it works :-)

ahayzen-kdab (approved):
LGTM :-)

PR #1098, opened 2024-10-14
Update CI runners to use aqt install

BenFordTytherington:


.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -234,7 +237,7 @@ jobs:
             compiler_cache_path: /home/runner/.cache/sccache
             clang_format_path: /home/runner/.local/bin/clang-format
             cargo_dir: ~/.cargo
-            packages-extra: >-
+            packages-extra: >- # Can all these packages go?
                 qtbase5-dev
                 qtdeclarative5-dev
                 qtquickcontrols2-5-dev
```
ahayzen-kdab, line 243: yes they can all go

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -255,7 +261,7 @@ jobs:
             compiler_cache_path: /home/runner/.cache/sccache
             clang_format_path: /home/runner/.local/bin/clang-format
             cargo_dir: ~/.cargo
-            packages-extra: >-
+            packages-extra: >- # Can all these packages go, apart from last 3?
```
ahayzen-kdab, line 264: yes

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -224,6 +224,9 @@ jobs:
           - name: Ubuntu 22.04 (gcc) Qt5
             os: ubuntu-22.04
             qt_version: 5
+            aqt_version: '5.15.2'
+            arch: 'gcc_64'
+            host: 'linux'
```
ahayzen-kdab, line 229: i would maybe prefix all of these with `aqt` so `aqt_host` and `aqt_arch` etc to avoid any confusion or leaking into setting other things

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -245,6 +248,9 @@ jobs:
           - name: Ubuntu 22.04 (gcc) Qt6
             os: ubuntu-22.04
             qt_version: 6
+            aqt_version: '6.7.3'
+            arch: 'gcc_64'
```
ahayzen-kdab, line 252: looks like this is now maybe `linux_gcc_64` for Qt 6 ? Qt 5 is still `gcc_64` according to https://ddalcino.github.io/aqt-list-server/

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -335,25 +328,20 @@ jobs:
...
         # but still leave room for the tools cache
         SCCACHE_CACHE_SIZE: 600M
 
     steps:
-    # On Windows, the D drive that the workspace is on by default runs out of space when
-    # vcpkg builds Qt, so the repo needs to be cloned to the C drive. GitHub's checkout
-    # action does not support cloning to a path outside GITHUB_WORKSPACE:
     # https://github.com/actions/checkout/issues/197
     - name: "Clone Git repository"
       uses: LebedevRI/checkout@issue197
```
ahayzen-kdab, line 347: change this to `actions/checkout@v4`

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -335,28 +328,23 @@ jobs:
...
         # With GitHub caches we have a 10 GB limit / 6 conditions = 1666 MB
         # Allow a larger cache size so that code in branches can be cached
         # but still leave room for the tools cache
         SCCACHE_CACHE_SIZE: 600M
 
     steps:
-    # On Windows, the D drive that the workspace is on by default runs out of space when
-    # vcpkg builds Qt, so the repo needs to be cloned to the C drive. GitHub's checkout
-    # action does not support cloning to a path outside GITHUB_WORKSPACE:
     # https://github.com/actions/checkout/issues/197
```
ahayzen-kdab, line 345: oops also remove this line linking to the github issue as it's not relevant anymore

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -449,40 +439,31 @@ jobs:
...
         cmake
         -D USE_QT5=${{ matrix.qt_version == 5 }}
         -D CMAKE_BUILD_TYPE=Release
-        -D VCPKG=${{ matrix.vcpkg }}
-        -D VCPKG_HOST_TRIPLET=${{ matrix.vcpkg_triplet }}
-        -D VCPKG_BUILDTREES_ROOT=${{ matrix.vcpkg_buildtrees_root  }}
-        -D VCPKG_TARGET_TRIPLET=${{ matrix.vcpkg_triplet }}
         -G Ninja
         -S . -B build
-      working-directory: ${{ matrix.workspace }}
```
LeonMatthesKDAB, line 467: Why has this been removed? Doesn't this still need to run in the workspace?

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -449,40 +439,31 @@ jobs:
...
       run: ctest ${{ matrix.ctest_args }} -C Release -T test --output-on-failure --parallel ${{ matrix.cores }}
-      working-directory: ${{ matrix.workspace }}/build
+      working-directory: ./build
       env:
         # Use the version of clang-format from pip
         CLANG_FORMAT_BINARY: ${{ matrix.clang_format_path }}
         RUSTC_WRAPPER: sccache
         QT_QPA_PLATFORM: ${{ matrix.qt_qpa_platform }}
         QT_SELECT: qt${{ matrix.qt_version }}
+        DYLD_FRAMEWORK_PATH: ${{ matrix.dyld_framework_path }}
```
LeonMatthesKDAB, line 466: This is needed so it can find the dynamic libraries from Qt on MacOS, right?
Is this still needed after the fixes by @ahayzen-kdab ?

.github/workflows/github-cxx-qt-tests.yml
ahayzen-kdab: Before we had to set a custom workspace as we needed to build vcpkg in a custom location otherwise we would run out of disk space. Now we can use the default folder where everything is normally checked out (note how we don't need the custom checkout task anymore) to and all tasks are by default executed in. (note that the WASM runner is already doing this).

So this was a workaround we had before for vcpkg.

LeonMatthesKDAB: Ah, nevermind, we don't need to specify the workspace explicitly anymore without vcpkg, because we're not running out of space in the default directory.
Makes sense.

.github/workflows/github-cxx-qt-tests.yml
ahayzen-kdab: Right, this is needed for `cargo test` to work in certain places where there is linking going on to dyn libs, but they are not going through CxxQtBuilder (which sets up the rpath). Think of this as effectively LD_LIBRARY_PATH.

LeonMatthesKDAB (commented):
Mostly looks good, some small comments/questions left :)

README.md
```diff
@@ -116,12 +116,8 @@ example does not link with GNU ld.bfd which is the default linker on most Linux
 installing [mold](https://github.com/rui314/mold), [lld](https://lld.llvm.org/), or GNU ld.gold
 (from GNU binutils but may be separate package) is required on Linux.
 
-On Windows and macOS vcpkg can be used by adding `-D VCPKG=ON` to the CMake configure step to
-automatically download release mode packages from GitHub Packages (this will take several minutes the first time you run CMake).
-Note that debug symbols are not built in these packages.
-
 CXX-Qt defaults to building with Qt6. If you want to build with Qt5 when both are installed,
-or you want to tell vcpkg to use Qt5, add `-D USE_QT5=ON` to the CMake configure step.
+or you want to tell aqt_install to use Qt5, add `-D USE_QT5=ON` to the CMake configure step.
```
ahayzen-kdab, line 120: i think this paragraph should be
```
CXX-Qt defaults to building with Qt6. If you want to build with Qt5 when both are installed add `-D USE_QT5=ON` to the CMake configure step.
```

BenFordTytherington: Fixed

ahayzen-kdab:
This looks good to me now :-)

.github/workflows/github-cxx-qt-tests.yml
LeonMatthesKDAB: Ah, so it's needed for the tests that use qt types, but use CXX builder, not CxxQtBuilder?
That makes sense.

LeonMatthesKDAB (approved):

PR #1099, opened 2024-10-15
Apple framework investigation

ahayzen-kdab:


ahayzen-kdab:
Landed as part of #1098

PR #1100, opened 2024-10-15
cxx-qt: use cxx-qt-build and reexport from cxx-qt-lib

ahayzen-kdab:
This removes a build depends of cxx-qt-build on cxx-qt

### DIFF
--- a/crates/cxx-qt/build.rs
+++ b/crates/cxx-qt/build.rs
@@ -3,60 +3,55 @@
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
-use std::{fs::File, io::Write};
+use std::path::PathBuf;
+
+use cxx_qt_build::{CxxQtBuilder, Interface};
+
+fn header_dir() -> PathBuf {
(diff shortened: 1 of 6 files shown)


crates/cxx-qt/build.rs
```diff
@@ -3,60 +3,55 @@
...
+}
 
 fn main() {
-    let qtbuild = qt_build_utils::QtBuild::new(vec!["Core".to_owned()])
-        .expect("Could not find Qt installation");
+    write_headers();
+
+    let interface = Interface::default()
+        .export_include_prefixes([])
+        .export_include_directory(header_dir(), "cxx-qt");
```
ahayzen-kdab, line 38: ah this should maybe be `rexport_dependency` ?

LeonMatthesKDAB (approved):
Mostly looks good.
One thing our call with Jonathan reminded me of is that the links key also enforces that there can only be one major version of cxx-qt (and cxx-qt-lib) within a given project.
That's not ideal, but may be something we have to live with for now...

PR #1101, opened 2024-10-15
cargo: use cargo:: instead of cargo: to match documentation

ahayzen-kdab:
We have been mixing eg cargo:rerun-if-changed and
cargo::rerun-if-changed. According to these docs it should be with a double colon.
https://doc.rust-lang.org/cargo/reference/build-scripts.html


BenFordTytherington (approved):
LGTM :)

BenFordTytherington (approved):
LGTM :)

Issue #1103, opened 2024-10-16
Version the links keys


ahayzen-kdab:
This could then allow for differing versions of our crates in the same project.

As raised in this comment https://github.com/KDAB/cxx-qt/pull/1100#pullrequestreview-2373087584

PR #1104, opened 2024-10-18
Add versions to link properties in manifest

BenFordTytherington:
Fix for #1103

### DIFF
--- a/crates/cxx-qt-lib-extras/Cargo.toml
+++ b/crates/cxx-qt-lib-extras/Cargo.toml
@@ -11,7 +11,7 @@ license.workspace = true
 description = "Extra Qt types for integrating `cxx-qt` crate with `cxx` that are not available in `cxx-qt-lib`"
 repository.workspace = true
 
-links = "cxx-qt-lib-extras"
+links = "cxx-qt-lib-extras-0-6-1"
 
 [dependencies]
 cxx.workspace = true
--- a/crates/cxx-qt-lib-extras/build.rs
+++ b/crates/cxx-qt-lib-extras/build.rs
@@ -46,7 +46,7 @@ fn main() {
         // Disable exporting the standard include directory, as we are exporting custom headers
         .export_include_prefixes([])
         .export_include_directory(header_dir(), "cxx-qt-lib-extras")
-        .reexport_dependency("cxx-qt-lib");
+        .reexport_dependency("cxx-qt-lib-0-6-1");
 
     let mut builder = CxxQtBuilder::library(interface)
         .qt_module("Gui")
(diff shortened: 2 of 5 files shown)


crates/cxx-qt-lib-extras/Cargo.toml
```diff
@@ -11,7 +11,7 @@ license.workspace = true
 description = "Extra Qt types for integrating `cxx-qt` crate with `cxx` that are not available in `cxx-qt-lib`"
 repository.workspace = true
 
-links = "cxx-qt-lib-extras"
+links = "cxx-qt-lib-extras-1-6-0"
```
ahayzen-kdab, line 14: we should follow semver order, so 0.6.1 becomes `-0-6-1`

crates/cxx-qt-lib-extras/Cargo.toml
```diff
@@ -11,7 +11,7 @@ license.workspace = true
 description = "Extra Qt types for integrating `cxx-qt` crate with `cxx` that are not available in `cxx-qt-lib`"
 repository.workspace = true
```
ahayzen-kdab, line 12: the build fails because i believe you need to update the reexport links. Search for 

`.reexport_dependency` and update those names in the `Interface` usages

crates/cxx-qt-lib-extras/build.rs
```diff
@@ -46,7 +46,7 @@ fn main() {
         // Disable exporting the standard include directory, as we are exporting custom headers
         .export_include_prefixes([])
         .export_include_directory(header_dir(), "cxx-qt-lib-extras")
-        .reexport_dependency("cxx-qt-lib");
+        .reexport_dependency("cxx-qt-lib-0-6-1");
```
LeonMatthesKDAB, line 49: Ah, this is a problem!
With Cargos dependency resolver, we may end up with cxx-qt-lib `0.6.2` as the actual dependency here, as it can choose another compatible version.
In which case, reexport_dependency would no longer work!

And we have the same issue with future minor versions, where Cargo may choose a newer version...
So maybe we only include the part of the version in the links key that actually includes a "breaking change".
E.g. `cxx-qt-lib-0.6`.
However, that will lead to issues if two downstream dependencies enforce exact minor versions that then will have conflicting dependencies...

The other route in which we could solve this issue is to use another name in the `reexport_dependency`.
I guess we could use the `name` from the manifest instead :thinking: 

Any opinions @BenFordTytherington @ahayzen-kdab ?

LeonMatthesKDAB (changes requested):
I think we may run into issues with `reexport_dependency` if multiple compatible minor versions exist...

crates/cxx-qt-lib-extras/build.rs
ahayzen-kdab: @LeonMatthesKDAB Hmm so maybe we reexplore this with the build system refactor in 0.8?

LeonMatthesKDAB: Yes, we still don't have a firm answer on what takes care of exporting dependencies anyway.
Maybe that will need to be its own build step...

PR #1108, opened 2024-10-21
Feature explicit case conversion

BenFordTytherington:
Add an attribute for extern blocks which specifies all the items inside can have their names automatically converted to the correct case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- `#[auto_case]` attribute for `extern` blocks, which will convert function names to the appropriate case
+- `#[auto_cxx_name]` and `#[auto_rust_name]` attributes for `extern` blocks, which will convert the case of names, automatically camelCase for cxx, and snake_case for rust
 - Support for further types: `QLine`, `QLineF`, `QImage`, `QPainter`, `QFont`, `QPen`, `QPolygon`, `QPolygonF`, `QRegion`, `QAnyStringView`
 - `internal_pointer_mut()` function on `QModelIndex`
 - `c_void` in CXX-Qt-lib for easy access to `void *`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `#[auto_case]` attribute for `extern` blocks, which will convert function names to the appropriate case
 - Support for further types: `QLine`, `QLineF`, `QImage`, `QPainter`, `QFont`, `QPen`, `QPolygon`, `QPolygonF`, `QRegion`, `QAnyStringView`
 - `internal_pointer_mut()` function on `QModelIndex`
 - `c_void` in CXX-Qt-lib for easy access to `void *`
--- a/book/src/bridge/attributes.md
+++ b/book/src/bridge/attributes.md
@@ -36,4 +36,10 @@ For [`#[qproperty]`](./extern_rustqt.md#properties), a CXX or Rust name can be p
 
 > **⚠️ Deprecation warning**:
 > CXX-Qt <0.6 did automatic case conversion if no `#[cxx_name = "..."]` or `#[rust_name = "..."]` is specified.
-> Starting with CXX-Qt 0.7, this is no longer the case!
+> Starting with CXX-Qt 0.7, this is no longer the case! Automatic case conversion will be opt-in instead.
+
+### Automatic case conversion
+
+The `#[auto_cxx_name]` and `#[auto_rust_name]` attributes can be used to automatically rename cxx and rust names.
+These are placed at a block level on `extern "RustQt"` or `extern "C++Qt"` blocks, and will automatically case convert the items inside, unless they specify either a `rust_name` or `cxx_name`.
+By default `#[auto_cxx_name]` will generate a camelCase conversion for`cxx_name` and `#[auto_rust_name]` will generate a snake_case conversion for `rust_name`.
--- a/book/src/bridge/extern_cppqt.md
+++ b/book/src/bridge/extern_cppqt.md
@@ -29,6 +29,8 @@ A bridge module may contain zero or more `extern "C++Qt"` blocks.
 This complements the [`extern "C++"` CXX section](https://cxx.rs/extern-c++.html)
 but allows for declaring Qt specific features on C++ types.
 
+Automatically converting to camel or snake case can be done through an [attribute](./attributes.md#automatic-case-conversion) at the block level.
+
 ## `QObject`s
 
 QObject types that are defined in C++ can be made available to Rust, by declaring them as [opaque types](https://cxx.rs/extern-c++.html#opaque-c-types) with a `#[qobject]` attribute.
--- a/book/src/bridge/extern_rustqt.md
+++ b/book/src/bridge/extern_rustqt.md
@@ -30,6 +30,8 @@ A bridge module may contain zero or more `extern "RustQt"` blocks.
 This complements the [`extern "Rust"` CXX section](https://cxx.rs/extern-rust.html)
 but allows for declaring Qt specific features on C++ types.
 
+Automatically converting to camel or snake case can be done through an [attribute](./attributes.md#automatic-case-conversion) at the block level.
+
 ## `QObject`s
 
 The `#[qobject]` attribute may be placed on a type alias to generate a [`QObject`](https://doc.qt.io/qt-6/qobject.html) type in C++.
--- a/crates/cxx-qt-gen/src/naming/name.rs
+++ b/crates/cxx-qt-gen/src/naming/name.rs
@@ -126,13 +126,11 @@ impl Name {
             })
             .transpose()?;
 
-        if let Some(case) = auto_case.cxx {
-            if cxx_name.is_none() {
+        if cxx_name.is_none() && rust_name.is_none() {
+            if let Some(case) = auto_case.cxx {
                 cxx_name = Some(ident.to_string().to_case(case));
             }
-        }
-        if let Some(case) = auto_case.rust {
-            if rust_name.is_none() {
+            if let Some(case) = auto_case.rust {
                 rust_name = Some(format_ident!("{}", ident.to_string().to_case(case)));
             }
         }
@@ -151,7 +149,7 @@ impl Name {
         let cxx_ident = if cxx.is_none() && rust.is_some() {
             Some(self.rust.to_string())
         } else {
-            cxx.clone()
+            cxx
         };
         Self {
             rust: rust.unwrap_or_else(|| self.rust.clone()),
@@ -218,7 +216,7 @@ impl Name {
     /// 2. Any attributes like cxx_name and namespace
     /// 3. The rust_qualified path to access the function (if not needed use _ during destructuring)
     pub fn into_cxx_parts(self) -> (Ident, Vec<Attribute>, Path) {
-        let rust_qualified = self.rust_qualified().clone();
+        let rust_qualified = self.rust_qualified();
         let cxx_name: Option<Attribute> = self.cxx.map(|cxx| {
             syn::parse_quote! { #[cxx_name = #cxx] }
         });
--- a/crates/cxx-qt-gen/src/naming/name.rs
+++ b/crates/cxx-qt-gen/src/naming/name.rs
@@ -3,9 +3,9 @@
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
-use crate::parser::AutoCase;
+use crate::parser::CaseConversion;
 use crate::syntax::{attribute::attribute_get_path, expr::expr_to_string};
-use convert_case::{Case, Casing};
+use convert_case::Casing;
 use quote::format_ident;
 use syn::{spanned::Spanned, Attribute, Error, Ident, Path, Result};
 
@@ -93,7 +93,7 @@ impl Name {
         attrs: &[Attribute],
         parent_namespace: Option<&str>,
         module: Option<&Ident>,
-        auto_case: AutoCase,
+        auto_case: CaseConversion,
     ) -> Result<Self> {
         // Find if there is a namespace (for C++ generation)
         let mut namespace = if let Some(attr) = attribute_get_path(attrs, &["namespace"]) {
@@ -116,7 +116,7 @@ impl Name {
             .transpose()?;
 
         // Find if there is a rust_name mapping
-        let rust_name = attribute_get_path(attrs, &["rust_name"])
+        let mut rust_name = attribute_get_path(attrs, &["rust_name"])
             .map(|attr| -> Result<_> {
                 Ok(format_ident!(
                     "{}",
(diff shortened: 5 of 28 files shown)


crates/cxx-qt-gen/src/naming/name.rs
```diff
@@ -123,6 +126,15 @@ impl Name {
             })
             .transpose()?;
 
+        match auto_case {
+            AutoCase::CamelCase => {
```
ahayzen-kdab, line 130: @LeonMatthesKDAB did we want to have camelCase and snake_case conversion depending on the block type?

eg a RustQt block you write from the perspective of Rust and we auto convert to C++. And a C++Qt block you write from the perspective of C++ and you auto convert to Rust

```rust
#[auto_case]
unsafe extern "RustQt" {
  fn function_name(); // ends up as functionName in C++
}

#[auto_case]
unsafe extern "C++Qt" {
  fn functionName(); // ends up as function_name in Rust?! (but this is means the rust name is magic?)
}
```

Or do we already want to introduce the type of the conversion, then leave it up to the developer to decide the direction ?

```rust
#[auto_case = Camel]
unsafe extern "RustQt" {
  fn function_name(); // ends up as functionName in C++
}

#[auto_case = Snake]
unsafe extern "C++Qt" {
  fn functionName(); // ends up as function_name in C++
}
```

And when do we do the auto conversion ? Do we do it when we see there are no custom cxx_name and rust_name ?

How do we know where to apply it to ? Do we always do it to the `cxx_name` and never the rust name ?


Or is the whole thing just `#[auto_case]` which just converts any Rust name to a camel case name in C++ ? And never changes the Rust name ? :thinking:

BenFordTytherington: My plan was going to be to include both, so like you said camel for rust blocks, and snake for c++ blocks, but then also allowing the developer to specify if they wanted, so that auto_case would be a shorthand.

ahayzen-kdab: Right, i just wonder if any of this is confusing if we start modifying effectively the rust_name, or maybe that is fine :thinking: 

As then the question becomes

```rust
#[auto_camel]
unsafe extern "RustQt" {
   fn my_function1();  // becomes cxx_name = myFunction1
}
#[auto_camel]
unsafe extern "C++Qt" {
   fn myFunction2();  // becomes rust_name my_function2
}
```

But then this is curious, if you specify the Camel / Snake as opposites.
```rust
#[auto_camel = Snake]
unsafe extern "RustQt" {
   fn my_function1();
   fn myFunction3();  // becomes cxx_name = my_function3
}
#[auto_camel = Camel]
unsafe extern "C++Qt" {
   fn myFunction2();
   fn my_Function4();  // becomes rust_name = myFunction4
}
```

ahayzen-kdab: Another route to avoid the confusion over whether the "auto" affects the cxx or rust name is to encode that into the attribute. So you could have `auto_cxx_case` and `auto_rust_case`.

Then this at least makes sense with what it does and gives you full flexibility
```rust
#[auto_cxx_case = Snake]
unsafe extern "RustQt" {
   fn my_function1();
   fn myFunction3();  // becomes cxx_name = my_function3
}
#[auto_rust_case = Camel]
unsafe extern "C++Qt" {
   fn myFunction2();
   fn my_Function4();  // becomes rust_name = myFunction4
}
```

ahayzen-kdab: So then we could start with `#[auto_cxx_case]` and `#[auto_rust_case]` which default to Camel and Snake for now. If we need to pick a non-default case later we can still do this without breaking API.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `#[auto_case]` attribute for `extern` blocks, which will convert function names to the appropriate case
```
ahayzen-kdab, line 22: needs to change to `auto_cxx_case` / `auto_rust_case`

crates/cxx-qt-gen/src/naming/name.rs
```diff
@@ -123,6 +126,17 @@ impl Name {
...
+        if let Some(case) = auto_case.cxx {
+            if cxx_name.is_none() {
+                cxx_name = Some(ident.to_string().to_case(case));
+            }
+        }
+        if let Some(case) = auto_case.rust {
+            if rust_name.is_none() {
+                rust_name = Some(format_ident!("{}", ident.to_string().to_case(case)));
+            }
+        }
```
ahayzen-kdab, line 138: Do we only auto case if there is no cxx name *and* no rust name in all cases?

Or does it apply when only the name we are affecting is missing ? :thinking:

crates/cxx-qt-gen/src/parser/mod.rs
```diff
@@ -21,15 +21,71 @@ use crate::{
...
+    pub cxx: Option<Case>,
+    pub rust: Option<Case>,
+}
+
+/// Used to match the auto_case attributes and turn it into a Case to convert to
+fn meta_to_case(attr: &Attribute, default: Case) -> Result<Case> {
+    match &attr.meta {
+        Meta::Path(_) => Ok(default),
+        Meta::NameValue(case) => match &case.value {
+            Expr::Path(expr_path) => match expr_path.path.require_ident()?.to_string().as_str() {
```
ahayzen-kdab, line 45: You might be able to remove the `.to_string().as_str()`, as `Ident` can be compared with a str in certain places? Unsure if that works here?

```
impl<T> PartialEq<T> for Ident
where
    T: AsRef<str> + ?Sized,
```

crates/cxx-qt-gen/src/parser/mod.rs
```diff
@@ -21,15 +21,71 @@ use crate::{
...
+    /// Parses both `auto_cxx_case` and `auto_cxx_case = Camel`
+    pub fn from_attrs(attrs: &BTreeMap<&str, &Attribute>) -> Result<Self> {
+        let rust = attrs
+            .get("auto_rust_case")
+            .map(|attr| meta_to_case(attr, Case::Snake))
+            .transpose()?;
+        let cxx = attrs
+            .get("auto_cxx_case")
+            .map(|attr| meta_to_case(attr, Case::Camel))
+            .transpose()?;
```
ahayzen-kdab, line 83: spoke with Leon, can we change the name of the attribute to `auto_rust_name` and `auto_cxx_name` ? As it's then clearer that this affects the `rust_name` and `cxx_name`

crates/cxx-qt-gen/src/naming/name.rs
ahayzen-kdab: Spoke with Leon, lets go with only when both are none for now

crates/cxx-qt-gen/src/parser/mod.rs

crates/cxx-qt-gen/src/naming/name.rs
BenFordTytherington: Done

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
```
ahayzen-kdab, line 20: looks good as discussed an example and docs in book would be good

book/src/bridge/extern_rustqt.md
```diff
@@ -30,6 +30,13 @@ A bridge module may contain zero or more `extern "RustQt"` blocks.
 This complements the [`extern "Rust"` CXX section](https://cxx.rs/extern-rust.html)
 but allows for declaring Qt specific features on C++ types.
 
+## Automatic case conversion
```
ahayzen-kdab, line 33: this is not specific to the `RustQt` block and should work on the `C++Qt` block, maybe merge this text with the `attributes` text and link to that header from somewhere in here?

examples/qml_features/tests/tst_naming.qml
```diff
@@ -24,5 +24,6 @@ TestCase {
         compare(obj.numberProp, 0);
         obj.increment();
         compare(obj.numberProp, 1);
+        compare(obj.getGreeting(), 42)
```
ahayzen-kdab, line 27: shouldn't this be `getNum` ?

book/src/bridge/extern_rustqt.md
```diff
@@ -30,6 +30,13 @@ A bridge module may contain zero or more `extern "RustQt"` blocks.
...
+
+Automatically converting to camel or snake case can be done through the `#[auto_cxx_name]` and `#[auto_rust_name]` attributes.
+These attributes are placed on an `extern` block, and apply to all elements inside it, provided they don't specify rust or cxx names.
+`#[auto_cxx_name]` will convert the rust identifier to camelCase and use that for the cxx name.
+`#[auto_rust_name]` will convert the rust identifier to snake_case and use that for the rust name.
```
ahayzen-kdab, line 38: then as we are merging sections, do you mention that you can do `#[auto_cxx_name = Snake]` as well i wonder, which could allow for further cases in the future. Or do we ignore that ? :thinking:

book/src/bridge/attributes.md
```diff
@@ -36,4 +36,9 @@ For [`#[qproperty]`](./extern_rustqt.md#properties), a CXX or Rust name can be p
 
 > **⚠️ Deprecation warning**:
 > CXX-Qt <0.6 did automatic case conversion if no `#[cxx_name = "..."]` or `#[rust_name = "..."]` is specified.
-> Starting with CXX-Qt 0.7, this is no longer the case!
+> Starting with CXX-Qt 0.7, this is no longer the case! Automatic case conversion will be opt-in instead.
+
+### Automatic case conversion
+The `#[auto_cxx_name]` and `#[auto_rust_name]` attributes can be used to automatically rename cxx and rust names.
+These are placed at a block level on extern blocks, and will automatically case convert the items inside, unless they specify either rust or cxx name.
```
ahayzen-kdab, line 43: might need to note it only works on `extern "C++Qt"` / `extern "RustQt"` blocks?

book/src/bridge/extern_rustqt.md
BenFordTytherington: Support is there, but it's kinda superfluous for now I feel so I left it out. I just can't see any case where you would need to convert to the other case than is done by default, but leon suggested leaving that door open so I implemented it for future use.

book/src/bridge/extern_cppqt.md
```diff
@@ -29,6 +29,10 @@ A bridge module may contain zero or more `extern "C++Qt"` blocks.
 This complements the [`extern "C++"` CXX section](https://cxx.rs/extern-c++.html)
 but allows for declaring Qt specific features on C++ types.
 
+## Automatic case conversion
+
+Automatically converting to camel or snake case can be done through an attribute for extern "RustQt" and "C++Qt" blocks, see [here](./attributes.md#automatic-case-conversion)
```
ahayzen-kdab, line 34: Would change this to the following and remove the header, so it's just a sentence in the previous block

```
Automatically converting to camel or snake case can be enabled through an [attribute](./attributes.md#automatic-case-conversion).
```

book/src/bridge/extern_rustqt.md
```diff
@@ -30,6 +30,10 @@ A bridge module may contain zero or more `extern "RustQt"` blocks.
 This complements the [`extern "Rust"` CXX section](https://cxx.rs/extern-rust.html)
 but allows for declaring Qt specific features on C++ types.
 
+## Automatic case conversion
+
+Automatically converting to camel or snake case can be done through an attribute for "RustQt" and "C++Qt" blocks, see [here](./attributes.md#automatic-case-conversion)
```
ahayzen-kdab, line 35: Would change this to the following and remove the header, so it's just a sentence in the previous block

```
Automatically converting to camel or snake case can be enabled through an [attribute](./attributes.md#automatic-case-conversion).
```

book/src/bridge/attributes.md
ahayzen-kdab: Would move the part about it only working on `extern "C++Qt"` and `extern "RustQt"` blocks to this text

book/src/bridge/extern_rustqt.md
ahayzen-kdab: Right, i need `ScreamingSnake` everywhere :-D

book/src/bridge/attributes.md
```diff
@@ -36,4 +36,10 @@ For [`#[qproperty]`](./extern_rustqt.md#properties), a CXX or Rust name can be p
 
 > **⚠️ Deprecation warning**:
 > CXX-Qt <0.6 did automatic case conversion if no `#[cxx_name = "..."]` or `#[rust_name = "..."]` is specified.
-> Starting with CXX-Qt 0.7, this is no longer the case!
+> Starting with CXX-Qt 0.7, this is no longer the case! Automatic case conversion will be opt-in instead.
+
+### Automatic case conversion
+
+The `#[auto_cxx_name]` and `#[auto_rust_name]` attributes can be used to automatically rename cxx and rust names.
+These are placed at a block level on extern "RustQt" or "C++Qt" blocks, and will automatically case convert the items inside, unless they specify either rust or cxx name.
```
ahayzen-kdab, line 44: I would write the block names fully in code blocks so
```
These are placed at a block level on `extern "RustQt"` or `extern "C++Qt"` blocks
```

Then I would change the last part to
```
unless they specify either a `rust_name` or `cxx_name`.
```

book/src/bridge/extern_cppqt.md
```diff
@@ -29,6 +29,8 @@ A bridge module may contain zero or more `extern "C++Qt"` blocks.
 This complements the [`extern "C++"` CXX section](https://cxx.rs/extern-c++.html)
 but allows for declaring Qt specific features on C++ types.
 
+Automatically converting to camel or snake case can be done through an [attribute](./attributes.md#automatic-case-conversion) for "RustQt" and "C++Qt" blocks.
```
ahayzen-kdab, line 32: I would remove the `for "RustQt" and "C++Qt" blocks` as this page is specifically for `C++Qt` blocks anyway and then the attribute page explains which blocks this works for.

book/src/bridge/extern_rustqt.md
```diff
@@ -30,6 +30,8 @@ A bridge module may contain zero or more `extern "RustQt"` blocks.
 This complements the [`extern "Rust"` CXX section](https://cxx.rs/extern-rust.html)
 but allows for declaring Qt specific features on C++ types.
 
+Automatically converting to camel or snake case can be done through an [attribute](./attributes.md#automatic-case-conversion) for "RustQt" and "C++Qt" blocks.
```
ahayzen-kdab, line 33: I would remove the `for "RustQt" and "C++Qt" blocks` as this page is specifically for `RustQt` blocks anyway and then the attribute page explains which blocks this works for.

examples/qml_features/tests/tst_naming.qml
```diff
@@ -24,5 +24,6 @@ TestCase {
         compare(obj.numberProp, 0);
         obj.increment();
         compare(obj.numberProp, 1);
+        compare(obj.getNum(), 42)
```
ahayzen-kdab, line 27: while editing things add a semi-colon at the end of the line :-D

ahayzen-kdab (approved):
LGTM :-)

PR #1110, opened 2024-10-25
Added MSRV info to crates

BenFordTytherington:
Adding an MSRV to crates, in order to enforce an optional clippy lint which prevents newer features breaking things.

### DIFF
--- a/.github/workflows/github-cxx-qt-tests.yml
+++ b/.github/workflows/github-cxx-qt-tests.yml
@@ -42,6 +42,7 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
+      - run: rustup default 1.77.2
       - run: rustup component add rustfmt
       - run: cargo fmt --all --check --verbose
 
@@ -72,8 +73,9 @@ jobs:
         uses: actions/checkout@v4
       - name: Setup toolchain
         run: |
+          rustup default 1.77.2
           cargo install grcov
-          rustup component add llvm-tools
+          rustup component add llvm-tools rustfmt
       - name: build
         env:
           RUSTFLAGS: -Cinstrument-coverage
@@ -106,7 +108,8 @@ jobs:
         uses: actions/checkout@v4
       - name: Setup toolchain
         run: |
-          rustup component add clippy rustfmt
+          rustup default 1.77.2
+          rustup component add rustfmt
           rustup target add wasm32-unknown-emscripten
       - name: "Rust tools cache"
         uses: actions/cache@v4
@@ -116,13 +119,14 @@ jobs:
             ~/.cargo/bin/sccache
             ~/.cargo/bin/mdbook
             ~/.cargo/bin/mdbook-linkcheck
-          key: ubuntu-24.04_sccache-0.7.6_mdbook-0.4.36_mdbook-linkcheck-0.7.7
+          key: ubuntu-24.04_sccache-0.8.2_mdbook-0.4.36_mdbook-linkcheck-0.7.7
       - name: "Build Rust tools"
         if: steps.rust-tools-cache.outputs.cache-hit != 'true'
         # Do not build with storage backends enabled, we only need local
         run: |
-          cargo install --no-default-features sccache
-          cargo install mdbook mdbook-linkcheck
+          cargo install --no-default-features --version 0.8.2 sccache
+          cargo install --version 0.4.36 mdbook 
+          cargo install --version 0.7.7 mdbook-linkcheck
       # We want our compiler cache to always update to the newest state.
       # The best way for us to achieve this is to **always** update the cache after every landed commit.
       # That way it will closely follow our development.
@@ -366,10 +370,14 @@ jobs:
     - name: "Clone Git repository"
       uses: actions/checkout@v4
     # Ensure clippy and rustfmt is installed, they should come from github runner
+    # clippy version needs to be 1.78.0 for the MSRV lint
     #
     # Note we still need rustfmt for the cxx-qt-gen tests
     - name: "Install Rust toolchain"
-      run: rustup component add clippy rustfmt
+      run: |
+        rustup toolchain add 1.78.0 --component clippy
+        rustup default 1.77.2
+        rustup component add rustfmt
 
     - name: "Rust tools cache"
       uses: actions/cache@v4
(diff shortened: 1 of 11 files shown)


ahayzen-kdab:
> MSRVs were found using https://github.com/foresterre/cargo-msrv, but when trying to run cargo test on the project-wide MSRV, it failed. I ran cargo test on 1.77.2 which is the highest MSRV for any of the crates, and it seemed that any rust version before 1.80.0 broke the generation tests. The diff showed that all the output simply didn't happen, so all tests passed except for the generation ones. This may be an issue with my locally installed toolchains however, but worth exploring.

For me the generation tests work with 1.77.2, i suspect maybe you didn't have rustfmt installed with that toolchain enabled? As note that we call out to that. Or something similar ?

crates/cxx-qt-build/Cargo.toml
```diff
@@ -11,6 +11,7 @@ edition.workspace = true
 license.workspace = true
 description = "Helpers for integrating `cxx-qt` crate into a Cargo build"
 repository.workspace = true
+rust-version = "1.70.0"
```
ahayzen-kdab, line 14: can you add a `clippy.toml` in the root folder.

In this you could specify the MSRV in clippy so that it doesn't warn for new clippy lints that aren't "available" yet
https://doc.rust-lang.org/stable/clippy/configuration.html#specifying-the-minimum-supported-rust-version

And as discussed before enable the `incompatible_msrv` lint too https://rust-lang.github.io/rust-clippy/master/#incompatible_msrv

Cargo.toml
```diff
@@ -55,3 +56,6 @@ syn = { version = "2.0", features = ["extra-traits", "full"] }
 quote = "1.0"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+
+[workspace.lints.clippy]
+incompatible_msrv = "deny"
```
ahayzen-kdab, line 61: might need

```
[lints]
workspace = true
```
in other cargo tomls?

Cargo.toml
```diff
@@ -22,6 +22,7 @@ members = [
     "tests/qt_types_standalone/rust",
 ]
 resolver = "2"
+msrv = "1.77.2"
```
ahayzen-kdab, line 25: not sure that this is used in the cargo.toml? this would be rust-version ?

Then we need to decide if we have one version across all our crates and inherit it or whether we have a minimum per crate

Cargo.toml
BenFordTytherington: Added that I'm pretty sure

crates/cxx-qt-build/Cargo.toml
ahayzen-kdab, line 14: should also use `.workspace = true`

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -361,7 +361,9 @@ jobs:
     #
     # Note we still need rustfmt for the cxx-qt-gen tests
     - name: "Install Rust toolchain"
-      run: rustup component add clippy rustfmt
+      run: |
+        rustup default 1.77.2
```
ahayzen-kdab, line 365: you'll need to do this for other tasks like coverage / build_wasm to be consistent, but lets see what CI thinks of this so far

Cargo.toml
```diff
@@ -28,6 +28,7 @@ edition = "2021"
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/KDAB/cxx-qt/"
 version = "0.6.1"
+rust-version = "1.77.2" # Clippy needs to >= 1.80.0 for a lint, but the package can be 1.77.2
```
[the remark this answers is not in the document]
BenFordTytherington: oops :)

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -376,8 +378,8 @@ jobs:
       if: steps.rust-tools-cache.outputs.cache-hit != 'true'
```
ahayzen-kdab, line 378: you'll need to update the cache line above to "break" the cache, it should become

```
key: ${{ matrix.os }}_sccache-0.7.7_mdbook-0.4.36_mdbook-linkcheck-0.7.7
```

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -376,8 +378,8 @@ jobs:
...
       # Do not build with storage backends enabled, we only need local
       run: |
-        cargo install --no-default-features sccache
-        cargo install mdbook mdbook-linkcheck
+        cargo install --no-default-features --version 0.7.7 sccache
+        cargo install --version 0.4.36 mdbook mdbook-linkcheck
```
ahayzen-kdab, line 382: `mdbook` and `mdbook-linkcheck` have different versions they need to be on separate lines
```
cargo install --version 0.4.36 mdbook
cargo install --version 0.7.7 mdbook-linkcheck
```

ahayzen-kdab:
We appear to be hitting https://github.com/mozilla/sccache/issues/2092 on macOS as we have reached the maximum cache size and it cannot remove old entries. Will need to debug if there is a way around this :-/

clippy.toml
```diff
@@ -0,0 +1,5 @@
+# SPDX-FileCopyrightText: 2024 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+# SPDX-FileContributor: Ben Ford <ben.ford@kdab.com>
+#
+# SPDX-License-Identifier: MIT OR Apache-2.0
+msrv = "1.78.0"
```
ahayzen-kdab, line 5: maybe add a comment why this is higher than the other rust-version (because the clippy lint wasn't introduced until this version etc)

ahayzen-kdab (approved):
LGTM :-)

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -106,6 +108,8 @@ jobs:
         uses: actions/checkout@v4
       - name: Setup toolchain
         run: |
+          rustup toolchain add 1.78.0
```
ahayzen-kdab, line 111: lets see how long this takes but it's unused on windows/macos

ahayzen-kdab: actually it's unused here build-wasm doesn't run clippy?

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -369,7 +374,10 @@ jobs:
     #
     # Note we still need rustfmt for the cxx-qt-gen tests
     - name: "Install Rust toolchain"
-      run: rustup component add clippy rustfmt
+      run: |
+        rustup toolchain add 1.78.0
```
ahayzen-kdab, line 378: lets see how long this takes but it's unused on windows/macos

ahayzen-kdab: It's pretty fast...

ahayzen-kdab (approved):
LGTM :-)

PR #1113, opened 2024-10-30
cxx-qt-gen: ensure that qproperties follow auto casing too

ahayzen-kdab:


crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -242,16 +249,60 @@ mod tests {
...
+        // rust_name to camel
+        {
+            let auto_case = CaseConversion {
+                cxx: None,
+                rust: Some(Case::Camel),
+            };
+            let property = mock_property_case(input, auto_case);
+            assert_eq!(property.name.cxx_unqualified(), "my_name");
+            assert_eq!(property.name.rust_unqualified(), "myName");
+        }
```
LeonMatthesKDAB, line 299: Add a test case for both auto_cxx_name and auto_rust_name here?
Or do we currently not support this?

LeonMatthesKDAB (commented):
Mostly looks good, do we want to test `auto_cxx_name` + `auto_rust_name`?
Otherwise can be approved from my side :)

crates/cxx-qt-gen/src/parser/property.rs
ahayzen-kdab: I'll add a test for this

PR #1114, opened 2024-10-30
book: Add missing docs for QT_MODULES argument

LeonMatthesKDAB:


crates/qt-build-utils/src/lib.rs
```diff
@@ -468,7 +468,7 @@ impl QtBuild {
 
             // Ensure that any framework paths are set to -F
             for framework_path in self.framework_paths() {
-                builder.flag_if_supported(format!("-F{}", framework_path.display()));
+                builder.flag_if_supported(&format!("-F{}", framework_path.display()));
```
ahayzen-kdab, line 471: can you check this is really needed, as i've been building fine on 1.77.2 and on a recent stable rustc :thinking:

ahayzen-kdab: As discussed can you bump the version of `cc` in our root cargo.toml to 1.0.100 (and check that works), as looks like that changed here https://github.com/rust-lang/cc-rs/commit/a1e291a4f5bb7ef7f75bb6606ad2149827721946

ahayzen-kdab (approved):
LGTM :-)

PR #1116, opened 2024-10-30
cargo: ensure version is set for all crates

ahayzen-kdab:
Otherwise you get

```
    Updating crates.io index
error: all dependencies must have a version specified when publishing.
dependency `cxx-qt` does not specify a version
Note: The published dependency will use the version from crates.io,
the `path` specification will be removed from the dependency declaration.
```


BenFordTytherington (approved):
LGTM

PR #1122, opened 2024-11-15
cxx-qt-gen: allow for unused unsafe blocks in the CXX block

ahayzen-kdab:
Otherwise we get build warnings as CXX has removed this attribute with changes in CXX 1.0.130


ahayzen-kdab:
Instead using this route #1131 .

PR #1124, opened 2024-11-15
cxx-qt-gen: allow for unused unsafe blocks and resolve no-exception issues

ahayzen-kdab:


ahayzen-kdab:
Instead using this route #1131 .

PR #1127, opened 2024-11-17
Add from(&String) to QAnyStringView

nicolasfella:
Makes constructing a QAnyStringView from String slightly less verbose


LeonMatthesKDAB:
@nicolasfella I'm all for making this convenient.
However, I'm wondering why the `From<&str>` implementation isn't sufficient...
String should deref to `&str`, so in theory `QAnyStringView::from(&my_string)` should do the trick.
And I assume the same works for `my_string.into()`.

Can you share the exact code that this get's rid off.

redstrate:
@nicolasfella ping?

LeonMatthesKDAB:
Closing due to inactivity, feel free to reopen when development restarts.

Issue #1120, opened 2024-11-08
weird symlinking error on windows?


CelestialCrafter:
hi! im getting an error while building my project on windows, but it builds fine on linux.

```
C:\Users\Celestial\Documents\fenlu\nodes\sink-frontend>cargo build
   Compiling cxx-qt v0.7.0
error: failed to run custom build command for `cxx-qt v0.7.0`

Caused by:
  process didn't exit successfully: `C:\Users\Celestial\Documents\fenlu\nodes\sink-frontend\target\debug\build\cxx-qt-d42316ea98c17eb5\build-script-build` (exit code: 101)
  --- stdout
  cargo::rerun-if-changed=include/
  cargo::rerun-if-changed=include/connection.h
  cargo::rerun-if-changed=include/signalhandler.h
  cargo::rerun-if-changed=include/thread.h
  cargo::rerun-if-changed=include/threading.h
  cargo::rerun-if-changed=include/type.h
  cargo::rerun-if-changed=src/connection.rs
  cargo::rerun-if-changed=src/connection.cpp

  --- stderr
  thread 'main' panicked at C:\Users\Celestial\.cargo\registry\src\index.crates.io-6f17d22bba15001f\cxx-qt-build-0.7.0\src\lib.rs:1026:25:
  Failed to create symlink `cxx-qt` for export_include_directory: C:\Users\Celestial\Documents\fenlu\nodes\sink-frontend\target\debug\build\cxx-qt-2f7fe7d391714b2a\out\include\cxx-qt
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

the directory exists, and the error isnt that detailed, so i dont really know what to do.
here's my build.rs:

```rust
use cxx_qt_build::{CxxQtBuilder, QmlModule};

fn main() {
    CxxQtBuilder::new()
        .qt_module("Network")
        .qml_module(QmlModule {
            uri: "sinkfrontend",
            rust_files: &["src/qt/media.rs", "src/qt/actions.rs",],
            qml_files: &["qml/main.qml", "qml/Media.qml", "qml/MediaBackground.qml", "qml/MediaDetails.qml", "qml/CustomScrollGridView.qml", "qml/ActionsContextMenu.qml"],
            ..Default::default()
        })
        .build();
}
```

thank you!

CelestialCrafter:
contents of directory:
```
[...]

PR #1136, opened 2024-11-27
On non-Unix platforms, use deep copying rather than symlinking

jnbooth:
Addresses #1120. cxx-qt-build's current symlinking works on Unix, fails on Windows unless Developer Mode is enabled, and panics on anything else. This PR modifies the non-Unix behavior to use deep file copying rather than symlinking.

I have tested this on Windows 11 and MacOS Sequoia 15.1.1.

**Note:** I'm not sure if this PR means these lines should be removed:

https://github.com/KDAB/cxx-qt/blob/f5afefecf837e199848b99d3317a623a64767c5a/README.md?plain=1#L87-L93

### DIFF
--- a/crates/cxx-qt-build/src/dir.rs
+++ b/crates/cxx-qt-build/src/dir.rs
@@ -8,7 +8,7 @@
 use crate::{crate_name, module_name_from_uri};
 use std::io::Result;
 use std::{
-    env,
+    env, fs,
     path::{Path, PathBuf},
 };
 
@@ -80,4 +80,51 @@ pub(crate) fn out() -> PathBuf {
 
 pub(crate) fn is_exporting() -> bool {
     export().is_some()
+}
+
+/// Two dependencies may be reexporting the same shared dependency, which will
+/// result in conflicting symlinks.
+/// Try detecting this by resolving the symlinks and checking whether this leads us
+/// to the same paths. If so, it's the same include path for the same prefix, which
+/// is fine.
+#[cfg(unix)]
+pub(crate) fn symlinks_conflict(
+    symlink: impl AsRef<Path>,
+    target: impl AsRef<Path>,
+) -> Result<bool> {
+    if !symlink.as_ref().exists() {
+        return Ok(false);
+    }
+    let symlink = fs::canonicalize(symlink)?;
+    let target = fs::canonicalize(target)?;
+    Ok(symlink == target)
+}
+
+pub(crate) fn symlink_or_copy_directory(
+    source: impl AsRef<Path>,
+    dest: impl AsRef<Path>,
+) -> Result<()> {
+    #[cfg(unix)]
+    let result = std::os::unix::fs::symlink(source, dest);
+
+    #[cfg(not(unix))]
+    let result = copy_dir_all(source.as_ref(), dest.as_ref());
+
+    result
+}
+
+#[cfg(not(unix))]
+fn copy_dir_all(source: &Path, dest: &Path) -> Result<()> {
+    fs::create_dir_all(dest)?;
+    for entry in fs::read_dir(source)? {
+        let entry = entry?;
+        let target_path = entry.path();
+        let link_path = dest.join(entry.file_name());
+        if entry.file_type()?.is_dir() {
+            copy_dir_all(&target_path, &link_path)?;
+        } else {
+            fs::copy(&target_path, &link_path)?;
+        }
+    }
+    Ok(())
 }
--- a/crates/cxx-qt-build/src/dir.rs
+++ b/crates/cxx-qt-build/src/dir.rs
@@ -12,6 +12,13 @@ use std::{
     path::{Path, PathBuf},
 };
 
+/// On Unix platforms, included files are symlinked into destination folders.
+/// On non-Unix platforms, due to poor support for symlinking, included files are deep copied.
+#[cfg(unix)]
+pub const INCLUDE_VERB: &str = "create symlink";
+#[cfg(not(unix))]
+pub const INCLUDE_VERB: &str = "deep copy files";
+
 // Clean a directory by removing it and recreating it.
 pub(crate) fn clean(path: impl AsRef<Path>) -> Result<()> {
     let result = std::fs::remove_dir_all(&path);
--- a/crates/cxx-qt-build/src/dir.rs
+++ b/crates/cxx-qt-build/src/dir.rs
@@ -136,6 +136,7 @@ fn deep_copy_directory(source: &Path, dest: &Path) -> Result<bool> {
     Ok(true)
 }
 
+#[cfg(not(unix))]
 fn files_conflict(source: &Path, dest: &Path) -> Result<bool> {
     use fs::File;
     use std::io::{BufRead, BufReader};
--- a/crates/cxx-qt-build/src/dir.rs
+++ b/crates/cxx-qt-build/src/dir.rs
@@ -136,25 +136,27 @@ fn deep_copy_directory(source: &Path, dest: &Path) -> Result<bool> {
     Ok(true)
 }
 
-#[cfg(not(unix))]
 fn files_conflict(source: &Path, dest: &Path) -> Result<bool> {
     use fs::File;
-    use std::io::Read;
-    let mut source = File::open(source)?;
-    let mut dest = File::open(dest)?;
+    use std::io::{BufRead, BufReader};
+    let source = File::open(source)?;
+    let dest = File::open(dest)?;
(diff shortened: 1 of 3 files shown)


LeonMatthesKDAB:
@jnbooth thank you for working on this.

Unfortunately the `symlinks_conflict` function won't work when a directory is copied, which is what CI is complaining about.
Basically, multiple dependencies may re-export the headers from e.g. `cxx-qt-lib`.
When we combine these dependencies in the main application, there should only be a single set of headers for `cxx-qt-lib`.
With the previous symlinks, this was rather simple to check, as in the end after resolving the symlinks they paths all point to the same headers that were initially created by cxx-qt-lib.

At the moment, I see two ways to fix this:
1. When re-exporting dependencies, propagate the path to the original dependencies header files. Then when detecting the dependencies in cxx-qt-build, we need to store that original path and resolve conflicts that way.
2. Adapt symlinks_conflict to recursively diff the directory and check the actual files inside on non-unix systems.

While option 1 is faster, option 2 is simpler, so option 2 is probably the better way forward at the moment.

There's also the option to remove the check entirely, but I'm a bit hesitant to do that, as accessing the wrong headers from a duplicate dependency sounds like a really bad time and super hard to detect without this check...

jnbooth:
@LeonMatthesKDAB Thank you for taking a look at this so quickly! I'm taking most of today off for Thanksgiving, but I'll get to work on implementing those changes tomorrow.

jnbooth:
@LeonMatthesKDAB  Ready for re-review. Here's how I handled the diffing:

https://github.com/KDAB/cxx-qt/blob/331829f88a9148e3c0db5894c305645b263012c0/crates/cxx-qt-build/src/dir.rs#L138-L154

I think that should be fairly efficient. Using 2*1024 bytes of buffer seems safe enough. I noticed that [in std](https://doc.rust-lang.org/src/std/sys_common/io.rs.html), the default buffer size for BufReader is normally 8*1024 but set to only 512 for "bare metal" platforms, but I think those probably aren't relevant to cxx-qt.

crates/cxx-qt-build/src/dir.rs
```diff
@@ -81,3 +81,75 @@ pub(crate) fn out() -> PathBuf {
...
+    loop {
+        let source_n = source.read(&mut source_buf)?;
+        let dest_n = dest.read(&mut dest_buf)?;
+        if source_n == 0 && dest_n == 0 {
+            return Ok(false);
+        }
+        if &source_buf[..source_n] != &dest_buf[..dest_n] {
+            return Ok(true);
+        }
+    }
```
LeonMatthesKDAB, line 154: This code feels quite C-esque, with the indexing and manual buffer management.

I prefer to see Rusts `ReadBuffer` and `Bytes` iterator here, which should do pretty much the same but express the intent more clearly.
See also the example code here: https://users.rust-lang.org/t/efficient-way-of-checking-if-two-files-have-the-same-content/74735

Please refactor this to simply iterate and compare the zipped bytes.
The check for metadata path is still useful to keep though, as that offers a fast path to failure.

crates/cxx-qt-build/src/lib.rs
```diff
@@ -46,6 +46,13 @@ use cxx_qt_gen::{
     GeneratedRustBlocks, Parser,
 };
 
+// On Unix platforms, included files are symlinked into destination folders.
+// On non-Unix platforms, due to poor support for symlinking, included files are deep copied.
+#[cfg(unix)]
+const INCLUDE_VERB: &str = "create symlink";
+#[cfg(not(unix))]
+const INCLUDE_VERB: &str = "deep copy files";
```
LeonMatthesKDAB, line 54: nice way of expressing this :+1: 

However, a small nitpick: Please move this constant into the `dir` module, as it's tightly coupled with the behavior of the `symlink_or_copy_directory` function.

crates/cxx-qt-build/src/lib.rs
```diff
@@ -630,49 +637,28 @@ impl CxxQtBuilder {
...
-                    std::fs::canonicalize(symlink).expect("Failed to canonicalize symlink!");
-                let target =
-                    std::fs::canonicalize(target).expect("Failed to canonicalize symlink target!");
-                if symlink != target {
+            let source = dependency_root.join(include_prefix);
+            let dest = header_root.join(include_prefix);
+
+            match dir::symlink_or_copy_directory(source, dest) {
+                Ok(true) => (),
+                Ok(false) => {
```
LeonMatthesKDAB, line 653: This cleans up our old code quite nicely, awesome! :blush:

crates/cxx-qt-build/src/lib.rs
```diff
@@ -1019,17 +1005,18 @@ impl CxxQtBuilder {
...
-                        panic!(
-                            "Failed to create symlink `{}` for export_include_directory: {}",
-                            symlink,
-                            header_dir.to_string_lossy()
-                        )
-                    });
-            }
+        let Some(interface) = &self.public_interface else {
+            return;
+        };
```
LeonMatthesKDAB, line 1010: Huh, I had no idea this was a thing :sweat_smile: 

Apparently this has been stable since Rust 1.65: https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html
The rust-version in our main Cargo.toml says we depend on Rust 1.77.2, so we can use this feature, nice!

LeonMatthesKDAB (commented):
@jnbooth Thank you for keeping on top of this.

I'm quite happy with how the PR is turning out.
There's some small things left, especially the `files_conflict` function feels a bit more C-like than I would want.

Once those are resolved I'm happy to approve this :)

crates/cxx-qt-build/src/dir.rs
jnbooth: Hmm. I agree with this comment on that thread:

>I'm not sure if `.bytes()` iterator is the best for it, since handling of `io::Error` adds overhead and can't be optimized out. Reading in larger chunks should improve it.

What would you think about using the `BufReader` buffers themselves for comparison, as in [the latest commit](https://github.com/KDAB/cxx-qt/pull/1136/files#diff-b3d0f156e4bfb7954ff4537ae918929c3edfe9356658c3a8d7b99bb2f360ee2f)? Asking Rust to do an equality check on two slices definitely seems more efficient than equality-checking byte by byte, unwrapping a `Result` every time.

crates/cxx-qt-build/src/dir.rs
```diff
@@ -8,10 +8,17 @@
...
+    env, fs,
     path::{Path, PathBuf},
 };
 
+/// On Unix platforms, included files are symlinked into destination folders.
+/// On non-Unix platforms, due to poor support for symlinking, included files are deep copied.
+#[cfg(unix)]
+pub const INCLUDE_VERB: &str = "create symlink";
+#[cfg(not(unix))]
+pub const INCLUDE_VERB: &str = "deep copy files";
```
ahayzen-kdab, line 20: These should likely be `pub(crate)` i guess? As they are only used internally ?

LeonMatthesKDAB:
@jnbooth sorry for the delay, took me a bit longer than I hoped to re-review this patch.

Unfortunately I also have to correct this statement:

> Ah, and side-note: The note from the README can be removed with this patch :) This was the only reason we had it IIRC.

Indeed I did not recall correctly, because while **depending on** CXX-Qt now no longer requires symbolic links, the **repository** still contains some.
The `qml_minimal` example and the `cargo_without_cmake` share some code, which we do via symlinks to avoid duplicating files.
So can you please restore the note from the README :pray: 
I'm sorry about the extra work...

crates/cxx-qt-build/src/dir.rs
LeonMatthesKDAB: Hm, this actually seems like a good compromise between readability and indexing complexity.

If you can modify the method this way I can approve the PR :)

jnbooth: Done! Excited to see this in.

jnbooth:
@LeonMatthesKDAB Is there anything else I should do for this PR?

LeonMatthesKDAB (approved):
@jnbooth Again, sorry for the delay. The holidays had me firmly in their grasp :santa: :relaxed: 

Thank you for the contribution.

rnayabed:
Nice to see this has already been fixed. I was unable to get even `qml_minimal` (assuming it is the same as the one described in the [CXX-Qt Book](https://kdab.github.io/cxx-qt/book/) on Windows. I see the README of this repo does point it out, but not the book.

PR #1139, opened 2024-12-10
cxx-qt-lib: Add binding for QDateTime::fromString

redstrate:
This gives cxx-qt users a way to parse QDateTime from a QString. This is using one of the QDateTime::fromString overloads.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/KDAB/cxx-qt/compare/v0.7.0...HEAD)
 
+### Added
+
+- `QDateTime::from_string` to parse `QDateTime` from a `QString`.
+
 ### Fixed
 
 - Build warnings due to unused unsafe blocks since CXX 1.0.130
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - QObject subclasses can now inherit from other CXX-Qt generated QObject classes
 - `BUILD_WASM` CMake option to support WebAssembly builds and a book page for building for WASM
 - Add support for cxx_name and rust_name on qproperty attributes which applies to the QProperty generated as well as functions
+- `QDateTime::from_string` to parse `QDateTime` from a `QString`.
```
ahayzen-kdab, line 41: 0.7.0 is released, this needs to be moved into the unreleased section :-)

crates/cxx-qt-lib/src/core/qdatetime.rs
```diff
@@ -298,6 +301,16 @@ impl QDateTime {
         ffi::qdatetime_from_secs_since_epoch(secs, time_zone)
     }
 
+    /// Returns the datetime represented in the string as a QDateTime using the format given, or None if this is not possible.
+    pub fn from_string(string: &ffi::QString, format: ffi::DateFormat) -> Option<Self> {
```
ahayzen-kdab, line 305: Always wonder if we should be calling it `string` or `qstring` in our API of cxx-qt-lib :thinking:

redstrate: Let's go with `from_string`, reference: QDate already has `from_string` methods :)

CHANGELOG.md
redstrate: Oops, done :)

LeonMatthesKDAB (approved):

Issue #1133, opened 2024-11-18
Pass in the cfg_evaluator to cxx-qt-gen and consider when generating C++


ahayzen-kdab:
Along with #1132 once we are handling cfg everywhere we need to also ensure that our C++ calls the cfg_evaluator from the build system so that it can determine which headers / source need to be generated.

Otherwise if you have

```rust
extern "RustQt" {
   #[qobject]
   type MyObject = ....;

   #[cfg(...)]
   #[cxx_name = "method"]
   fn method_a(&MyObject);

   #[cfg(...)]
   #[cxx_name = "method"]
   fn method_b(&MyObject);
}
```

This fails as on the C++ side it generates two functions of `void method() const noexcept` and `void method() const`.

ahayzen-kdab:
WIP here #1144

LeonMatthesKDAB:
I believe we can close this since #1168 is merged now.
@ahayzen-kdab correct?

Issue #1132, opened 2024-11-18
Consider a generic way to support pass through attributes in cxx-qt-gen


ahayzen-kdab:
We need to be able to pass through attributes such as

- cfg
- doc
- allow
- deny
- etc

On many different elements in the original bridge macro. We should consider if there is a more generic way of doing this in the parser and generator phases with a generic type / mapping. Also see what attributes CXX is supporting https://github.com/dtolnay/cxx/blob/547a0edcf79cca2dad3f835ba736cfb85e93b1e5/syntax/attrs.rs#L286  https://github.com/dtolnay/cxx/blob/547a0edcf79cca2dad3f835ba736cfb85e93b1e5/macro/src/derive.rs#L17

PR #1140, opened 2024-12-10
cxx-qt-lib: Add binding for QQmlApplicationEngine::singletonInstance

redstrate:
This allows accessing QObject singleton instances registered in the QML engine (using #[qml_singleton]) from the Rust side.

### DIFF
--- a/crates/cxx-qt-lib/src/qml/qqmlapplicationengine.cpp
+++ b/crates/cxx-qt-lib/src/qml/qqmlapplicationengine.cpp
@@ -22,5 +22,16 @@ qqmlapplicationengineAsQQmlEngine(QQmlApplicationEngine& engine)
   return static_cast<QQmlEngine&>(engine);
 }
 
+#if (QT_VERSION >= QT_VERSION_CHECK(6, 5, 0))
+void*
+qqmlapplicationengineSingletonInstance(QQmlApplicationEngine& engine,
+                                       QAnyStringView uri,
+                                       QAnyStringView typeName)
+{
+  return reinterpret_cast<void*>(
+    engine.singletonInstance<QObject*>(uri, typeName));
+}
+#endif
+
 }
 }
--- a/crates/cxx-qt-lib/src/qml/qqmlapplicationengine.rs
+++ b/crates/cxx-qt-lib/src/qml/qqmlapplicationengine.rs
@@ -68,6 +68,9 @@ mod ffi {
 
     #[namespace = "rust::cxxqtlib1"]
     unsafe extern "C++" {
+        include!("cxx-qt-lib/common.h");
+        type c_void = crate::c_void;
+
         #[doc(hidden)]
         #[rust_name = "qqmlapplicationengine_new"]
         fn qqmlapplicationengineNew() -> UniquePtr<QQmlApplicationEngine>;
@@ -79,6 +82,24 @@ mod ffi {
         ) -> Pin<&mut QQmlEngine>;
     }
 
+    #[cfg(any(cxxqt_qt_version_at_least_7, cxxqt_qt_version_at_least_6_5))]
+    unsafe extern "C++" {
+        include!("cxx-qt-lib/qanystringview.h");
+        type QAnyStringView<'a> = crate::QAnyStringView<'a>;
+    }
+
+    #[namespace = "rust::cxxqtlib1"]
+    unsafe extern "C++" {
+        #[doc(hidden)]
+        #[rust_name = "qqmlapplicationengine_singleton_instance"]
+        #[cfg(any(cxxqt_qt_version_at_least_7, cxxqt_qt_version_at_least_6_5))]
+        fn qqmlapplicationengineSingletonInstance(
+            ptr: Pin<&mut QQmlApplicationEngine>,
+            uri: QAnyStringView,
+            typeName: QAnyStringView,
+        ) -> *mut c_void;
+    }
+
     // QQmlApplicationEngine is not a trivial to CXX and is not relocatable in Qt
     // as the following fails in C++. So we cannot mark it as a trivial type
     // and need to use references or pointers.
@@ -86,6 +107,8 @@ mod ffi {
     impl UniquePtr<QQmlApplicationEngine> {}
 }
 
+#[cfg(any(cxxqt_qt_version_at_least_7, cxxqt_qt_version_at_least_6_5))]
+use crate::QAnyStringView;
 use crate::QQmlEngine;
 use core::pin::Pin;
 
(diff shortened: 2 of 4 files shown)


crates/cxx-qt-lib/src/qml/qqmlapplicationengine.cpp
```diff
@@ -22,5 +22,11 @@ qqmlapplicationengineAsQQmlEngine(QQmlApplicationEngine& engine)
   return static_cast<QQmlEngine&>(engine);
 }
 
+void*
+qqmlapplicationengineSingletonInstance(QQmlApplicationEngine& engine, QAnyStringView uri, QAnyStringView typeName)
+{
+  return reinterpret_cast<void*>(engine.singletonInstance<QObject*>(uri, typeName));
```
redstrate, line 28: I kind of hate everything about this, but this somehow works. Suggestions appreciated, since I don't know of any other method supported in cxx-qt that has a template parameter.

crates/cxx-qt-lib/src/qml/qqmlapplicationengine.rs
```diff
@@ -79,13 +82,33 @@ mod ffi {
         ) -> Pin<&mut QQmlEngine>;
     }
 
+    #[cfg(any(cxxqt_qt_version_at_least_7, cxxqt_qt_version_at_least_6_5))]
+    unsafe extern "C++" {
```
redstrate, line 86: This #[cfg] doesn't work I guess, so Qt5 and <6.5 builds will fail. An easy solution would be to separate this into a different cxx::bridge as suggested in [this issue](https://github.com/dtolnay/cxx/issues/989) but that seems extreme?

ahayzen-kdab: Fully support for cfg flags everywhere comes in #1133 and #1132 which i plan to look at next.

For now you could try instead of putting the cfg flag on the extern block itself put it on the method as this is passed through to CXX.

crates/cxx-qt-lib/src/qml/qqmlapplicationengine.cpp
ahayzen-kdab: Hmm, once we have up/down casting and QObject already as a type #562 then this would be much easier. We would like this have this implemented in this cycle as it seems useful in multiple places.

It would be return a pointer to a QObject, then use the downcast_ptr method to get your original type back.

crates/cxx-qt-lib/src/qml/qqmlapplicationengine.rs
```diff
@@ -101,4 +124,21 @@ impl QQmlApplicationEngine {
...
+        self: Pin<&mut Self>,
+        uri: QAnyStringView,
+        type_name: QAnyStringView,
+    ) -> Option<&mut T> {
+        unsafe {
+            let ptr = ffi::qqmlapplicationengine_singleton_instance(self, uri, type_name);
+            if ptr.is_null() {
+                return None;
+            }
+            Some(&mut *(ptr as *mut T))
```
LeonMatthesKDAB, line 141: Hm, this is definitely *not* safe to do here!

As you could basically "spoof" T to be different from the actual type that the singleton has.

In the most extreme case, I could just do:
```
engine.singleton_instance::<i32>(...);
```

We could mark the function itself as `unsafe` to move that responsibility up one level, but it's still not great...

crates/cxx-qt-lib/src/qml/qqmlapplicationengine.cpp
LeonMatthesKDAB: Hm, right, this could be a decent way to solve this issue :thinking:

akiselev:
For what it's worth, I implemented this using a template helper:

```cpp
template <typename T>
T *qqmlengineSingletonInstance(QQmlEngine &engine, const QString &module, const QString &name)
{
    auto ptr =
        engine.template singletonInstance<T *>(
            QAnyStringView(module), QAnyStringView(name));
    return ptr;
}
```

Then in the C++ bridge:

```rust
    #[rust_name = "get_theme_singleton"]
    unsafe fn qqmlengineSingletonInstance(
        engine: Pin<&mut QQmlEngine>,
        module: &QString,
        name: &QString,
    ) -> *mut ThemeSingleton;
```

and finally the safe Rust wrapper:

```rust
    fn get_theme_singleton(self: Pin<&mut Self>) -> Option<Pin<&mut ThemeSingleton>> {
        unsafe {
            // This is safe because we're checking if it's null and only handing out a pinned mut ref
            // Qt does its own type safety checks here too
            let ptr = get_theme_singleton(self.as_qqmlengine());
            ptr.as_mut().map(|p| Pin::new_unchecked(p))
        }
    }
```

crates/cxx-qt-lib/src/qml/qqmlapplicationengine.rs
redstrate: I have added some lifetime checks to improve it a bit, but yes this is inherently unsafe due to a lack of type-checking. Note that once we _have_ the ability to do type checking (like in #562, or akiselev's comment), this should stay unsafe.

crates/cxx-qt-lib/src/qml/qqmlapplicationengine.rs
redstrate: That does work, but the QAnyStringView inclusion is hitting the same problems as seen in #1141 :(

crates/cxx-qt-lib/src/qml/qqmlapplicationengine.rs
redstrate: I'll leave it up to your discretion whether we should include this inherently-unsafe version now, and then have a proper solution in tandem with #562. IMO, it would be really useful right now :-)

LeonMatthesKDAB: Hm, so I dug a bit deeper into the topic of downcasting in C++.

The way this implementation works is that it basically just reinterpret_casts the base type to the derived type, which is done by changing the pointer type in Rust. (not actually the reinterpret_cast in C++, as that basically "upcasts" to `void*`).

Of course, this has the problem of whether the type is actually of the derived class.
This is a check that I'd be willing to simply leave to `unsafe` code to deal with and is very similar to using `static_cast` for downcasting.

However, compared to a `static_cast` for downcasting, this approach has another problem, which is the undefined memory-layout of a derived C++ class.
As far as I can gather, it is not specified that the address of the derived type is the same as the address of the base type when downcasting, as this seems up to the compiler to decide.

According to: https://stackoverflow.com/a/6354558 , this seems to mostly be the case when using single-inheritance and definitely not the case for multi-inheritance.
However, I can't find a guarantee (even for single-inheritance) anywhere.
So it can be not just that the type you're casting to is incorrect (which is feasible to guarantee at the call-site), but it's also possible that the cast we're doing will simply give you back garbled memory, depending on the compilers internal decisions about the memory layout.

To me this is a level of risk I'm not very comfortable with, as for the caller this is almost impossible to guarantee.

So I advocate for ensuring we have some kind of downcasting support first before attempting to support this, even though it's a bummer to not have it right now.

The solution presented by @akiselev is also very intruiging and could be a good solution if it turns out downcasting support in general is not feasible.

redstrate: I'm fine with waiting for downcast support, since there's other stuff like #1153 that's also blocking the project I'm using this function in anyway. Converting this to a draft in the meantime

crates/cxx-qt-lib/src/qml/qqmlapplicationengine.rs
ahayzen-kdab: See my comment in #1141 i think this should work now, we fixed cfg not working on type aliases in CXX

PR #1141, opened 2024-12-10
cxx-qt-lib: Add binding for QQmlApplicationEngine::loadFromModule

redstrate:
This function was added in Qt 6.5, as an alternative from loading files by file path.

### DIFF
--- a/crates/cxx-qt-lib/src/qml/qqmlapplicationengine.rs
+++ b/crates/cxx-qt-lib/src/qml/qqmlapplicationengine.rs
@@ -6,6 +6,12 @@
 #[cxx::bridge]
 mod ffi {
     unsafe extern "C++" {
+        #[cfg(any(cxxqt_qt_version_at_least_7, cxxqt_qt_version_at_least_6_5))]
+        include!("cxx-qt-lib/qanystringview.h");
+        type QAnyStringView<'a> = crate::QAnyStringView<'a>;
+    }
+
+    unsafe extern "C++" {
         include!("cxx-qt-lib/qstring.h");
         type QString = crate::QString;
         include!("cxx-qt-lib/qstringlist.h");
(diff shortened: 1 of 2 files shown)


redstrate:
This will fall into the same fallacy as described here: https://github.com/KDAB/cxx-qt/pull/1140#discussion_r1877062884 due to using QAnyStringView.

crates/cxx-qt-lib/src/qml/qqmlapplicationengine.rs
```diff
@@ -12,6 +12,8 @@ mod ffi {
         type QStringList = crate::QStringList;
         include!("cxx-qt-lib/qurl.h");
         type QUrl = crate::QUrl;
+        include!("cxx-qt-lib/qanystringview.h");
+        type QAnyStringView<'a> = crate::QAnyStringView<'a>;
```
ahayzen-kdab, line 16: you need the cfg flags also here, although unsure if they work on include macros?

As this is a `cxx::bridge` rather than a `cxx_qt::bridge` I think the cfg flag should work on an `extern block`. It's only CXX-Qt bridges where we have the limitation of support only on certain items #1133

redstrate: I tried a ton of solutions, and it doesn't seem like extern blocks support cfg macros. It would get rid of the include, but still try to parse the `type`. I have looked into substituting a dummy type which kinda works, but it's working around missing functionality that should really be in CXX..

ahayzen-kdab: Aha, as part of  #1168 we improved CXX-Qt's side but found some issues with CXX itself. One of them we fixed as cfg attributes not working on type aliases ( https://github.com/dtolnay/cxx/pull/1464 ).

Can you try this again with the newer CXX version and with the following i think it might work now :pray: 

```
    unsafe extern "C++" {
        #[cfg(any(cxxqt_qt_version_at_least_7, cxxqt_qt_version_at_least_6_5))]
        include!("cxx-qt-lib/qanystringview.h");
        #[cfg(any(cxxqt_qt_version_at_least_7, cxxqt_qt_version_at_least_6_5))]
        type QAnyStringView<'a> = crate::QAnyStringView<'a>;
    }
```

PR #1142, opened 2024-12-10
cxx-qt-lib: Add binding for QQmlApplicationEngine::addImageProvider

redstrate:
Upstreamed from the patch in #785, this adds a way to add image providers written in Rust to the QML engine.

### DIFF
--- a/crates/cxx-qt-lib/src/qml/qqmlapplicationengine.rs
+++ b/crates/cxx-qt-lib/src/qml/qqmlapplicationengine.rs
@@ -12,9 +12,19 @@ mod ffi {
         type QStringList = crate::QStringList;
         include!("cxx-qt-lib/qurl.h");
         type QUrl = crate::QUrl;
+        include!(<QQmlImageProviderBase>);
+        type QQmlImageProviderBase;
+        include!(<QtQuick/QQuickImageProvider>);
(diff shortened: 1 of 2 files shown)


redstrate:
WIP, I would like to have some sort of example. Someone in the KDE community asked about this, so we should probably upstream the patch to cxx-qt.

crates/cxx-qt-lib/src/qml/qqmlapplicationengine.rs
```diff
@@ -101,4 +119,17 @@ impl QQmlApplicationEngine {
     pub fn new() -> cxx::UniquePtr<Self> {
         ffi::qqmlapplicationengine_new()
     }
+
+    /// Sets the provider to use for images requested via the image: url scheme, with host proveri_id. The QQmlEngine takes ownership of provider.
+    pub unsafe fn add_image_provider(self: Pin<&mut QQmlApplicationEngine>, provider_id: &QString, provider: QQmlImageProviderBasePointer) {
+        let ptr = match provider {
```
ahayzen-kdab, line 125: this is another case where having up/down casting would help out #562 (something we hope for this cycle). As then the method could just support the QQmlImageProviderBase* and in Rust you do upcast_ptr() when calling the method.

PR #1144, opened 2024-12-18
cxx-qt-gen: add support for passing in CfgEvaluator

ahayzen-kdab:


crates/cxx-qt-gen/src/syntax/cfg.rs
```diff
@@ -0,0 +1,159 @@
...
+        if let CfgExpr::Unconditional = self {
+            *self = expr;
+        } else if let CfgExpr::All(list) = self {
+            list.push(expr);
+        } else {
+            let prev = mem::replace(self, CfgExpr::Unconditional);
+            *self = CfgExpr::All(vec![prev, expr]);
+        }
+    }
+}
```
ahayzen-kdab, line 34: it's unclear what codecov is complaining about on these lines :thinking:

ahayzen-kdab:
Going to use #1168 instead

Issue #1135, opened 2024-11-27
QML Modules defined in external crates don't work


akiselev:
Repro: https://github.com/akiselev/cxx-qt-module-repro

Current output:

```
Default MyObjectRust MyObjectRust { number: 0, string:  }
QQmlApplicationEngine failed to load component
file:///.../cxx-qt-module-repro/qml/main.qml:13:1: module "CxxQt.Repro" is not installed
```

@LeonMatthes  said in Zulip:

* All your crates should use CXX-Qt 0.7 (or git main)
* All library crates need to use the CxxQtBuilder::library function
* All library crates need to have a links key set in their Cargo.toml
* Your main executable needs to build with a normal CxxQtBuilder::new, even if it doesn't actually need to build anything itself.

I think I've got all that covered but the repro still fails to load the module correct. In my own project, this works with CMake but not Cargo.

LeonMatthesKDAB:
@akiselev Thank you for taking the time to report the issue and especially to create the reproducable example. That has really helped with troubleshooting.

I was able to reproduce the issue and to find the root cause.
QML modules actually build and link their own initializers, but don't export them to downstream dependencies.

With CMake, this isn't necessarily a problem, as you import the QML module separately via `cxxqt_import_qml_module`, which then includes the object file in the build system (though you then have to import the crate that contains the qml module separately as well).

With Cargo-only builds, our assumption was that linking the object file directly would cause it to not be optimized out at the linking stage, even for dependencies.
However, this assumption seems to have been incorrect, as Rust seems to still link the dependencies into a static library or something similar before building the final executable.
This then ends up optimizing the initializer out again.
 cc @ahayzen-kdab 
 
 While the fix for cargo-only builds is quite simple (see: https://github.com/KDAB/cxx-qt/tree/7952f55c1529df8d8a67b2278ea12d67f885412d ), a solution that also fixes the CMake integration needs a bit more thought.

akiselev:
> While the fix for cargo-only builds is quite simple (see: https://github.com/KDAB/cxx-qt/tree/7952f55c1529df8d8a67b2278ea12d67f885412d ), a solution that also fixes the CMake integration needs a bit more thought.

@LeonMatthesKDAB is there more to this fix than what I am seeing in the commit?

When trying your modifications locally with my repro repo, I get the following error:

```
[...]

Issue #1125, opened 2024-11-15
CXX-Qt-build: Goal for 0.8


LeonMatthesKDAB:
**Problem Statement**
Currently, our build system suffers from some issues.
1. Building a QML module uses the same cc_builder as the main build
2. Regenerating the header files requires a full rebuild
    * Problematic for tooling like clangd that depends on up-to-date headers
3. CMake has limited control over the build process.

**Proposed solution**
CXX actually solves this by explicitly leaving the build configuration to the respective build system in all languages.
For Cargo, CXX relies on the cc crate, for CMake it allows generating the files with [`corossion_add_cxxbridge`](https://corrosion-rs.github.io/corrosion/ffi_bindings.html#cxx-integration).
CMake then takes care of the actual building and linking.

We want to keep the advantage of our current build system approach, that deduplicates code between Cargo and CMake and allows implementing features like QML Modules for Qt5, but we also want to be prepared to give users more control when needed.

To achieve this, we aim to explicitly split the code generation and the build process itself.

The goal is that cxx_qt_build, like cxx_build just returns a pre-configured compiler, that the users are then free to do what they like with.
However, as Qt extends C++, it also extends the compiler capabilities.
To reflect that, we need to extend cc::Builder with our own QtBuilder (which we basically almost do).

The QtBuilder can then just be a direct wrapper of the cc::Builder, with additional features like `moc`, `qrc`, etc.
We also plan to move this QtBuilder fully into qt_build_utils, so that it is reusable outside CXX-Qt, and should only be Qt-specific, not CXX-Qt specific.

**Example**
This is an example of what the new API might look like:
```rust
fn main() {
    let builder : qt_build_utils::QtBuild = cxx_qt_build::bridge("src/lib.rs");
    builder
         .qrc("../")
         .compile_qml_module(QmlModule {
            uri: "...",
            version_major: "...",
            // ...
         });
}
```
This makes the difference between generating the bridge files (cxx_qt_build::bridge) and configuring the build process itself (QtBuild) clearer and makes everything more reusable.

**Open Questions**
We're not yet entirely clear about how this all plays together with initializers that are passed between crates and the export to CMake.
Both of these things should ideally only happen once per crate, not per QtBuild instance, so it's unclear what exactly needs to be responsible for that.

ahayzen-kdab:
[...]

PR #1146, opened 2024-12-19
Add support for C++ QML QObjects in CxxQtBuilder

akiselev:
This PR is a WIP to add support for registering classes implemented in C++ with a QML module so that they can be included in Cargo-only builds. Instead of just running `moc` on a header, it also includes the generated metatype information which is needed to register the C++ QML elements.

This first implementation is one I wrote to unblock my own project so I'd appreciate some advice for how to integrate this properly into `cxx-qt-build`. It should probably be another property on QmlModule like so, but I couldn't figure out how to best implement it that way:

```rust
    let module: QmlModule<&str, &str> = QmlModule {
        uri: "Clippo.Theme",
        rust_files: &[
            "src/theme.rs",
        ],
        cpp_files: &[
            ("src/ThemeManager.h", "src/ThemeManager.cpp),
        ],
        ..Default::default()
    };
```

With @LeonMatthesKDAB's temporary fix found in #1135, it correctly includes C++ QML elements in modules built with Cargo only.

### DIFF
--- a/crates/cxx-qt-build/src/lib.rs
+++ b/crates/cxx-qt-build/src/lib.rs
@@ -358,6 +358,7 @@ pub struct CxxQtBuilder {
     public_interface: Option<Interface>,
     include_prefix: String,
     initializers: Vec<String>,
+    qml_qobjects: Vec<(PathBuf, PathBuf)>,
 }
 
 impl CxxQtBuilder {
@@ -398,6 +399,7 @@ impl CxxQtBuilder {
             initializers: vec![],
             public_interface: None,
             include_prefix: crate_name(),
+            qml_qobjects: vec![],
         }
     }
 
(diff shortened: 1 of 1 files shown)


crates/cxx-qt-build/src/lib.rs
```diff
@@ -532,6 +534,22 @@ impl CxxQtBuilder {
         self
     }
 
+    /// Specify a C++ header and source file containing a Q_OBJECT macro to run [moc](https://doc.qt.io/qt-6/moc.html) on.
+    /// Unlike [CxxQtBuilder::qobject_header], it includes the generated metatypes.json, so that C++ classes can be included
+    /// in QML modules.
+    pub fn qml_qobject(
```
redstrate, line 540: `cpp_qobject` ? I guess this could have other applications outside of QML

redstrate (commented):
I think this deserves a mention in CHANGELOG.md, can you add it to the MR?

LeonMatthesKDAB:
Hi @akiselev ,
as you mentioned, the best way to integrate this is indeed as a property on QmlModule.
The design of the QmlModule type is unfortunately a bit overly complicated, as it uses two template types, which we really don't want to expand.
My suggestion is to just hijack the existing `A` type, like we do for qrc_files.

I'm okay with the "list of tuples" approach you have in your initial comment.
We could even consider just a plain list of cpp_files, as we should be able to differentiate between .cpp and .h files by the extension.

Please note that for version 0.8 we are planning a larger refactor of the build system, which will likely end up supporting what you're currently asking anyway due to the nature of the proposed architecture.
See also: https://github.com/KDAB/cxx-qt/issues/1125
(basically: The qobject_header call should in future take care of registering automatically, if you're buidling a Qml module).

ahayzen-kdab:
Note that there is an example of this with CustomObject being a C++ header with a QML_ELEMENT that is used within a QML module build here https://github.com/KDAB/cxx-qt/blob/f2b0b5be4195e198b392f114f7ab89ed6bfbe6e0/examples/qml_features/rust/build.rs#L56

PR #1147, opened 2024-12-19
Add QJSValue, QJSEngine, and related types to cxx-qt-lib-extras with QJSValue serde support

akiselev:
I made serde a default feature of `cxx-qt-lib-extras` to make sure tests run in CI, can update before merge

`QJSEngine` is the entry point for manipulating the Javascript engine including evaluating arbitrary Javascript, creating new objects for the serializer, etc.

`QJSValue` allows accessing and manipulating Javascript values

`QJSValueList` is used for call Javascript functions from Rust but I haven't gotten around to the full implementation yet

`QJSValueIterator` is used by the (de)serializer support to iterate over objects and arrays.

### DIFF
--- a/crates/cxx-qt-lib-extras/Cargo.toml
+++ b/crates/cxx-qt-lib-extras/Cargo.toml
@@ -17,12 +17,14 @@ rust-version.workspace = true
 cxx.workspace = true
 cxx-qt.workspace = true
 cxx-qt-lib = { workspace = true, features = ["qt_full"] }
+serde = { version = "1", features=["derive"], optional = true }
 
 [build-dependencies]
 cxx-qt-build.workspace = true
 
 [features]
-default = []
+default = ["serde"]
+serde = ["dep:serde"]
 link_qt_object_files = ["cxx-qt-build/link_qt_object_files"]
 
 [lints]
--- a/crates/cxx-qt-lib-extras/include/qml/qjsengine.h
+++ b/crates/cxx-qt-lib-extras/include/qml/qjsengine.h
@@ -1,0 +1,81 @@
+#pragma once
+
+#include <memory>
+
+#include <QtQml/QJSEngine>
+
+#include "rust/cxx.h"
+
+namespace rust
+{
+    namespace cxxqtlib2
+    {
+        ::std::unique_ptr<QJSEngine>
+        qjsengineNew();
+
+        template <typename T>
+        ::std::unique_ptr<QJSValue>
+        jsengineNewArray(T &engine, quint32 length)
+        {
+            auto ptr = ::std::make_unique<QJSValue>(engine.newArray(length));
+            Q_ASSERT(ptr != nullptr);
+
+            return ptr;
+        }
+
+        template <typename T>
+        ::std::unique_ptr<QJSValue>
+        jsengineNewObject(T &engine)
+        {
+            auto ptr = ::std::make_unique<QJSValue>(engine.newObject());
+            Q_ASSERT(ptr != nullptr);
+
+            return ptr;
+        }
+
+        template <typename T>
+        ::std::unique_ptr<QJSValue>
+        jsengineNewQObject(T &engine, QObject &object)
+        {
+            auto ptr = ::std::make_unique<QJSValue>(engine.newQObject(&object));
+            Q_ASSERT(ptr != nullptr);
+
+            return ptr;
+        }
+
+        template <typename T>
+        ::std::unique_ptr<QJSValue>
+        jsengineEvaluate(
+            T &engine,
+            const QString &program,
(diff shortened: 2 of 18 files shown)


akiselev:
I actually think this should be in cxx-qt-lib proper with `QQmlEngine` because that's going to be the main access point for `QJSValue`. QML elements can  get their `QQmlEngine` instance via a new trait `AsQQmlEngine` that should be autoimplemented by the cxx-bridge macro and use the resulting engine instance for serialization.

ahayzen-kdab:
Looks interesting! I would be interested in usecases, as i've hardly used any of the QJS* APIs over multiple years of QML dev :sweat_smile: as it usually points that your backend is having knowledge of your frontend which is not generally a good idea.

-extras is a good place for this to start, but as you say some of the API is also in QQmlEngine so may need some thought if that could work in -extras or not.

akiselev:
That was just the easiest way to get JSON in and out of Javascript for a multiprocess extension system with user defined types. It also helped with build times since I didn’t need to run cxx-qt bridge generation and cc for a bunch of QML types that were basically just dumb records.

The vast majority of my backend doesn’t know anything about QML or JS but there’s a lot of glue code that handles request/responses for the backend “services” that implement logic

I think this also opens the door to auto-(de)serializing in invokables. For example (on my phone so this rough)

```rust
#[qinvokable]
#[deserialize(bar)]
#[serialize_return]
fn foo(&self, bar: RustArgType) -> RustReturnType;
```

ahayzen-kdab:
Ah interesting so it's being used like an extended `QVariantMap` with some nice serde fun :-)

crates/cxx-qt-lib-extras/include/qml/qjsengine.h
```diff
@@ -0,0 +1,81 @@
...
+#include <QtQml/QJSEngine>
+
+#include "rust/cxx.h"
+
+namespace rust
+{
+    namespace cxxqtlib2
+    {
+        ::std::unique_ptr<QJSEngine>
+        qjsengineNew();
```
LeonMatthesKDAB, line 14: It should be possible to replace this with our make_unique template from cxx-qt-lib.

Take a look at the `qpainter_init_default()` function in qpainter.rs for how to use this.

LeonMatthesKDAB (commented):
Hi @akiselev ,
this PR looks super interesting. Like Andrew mentioned, I also haven't used much of QJSValue, but especially the serialize/deserialize aspect looks intriguing.

As is [policy with -lib-extras](https://kdab.github.io/cxx-qt/book/internals/crate-organization.html#cxx-qt-lib-extras), I'm happy to approve this PR, as long as "it compiles", which at the moment includes passing CI.

Please fix the formatting and add the appropriate licensing.

I left a comment, which you can also feel free to ignore, that was just a suggestion for a small trick you may not be familiar with yet :)

crates/cxx-qt-lib-extras/Cargo.toml
```diff
@@ -17,12 +17,14 @@ rust-version.workspace = true
...
 cxx-qt.workspace = true
 cxx-qt-lib = { workspace = true, features = ["qt_full"] }
+serde = { version = "1", features=["derive"], optional = true }
 
 [build-dependencies]
 cxx-qt-build.workspace = true
 
 [features]
-default = []
+default = ["serde"]
```
redstrate, line 26: nitpick: I'm unsure if serde should be a default here

crates/cxx-qt-lib-extras/src/qml/qjsvaluelist.cpp
```diff
@@ -0,0 +1,26 @@
+// #include "cxx-qt-lib-extras/qapplication.h"
```
redstrate, line 1: leftover include?

crates/cxx-qt-lib-extras/src/qml/qjsvalueiterator.cpp
```diff
@@ -0,0 +1,25 @@
+// #include "cxx-qt-lib-extras/qapplication.h"
```
redstrate, line 1: ditto

crates/cxx-qt-lib-extras/src/qml/qjsvalue.cpp
```diff
@@ -0,0 +1,124 @@
+// #include "cxx-qt-lib-extras/qapplication.h"
```
redstrate, line 1: same

redstrate (commented):
The CI can be quieted by including REUSE metadata, see the other files with SPDX for reference. Then run it over with `cargo fmt` and then `clang-format -i <your cpp/h file>`.

PR #1149, opened 2024-12-25
cxx-qt-lib: Add support for QUuid

jnbooth:
Adds a shared type for [QUuid](https://doc.qt.io/qt-6/quuid.html), which can optionally be converted to and from the type provided by the [uuid](https://crates.io/crates/uuid) crate. Since it's expensive to convert between Strings and QStrings, QUuids provide a convenient no-allocation alternative for identifiers.


crates/cxx-qt-lib/src/core/quuid.rs
```diff
@@ -0,0 +1,386 @@
...
+                b0, b1, b2, b3, b4, b5, b6, b7, b8, b9, ba, bb, bc, bd, be, bf,
+            ]
+        }
+        #[cfg(target_endian = "little")]
+        {
+            [
+                b3, b2, b1, b0, b5, b4, b7, b6, b8, b9, ba, bb, bc, bd, be, bf,
+            ]
+        }
+    }
```
jnbooth, line 220: Alternatively, using `std::mem::transmute`:
```rs
    const fn to_big_endian(self) -> Self {
        #[cfg(target_endian = "big")]
        {
            self
        }
        #[cfg(target_endian = "little")]
        {
            Self {
                data1: self.data1.swap_bytes(),
                data2: self.data2.swap_bytes(),
                data3: self.data3.swap_bytes(),
                data4: self.data4,
            }
        }
    }

    /// Creates a UUID from its representation as a byte array in big endian.
    pub const fn from_bytes(bytes: [u8; 16]) -> Self {
        unsafe { mem::transmute::<[u8; 16], Self>(bytes) }.to_big_endian()
    }

    /// Returns the memory representation of this UUID as a byte array in big-endian byte order.
    pub const fn to_bytes(self) -> [u8; 16] {
        unsafe { mem::transmute::<Self, [u8; 16]>(self.to_big_endian()) }
    }
```
It's simpler and more efficient, but at the cost of using `unsafe`, which is why I figured you'd prefer the PR's approach instead.

crates/cxx-qt-lib/src/core/quuid.rs
```diff
@@ -0,0 +1,386 @@
...
+    fn from(value: QUuid) -> Self {
+        Self::from_fields(value.data1, value.data2, value.data3, &value.data4)
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    const NAMESPACE_DNS: &QUuid = &QUuid::from_u128(0x6ba7b8109dad11d180b400c04fd430c);
```
jnbooth, line 280: Those `const fn` conversions make it easy to declare constants like this one.

crates/cxx-qt-lib/src/core/qvector/qvector_quuid.rs
```diff
@@ -0,0 +1,92 @@
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
```
redstrate, line 2: wrong contributor, it should be you yes?

crates/cxx-qt-lib/src/core/qvariant/qvariant_quuid.rs
```diff
@@ -0,0 +1,37 @@
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
```
redstrate, line 2: ditto, wrong contributor

crates/cxx-qt-lib/src/core/quuid.rs
redstrate: Personally I like the transmute option more. I don't really like the hand-rolled option as that just seems easy to mess up. I wonder if to_big_endian needs/should be public API though?

crates/cxx-qt-lib/src/core/quuid.rs
```diff
@@ -0,0 +1,386 @@
...
+        /// Time-based, by using timestamp, clock sequence, and MAC network card address (if
+        /// available) for the node sections
+        Time = 1,
+        /// DCE Security version, with embedded POSIX UUIDs
+        EmbeddedPOSIX = 2,
+        /// Name-based, by using values from a name for all sections
+        Md5 = 3,
+        /// Random-based, by using random numbers for all sections
+        Random = 4,
+        Sha1 = 5,
```
redstrate, line 38: SHA1 not good enough to get it's own documentation? :D

crates/cxx-qt-lib/src/core/quuid.cpp
```diff
@@ -0,0 +1,71 @@
+// clang-format off
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// clang-format on
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
```
redstrate, line 4: I think it can just be you as a contributor to this code, right? Unless there's some backstory to this MR I'm not getting

crates/cxx-qt-lib/src/core/qset/qset_quuid.rs
```diff
@@ -0,0 +1,70 @@
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
```
redstrate, line 2: ditto, you here please :)

crates/cxx-qt-lib/include/core/quuid.h
```diff
@@ -0,0 +1,34 @@
+// clang-format off
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// clang-format on
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
```
redstrate, line 4: ditto

redstrate:
This looks pretty in-depth, thanks! I think this is definitely worthy of a mention in CHANGELOG.md, you should add it there in this PR.

I also started up the CI workflow, so we can see if that succeeds.

crates/cxx-qt-lib/src/core/qvector/qvector_quuid.rs
jnbooth: Autogenerated from https://github.com/cxx-qt-lib/cxx-qt/blob/d79dd4c39db6fbd5e29cd38caed095c2a7022f91/crates/cxx-qt-lib/src/core/qvector/generate.sh#L16

crates/cxx-qt-lib/src/core/qvariant/qvariant_quuid.rs
jnbooth: Autogenerated from https://github.com/KDAB/cxx-qt/blob/main/crates/cxx-qt-lib/src/core/qvector/generate.sh#L16

crates/cxx-qt-lib/src/core/quuid.rs
jnbooth: Taken directory from the [Qt QUUID docs](https://doc.qt.io/qt-6/quuid.html#Version-enum)

crates/cxx-qt-lib/src/core/qset/qset_quuid.rs
jnbooth: Autogenerated from https://github.com/KDAB/cxx-qt/blob/main/crates/cxx-qt-lib/src/core/qset/generate.sh#L16

crates/cxx-qt-lib/src/core/quuid.rs
jnbooth: I would be nervous about making `to_big_endian` part of the public API, because it's only a utility for `from_bytes` and `to_bytes`. If we decided to change our implementation of those functions, I personally think we should be able to get rid of `to_big_endian` without it being a breaking change.

crates/cxx-qt-lib/src/core/qvector/qvector_quuid.rs
jnbooth: It might be a good idea to add a .gitattributes to [mark these files as autogenerated](https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github).

redstrate: Ah then I'll drop the comments for now. Mind opening an issue about looking into .gitattributes?

crates/cxx-qt-lib/src/core/quuid.rs
redstrate: ah, that's fine then IMO since they omit it

crates/cxx-qt-lib/src/core/qvector/qvector_quuid.rs
jnbooth: Sure thing! #1150

crates/cxx-qt-lib/src/core/quuid.rs
```diff
@@ -0,0 +1,382 @@
...
+        }
+        #[cfg(target_endian = "little")]
+        {
+            Self {
+                data1: self.data1.swap_bytes(),
+                data2: self.data2.swap_bytes(),
+                data3: self.data3.swap_bytes(),
+                data4: self.data4,
+            }
+        }
```
LeonMatthesKDAB, line 205:
```suggestion
        Self {
            data1: self.data1.to_be(),
            data2: self.data2.to_be(),
            data3: self.data3.to_be(),
            data4: self.data4,
        }
```

crates/cxx-qt-lib/src/core/quuid.rs
```diff
@@ -0,0 +1,382 @@
...
+    }
+
+    /// Creates a UUID from its representation as a byte array in big endian.
+    pub const fn from_bytes(bytes: [u8; 16]) -> Self {
+        unsafe { mem::transmute::<[u8; 16], Self>(bytes) }.to_big_endian()
```
LeonMatthesKDAB, line 210: I have to admit it took me quite a while to understand this.

The function takes an array of big-endian bytes and then calls `to_big_endian` on it. That is quite confusing, as it's already big-endian.
I believe this still turns out to be correct, as the internal byte order is actually the host endianness, so calling to_big_endian will actually go from big to little endian, but only if the host order is little endian, which is exactly what's needed. But it's definitely not obvious that this is whats happening.

So at the very least this needs an explanation comment.

crates/cxx-qt-lib/src/core/quuid.rs
LeonMatthesKDAB: I agree that `to_big_endian` should remain private API.

A QUuid always uses the target endianness, so to_big_endian is already a kind of hack, as it sort off invalidates the QUuid.
It is still very useful for conversion to/from bytes though.

LeonMatthesKDAB (commented):
Hi @jnbooth ,
thank you for taking the time to work on this.
UUIDs are an important topic and having QUuid in cxx-qt-lib would be great!

I have left some minor nitpicks, but the most important problem is going to be Qt5 support, as QAnyStringView isn't supported on Qt5.
For now I'm okay to only support QString so we can merge this PR earlier.

crates/cxx-qt-lib/src/core/quuid.cpp
```diff
@@ -0,0 +1,70 @@
...
+namespace rust {
+namespace cxxqtlib1 {
+QUuid
+quuidNewV3(const QUuid& ns, ::rust::Slice<const ::std::uint8_t> slice)
+{
+  return QUuid::createUuidV3(ns, byteView(slice));
+}
+
+QUuid
+quuidNewV4()
```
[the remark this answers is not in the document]
jnbooth: I've renamed these functions to match [the Qt names](https://doc.qt.io/qt-6/quuid.html#static-public-members), so `create_uuid_v3`, `create_uuid`, and `create_uuid_v5`. Is that the change you were looking for?

crates/cxx-qt-lib/src/core/quuid.rs
jnbooth: I've added explaining documentation and directly copied over the naming convention and documentation for [u8::to_be](https://doc.rust-lang.org/std/primitive.u8.html#method.to_be), so that it will be more familiar to people who have worked with endian conversions. Do these changes make it sufficiently easy to understand? Is there anything else I should add?

crates/cxx-qt-lib/src/core/quuid.rs
```diff
@@ -0,0 +1,382 @@
...
+
+    unsafe extern "C++" {
+        include!("cxx-qt-lib/qbytearray.h");
+        type QByteArray = crate::QByteArray;
+    }
+
+    unsafe extern "C++" {
+        include!("cxx-qt-lib/qstring.h");
+        type QString = crate::QString;
+        type QAnyStringView<'a> = crate::QAnyStringView<'a>;
```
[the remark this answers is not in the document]
jnbooth: Makes sense. I've split it up into separate conversions from QStrings, Rust strs, and QByteArrays. Honestly, this way is cleaner anyway, because it's explicit about the types that can be converted rather than `T: Into<QAnyStringView>`.

crates/cxx-qt-lib/src/core/quuid.cpp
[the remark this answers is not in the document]
LeonMatthesKDAB: Thank you, that works for me.

The rename would have been okay as well if there is a benefit to the rename that out-weighs the downside of having to look up the new name.
Otherwise I prefer just sticking with the Qt naming (in snake_case) so that it's easy to find the corresponding function name if you're used to the Qt API.

crates/cxx-qt-lib/src/core/quuid.rs
LeonMatthesKDAB: Yes, thank you.

crates/cxx-qt-lib/src/core/quuid.rs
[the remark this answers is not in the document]
LeonMatthesKDAB: Thank you, just accepting `Str` and converting to QAnyStringView in C++ is a good solution :+1:

LeonMatthesKDAB (approved):
Thank you for the contribution @jnbooth :tada:

PR #1151, opened 2024-12-28
implement Deref and DerefMut for QStringList to QList<QString>

jnbooth:
Per @LeonMatthesKDAB's [comment](https://github.com/KDAB/cxx-qt/pull/761#discussion_r1447611935), implements `Deref` and `DerefMut` for `QStringList` so it isn't necessary to reimplement the `QList` API.


LeonMatthesKDAB (approved):
Looks good to me.

Note that this is somewhat related to our recent efforts with upcasting.
However, for convenience it probably makes sense to use Deref/DerefMut for this, even if we end up adding an `Inherits` implementation later.

@BenFordTytherington the qstringlistAsQListQStringRef method may be replacable with a a call to `upcast`, so is a target for de-duplication once we have proper Inheritance..

PR #1154, opened 2025-01-12
Add `serde` support for cxx-qt-lib using proxy types and deriving

### LINKED ISSUE #943: Add `serde` support for cxx-qt-lib using `#[serde(into=...,from=...)]`

jnbooth:
Per #943, adds serialization and deserialization for a number of types, using various strategies.

Using `#[derive(Serialize, Deserialize)]`:
- `QLine` / `QLineF`
- `QMargins` / `QMarginsF`
- `QPoint` / `QPointF`
- `QRect` / `QRectF`
- `QSize` / `QSizeF`
- `QVector2D` / `QVector3D` / `QVector4D`

Proxied to `QString`:
- `QDate`, `QDateTime`, and `QTime` (ISO-8601)
- `QFont`
- `QUrl`
- `QUuid`
- `QColor` (hex encoding)

Using a macro for serializing and deserializing Qt sequential containers:
- `QList`
- `QPolygon` / `QPolygonF`
- `QSet`
- `QStringList`
- `QVector`

The goal of this PR is to avoid handwritten serde logic as much as possible in order to minimize maintenance cost. `#[derive(Serialize, Deserialize)]` does expose field names, but that is [intended behavior](https://github.com/KDAB/cxx-qt/issues/943#issuecomment-2074978777). (If it's a problem, we can use [#[serde(rename)]](https://serde.rs/field-attrs.html#rename).) `QString` proxies are used for `QColor`, `QFont`, `QUuid`, `QUrl`, and date/time types because a) it offloads ser/de logic to Qt, and b) the `chrono` and `uuid` crates use string proxies for ser/de anyway, so if we used them as proxies, it'd have the same effect. Ser/de has not been added for `QHash`/`QMap` because they currently use `QVariant` values, which aren't serializable except by turning them into non-human-readable byte arrays. If we add more `QHashPair` and `QMapPair` specializations, that could be worth revisiting.

In order to write round-trip unit tests, there needs to be a dev-dependency for a serde format. Currently the unit tests use [serde_json](https://crates.io/crates/serde_json).

**Note:** This PR also adds the missing`QSet::reserve(&mut self, size: isize)` function in order to reserve capacity before deserialization. That's where most of the changed files are coming from. It also adds a few missing `Debug` implementations so that `assert_eq!` works with those types.

Closes #943


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `QDateTime::from_string` to parse `QDateTime` from a `QString`.
 - Support for further types: `QUuid`
+- Serde support for further types: `QByteArray`, `QColor`, `QDate`, `QDateTime`, `QLine`, `QLineF`, `QList`, `QMargins`, `QMarginsF`, `QPoint`, `QPointF`, `QPolygon`, `QPolygonF`, `QRect`, `QRectF`, `QSet`, `QSize`, `QSizeF`,  `QStringList`, `QVector`, `QVector2D`, `QVector3D`, `QVector4D`, `QTime`, `QUrl`, `QUuid`
```
redstrate, line 24: We should also mention QSet::reserve here too

crates/cxx-qt-lib/src/serde_impl.rs
```diff
@@ -0,0 +1,190 @@
+// SPDX-FileCopyrightText: 2022 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
```
redstrate, line 1: nit: this should be 2025, unless you invented a time machine :)

redstrate (commented):
IMO it would be nice to have some of these useful doc and other additions unrelated to serde now, but I'll leave that up to Andrew & Leon if they want that separated :)

redstrate:
This should fix #943 right?

jnbooth:
@redstrate I believe so.

crates/cxx-qt-lib/src/serde_impl.rs
```diff
@@ -0,0 +1,190 @@
+// SPDX-FileCopyrightText: 2025 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
```
LeonMatthesKDAB, line 1: Hm, the implementations in this file probably work, however, as mentioned in #943 , the goal with our serde support is to keep maintenance effort low.

We should try to replace these implementations with `from` and `into` implementations wherever possible.

LeonMatthesKDAB (commented):
Hi @jnbooth , thank you for the contribution!

I haven't yet had the time to give this PR a full review, as it's quite large.

However, what I noticed is that there are more manual Deserialize/Serialize implementations than I would like to support in the future.
It is good that some of these already proxy through e.g. QString, however it would be best if we can use serdes `from` and `into` attributes for this, as we then can get rid of a lot of boilerplate.

See also: https://serde.rs/container-attrs.html#from

We may even want to implement e.g. `From` and `Into` for `Vec<>`, `Hashmap<>`, etc. to get rid of the manual container implementations.

jnbooth:
@LeonMatthesKDAB Assuming this is what you wanted me to do, I've replaced serde implementations with the following implementations:

- `From<&QFont> for QString` + `From<QFont> for QString`
- `TryFrom<&QString> for QFont` + `TryFrom<QString> for QFont`
- `From<QString> for QUrl`
- `From<&QUrl> for QString` + `From<QUrl> for QString`
- `From<QString> for QUuid`
- `From<QString> for QColor`
- `From<&QColor> for QString` + `From<QColor> for QString`
- `From<&QByteArray> for Vec<u8>` + `From<QByteArray> for Vec<u8>`
- `From<Vec<u8>> for QByteArray`
- `TryFrom<&QString> for QDate` + `TryFrom<QString> for QDate`
- `From<&QDate> for QString` + `From<QDate> for QString`
- `TryFrom<&QString> for QDateTime` + `TryFrom<QString> for QDateTime`
- `From<&QDateTime> for QString` + `From<QDateTime> for QString`
- `TryFrom<&QString> for QTime` + `TryFrom<QString> for QTime`
- `From<&QTime> for QString` + `From<QTime> for QString`

However, I think there are some drawbacks to this approach that might make it worth reverting.
1. That's a lot of new trait implementations.
2. All those implementations are tightly coupled to (de)serialization. If someone submits a pull request that changes any of those implementations, it will break deserialization. That's a lot of code fragility.
3. Since serde's `from` and `into` attributes use `.clone()`, every call to serialization and deserialization now starts with `.clone()`. It's especially bad in the cases of `QString` and `Vec<u8>`, which now allocate unnecessary strings and buffers.
4. This isn't less code to maintain. It's more.

crates/cxx-qt-lib/src/core/qstringlist.rs
```diff
@@ -215,4 +215,13 @@ mod test {
...
     }
+
+    #[cfg(feature = "serde")]
+    #[test]
+    fn qstringlist_serde() {
+        let mut qstringlist = QStringList::default();
+        qstringlist.append(QString::from("element 1"));
+        qstringlist.append(QString::from("element 2"));
+        assert_eq!(crate::serde_impl::roundtrip(&qstringlist), qstringlist)
+    }
```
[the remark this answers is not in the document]
jnbooth: It's implemented using the containers macro in serde_impl.rs.

crates/cxx-qt-lib/src/core/qtime.rs
```diff
@@ -140,11 +140,31 @@ impl QTime {
...
 
+    /// Returns the QTime represented by the string, using the format given, or None if the string cannot be parsed.
+    pub fn from_string_opt(string: &ffi::QString, format: &ffi::QString) -> Option<Self> {
+        let time = ffi::qtime_from_string(string, format);
+        if time.is_valid() {
+            Some(time)
+        } else {
+            None
+        }
+    }
```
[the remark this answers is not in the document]
jnbooth: (Side note: I've noticed in libraries such as chrono that if a function returns `Option<Self>` rather than `Self`, they add `_opt` to the name.)

LeonMatthesKDAB:
@jnbooth We posted an idea on how to get this merged on Zulip:
https://cxx-qt.zulipchat.com/#narrow/channel/426350-cxx-qt-lib/topic/serde.20support/near/500690173

Let me know if you want me to copy the discussion here on Github.

jnbooth:
@LeonMatthesKDAB Thanks for making that channel! No need to copy it here, I can participate in it directly.

PR #1156, opened 2025-01-14
qt-build-utils: Add an option to use CMake instead of QMake

### LINKED ISSUE #1153: CXX-Qt does not work in split-Qt installations
I was working on integrating Rust and a CXX-Qt application into [KDE's Craft](https://community.kde.org/Craft) build/package system, specifically for building an Android application. I managed to get somewhat far, but I'm now hitting a brick wall. This is half-question, half-notes to myself.

### Selected qmake binary

I needed to set `QMAKE` manually, which we _do_ have in the documentation which is extremely helpful. However, it's weird that the qmake you find in CMake (through the `Qt::qmake` target) can be completely different than what cxx-qt uses by default. In Craft, we have two `qmake` binaries: one for the host (x86_64) and one belonging to the actual Qt installation used for cross-compilation.

The `Qt::qmake` targets points to the correct qmake to use, but cxx-qt picks the host's version which is very wrong and ends up pointing the wrong Qt.

### The Qt installation is split

In Craft, we split Qt into separate packages that are installed separately. This is the same split as in git, so `qtbase` contains `QtCore` and it's sister modules and `qtdeclarative` contains `QtQml`, `QtQuickControls2`, and so on. You can start to see where the problem is.

In cxx-qt, we use `qmake` to query where the headers are with the `QT_INSTALL_HEADERS` qmake built-in. Except that this only has one path(?), which is the one from qtbase. This means that it's currently impossible to build cxx-qt out of the box, as far as I'm aware.

```shell
CRAFT: user@5d0ff04fdbc2:~/CraftRoot$ ls /home/user/CraftRoot/build/libs/qt6/qtbase/image-MinSizeRel-6.8.0/include
QtConcurrent  QtDeviceDiscoverySupport  QtExamplesAssetDownloader  QtGui           QtNetwork  QtOpenGLWidgets  QtSql   QtWidgets
QtCore        QtExampleIcons            QtFbSupport                QtInputSupport  QtOpenGL   QtPrintSupport   QtTest  QtXml
CRAFT: user@5d0ff04fdbc2:~/CraftRoot$ ls /home/user/CraftRoot/build/libs/qt6/qtdeclarative/image-MinSizeRel-6.8.0/include
QtLabsAnimation        QtQmlAssetDownloader  QtQmlNetwork                    QtQuickControls2FluentWinUI3StyleImpl  QtQuickControls2UniversalStyleImpl  QtQuickTemplates2
QtLabsFolderListModel  QtQmlCompiler         QtQmlToolingSettings            QtQuickControls2Fusion                 QtQuickControlsTestUtils            QtQuickTest
QtLabsPlatform         QtQmlCore             QtQmlTypeRegistrar              QtQuickControls2FusionStyleImpl        QtQuickDialogs2                     QtQuickTestUtils
QtLabsQmlModels        QtQmlDebug            QtQmlWorkerScript               QtQuickControls2Imagine                QtQuickDialogs2QuickImpl            QtQuickVectorImage
QtLabsSettings         QtQmlDom              QtQmlXmlListModel               QtQuickControls2ImagineStyleImpl       QtQuickDialogs2Utils                QtQuickVectorImageGenerator
QtLabsSharedImage      QtQmlIntegration      QtQuick                         QtQuickControls2Impl                   QtQuickEffects                      QtQuickWidgets
QtLabsWavefrontMesh    QtQmlLocalStorage     QtQuickControls2                QtQuickControls2Material               QtQuickLayouts
QtPacketProtocol       QtQmlMeta             QtQuickControls2Basic           QtQuickControls2MaterialStyleImpl      QtQuickParticles
QtQml                  QtQmlModels           QtQuickControls2BasicStyleImpl  QtQuickControls2Universal              QtQuickShapes
```
(cxx-qt-build only includes `/home/user/CraftRoot/build/libs/qt6/qtbase/image-MinSizeRel-6.8.0/include`, so stuff like `QQmlApplication` can't build out of the box.)

In cxx-kde-frameworks, we have started using [cmake-package-rs](https://github.com/danvratil/cmake-package-rs) to use the existing CMake modules from KDE Frameworks, I wonder if that'll work here too? That way we can depend less on qmake too.

redstrate:
This is needed for scenarios where the CMake packaging for Qt is better than what QMake can handle, e.g. split Qt installations where the modules live in different directories.

A new optional feature is added to cxx-qt-build to prefer CMake for most operations related to linking/including Qt modules. QMake is still the default.

Fixes #1153.

### DIFF
--- a/crates/qt-build-utils/Cargo.toml
+++ b/crates/qt-build-utils/Cargo.toml
@@ -17,6 +17,7 @@ rust-version.workspace = true
 cc.workspace = true
 versions = "6.3"
 thiserror.workspace = true
+cmake-package = { version = "0.1.5", optional = true }
 
 [features]
 # When Cargo links an executable, whether a bin crate or test executable,
@@ -30,6 +31,10 @@ thiserror.workspace = true
 #
 # When linking Qt dynamically, this makes no difference.
 link_qt_object_files = []
+
+# Prefer using CMake to find Qt modules, instead of using qmake.
+# This may be desirable in certain installations.
+cmake = ["dep:cmake-package"]
 
 [lints]
 workspace = true
--- a/crates/qt-build-utils/src/lib.rs
+++ b/crates/qt-build-utils/src/lib.rs
@@ -558,6 +558,30 @@ impl QtBuild {
     /// Get the include paths for Qt, including Qt module subdirectories. This is intended
     /// to be passed to whichever tool you are using to invoke the C++ compiler.
     pub fn include_paths(&self) -> Vec<PathBuf> {
+        #[cfg(feature = "cmake")]
+        {
+            let Ok(package) = cmake_package::find_package(format!("Qt{}", self.version.major))
+                .components(self.qt_modules.clone())
+                .find()
+            else {
+                return Vec::default();
+            };
+
+            let mut paths = Vec::new();
+            for qt_module in &self.qt_modules {
+                if let Some(target) = package.target(format!("Qt::{}", qt_module)) {
+                    paths.extend(target.include_directories);
+                }
+            }
+
+            return paths
+                .iter()
+                .map(PathBuf::from)
+                // Only add paths if they exist
+                .filter(|path| path.exists())
+                .collect();
+        }
+
         let root_path = self.qmake_query("QT_INSTALL_HEADERS");
         let lib_path = self.qmake_query("QT_INSTALL_LIBS");
         let mut paths = Vec::new();
@@ -591,6 +615,27 @@ impl QtBuild {
     /// Lazy load the path of a Qt executable tool
     /// Skip doing this in the constructor because not every user of this crate will use each tool
     fn get_qt_tool(&self, tool_name: &str) -> Result<String, ()> {
+        #[cfg(feature = "cmake")]
+        {
+            // QML tools live in the QML target, of course
+            let tools_component = if tool_name.contains("qml") {
+                "QmlTools"
+            } else {
+                "CoreTools"
+            };
+
+            let Ok(package) = cmake_package::find_package(format!("Qt{}", self.version.major))
+                .components([tools_component.into()])
+                .find()
+            else {
+                return Err(());
+            };
+            let Some(target) = package.target(format!("Qt6::{}", tool_name)) else {
+                return Err(());
+            };
+            return target.location.ok_or(());
+        }
+
         // "qmake -query" exposes a list of paths that describe where Qt executables and libraries
         // are located, as well as where new executables & libraries should be installed to.
         // We can use these variables to find any Qt tool.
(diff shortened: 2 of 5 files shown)


crates/qt-build-utils/Cargo.toml
```diff
@@ -17,6 +17,7 @@ rust-version.workspace = true
 cc.workspace = true
 versions = "6.3"
 thiserror.workspace = true
+cmake-package = { path = "/home/josh/sources/cmake-package-rs", optional = true }
```
redstrate, line 20: Temporary hack, waiting for https://github.com/danvratil/cmake-package-rs/pull/48 to land upstream.

redstrate: Fixed, 0.1.5 is released with my changes

crates/qt-build-utils/src/lib.rs
```diff
@@ -558,6 +558,30 @@ impl QtBuild {
...
+            };
+
+            let mut paths = Vec::new();
+            for qt_module in &self.qt_modules {
+                if let Some(target) = package.target(format!("Qt::{}", qt_module)) {
+                    paths.extend(target.include_directories);
+                }
+            }
+
+            return paths
```
ahayzen-kdab, line 577: this makes the code after it unreachable

crates/qt-build-utils/src/lib.rs
```diff
@@ -591,6 +615,27 @@ impl QtBuild {
...
+            let Ok(package) = cmake_package::find_package(format!("Qt{}", self.version.major))
+                .components([tools_component.into()])
+                .find()
+            else {
+                return Err(());
+            };
+            let Some(target) = package.target(format!("Qt6::{}", tool_name)) else {
+                return Err(());
+            };
+            return target.location.ok_or(());
```
ahayzen-kdab, line 636: same here

crates/qt-build-utils/src/lib.rs
ahayzen-kdab: I think this is really interesting, i wonder if the approach here could be.

- Make the qmake mode a feature
- Have the fallbacks all by `unimplemented!()`
- Default qt-build-utils to enable the qmake feature
- Potentially change things to be trait driven for implementations here? (see below)
- Add code for a CMake feature
- Potentially change the default in the future easily

Whether as part of the refactor @LeonMatthesKDAB we change the code of qt-build-utils to be trait driven, so instead of every function having a two cfg blocks and a unimplemented instead there is a trait that is implemented by either CMake/QMake implementations ? Then the question is whether the feature is even needed and both are always built, but maybe the feature chooses the "default" implementation and you can still manually force one of them...

ahayzen-kdab:
So FWIW as part of the build system refactor we now have the `qmake` implementation split out. We now have

- `QtInstallation` trait https://github.com/KDAB/cxx-qt/blob/main/crates/qt-build-utils/src/installation/mod.rs
- `QtInstallationQMake` implementation https://github.com/KDAB/cxx-qt/blob/main/crates/qt-build-utils/src/installation/qmake.rs

This code is still evolving as we continue to refactor things, and for now cxx-qt-build is going to be using the qmake one. But at least we can start to see a clear path forward once the refactor is complete where you might have a cmake QtInstallation implementation, then a feature on qt-build-utils, and either a feature or opt-in somewhere in cxx-qt-build potentially (but unsure how this would interact if mixing crates).

PR #1157, opened 2025-01-16
QImage: Add conversion from image::RgbaImage

LeonMatthesKDAB:
Because the image crate changes versions somewhat frequently, the
features in CXX-Qt are added with explicit image crate versions.

### DIFF
--- a/crates/cxx-qt-lib/Cargo.toml
+++ b/crates/cxx-qt-lib/Cargo.toml
@@ -31,13 +31,19 @@ time = { version = "0.3.20", optional = true }
 url = { version = "2.3", optional = true }
 uuid = { version = "1.1.0", optional = true }
 serde = { version = "1", features=["derive"], optional = true }
+# Note: The image crate is not yet at 1.0 and a new version is released regularly.
+# To avoid a breaking change at each update, we don't specify a default version, and make the versions explicit.
+# Once 1.0 is released, we can add a dependency on `image`, which would then be `image = "1"`
+# Also, disable the default features, we don't need any, so they would be included unnecessarily and raise the MSRV unnecessarily.
+image-v0-24 = { version = "0.24", optional = true, package = "image", default-features = false }
+image-v0-25 = { version = "0.25", optional = true, package = "image", default-features = false }
 
 [build-dependencies]
 cxx-qt-build.workspace = true
 qt-build-utils.workspace = true
 
 [features]
-full = ["qt_full", "serde", "url", "uuid", "time", "rgb", "http", "chrono", "bytes"]
+full = ["qt_full", "serde", "url", "uuid", "time", "rgb", "http", "chrono", "bytes", "image-v0-24", "image-v0-25"]
 default = []
 
 qt_full = ["qt_gui", "qt_qml", "qt_quickcontrols"]
@@ -53,6 +59,8 @@ time = ["dep:time"]
 url = ["dep:url"]
 serde = ["dep:serde"]
 uuid = ["dep:uuid"]
+image-v0-24 = ["dep:image-v0-24"]
+image-v0-25 = ["dep:image-v0-25"]
 link_qt_object_files = ["cxx-qt-build/link_qt_object_files"]
 
 [lints]
--- a/crates/cxx-qt-lib/src/gui/qimage.rs
+++ b/crates/cxx-qt-lib/src/gui/qimage.rs
@@ -83,6 +83,8 @@ mod ffi {
         include!("cxx-qt-lib/qsizef.h");
         #[allow(dead_code)]
         type QSizeF = crate::QSizeF;
+        type QImageCleanupFunction = super::QImageCleanupFunction;
+        type uchar;
 
         /// Returns true if all the colors in the image are shades of gray
         #[rust_name = "all_gray"]
@@ -236,6 +238,7 @@ mod ffi {
         include!("cxx-qt-lib/common.h");
         type QImageFormat;
         type QImageInvertMode;
+        type c_void = crate::c_void;
 
         #[doc(hidden)]
         #[rust_name = "qimage_init_default"]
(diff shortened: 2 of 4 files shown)


crates/cxx-qt-lib/Cargo.toml
```diff
@@ -53,6 +59,8 @@ time = ["dep:time"]
 url = ["dep:url"]
 serde = ["dep:serde"]
 uuid = ["dep:uuid"]
+imagev24 = ["dep:imagev24"]
+imagev25 = ["dep:imagev25"]
```
ahayzen-kdab, line 63: this does continue to raise the fun in the Rust ecosystem due to the orphan rules around which crate should implement things and how many different features we should support for the types. Would still be nice if developers could easily add their own implementations easily

crates/cxx-qt-lib/src/gui/qimage.rs
```diff
@@ -332,6 +371,73 @@ impl QImage {
...
+
+    /// Constructs a QImage from the raw data inside a `Vec<u8>`.
+    ///
+    /// # Safety
+    /// This function is unsafe because it requires that the data matches the given QImageFormat,
+    /// width and height.
+    ///
+    /// It is a convenience function around [Self::from_raw_parts_mut].
+    pub unsafe fn from_raw_bytes(
+        data: Vec<u8>,
```
ahayzen-kdab, line 427: should this be a `&[u8]` as the Qt docs say it is read-only and in theory the lifetime out data should outlive the lifetime of the QImage.

But here we don't have any lifetime on QImage, similar to the QByteArray situation it raises the question if the Rust API should have two structs like a QImage and QImageRef in the future. Maybe we should raise an issue somewhere to discuss this and list the places where this could be useful?

crates/cxx-qt-lib/src/gui/qimage.rs
```diff
@@ -332,6 +371,73 @@ impl QImage {
...
+            QImageCleanupFunction(image_cleanup_function),
+            image_cleanup_data,
+        )
+    }
+
+    /// Constructs a QImage from the raw data inside a `Vec<u8>`.
+    ///
+    /// # Safety
+    /// This function is unsafe because it requires that the data matches the given QImageFormat,
+    /// width and height.
```
ahayzen-kdab, line 423: and that the data lives long enough?

ahayzen-kdab:
Looks good, small questions and conflicts

crates/cxx-qt-lib/src/gui/qimage.rs
```diff
@@ -332,6 +371,73 @@ impl QImage {
...
+    ///
+    /// # Safety
+    /// For details on safety see the [Qt documentation](https://doc.qt.io/qt-6/qimage.html#QImage-7)
+    pub unsafe fn from_raw_parts(
+        data: *const ffi::uchar,
+        width: i32,
+        height: i32,
+        format: QImageFormat,
+        image_cleanup_function: extern "C" fn(x: *mut ffi::c_void),
+        image_cleanup_data: *mut ffi::c_void,
```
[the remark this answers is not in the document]
LeonMatthesKDAB: Renamed it to `cleanup_function` and `cleanup_info` to be consistent with Qt.

crates/cxx-qt-lib/src/gui/qimage.rs
LeonMatthesKDAB: This isn't necessary here. As this function takes a `Vec<u8>` the QImage gets ownership of the data.

crates/cxx-qt-lib/src/gui/qimage.rs
LeonMatthesKDAB: This function explictily takes a `Vec<u8>` so that we can pass the ownership of the data to the QImage, because we can't model the lifetime otherwise.

I also don't think it's really necessary to support a non-owning version of QImage, as the point of QImage is mostly to transfer ownership to Qt.
And as soon as we pass the FFI bridge, the lifetimes don't really mean much anyway.
So I'd rather just support this owning version then start to introduce lifetimes.

Also, side note: this version can actually modify the data, as it uses the non-const version of the constructor: https://doc.qt.io/qt-6/qimage.html#QImage-8 (i.e. `from_raw_parts_mut`).

crates/cxx-qt-lib/Cargo.toml
LeonMatthesKDAB: Yeah, this part is a bit annoying.
But I guess this is the best way forward for now?

ahayzen-kdab: yup was more an observation

crates/cxx-qt-lib/src/gui/qimage.rs
ahayzen-kdab: Ah i see so you are using the cleanup function to free it

crates/cxx-qt-lib/src/gui/qimage.rs
ahayzen-kdab: :+1:

ahayzen-kdab:
Needs some conflicts resolving otherwise looks good i think.

crates/cxx-qt-lib/src/gui/qimage.rs
```diff
@@ -332,6 +371,73 @@ impl QImage {
...
+    ///
+    /// It is a convenience function around [Self::from_raw_parts_mut].
+    pub unsafe fn from_raw_bytes(
+        data: Vec<u8>,
+        width: i32,
+        height: i32,
+        format: QImageFormat,
+    ) -> Self {
+        extern "C" fn delete_boxed_vec(boxed_vec: *mut ffi::c_void) {
+            drop(Box::new(boxed_vec as *mut Vec<u8>));
```
milianw, line 433: this looks bogus to me, can you explain it please? isn't `Box::new()` like a `std::make_unique`, meaning it does a new memory allocation? Or is this a misnomer and it actually adopts the ptr that gets passed in, so more like `std::unique_ptr<T>(...)`?

LeonMatthesKDAB: Very good catch, thank you!

It needs to be `Box::from_raw`. Should be fixed now.

ahayzen-kdab: Curious that valgrind didn't spot this, or are only the Qt unit tests not the `cargo test` run under valgrind ? Maybe this should be added to the qt_types_standalone or we should run the `cargo test` (even just for cxx-qt-lib) under valgrind ? :thinking:

LeonMatthesKDAB: I just checked, it indeed did not find the issue.
It seems to be that we really don't run `cargo test` under valgrind, but just the tests that we run via C++ (e.g. qml_features_test, etc.)

LeonMatthesKDAB: We could try using: https://github.com/jfrimmel/cargo-valgrind
But that would probably be better served in a different PR, as that means we might want to remove/reduce the qt_types_standalone tests as well :thinking:

milianw: as an outsider: can you use sanitizers instead? they are much faster and also have leak checking built-in - that should be preferred over valgrind nowadays

ahayzen-kdab (approved):
Looks good, @LeonMatthesKDAB would still be good to raise an issue around using sanitizers or valgrind of some kind on the cxx-qt-lib tests.

crates/cxx-qt-lib/src/gui/qimage.rs
LeonMatthesKDAB: Done as #1177

ahayzen-kdab (dismissed):
LGTM :-)

PR #1159, opened 2025-01-16
Implement automatic implementations of up and downcasting

BenFordTytherington:


crates/cxx-qt-gen/src/generator/rust/mod.rs
```diff
@@ -95,6 +97,29 @@ impl GeneratedRustBlocks {
...
+    let includes = qobjects
+        .iter()
+        .any(|obj| obj.has_qobject_macro && obj.base_class.is_none());
+    if includes {
+        GeneratedRustFragment {
+            cxx_mod_contents: vec![parse_quote! {
+                extern "C++" {
+                    #[doc(hidden)]
+                    #[namespace=""]
+                    type QObject = cxx_qt::qobject::QObject;
```
ahayzen-kdab, line 110: what happens if you have the same type twice in CXX :thinking: as wondering if the developer would ever write `type QObject` themselves too ...

crates/cxx-qt-gen/src/generator/rust/qobject.rs
```diff
@@ -86,6 +83,64 @@ impl GeneratedRustFragment {
...
+        let base = structured_qobject
+            .declaration
+            .base_class
+            .as_ref()
+            .map(|name| type_names.lookup(name))
+            .transpose()?
+            .cloned()
+            .unwrap_or(
+                Name::new(format_ident!("QObject")).with_module(parse_quote! {::cxx_qt::qobject}),
+            ); // TODO! is this default module here causing the issues in the threading examples
```
ahayzen-kdab, line 97: is this still a problem? or did you solve it?

crates/cxx-qt/src/lib.rs
```diff
@@ -9,10 +9,13 @@
 //!
 //! See the [book](https://kdab.github.io/cxx-qt/book/) for more information.
 
+use std::ops::Deref;
+use std::pin::Pin;
 use std::{fs::File, io::Write, path::Path};
 
 mod connection;
 mod connectionguard;
+pub mod qobject;
```
ahayzen-kdab, line 18: do we want developers to go `cxx_qt::qobject::QObject` or `cxx_qt::QObject` ? At the moment for cxx-qt-lib everything is in the root module, but we wondered if they should be split by core/gui/quick etc, wonder what we do here?

crates/cxx-qt/src/qobject.rs
```diff
@@ -0,0 +1,22 @@
...
+// SPDX-FileContributor: Ben Ford <ben.ford@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! QObject module
+
+#[cxx::bridge]
+mod ffi {
+    unsafe extern "C++" {
+        // TODO! Implement QObject wrapper properly
```
ahayzen-kdab, line 11: Right and then we reach the `QObject::thread() -> QThread` method, which @LeonMatthesKDAB might help with that threading idea :-)

crates/cxx-qt-gen/src/generator/rust/mod.rs
LeonMatthesKDAB: Maybe check if TypeNames already contains `QObject`?

crates/cxx-qt/src/qobject.rs
LeonMatthesKDAB: I should note that we likely will implement QObject via an extension trait in cxx-qt-lib, as many of the methods rely on types that we only get via cxx-qt-lib, and we don't want cxx-qt to depend on -lib.

ahayzen-kdab: Nice :-)

crates/cxx-qt-gen/src/generator/rust/mod.rs
ahayzen-kdab: But then what if a developer does `type QObject = mycrate::MyQObjectImpl;` instead of to `cxx_qt::qobject::QObject` ? :thinking: 

Elsewhere we have been doing this for types that we inject into the bridge in cxx-qt-gen
```
        #[doc(hidden)]
        #[namespace = "Qt"]
        #[rust_name = "CxxQtConnectionType"]
        #[allow(dead_code)]
        type ConnectionType = cxx_qt::ConnectionType;
```

Eg we provide a different Rust name so they don't collide and the type is still the same.

Maybe we should just do that here too ?

So use

```
#[rust_name = "CxxQtQObject"]
type QObject = ...;
```

Then when in the bridge refer to it as `CxxQtQObject` and outside the bridge it can still be `cxx_qt::qobject::QObject` as usual

crates/cxx-qt/src/lib.rs
```diff
@@ -121,10 +124,62 @@ pub trait Threading: Sized {
...
+        let this = self.deref() as *const Self;
+        unsafe {
+            let base = Self::upcast_ptr(this) as *mut T;
+            Pin::new_unchecked(&mut *base)
+        }
+    }
+}
+
+/// Trait for downcasting to a subclass, provided the subclass implements Upcast to this type
+pub trait Downcast: Sized {
```
ahayzen-kdab, line 168: should we have the downcast pint / mut variants here too ?

crates/cxx-qt-lib/src/core/qstringlist.cpp
```diff
@@ -36,18 +36,6 @@ static_assert(QTypeInfo<QStringList>::isRelocatable);
 namespace rust {
 namespace cxxqtlib1 {
 
-const QList<QString>&
-qstringlistAsQListQStringRef(const QStringList& list)
```
ahayzen-kdab, line 40: there will be a header verison that needs removing too in `qstringlist.h`

crates/cxx-qt-gen/src/generator/rust/mod.rs
ahayzen-kdab: Maybe just didn't know if we are hiding "magic" that the QObject type is implicitly there. And we have found a way around this before with the ConnectionType so thought we could copy that :thinking: 

As then the type externally to the bridge is still the same, but it allows the dev to write `type QObject = ...` themselves too :thinking: But this is a general problem we have, what happens if the same type is defined twice.

BenFordTytherington: In this case, CXX errors with `type ... is defined multiple times`, so are we just going to implicitly include it and write a doc that it doesn't need to be included manually?

crates/cxx-qt/src/lib.rs
BenFordTytherington: Added

crates/cxx-qt-lib/src/core/qstringlist.cpp
BenFordTytherington: Removed

crates/cxx-qt-lib/src/core/qstringlist.rs
```diff
@@ -180,17 +193,30 @@ impl From<&QStringList> for QList<QString> {
     }
 }
 
+#[cfg(cxxqt_qt_version_at_least_6)]
```
ahayzen-kdab, line 196: ugh it's unfortunate that Qt 5 seems different enough that the casting doesn't work.

Wonder how annoying it becomes to not be able to at least deref to a QList<QString> as yes then this is what is causing things like .append / .get to fail in the tests (and there are further tests too) that will only work with Qt 6.

So maybe that points that this would be painful without deref ability and you should bring back the old behaviour but only for Qt 5. So you need to use `cfg(cxxqt_qt_version_major = "5")` on the Rust side and `#if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))` on the C++ side

crates/cxx-qt-lib/src/core/qstringlist.rs
```diff
@@ -180,17 +193,30 @@ impl From<&QStringList> for QList<QString> {
...
 impl Deref for QStringList {
```
ahayzen-kdab, line 197: side question, do we need the manual Deref when we have Upcast already ? Or can that be templated somehow if the type implements Upcast<T> ?

crates/cxx-qt-lib/src/gui/qguiapplication.rs
```diff
@@ -98,7 +116,19 @@ mod ffi {
     impl UniquePtr<QGuiApplication> {}
 }
 
-pub use ffi::QGuiApplication;
+pub use ffi::{
+    downcast_qcoreapplication, upcast_qguiapplication, QCoreApplication, QGuiApplication,
+};
+
+impl Upcast<QCoreApplication> for QGuiApplication {
```
ahayzen-kdab, line 123: nice, `QQmlApplicationEngine` -> `QQmlEngine` would be another one

crates/cxx-qt-lib/src/core/qstringlist.rs
BenFordTytherington: I think this was something me and Leon discussed, there's already a blanket Deref implementation for QObjects which is used for the inner rust type I believe, so doing this would only be useful for a certain subset of objects I guess.

crates/cxx-qt/src/lib.rs
LeonMatthesKDAB: I'm okay with exporting it as `cxx_qt::QObject`.
I believe this was done to go in the direction that we want cxx-qt-lib to go, which is to start using submodules more.

But that's indeed inconsistent at the moment, so let's do a top-level export and fix it if we ever get around to exporting things from cxx-qt-lib via submodules.

crates/cxx-qt-lib/src/qml/qqmlapplicationengine.cpp
```diff
@@ -16,11 +16,5 @@ qqmlapplicationengineNew()
   return ::std::make_unique<QQmlApplicationEngine>();
 }
 
-QQmlEngine&
-qqmlapplicationengineAsQQmlEngine(QQmlApplicationEngine& engine)
```
ahayzen-kdab, line 20: there should be a header side of this to remove too i think ?

crates/cxx-qt/include/casting.h
```diff
@@ -0,0 +1,26 @@
+// SPDX-FileCopyrightText: 2025 Klarälvdalens Datakonsult AB, a KDAB Group
+// company <info@kdab.com> SPDX-FileContributor: Ben Ford <ben.ford@kdab.com>
```
ahayzen-kdab, line 2: clang format likes to reformat SPDX lines instead use
```
// clang-format off
// SPDX-FileCopyrightText: 2025 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
// clang-format on
// SPDX-FileContributor: Ben Ford <ben.ford@kdab.com>
//
```

crates/cxx-qt-lib/src/core/qstringlist.rs
```diff
@@ -25,6 +27,21 @@ mod ffi {
         include!("cxx-qt-lib/qstringlist.h");
         type QStringList = super::QStringList;
 
+        include!("cxx-qt/casting.h");
+
+        #[doc(hidden)]
+        #[rust_name = "upcast_qstringlist"]
+        #[cxx_name = "upcastPtr"]
+        #[namespace = "rust::cxxqt1"]
+        unsafe fn upcast(thiz: *const QStringList) -> *const QList_QString;
```
ahayzen-kdab, line 36: typo should be `this`

crates/cxx-qt-lib/src/qml/qqmlapplicationengine.rs
```diff
@@ -88,13 +97,30 @@ mod ffi {
...
+
+    #[cfg(cxxqt_qt_version_at_least_6)]
+    unsafe fn from_base_ptr(base: *const QQmlEngine) -> *const Self {
+        ffi::downcast_qqml_engine(base)
+    }
+
+    #[cfg(cxxqt_qt_version_major = "5")]
+    unsafe fn from_base_ptr(_base: *const QQmlEngine) -> *const Self {
+        std::ptr::null()
+    }
```
ahayzen-kdab, line 117: why can't Qt 5 downcast ? It supposedly has a virtual on the deconstructor of QQmlEngine

ahayzen-kdab (commented):
Overall looks good, would be nice in a future PR to add support for `#[base = T]` to extern "C++Qt" types too so we can do the same tricks of up/down casting there

crates/cxx-qt-lib/src/core/qstringlist.rs
BenFordTytherington: Did we not need this to be thiz as it's a keyword in cpp?

ahayzen-kdab: Aha i see so it's deliberate to avoid collisions :-)

crates/cxx-qt/src/lib.rs
ahayzen-kdab, line 18: this module doesn't need to be `pub` now as we are reexporting

BenFordTytherington: Done

ahayzen-kdab (approved):
Looks good :-) @LeonMatthesKDAB any comments ? Ben is going to look at further changes like adding more docs to the book / adding support for base to extern C++Qt so that casing can be done on those types.

But i think this is probably far enough to land this part already?

book/src/bridge/traits.md
```diff
@@ -26,3 +26,7 @@ For further documentation, refer to the documentation of the individual traits:
 - [Constructor](https://docs.rs/cxx-qt/latest/cxx_qt/trait.Constructor.html) - custom constructor
 - [Initialize](https://docs.rs/cxx-qt/latest/cxx_qt/trait.Initialize.html) - execute Rust code when the object is constructed
 - [Threading](https://docs.rs/cxx-qt/latest/cxx_qt/trait.Threading.html) - marker trait whether CXX-Qt threading should be enabled
+- `Upcast` - Allows a type to access its parent class if there is one
```
ahayzen-kdab, line 29: can we add a link here to the docs.rs link ?

book/src/bridge/traits.md
```diff
@@ -26,3 +26,7 @@ For further documentation, refer to the documentation of the individual traits:
...
+  - ⚠️ Should not be implemented manually by the user
```
ahayzen-kdab, line 30: In theory for extern C++Qt / RustQt types yes the user/developer should not implement. However if you are building your own bindings like cxx-qt-lib you may need to manually implement them ?

Unsure how we communicate / phrase this :thinking:

book/src/concepts/casting.md
```diff
@@ -0,0 +1,39 @@
+<!--
+SPDX-FileCopyrightText: 2025 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+SPDX-FileContributor: Ben Ford <ben.ford@kdab.com>
+
+SPDX-License-Identifier: MIT OR Apache-2.0
+-->
+
+# Casting
```
ahayzen-kdab, line 8: wonder if we note somewhere how this only is implemented for RustQt types for now that have qobject attribute and optionally the base attribute.

As in a future PR support will be added for C++Qt (hopefully the next PR ;-) )

book/src/concepts/casting.md
```diff
@@ -0,0 +1,39 @@
...
+To do this, use the `Downcast` trait like so `use cxx_qt::Downcast`.
+The child can then be accessed in the same manner, with the following methods
+
+| Self Type        | Method           |
+|------------------|------------------|
+| `&self`          | `downcast()`     |
+| `&mut self`      | `downcast_mut()` |
+| `Pin<&mut self>` | `downcast_pin()` |
+
+These will return an `Option<T>`, as it is possible that downcasting will not be available,
```
LeonMatthesKDAB, line 38:
```suggestion
These will return an `Option<T>`, as it is possible that downcasting will fail, if the type is not actually of the given subclass,
```
The downcast is always "available", it will just fail if the type is not right.

LeonMatthesKDAB (commented):
Only some nitpicks left from my side.

Feel free to merge after these and the comments from @ahayzen-kdab are fixed.
Another really good step forward! :partying_face:

ahayzen-kdab (approved):
LGTM, support for C++Qt will be in another change and extending the QObject type etc. @LeonMatthesKDAB any objections ?

PR #1160, opened 2025-01-16
qt-build-utils: Add support for including headers from private modules

### LINKED ISSUE #955: Add private include paths
I'm trying to bind [QT Advanced Docking System](https://github.com/githubuser0xFFFF/Qt-Advanced-Docking-System/) but i'm getting the following error: 

```
Qt-Advanced-Docking-System/src/ads_globals.cpp:45:10: fatal error: qpa/qplatformnativeinterface.h: No such file or directory
    45 | #include <qpa/qplatformnativeinterface.h>
       |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

That header is in `QtGui/6.7.0/QtGui/qpa/qplatformnativeinterface.h` so I have a very hacky solution below that builds, but it'd be nice to have official support.

```rust
for path in qtbuild.include_paths() {
    let path = path.to_str().unwrap();
    if path.contains("QtGui") {
        cc.include(format!("{}/{}/QtGui", path, version));
    }
}
```

redstrate:
This allows you to include headers from private modules, e.g. "qpa/qplatformnativeinterface.h" from GuiPrivate. It does this by adding a special case for these modules, and appends the correct path.

Fixes #955

### DIFF
--- a/crates/qt-build-utils/src/lib.rs
+++ b/crates/qt-build-utils/src/lib.rs
@@ -489,6 +489,10 @@ impl QtBuild {
         };
 
         for qt_module in &self.qt_modules {
+            if qt_module.contains("Private") {
+                continue;
+            }
+
             let framework = if is_apple_target() {
                 Path::new(&format!("{lib_path}/Qt{qt_module}.framework")).exists()
             } else {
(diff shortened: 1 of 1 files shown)


crates/qt-build-utils/src/lib.rs
```diff
@@ -489,6 +489,10 @@ impl QtBuild {
         };
 
         for qt_module in &self.qt_modules {
+            if qt_module.contains("Private") {
```
ahayzen-kdab, line 492: Could you add a comment explaining how Private does not have a separate library and is just the normal library name.

I assume there is not a usecase of only putting like CorePrivate and *not* Core ? As this code currently requires you to specify the modules Core *and* CorePrivate, is this the same behaviour as with CMake ?

ahayzen-kdab:
Looks great overall, just some small questions

PR #1162, opened 2025-01-18
Implement Debug for all types that work with QDebug

jnbooth:
This PR adds `Debug` implementations to all types that a [QDebug](https://doc.qt.io/qt-6/qdebug.html) data stream accepts.

**Note:** Some types incorrectly used QDebug output to implement `Display` instead of `Debug`. This PR does not remove those implementations because that would be a breaking change, but they probably should be removed for the next major release.

### DIFF
--- a/crates/cxx-qt-lib/include/common.h
+++ b/crates/cxx-qt-lib/include/common.h
@@ -32,13 +32,6 @@ drop(T& value)
 
 template<typename T>
 QString
-toQString(const T& value)
-{
-  return value.toString();
-}
-
-template<typename T>
-QString
 toDebugQString(const T& value)
 {
   // We can't convert value directly into a string.
--- a/crates/cxx-qt-lib/include/common.h
+++ b/crates/cxx-qt-lib/include/common.h
@@ -32,6 +32,13 @@ drop(T& value)
 
 template<typename T>
 QString
+toQString(const T& value)
+{
+  return value.toString();
+}
+
+template<typename T>
+QString
 toDebugQString(const T& value)
 {
   // We can't convert value directly into a string.
--- a/crates/cxx-qt-lib/include/common.h
+++ b/crates/cxx-qt-lib/include/common.h
@@ -32,7 +32,7 @@ drop(T& value)
 
 template<typename T>
 QString
-toQString(const T& value)
+toDebugQString(const T& value)
 {
   // We can't convert value directly into a string.
   // However most Qt types are able to stream into a QDebug object such as
--- a/crates/cxx-qt-lib/src/core/qanystringview.rs
+++ b/crates/cxx-qt-lib/src/core/qanystringview.rs
@@ -61,10 +61,6 @@ mod ffi {
         #[doc(hidden)]
         #[rust_name = "QAnyStringView_len"]
         fn qanystringviewLen(string: &QAnyStringView) -> isize;
-
-        #[doc(hidden)]
-        #[rust_name = "QAnyStringView_to_qstring"]
-        fn toQString(string: &QAnyStringView) -> QString;
     }
 }
 
@@ -133,7 +129,7 @@ impl QAnyStringView<'_> {
 
 impl fmt::Display for QAnyStringView<'_> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}", ffi::QAnyStringView_to_qstring(self))
+        write!(f, "{}", self.to_qstring())
     }
 }
 
--- a/crates/cxx-qt-lib/src/core/qanystringview.rs
+++ b/crates/cxx-qt-lib/src/core/qanystringview.rs
@@ -7,6 +7,7 @@ use core::ffi::c_void;
 use core::marker::PhantomData;
 use core::mem::MaybeUninit;
 use cxx::{type_id, ExternType};
+use std::fmt;
 
 #[cxx::bridge]
 mod ffi {
--- a/crates/cxx-qt-lib/src/core/qanystringview.rs
+++ b/crates/cxx-qt-lib/src/core/qanystringview.rs
@@ -60,6 +60,10 @@ mod ffi {
         #[doc(hidden)]
         #[rust_name = "QAnyStringView_len"]
         fn qanystringviewLen(string: &QAnyStringView) -> isize;
+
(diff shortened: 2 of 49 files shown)


crates/cxx-qt-lib/src/core/qtimezone.rs
```diff
@@ -79,6 +111,13 @@ mod ffi {
         #[rust_name = "qtimezone_default"]
         fn qtimezoneDefault() -> UniquePtr<QTimeZone>;
         #[doc(hidden)]
+        #[rust_name = "qtimezone_display_name"]
+        fn qtimezoneDisplayName(
```
LeonMatthesKDAB, line 115: Is there a reason why this isn't wrapped as a member function?

If not, please replace it with a member function, then you can also remove the added code in the C++ header and source :)
e.g.:
```rust
        fn qtimezoneDisplayName(
            self: &QTimeZone,
            time_type: QTimeZoneTimeType,
            name_type: QTimeZoneNameType,
        ) -> QString;
```

crates/cxx-qt-lib/src/gui/qpolygon.cpp
```diff
@@ -32,3 +32,20 @@ static_assert(!::std::is_trivially_copy_constructible<QPolygon>::value);
...
+qpolygonAsQVectorQPointRef(const QPolygon& shape)
+{
+  return static_cast<const QVector<QPoint>&>(shape);
+}
+
+QVector<QPoint>&
+qpolygonAsQVectorQPointRef(QPolygon& shape)
+{
+  return static_cast<QVector<QPoint>&>(shape);
+}
```
LeonMatthesKDAB, line 48: I appreciate the effort here to also allow upcasting the QPolygon to it's base class `QList<QPoint>`.

Note that we will replace this implementation with the generalized version that @BenFordTytherington is working one once it lands:
https://github.com/KDAB/cxx-qt/pull/1159
For now we can keep it as it should be compatible API-wise

crates/cxx-qt-lib/src/core/qanystringview.rs
```diff
@@ -60,6 +61,10 @@ mod ffi {
         #[doc(hidden)]
         #[rust_name = "QAnyStringView_len"]
         fn qanystringviewLen(string: &QAnyStringView) -> isize;
+
+        #[doc(hidden)]
+        #[rust_name = "QAnyStringView_to_qstring"]
+        fn toQString(string: &QAnyStringView) -> QString;
```
LeonMatthesKDAB, line 67: There is already a `to_string` function on QAnyStringView (see line 34) which you can use.

crates/cxx-qt-lib/include/common.h
```diff
@@ -33,6 +33,13 @@ drop(T& value)
 template<typename T>
 QString
 toQString(const T& value)
```
LeonMatthesKDAB, line 35: This template should no longer be necessary.
If `toString()` is already a member function, we don't need to create a free function `toQString`, but can just access it directly via CXX:
i.e.:
```rust
// Maybe with #[doc(hidden)], depending on the situation
#[rust_name="to_qstring"]
fn toQString(self: &T);
```

Please remove this function and replace the existing references with direct member access.

LeonMatthesKDAB (changes requested):
Hi @jnbooth, thank you for this contribution!
It's great to see you're using CXX-Qt actively and I'm very happy to see the improvements that come out of it :smiley: 

This PR mostly looks good. I believe it can be simplified a little bit more, as the `toQString` function should actually not be necessary anymore, and there are a few nitpicks to take care of.

Then we should be able to merge this :partying_face:

crates/cxx-qt-lib/include/common.h
jnbooth: The idea of this template is to allow types to have more flexibility in how they define `to_qstring()`. Some types have extra parameters for `toString`. For others, we might not want to have a `to_qstring` method at all (rather than relying on `#[doc(hidden)]`. Or we might decide not to have methods like that at all and use `From<&Type> for QString`, which is more idiomatic. So by using this template, types aren't "locked in" to a particular implementation; we can keep the `Display` logic separate from everything else.

That said, I of course defer to you. What should be the rubric for whether or not to use `#[doc(hidden)]`?

crates/cxx-qt-lib/src/core/qtimezone.rs
jnbooth: It's because the [full method](https://doc.qt.io/qt-6/qtimezone.html#displayName)'s signature uses a `QLocale`, but we don't have a `QLocale`. Which causes this:

```
error: address of overloaded function 'displayName' does not match required type
'QString (QTimeZone::TimeType, QTimeZone::NameType) const'

  ::QString (::QTimeZone::*display_name$)(::QTimeZoneTimeType, ::QTimeZoneNameType) const
  = &::QTimeZone::displayName;

note: candidate function has different number of parameters (expected 2 but has 3)
```

LeonMatthesKDAB: Ah, that makes sense.

LeonMatthesKDAB:
Hi @jnbooth sorry for the late reply.

I'm still in favor of accessing the toString function directly.
They're part of Qts public API so adding them in `-lib` by directly wrapping them is fine.
We can always add `From`/`Into` implementations in addition.
Once that is changed and rebased I can approve the PR :)

jnbooth:
@LeonMatthesKDAB I've done that and merged in the latest changes, but now the tests seem to be broken and I don't know why...

ahayzen-kdab:
Appears CI is still failing due to a clippy warning
```
error: function `qfont_to_debug_qstring` is never used
   --> crates/cxx-qt-lib/src/gui/qfont.rs:363:23
    |
363 |         #[rust_name = "qfont_to_debug_qstring"]
    |                       ^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `-D dead-code` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(dead_code)]`
```

jnbooth:
@ahayzen-kdab Fixed!

LeonMatthesKDAB (approved):
Thank you for the contribution @jnbooth :rocket:

PR #1163, opened 2025-01-20
Add "Basics" tutorial

LeonMatthesKDAB:
This is the code developed during the second tutorial video:
https://www.youtube.com/watch?v=fdikaBcTgSM

It goes slightly beyond the minimal example, so add it as it's own
project.
It should also serve as a reference for backwards-compatibility.


ahayzen-kdab (approved):
LGTM :-)

PR #1164, opened 2025-01-21
Fix link_qt_object_files not working on Windows

jnbooth:
The current logic in qt-build-utils determines whether a file is an object file by testing whether it ends with ".o", but object files on Windows end in ".obj", resulting in numerous compiler warnings of "File path {} found in .prl file for {name}, but could not extract library base name to pass to linker command line". This PR fixes the issue.


LeonMatthesKDAB (approved):
Thank you @jnbooth for the contribution :)

PR #1167, opened 2025-01-23
CXX-Qt-build: Remove definitions from Interface

### LINKED ISSUE #1165: CXX-Qt-build: Remove ability to (re-)export definitions
We are not able to import definitions into CMake, so relying on definitions as compiler flags is problematic.

Instead, libraries that need certain definitions can generate a header file in their build script that includes the necessary definitions.
See also:
https://www.kdab.com/setting-defines-with-cmake/

LeonMatthesKDAB:
These definitions cannot be imported into CMake, as they are unknown at
configure time.
It is also best practice to keep definitions in a header file, as that
allows us to only include them where really needed, as pointed out by
this blog post: https://www.kdab.com/setting-defines-with-cmake/

This is the method that cxx-qt-lib now uses to configure its features.

Closes #1165


ahayzen-kdab (approved):
Nice! LGTM :-)

PR #1169, opened 2025-01-28
Use method calls to initialize dependencies

### LINKED ISSUE #1166: CXX-Qt-build: Refactor initializers to just call downstream dependencies
We noticed that recompiling the initializers can take quite a while, e.g. cxx-qt-lib registers quite a few types.

On small crates like qml_minimal, this actually takes up the vast majority of the build scripts runtime.

Instead, we should just compile the initializers into a callable function, which we then call from the downstream initializers.
That would drastically simplify the final initializer compilation.

It would also be a prerequisite for supporting something akin to [Q_INIT_RESOURCES](https://doc.qt.io/qt-5/qdir.html#Q_INIT_RESOURCE)/[Q_INIT_PLUGIN](https://doc.qt.io/qt-5/qtplugin.html#Q_IMPORT_PLUGIN).
As those macros would then just call the corresponding functions.

LeonMatthesKDAB:
This speeds up rebuilds, as we only need to link to upstream
dependencies initializers, not build their initializers every time.

It should also allow us to manually initialize certain libraries down
the line with something similar to Q_INIT_RESOURCE, etc.

Closes #1166


crates/cxx-qt-build/src/lib.rs
```diff
@@ -888,55 +886,92 @@ impl CxxQtBuilder {
...
+        std::fs::write(&initializers_file, initializer_code)
             .expect("Could not write initializers file");
+
+        // Build static initializers into a library that we link with whole-archive
+        // Then use object files just to trigger the initial call.
+        init_builder
+            .clone()
+            .file(initializers_file)
+            .link_lib_modifier("+whole-archive")
+            .compile(&format!("{}-initializers", crate_name()));
```
ahayzen-kdab, line 954: is the whole-archive for the cargo build or cmake build or both ?

and then why is this needed ? can these initialiser functions not be public exported functions that are built into a library. Then when linking the library the object file that is also links makes use of the public functions preventing them being optimised out ?

crates/cxx-qt-build/src/lib.rs
```diff
@@ -888,55 +886,92 @@ impl CxxQtBuilder {
...
+
+        let init_call = format!(
+            "extern \"C\" int {init_fun}();\nstatic const int do_{init_fun} = {init_fun}();",
```
ahayzen-kdab, line 957: `::std::int32_t` ?

crates/cxx-qt-lib/src/core/init.cpp
```diff
@@ -11,163 +11,102 @@
 #include <cxx-qt-lib/qset.h>
 #include <cxx-qt-lib/qvector.h>
 
-static const int register_QHash_i32_QByteArray =
+extern "C" int
```
ahayzen-kdab, line 14: `::std::int32_t` ?

crates/cxx-qt-lib/src/core/init.cpp
```diff
@@ -11,163 +11,102 @@
...
-static const int register_QVector_u8 =
   qRegisterMetaType<::QVector_u8>("QVector_u8");
-static const int register_QVector_u16 =
   qRegisterMetaType<::QVector_u16>("QVector_u16");
-static const int register_QVector_u32 =
   qRegisterMetaType<::QVector_u32>("QVector_u32");
-static const int register_QVector_u64 =
   qRegisterMetaType<::QVector_u64>("QVector_u64");
+
+  return 42;
```
ahayzen-kdab, line 111: why do you return 42 ? Seems to be a magic value that isn't used anywhere ? Can you not use `void` ? If you do need to return something to indicate success why not `bool` or true/false ? If this is a magic value then can we have a static const in cxx-qt or exposed from rust somewhere ?

crates/qt-build-utils/src/lib.rs
```diff
@@ -1030,9 +1031,12 @@ Q_IMPORT_PLUGIN({plugin_class_name});
...
     /// The path to the generated C++ file is returned, which can then be passed to [cc::Build::files](https://docs.rs/cc/latest/cc/struct.Build.html#method.file).
-    /// The compiled static library must be linked with [+whole-archive](https://doc.rust-lang.org/rustc/command-line-arguments.html#linking-modifiers-whole-archive)
-    /// or the linker will discard the generated static variables because they are not referenced from `main`.
-    pub fn qrc(&mut self, input_file: &impl AsRef<Path>) -> PathBuf {
+    /// This function also returns a String that contains the name of the resource initializer
+    /// function.
+    /// The build system must ensure that if the .cpp file is built into a static library, either
+    /// the `+whole-archive` flag is used, or the initializer function is called by the
+    /// application.
+    pub fn qrc(&mut self, input_file: &impl AsRef<Path>) -> (PathBuf, String) {
```
ahayzen-kdab, line 1039: wonder if in the refactor we should have a type for a ProcessedFile or something that then has input path, source, output path, any initialiser function etc wrapped up.

crates/cxx-qt-build/src/lib.rs
```diff
@@ -888,55 +886,92 @@ impl CxxQtBuilder {
...
-    fn generate_init_code(&self, initializers: &HashSet<PathBuf>) -> String {
-        initializers
+        let (declarations, calls): (Vec<_>, Vec<_>) = dependency_initializers
             .iter()
-            .map(|path| std::fs::read_to_string(path).expect("Could not read initializer file!"))
-            .chain(self.initializers.iter().cloned())
+            .map(|init_fun| {
+                (
+                    // declaration
+                    format!("extern \"C\" int {init_fun}();"),
```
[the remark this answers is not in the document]
LeonMatthesKDAB: Replaced with `bool`

crates/cxx-qt-lib/src/core/init.cpp
LeonMatthesKDAB: Replaced with `true`.

crates/cxx-qt-lib/src/core/init.cpp
```diff
@@ -11,169 +11,111 @@
 #include <cxx-qt-lib/qset.h>
 #include <cxx-qt-lib/qvector.h>
 
-static const int register_QHash_i32_QByteArray =
+extern "C" bool
+init_cxx_qt_lib_core()
+{
+  static bool initialized = false;
+  if (initialized) {
+    return 42;
```
ahayzen-kdab, line 19: aren't we returning a bool here?

crates/cxx-qt-lib/src/core/init.cpp
```diff
@@ -11,169 +11,111 @@
 #include <cxx-qt-lib/qset.h>
 #include <cxx-qt-lib/qvector.h>
 
-static const int register_QHash_i32_QByteArray =
+extern "C" bool
+init_cxx_qt_lib_core()
+{
+  static bool initialized = false;
```
ahayzen-kdab, line 17: did we say this might need to be an atomic bool ?

crates/cxx-qt-lib/src/gui/init.cpp
```diff
@@ -8,7 +8,16 @@
...
 #include <cxx-qt-lib/qvector.h>
 
-static const int register_QList_QColor =
+extern "C" bool
+init_cxx_qt_lib_gui()
+{
+  static bool initialized = false;
+  if (initialized) {
+    return 42;
+  }
```
ahayzen-kdab, line 17: same here

crates/cxx-qt/src/init.cpp
```diff
@@ -13,30 +13,42 @@
...
-// so that they can be used with QML in Qt 5
-
-static const int register_i8 =
+extern "C" bool
+init_cxx_qt_core()
+{
+  static bool initialized = false;
+  if (initialized) {
+    return true;
+  }
```
ahayzen-kdab, line 22: same here

crates/qt-build-utils/src/lib.rs
```diff
@@ -1067,7 +1097,22 @@ Q_IMPORT_PLUGIN({plugin_class_name});
...
+        let qt_6_5 = SemVer {
+            major: 6,
+            minor: 5,
+            ..SemVer::default()
+        };
+        let init_header = if self.version >= qt_6_5 {
+            // With Qt6.5 the Q_INIT_RESOURCE macro is in the QtResource header
+            "QtResource"
+        } else {
+            "QDir"
```
ahayzen-kdab, line 1109: These could both be `QtCore/QDir` and `QtCore/QtResource` i think, to make it obvious these are coming from the Qt Core module

crates/cxx-qt-lib/src/core/init.cpp
LeonMatthesKDAB: Turns out, there is a C++ standard function that does exactly what we need:
[std::call_once](https://en.cppreference.com/w/cpp/thread/call_once)

I'm shocked and awed by the fact that the C++ stdlib actually has an easy-to-use thing that does exactly what it says. 🤯

crates/cxx-qt-lib/src/core/init.cpp
LeonMatthesKDAB: Ah, yes, fixed.

crates/qt-build-utils/src/lib.rs
LeonMatthesKDAB: Done

ahayzen-kdab (approved):
LGTM :-)

PR #1172, opened 2025-02-02
Register QUuid container types with QML

jnbooth:
This is an addon to #1149 that adds `qRegisterMetaType` calls for `QList_QUuid`, `QSet_QUuid`, and `QVector_QUuid`. This should have been part of the original PR, but I didn't know about the `qRegisterMetaType` system in `cxx-qt-lib/src/core/init.cpp` at the time.


LeonMatthesKDAB (approved):
Approved.
That is actually a good reminder that we should probably auto-generate these initializers as well, as it's confusing that we have to do this manually where the rest is auto-generated...

PR #1173, opened 2025-02-03
Qvector header cleanup

QuentinLeeWeber:
The files were created because the Rust bridges, which are specific, included qvector, which is generic.
This, in turn, included all possible types of qvector, even though only the type required by the bridge was actually needed.
(Run-time: n^2)
With the change, each spefific bridge now has its own specific qvector that it can include.

The only downside is, that the user has to manually include the specific qvector header.


crates/cxx-qt-lib/src/core/qvector/generate.sh
```diff
@@ -9,18 +9,40 @@ set -e
 
 SCRIPT=$(realpath "$0")
 SCRIPTPATH=$(dirname "$SCRIPT")
+#INCLUDEPATH="~/cxx-qt/crates/cxx-qt-lib/include/core/qvector"
```
LeonMatthesKDAB, line 12: Remove commented code.

LeonMatthesKDAB (commented):
Small nitpick, otherwise looks good :)

LeonMatthesKDAB (approved):
LGTM

Issue #1171, opened 2025-01-31
Cannot define two signals in extern C++Qt blocks with the same name


ahayzen-kdab:
Eg the following fails even though the signals are on different objects.

```rust
    unsafe extern "C++Qt" {
        include!(<QtCore/QTimer>);

        #[qobject]
        type QTimer;

        #[qsignal]
        pub(self) fn timeout(self: Pin<&mut QTimer>);


        #[qobject]
        type QTimer2;

        #[qsignal]
        pub(self) fn timeout(self: Pin<&mut QTimer2>);
    }
```
fails with
```
called `Result::unwrap()` on an `Err` value: Error("The type name `timeout` is defined multiple times")
```

Note that this does work inside a `RustQt` block, it appears that signals in a C++Qt are not being combined with their QObject/self type when checking for duplicate names?

PR #1175, opened 2025-02-03
Add span to callables generation to improve error messages

BenFordTytherington:
Fix for https://github.com/KDAB/cxx-qt/issues/1171
- Stops adding signals to typenames
- Improves errors


ahayzen-kdab (approved):
LGTM :-) It seems unclear why the signals were added only in C++Qt to the TypeNames before, likely a mistake. So lets go with this as it doesn't appear to break anything and allows for signals with the same name on different QObjects.

PR #1178, opened 2025-02-05
Multi-Crate project - second try

### LINKED ISSUE #148: figure out and document how to link multiple cxx-qt modules into an application
Application developers will likely want to link multiple cxx-qt modules into applications. I can think of two approaches this could take:

1. Create a cargo workspace within the application repository with multiple lib crates, built as normal `rlib` crates. Link them together into one `staticlib` crate and link the `staticlib` crate into the C++ application. This is the approach Firefox takes: https://hg.mozilla.org/mozilla-central/rev/841c2247f57d, https://bugzilla.mozilla.org/show_bug.cgi?id=1163224
2. Write one big `staticlib` crate with separate modules using cxx-qt within it. I doubt this would be a good idea for build times because [Cargo parallelizes at the level of crates, not modules](https://en.pingcap.com/blog/rust-huge-compilation-units/).

### LINKED ISSUE #1135: QML Modules defined in external crates don't work

LeonMatthesKDAB:
A combination of #598 and #1169.

Closes #598
Closes #148
Closes #1135


examples/qml_multi_crates/rust/main/src/main.rs
```diff
@@ -0,0 +1,48 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    // In the test cfg there needs to be at least one test that calls the crate initialization.
+    // Otherwise linking will fail!
+    #[test]
+    fn init_dependencies() {
+        cxx_qt::init_crate!(qml_multi_crates);
```
ahayzen-kdab, line 45: Does the init crate also init the QML modules from it ? As where does `com.kdab.cxx_qt.demo.sub1` get initialised?

ahayzen-kdab: (assume from one of these as things are reexported?)

examples/qml_multi_crates/rust/main/src/lib.rs
```diff
@@ -0,0 +1,12 @@
...
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+mod main_object;
+
+// Ensure the symbols from the rlib dependencies end up
+// in the staticlib (if you use Rust symbols from these
+// crates in this crate, you can skip these `extern crate` statements).
+extern crate sub1;
+extern crate sub2;
```
ahayzen-kdab, line 12: wonder if we should have a further `sub3` crate that uses types from sub1 and sub2 to check that is possible. Or is that not possible ? :thinking: Can be in a separate later branch though...

book/src/SUMMARY.md
```diff
@@ -27,6 +27,7 @@ SPDX-License-Identifier: MIT OR Apache-2.0
   - [Shared types](./bridge/shared_types.md)
   - [Attributes](./bridge/attributes.md)
   - [Traits](./bridge/traits.md)
+- [Common Issues](./common-issues.md)
 - [For Contributors: CXX-Qt Internals](./internals/index.md)
   - [Build System](./internals/build-system.md)
```
ahayzen-kdab, line 32: does anything in the build system notes need updating ?

Cargo.toml
```diff
@@ -18,6 +18,9 @@ members = [
     "examples/qml_features/rust",
     "examples/qml_minimal/rust",
     "examples/qml_basics",
+    "examples/qml_multi_crates/rust/main",
+    "examples/qml_multi_crates/rust/sub1",
+    "examples/qml_multi_crates/rust/sub2",
```
ahayzen-kdab, line 23: should you add a changelog entry stating that this is now possible? (multi crate import/qml modules etc)

Cargo.toml
```diff
@@ -18,6 +18,9 @@ members = [
     "examples/qml_features/rust",
```
ahayzen-kdab, line 18: can you reword the WIP commit in the chain?

Cargo.toml
LeonMatthesKDAB: Done

book/src/SUMMARY.md
LeonMatthesKDAB: The existing ones already fit, but I've now moved them around a bit to fit better with our split of `problem space`/`solution space`.

examples/qml_multi_crates/rust/main/src/lib.rs
LeonMatthesKDAB: That's sort of what the main crate does though, right?
Or do you mean use types in Rust?

ahayzen-kdab: Right using the types, but lets wait for the pure virtual fun, as then we could have an abstract class and a concrete type in one crate. And then use the abstract version in the methods in another crate. Which would emulate some real world scenarios.

ahayzen-kdab (approved):
LGTM, lets go with this :-)

Be-ing:
Wow, good job finally figuring out this cursed puzzle!

book/src/common-issues.md
```diff
@@ -0,0 +1,79 @@
...
+```
+
+(where another_crate is replaced by the name of the dependency that isn't otherwise used).
+
+### 2. Include the initializers in your code
+
+Next we need to ensure the initializers can be found by the linker.
+
+If you followed step 1, modern linkers like `mold` or `lld` should already be able to link everything correctly.
+We encourage switching to such a linker if you're still using the (now deprecated) `ld.gold` on Linux.
```
Be-ing, line 49: I think `ld.bfd` is still the default linker on most distros?

book/src/common-issues.md
```diff
@@ -0,0 +1,79 @@
...
+/// # cxx_qt::init_qml_module!(another_crate);
+///
+/// X::do_something();
+/// ```
+struct X {}
+
+#[cfg(test)]
+mod tests {
+  #[test]
+  fn initialize_eependencies() {
```
Be-ing, line 74: Who needs dependencies when you can have eependencies? :grinning:

book/src/common-issues.md
Be-ing: I think it'd be worth mentioning here that CXX-Qt will automatically prefer lld or mold if they are installed, also that the linker can be configured for all Rust builds by putting

```
[target.x86_64-unknown-linux-gnu] # or whatever target
rustflags = [
    "-C", "link-arg=-fuse-ld=mold"
]
```

in ~/.cargo/config.toml.

I just tested the main branch with ld.bfd and ld.gold, and with the recent changes, `cargo build -p qml-minimal-no-cmake` links successfully. `cargo build -p qml_multi_crates` works with all linkers, but if I comment out `cxx_qt::init_crate!(qml_multi_crates);` then ld.bfd and ld.gold both fail.

PR #1182, opened 2025-02-13
ci: fix wasm cache using old key

ahayzen-kdab:


.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -232,7 +232,7 @@ jobs:
         continue-on-error: true
         run: |
           gh extension install actions/gh-actions-cache
-          gh actions-cache delete "Ubuntu 22.04 (wasm_32) Qt6_compiler_cache" --confirm
+          gh actions-cache delete "Ubuntu 24.04 (wasm_32) Qt6_compiler_cache" --confirm
```
LeonMatthesKDAB, line 235: The name of the cache is used in 3 different places, please extract it into an (environment) variable.
Then we can't have this issue in future.

ahayzen-kdab: done

LeonMatthesKDAB (approved):

PR #1185, opened 2025-02-13
ci: reenable sccache for macOS

ahayzen-kdab:


ahayzen-kdab:
We won't know if this has worked until it lands in main, makes a cache, and then we try to use it.

LeonMatthesKDAB:
The original issue on sccache has not yet been resolved.
Do you still want to try it anyway?

ahayzen-kdab:
Might be worth a try, as the original issue was not able to find the cause just that it's somewhere in the tokio threading code :shrug:

LeonMatthesKDAB (approved):

PR #1189, opened 2025-02-14
Add pure virtuals

BenFordTytherington:
Add `Pure` virtual functions API with a new attribute

### DIFF
--- a/crates/cxx-qt-gen/src/parser/method.rs
+++ b/crates/cxx-qt-gen/src/parser/method.rs
@@ -55,6 +55,8 @@ pub struct ParsedMethod {
     pub specifiers: HashSet<ParsedQInvokableSpecifiers>,
     /// Whether the method is qinvokable
     pub is_qinvokable: bool,
+    /// Whether the method is a pure virtual method
+    pub is_pure: bool,
     // No docs field since the docs should be on the method implementation outside the bridge
     // This means any docs on the bridge declaration would be ignored
     /// Cfgs for the method
@@ -117,12 +119,14 @@ impl ParsedMethod {
 
         // Determine if the method is invokable
         let is_qinvokable = attrs.contains_key("qinvokable");
+        let is_pure = attrs.contains_key("cxx_pure");
         let specifiers = ParsedQInvokableSpecifiers::from_attrs(attrs);
 
         Ok(Self {
             method_fields: fields,
             specifiers,
             is_qinvokable,
+            is_pure,
             cfgs,
         })
     }
--- a/crates/cxx-qt-gen/src/parser/method.rs
+++ b/crates/cxx-qt-gen/src/parser/method.rs
@@ -18,6 +18,7 @@ pub enum ParsedQInvokableSpecifiers {
     Final,
     Override,
     Virtual,
+    Pure,
 }
 
 impl ParsedQInvokableSpecifiers {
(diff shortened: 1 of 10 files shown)


examples/qml_features/rust/src/custom_base_class.rs
```diff
@@ -269,6 +285,15 @@ impl Default for qobject::State {
 
 /// A struct which inherits from QAbstractListModel
 #[derive(Default)]
+pub struct AbstractBaseClassRust {}
+
+impl qobject::AbstractBaseClass {
+    /// Virtual method for logging
+    pub fn log(self: &AbstractBaseClass) {}
+}
```
ahayzen-kdab, line 293: curious, as it's pure virtual in C++ we don't actually need/want an implementation.

I guess this needed to be here as we generate one of those magic Rust bridges where the implementation is on the Rust side by using extern "Rust" ?

Instead there should be no implementation in Rust, and probably just an extern "C++" block in the generated Rust.

Ah yes i can see that in the test_outputs...

crates/cxx-qt-gen/test_outputs/invokables.rs
```diff
@@ -98,6 +98,12 @@ mod ffi {
         #[doc(hidden)]
         fn invokable_virtual(self: &MyObject);
     }
+    extern "Rust" {
+        #[cxx_name = "invokable_pure_virtual"]
+        #[namespace = "cxx_qt::my_object"]
+        #[doc(hidden)]
+        fn invokable_pure_virtual(self: &MyObject);
+    }
```
ahayzen-kdab, line 106: as this is pure virtual with no implementation, we should make this `extern "C++"` as it's just a method in C++ and doesn't have an implementation in Rust

BenFordTytherington: Should be fixed now

LeonMatthesKDAB (commented):
The generated output looks good I think.

What may still be missing is a check that the method is virtual, if it's marked as pure.

`#[cxx_pure]` is only allowed together with `#[cxx_virtual]` or `#[cxx_override]`
And it's probably not compatible with `#[cxx_final]`.
Can you add checks for this...

We may also leave this check up to the C++ compiler, but please check that this the error message is reasonable.

ahayzen-kdab:
> The generated output looks good I think.
> 
> What may still be missing is a check that the method is virtual, if it's marked as pure.
> 
> `#[cxx_pure]` is only allowed together with `#[cxx_virtual]` or `#[cxx_override]` And it's probably not compatible with `#[cxx_final]`. Can you add checks for this...
> 
> We may also leave this check up to the C++ compiler, but please check that this the error message is reasonable.

Note Ben had already checked the C++ compiler warning, it comes back with this which seemed reasonable `initializer specified for non-virtual method 'void my_object::MyObject::incrementNumber()'`, so I thought initially leaving it up to C++ would be OK.

crates/cxx-qt-gen/src/parser/method.rs
```diff
@@ -26,6 +27,7 @@ impl ParsedQInvokableSpecifiers {
             ParsedQInvokableSpecifiers::Final => "cxx_final",
             ParsedQInvokableSpecifiers::Override => "cxx_override",
             ParsedQInvokableSpecifiers::Virtual => "cxx_virtual",
+            ParsedQInvokableSpecifiers::Pure => "cxx_pure",
```
ahayzen-kdab, line 30: can we add a changelog entry and update the book that i mentions these here https://kdab.github.io/cxx-qt/book/bridge/extern_rustqt.html#specifiers

ahayzen-kdab (approved):
LGTM :-)

PR #1191, opened 2025-02-17
ci: set the CARGO_BUILD_TARGET to reuse same cache

ahayzen-kdab:
Requires #1192

### DIFF
--- a/.github/workflows/github-cxx-qt-tests.yml
+++ b/.github/workflows/github-cxx-qt-tests.yml
@@ -346,6 +346,7 @@ jobs:
             cargo_dir: ~/.cargo
             rustc_wrapper: sccache
             build_type: Release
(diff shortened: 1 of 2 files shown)


.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -422,6 +429,8 @@ jobs:
         SCCACHE_CACHE_SIZE: 600M
         SCCACHE_LOG: debug
         SCCACHE_ERROR_LOG: ${{ matrix.sccache_log_path }}
+        # Ensure that we share the same target sub folder
```
ahayzen-kdab, line 432: Wonder how much sharing there can be between the builds, or if they are ultimately different anyway and need separate folders/caches. (cmake vs cargo)

PR #1194, opened 2025-02-19
Fix cargo test not using sccache correctly

LeonMatthesKDAB:


LeonMatthesKDAB:
Hm, this didn't do anything, as the env variables are actually already set in the environment via the workflow file.

But then why doesn't the sccache statistic show additional compile requests for C/C++ after running `cargo test`?

Before: https://github.com/KDAB/cxx-qt/actions/runs/13413232351/job/37467893158?pr=1194#step:15:75
After: https://github.com/KDAB/cxx-qt/actions/runs/13413232351/job/37467893158?pr=1194#step:17:75

LeonMatthesKDAB:
CI is currently fast enough, let's investigate this another time

PR #1199, opened 2025-02-26
cxx-qt-gen: allow for safe / unsafe extern "RustQt"

ahayzen-kdab:


LeonMatthesKDAB (approved):

LeonMatthesKDAB (changes requested):
I found some issues after all

LeonMatthesKDAB (approved):
This seems to behave correctly now.

PR #1204, opened 2025-03-01
Improve manual implementations of up- and downcasting

jnbooth:
This PR provides some minor improvements to #1159 for our own implementations of `Upcast`.

- Implements `Upcast<QVector<QPoint>>` for `QPolygon`, replacing the standalone functions previously used for `Deref` and `DerefMut`.
- Implements `Upcast<QVector<QPointF>>` for `QPolygonF`, replacing the standalone functions previously used for `Deref` and `DerefMut`.
- Implements `Upcast<QCoreApplication>` for `QGuiApplication` and adds `Deref` and `DerefMut`, replacing the standalone functions previously used for class methods.
- Implements `Upcast<QGuiApplication>` for `QApplication` and adds `Deref` and `DerefMut`, replacing the standalone functions previously used for class methods.
- Implements `Deref` and `DerefMut` for `QQmlApplicationEngine` to `QQmlEngine` using the existing upcasting implementation.
- Removes a null-pointer dereference inside `QStringList`'s implementation of `QList<QString>`.


crates/cxx-qt-lib-extras/src/gui/qapplication.rs
```diff
@@ -19,11 +21,26 @@ mod ffi {
...
+        include!("cxx-qt/casting.h");
+
+        #[doc(hidden)]
+        #[rust_name = "upcast_qapplication"]
+        unsafe fn upcastPtr(thiz: *const QApplication) -> *const QGuiApplication;
+
+        #[doc(hidden)]
+        #[rust_name = "downcast_qapplication"]
+        unsafe fn downcastPtr(base: *const QGuiApplication) -> *const QApplication;
+    }
```
ahayzen-kdab, line 42: This will not be required once #1202 lands as automatic up/down casting implementations will be done via a C++Qt block, such as

```rust
unsafe extern "C++Qt" {
   #[base = "QGuiApplication"]
   type QApplication;
}
```

crates/cxx-qt-lib-extras/src/gui/qapplication.rs
```diff
@@ -93,6 +110,24 @@ mod ffi {
...
+    unsafe fn upcast_ptr(this: *const Self) -> *const QGuiApplication {
+        ffi::upcast_qapplication(this)
+    }
+
+    unsafe fn from_base_ptr(base: *const QGuiApplication) -> *const Self {
+        ffi::downcast_qapplication(base)
+    }
+}
+
+impl Deref for QApplication {
```
ahayzen-kdab, line 123: I think it was discussed to do Deref automatically for Upcast before, but it was decided against, @LeonMatthesKDAB @BenFordTytherington can you remember the details?

crates/cxx-qt-lib/src/gui/qguiapplication.rs
```diff
@@ -21,26 +22,24 @@ mod ffi {
         type QVector_QByteArray = crate::QVector<QByteArray>;
         include!("cxx-qt-lib/qfont.h");
         type QFont = crate::QFont;
+        include!("cxx-qt-lib/qcoreapplication.h");
+        type QCoreApplication = crate::QCoreApplication;
 
         include!("cxx-qt-lib/qguiapplication.h");
         type QGuiApplication;
```
ahayzen-kdab, line 29: With #1202 this becomes

```rust
unsafe extern "C++Qt" {
   #[base = QCoreApplication]
   type QGuiApplication;
}
```

crates/cxx-qt-lib/src/gui/qpolygon.rs
```diff
@@ -30,6 +28,9 @@ mod ffi {
         #[allow(dead_code)]
         type QPolygonF = crate::QPolygonF;
 
+        include!("cxx-qt-lib/qvector.h");
+        type QVector_QPoint = crate::QVector<QPoint>;
+
         include!("cxx-qt-lib/qpolygon.h");
         type QPolygon = super::QPolygon;
```
ahayzen-kdab, line 35: can this also be done via the C++Qt base attribute?

crates/cxx-qt-lib/src/qml/qqmlapplicationengine.rs
```diff
@@ -13,23 +17,12 @@ mod ffi {
         include!("cxx-qt-lib/qurl.h");
         type QUrl = crate::QUrl;
 
+        include!("cxx-qt-lib/qqmlengine.h");
+        type QQmlEngine = crate::QQmlEngine;
+
         include!("cxx-qt-lib/qqmlapplicationengine.h");
         type QQmlApplicationEngine;
```
ahayzen-kdab, line 24: #1202 already changes this upcast to be done via base attribute

crates/cxx-qt/include/casting.h
```diff
@@ -22,7 +23,15 @@ const Sub*
...
-  return dynamic_cast<const Sub*>(base);
+  return qobject_cast<const Sub*>(base);
+}
+
+template<typename Sub, typename Base>
+const Sub*
+downcastPtrStatic(const Base* base)
+{
+  static_assert(std::is_base_of_v<Base, Sub>);
+  return static_cast<const Sub*>(base);
```
ahayzen-kdab, line 34: this doesn't check that the instance `base` is actually of type `Sub` ? As isn't that what the dynamic_cast was doing ?

crates/cxx-qt/include/casting.h
```diff
@@ -22,7 +23,15 @@ const Sub*
 downcastPtr(const Base* base)
 {
   static_assert(std::is_base_of_v<Base, Sub>);
-  return dynamic_cast<const Sub*>(base);
+  return qobject_cast<const Sub*>(base);
```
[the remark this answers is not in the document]
jnbooth: What we shouldn't use is `dynamic_cast`. There are only a few cases of Qt class inheritance that isn't based on a `QObject`, and those should use `static_cast`, as with `QPolygon` and `QStringList`. This is due to the complicated way that C++ handles RTTI and polymorphism. `dynamic_cast` only works if the class has at least one virtual method so it can check RTTI. For Qt, that means it wouldn't work unless it's a `QObject` anyway. That's why downcasting `QStringList` didn't work using `dynamic_cast`.

Given the #`[base]` pragma being added by #1202, I think we would probably want to break it down like this:

- If the class inherits `QObject`, use `#[base]`, which internally uses `qobject_cast`. This can be done both internally and by end users. Probably don't automatically implement `Deref` as it'd be too complicated. Plus, `QObject`s can't implement `DerefMut` because their mutable methods use `Pin<&mut Self>`.
- Otherwise, use `static_upcast` from those `extern` utility functions and implement `Deref` and `DerefMut`. This should only be done internally, because otherwise it might get added to something that isn't actually safe to cast.

crates/cxx-qt/include/casting.h
jnbooth: The static assert and inherent checks of `static_cast` handle that. There are no runtime checks for non-polymorphic classes, so what these checks do is assert that it's *always* valid to cast one to the other. For example, a `QStringList` inherits `QList<QString>` but doesn't add any fields, which means it's actually exactly the same as a `QList<QString>`—so it's always valid to cast a `QList<QString>` to a `QStringList`. Qt uses this pattern for a small handful of basic data types (i.e. `cxx::kind::Trivial`). For everything else, it uses `QObject` inheritance so that `qobject_cast` can be used.

crates/cxx-qt-lib-extras/src/gui/qapplication.rs
LeonMatthesKDAB: The reason we don't automatically support this is because of two reasons:

1. We already implement Deref for generated QObjects. They Deref to the inner Rust type.
2. There could be multiple-inheritance on the C++ side, in which case it would be unclear which of the base classes to deref to (though this is currently academic, we don't support it anyway).

For types like QApplication, none of these apply, so we can absolutely `impl Deref` via Upcasting.

crates/cxx-qt/include/casting.h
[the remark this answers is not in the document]
LeonMatthesKDAB: As we're talking about downcasting, `static_cast` is a no-go, we do not know if the type is actually of the subclass, so we need to check at runtime. `static_cast` would do the cast in any case, which is instant undefined behavior if the type is not actually of the subclass.

`dynamic_cast` is the only option we have for non-QObject types in that case, even given it's limitations.

@jnbooth you bring up some good points though, which we should integrate.

1. We can use [is_polymorphic](https://en.cppreference.com/w/cpp/types/is_polymorphic) to assert that a type actually has at least one virtual function so we can use dynamic_cast.
2. We can use template meta-programming with [is_base_of](https://en.cppreference.com/w/cpp/types/is_base_of) to detect if a type is derived from QObject and then use qobject_cast instead of dynamic_cast.

That way, we can still have a single templated method that uses the optimal cast in all cases.
@jnbooth If you want I can send you the required code, in case you don't want to deal with the template meta programming. Just let me know.

For types that don't meet these criteria, we likely don't want do support downcast?
I guess we could just always return `nullptr` there, which would turn into `None` in the Rust impl.

@ahayzen-kdab may need to respect this in #1202 . :thinking:

crates/cxx-qt/include/casting.h
LeonMatthesKDAB: @jnbooth you're right that in this case static_cast is okay.

It's pretty hacky but should work.
Just to make sure this doesn't accidentally break i.e. in Qt7, can you add a static_assert to assert that the size and alignment of Sub and Base are the same.
As that should be a good indication that the types are actually interchangeable.

LeonMatthesKDAB:
@jnbooth Thank you for the contribution, this is looking good.

Regarding the downcasting, I think we can get all those concerns implemented in a sane way with a bit of meta-programming that automatically selects between qobject_cast and dynamic_cast.

And keeping static_cast for the special cases is okay, though I think we should add another assertion just out of caution.

crates/cxx-qt/include/casting.h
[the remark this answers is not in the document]
jnbooth: To be clear, I'm not suggesting `static_cast` should be used in the general case or any generated code. I am only using it here, manually implemented, for cases where `static_cast` is *always* safe.

Are there any examples you have of types that might need to be downcasted that aren't QObjects, aside from the above special cases of classes that are always safe to `static_cast` (there are only a few of them)? I think it is probably a better idea to handle those few cases internally and use `qobject_cast` for everything else, because users can only use `QObject`s anyway. `dynamic_cast` is generally not a good idea, to the point that it is sometimes labeled "evil".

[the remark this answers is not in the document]
ahayzen-kdab: For other examples, you might have a class that is a gadget or something that is just a raw class in Qt like an implementation of `QQuickFramebufferObject::Renderer`. You may then take the base class as a parameter and want to downcast potentially to your class.

I agree where we have `#[qobject]` on the class we should be able to determine that we can use `qobject_cast`.

So i think we then potentially end up with the following
- `static_cast` when safe to do so (eg type is of same size and derived from the other etc)
- `qobject_cast` when QObject is a base class of the object
- `dynamic_cast` for anything else that does not have a QObject base or does not have same size or is not derived from the other etc

jnbooth:
@LeonMatthesKDAB I don't know how to do that kind of metaprogramming. Are there examples of it in the codebase that I could learn from? Alternatively, should I strip down part of this PR so it can be merged without that?

LeonMatthesKDAB:
@jnbooth This should do it:
https://github.com/LeonMatthesKDAB/cxx-qt/commit/fdd65e94eff925d796922aa36998c00bd7a409ae
In case you're interested, google SFINAE and enable_if_t, it's one of the crazier C++ concepts for sure.
But is quite powerful for cases like this.

I pushed this to a branch on my fork (non-qobject-casting) and rebased, feel free to cherry-pick or just use that branch for your PR.

My template meta-programming is a bit rusty nowadays, so @ahayzen-kdab please review carefully to check this is sane.

jnbooth:
Any updates on this? I think at minimum it would be good to remove the null pointer returns.

LeonMatthesKDAB (approved):

jnbooth:
I don't have approve permissions btw, so it is up to someone else to do that.

LeonMatthesKDAB:
Ah sorry @jnbooth this PR probably lost the race with another approval which outdated the auto-merge, this is an annoying issue with our workflow at the moment :roll_eyes: 
It should auto-merge now after the CI passes :crossed_fingers:

PR #1207, opened 2025-03-04
Fix: Only update caches when pushed to `main`

LeonMatthesKDAB:
Otherwise, backports would ruin the caches for all other branches that
target `main`.


ahayzen-kdab (approved):
LGTM :-)

ahayzen-kdab (approved):
LGTM :-)

PR #1209, opened 2025-03-05
create individual header files

QuentinLeeWeber:
The files were created because the Rust bridges, which are specific, included headers, which is generic. This, in turn, included all possible types, even though only the type required by the bridge was actually needed. (Run-time n^2)
With the change, each spefific bridge now has its own specific header that it can include.

The only downside is, that the user has to manually include the specific header.


ahayzen-kdab:
Looks good, going to attempt rebasing via github.

PR #1211, opened 2025-03-05
Clear header directory before overwriting it

QuentinLeeWeber:
Before writing the new header files, the header directory is now cleared. This change helps to avoid potential conflicts.

### DIFF
--- a/crates/cxx-qt-lib/build.rs
+++ b/crates/cxx-qt-lib/build.rs
@@ -46,6 +46,21 @@ fn write_headers_in(subfolder: &str) {
     }
 }
 
+fn clear_include_directory() {
+    if !header_dir().exists() {
+        return;
+    }
+    for entry in std::fs::read_dir(header_dir().as_path()).expect("Failed to read header directory")
+    {
+        let path = entry.expect("Failed to read path").path();
+        if path.is_dir() {
+            std::fs::remove_dir_all(path).expect("Failed to delete directory");
+        } else {
+            std::fs::remove_file(path).expect("Failed to delete file");
+        }
+    }
+}
+
 fn write_definitions_header() {
     // We cannot ensure that downstream dependencies set the same compile-time definitions.
     // So we generate a header file that adds those definitions, which will be passed along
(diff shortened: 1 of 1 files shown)


crates/cxx-qt-lib/build.rs
```diff
@@ -46,6 +46,21 @@ fn write_headers_in(subfolder: &str) {
     }
 }
```
ahayzen-kdab, line 47: side point, wonder if sccache uses the hash of the file or the mtime ?

Eg does modifying a file in the header dir (by deleting and writing the same contents again) cause a cache miss ? Or is it comparing the contents anyway.

Looking at the times in CI it seems to help with macOS and Windows Qt 5 specifically :thinking:

ahayzen-kdab: Now doing a clean run after rebasing things vs a run on main both with the same cache we seem to have.

| Runner | Before | After |
|------------|------------|---------|
| macOS 13 Qt 5 | 7m 42s | 7m 33s |
| macOS 14 Qt 6 | 6m 15s | 7m 28s |
| Ubuntu Qt 5 | 5m 50s | 5m 43s |
| Ubuntu Qt 6 | 6m 59s | 8m 4s |
| Windows Qt 5 | 10m 16s | 14m 9s |
| Windows Qt 6 MSVC2019 | 12m 19s | 12m 57s |
| Windows Qt 6 MSVC2022 | 13m 1s | 24m 57s |

Seems that this makes Qt 6 1min slower for Linux and macOS, then Windows is curious too. Would probably need more reruns to be sure of the timings.

But it does make me wonder if we should be more intelligent than just delete everything :thinking:

LeonMatthesKDAB: The thing with our CI is that in most cases, this change shouldn't affect it much.
Especially on Windows and macOS, we're doing only clean builds, so there is nothing to delete anyway, so this doesn't change anything.

The interesting case is the cargo_rerun test case under Linux, as that would show a difference in caching behavior.

PR #1212, opened 2025-03-05
CI: Bump rust version in Coverage runner

LeonMatthesKDAB:
Otherwise, the grcov crate fails to build, as symbolic-common now
requires a newer rust version

### DIFF
--- a/.github/workflows/github-cxx-qt-tests.yml
+++ b/.github/workflows/github-cxx-qt-tests.yml
@@ -83,14 +83,16 @@ jobs:
       #
       # See versions from the table in this link
       # https://github.com/taiki-e/cargo-llvm-cov?tab=readme-ov-file#get-coverage-of-cc-code-linked-to-rust-librarybinary
-      - name: Install llvm 17
+      - name: Install llvm 19
         run: |
-          sudo apt-get update && sudo apt-get install -y llvm-17
-          test -d /usr/lib/llvm-17/bin/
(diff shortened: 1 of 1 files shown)


.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -83,14 +83,16 @@ jobs:
...
+          sudo apt-get update && sudo apt-get install -y llvm-18
+          test -d /usr/lib/llvm-18/bin/
       - name: Setup toolchain
         run: |
           # Note that the llvm version needs to match, see the link above
-          rustup default 1.77.2
+          # grcov currently requires a newer version of the Rust compiler than CXX-Qts MSRV
+          # So use a newer version here than in the other tests
+          rustup default 1.81.0
           cargo install grcov
```
ahayzen-kdab, line 96: can we add a version to `grcov` so it doesn't break again? like we do with sccache / mdbook etc? And potentially use --locked too to avoid dependencies changing ?

```console
cargo install --locked --version 0.8.20 grcov
```

ahayzen-kdab: Or as it is still failing even with these changes, can you pin the old version instead?

LeonMatthesKDAB: Ah, fair, let me try that instead.

PR #1213, opened 2025-03-05
CI: Lock tool versions

LeonMatthesKDAB:
This should keep us from encounteing breakage due to rust version
changes, etc.

Closes #1212


ahayzen-kdab (approved):
LGTM :-)

PR #1216, opened 2025-03-06
Remove unused self parameters in Q*Application static methods

### LINKED ISSUE #1205: Unused self parameters in Q*Application methods
In `QCoreApplication`, `QGuiApplication`, and `QApplication`, static methods are handled in three different ways, arbitrarily:

1. No `self` parameter
2. `&self`
3. `Pin<&mut Self>`

The first approach is the only correct one. The second approach has an unused parameter, but at least it doesn't do anything with it. The third approach adds pinning and mutability logic that, again, is completely unused. It's passing around and wrapping and unwrapping a pointer that nothing happens to.

Obviously, removing those parameters would cause breaking changes, but I think it should probably be done in time for 1.0.

jnbooth:
Resolves #1205.

**Note:** This PR contains breaking changes.

### DIFF
--- a/crates/cxx-qt-lib/src/core/qcoreapplication.rs
+++ b/crates/cxx-qt-lib/src/core/qcoreapplication.rs
@@ -4,6 +4,8 @@
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
+use std::pin::Pin;
+
 use crate::{QByteArray, QString, QStringList, QVector};
 
 #[cxx_qt::bridge]
@@ -94,16 +96,17 @@ impl QCoreApplication {
     /// ensuring that it is searched for libraries first.
     /// If path is empty or already in the path list, the path list is not changed.
     pub fn add_library_path(path: &QString) {
+        // Locked by Qt's `libraryPathMutex`.
         ffi::qcoreapplication_add_library_path(path);
     }
 
     /// The name of this application
-    pub fn application_name() -> QString {
+    pub fn application_name(&self) -> QString {
         ffi::qcoreapplication_application_name()
     }
 
     /// The version of this application
-    pub fn application_version() -> QString {
+    pub fn application_version(&self) -> QString {
         ffi::qcoreapplication_application_version()
     }
 
@@ -115,6 +118,7 @@ impl QCoreApplication {
 
     /// Returns a list of paths that the application will search when dynamically loading libraries.
     pub fn library_paths() -> QStringList {
+        // Locked by Qt's `libraryPathMutex`.
         ffi::qcoreapplication_library_paths()
     }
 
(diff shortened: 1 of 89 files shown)


crates/cxx-qt-lib/src/core/qcoreapplication.rs
```diff
@@ -87,29 +86,29 @@ impl QCoreApplication {
...
+    pub fn application_version() -> QString {
+        ffi::qcoreapplication_application_version()
     }
 
     /// Enters the main event loop and waits until exit() is called,
     /// and then returns the value that was set to exit() (which is 0 if exit() is called via quit()).
-    pub fn exec(self: Pin<&mut Self>) -> i32 {
-        ffi::qcoreapplication_exec(self)
+    pub fn exec(&self) -> i32 {
+        ffi::qcoreapplication_exec()
```
ahayzen-kdab, line 106: Technically `exec` is static too :sweat_smile: although would create a strange API i guess.

And we cannot overload in Rust so we cannot have it be accessible via both :-/

So guess this is the "best" way unless there are other suggestions :thinking: 

https://doc.qt.io/qt-6/qcoreapplication.html#exec

LeonMatthesKDAB:
Sorry for not thinking of this earlier, but I just realized we may have a safety issue with these methods (which may be why we made them "members" initially).

Some of them seem to be not thread-safe!
E.g. [setActiveWindow](https://codebrowser.dev/qt6/qtbase/src/widgets/kernel/qapplication.cpp.html#_ZN19QApplicationPrivate15setActiveWindowEP7QWidget) just accesses the `QApplicationPrivate::active_window` without any locking.

So they are effectively only safe-to-call from a single thread (presumably the "main thread")
We'd have to make these methods `unsafe` or introduce some kind of locking ourselves.

The Qt docs also state that e.g. only one `QGuiApplication` should be active: https://doc.qt.io/qt-6/qguiapplication.html#QGuiApplication

So the way these could be made safe is by keeping them as member functions and enforcing in the constructor that only a single QGuiApplication can actually be constructed (and disabling Send+Sync).
We currently don't do that second part though.

I'm also open to other suggestions, especially as keeping member functions is inconvenient.

jnbooth:
>So they are effectively only safe-to-call from a single thread (presumably the "main thread")

I believe this is true of all GUI operations, not just the ones related to `QGuiApplication`. Per [KDAB's own documentation](https://www.kdab.com/the-eight-rules-of-multithreaded-qt/):

>Qt's GUI operations are not thread safe, so non-main threads cannot safely perform any GUI operation. That means no widgets, QtQuick, QPixmap, or anything that touches the window manager. There are some exceptions: GUI functions that only manipulate data but don't touch the window manager can be called on secondary threads, things like QImage and QPainter. Be careful though, as classes like QBitmap or QPixmap actually are not safe. Check the documentation for each API: if you don't see a note at the top of the documentation saying that the function is reentrant, it's not safe to be called except from the main thread.

`setActiveWindow` touches the window manager, so it would be included in that.

crates/cxx-qt-lib/src/core/qcoreapplication.rs
jnbooth: Yeah, I wasn't sure of the best approach here. The reason I settled on this is because if it's static, then

```rs
let app = QCoreApplication::new();
QCoreApplication::exec();
```

`app` will be marked by the linter as unused. And I'm not 100% confident that the compiler would know `app` still needs to be alive by the time of `QCoreApplication::exec()`.

LeonMatthesKDAB:
> > So they are effectively only safe-to-call from a single thread (presumably the "main thread")
> 
> I believe this is true of all GUI operations, not just the ones related to `QGuiApplication`. Per [KDAB's own documentation](https://www.kdab.com/the-eight-rules-of-multithreaded-qt/):
> 
> > Qt's GUI operations are not thread safe, so non-main threads cannot safely perform any GUI operation. That means no widgets, QtQuick, QPixmap, or anything that touches the window manager. There are some exceptions: GUI functions that only manipulate data but don't touch the window manager can be called on secondary threads, things like QImage and QPainter. Be careful though, as classes like QBitmap or QPixmap actually are not safe. Check the documentation for each API: if you don't see a note at the top of the documentation saying that the function is reentrant, it's not safe to be called except from the main thread.
> 
> `setActiveWindow` touches the window manager, so it would be included in that. Is it doable for cxx-qt to provide guarantees of thread safety where GUI operations are concerned? `Send` prevents moving between threads, but that doesn't stop someone from constructing an object outside of the main thread, among other things. It might need to be up to users.
> 
> EDIT: It could be possible to [throw a runtime error if code executes outside the main thread](https://stackoverflow.com/questions/35018713/detect-that-im-running-in-qt-gui-event-thread/35018825#35018825). That seems like a method of last resort, though.

Hm, with this kind of design by Qt, I think we may not get around doing runtime checks tbh.

In general, we should make sure that the objects that need to be used from the main thread are not `Send`, which should get us halfway there.
However, as you mentioned we need to enforce that you cannot create such an object outside the main thread...
I don't see a good way to do this apart from runtime checks and returning an `Err` if you call such a constructor outside the main thread (which might need to apply to QGuiApplication as well).

There is a method since Qt 6.8 that we can use to check this:
https://doc.qt.io/qt-6/qthread.html#isMainThread

For the static functions in `Q*Application`, we could then return an `Err` or just panic.

Unfortunately it uses some [private API internally](https://codebrowser.dev/qt6/qtbase/src/corelib/thread/qthread.cpp.html#_ZN7QThread12isMainThreadEv), so it would probably be hard to backport for Qt5 and Qt 6.7-.

More ideas are welcome though @ahayzen-kdab @jnbooth .

LeonMatthesKDAB:
Hm, I know that at least on MacOS, any event processing has to actually happen on the main thread, (e.g. the one that executes `main`).
The docs are a bit unclear on how this relates, as you can create at least a QCoreApplication on other threads.
Can you do the same with e.g. QGuiApplication?

It's possible that just checking the instances thread is good enough though.
However, there's a bit of a catch-22 here: `QCoreApplication::instance()` itself isn't thread-safe.
So using that to check whether we are on the right thread doesn't work, right?

jnbooth:
I've added (unused) `&self` and `self: Pin<&mut Self>` parameters to the functions that access the application instance. A few functions don't need them because they don't access the instance, and instead Qt uses a mutex to guard them:

- `add_library_path`: locked by Qt's `libraryPathMutex`
- `font`: locked by Qt's `applicationFontMutex`
- `library_path`: locked by Qt's `libraryPathMutex`
- `remove_library_path`: locked by Qt's `libraryPathMutex`
- `set_font`: locked by Qt's `applicationFontMutex`
- `set_library_paths`: locked by Qt's `libraryPathMutex`

The Q*Application structs do not implement `Send` or `Sync`, so that should be sufficient to keep them safe. 

>I know that at least on MacOS, any event processing has to actually happen on the main thread, (e.g. the one that executes main).

Does it? I don't see that in the documentation.

PR #1217, opened 2025-03-07
Add more signals to cxx_qt_lib

BenFordTytherington:
- Add objectCreated signal to QQmlApplicationEngine
- Add objectCreationFailed signal to QQmlApplicationEngine
- Fixed bug where type_name doesn't recognise QObject
- Fixed qualification of QCoreApplication in bridge

### DIFF
--- a/crates/cxx-qt-gen/src/generator/rust/signals.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/signals.rs
@@ -127,9 +127,9 @@ pub fn generate_rust_signal(
     if !signal.private {
         cxx_mod_contents.push(parse_quote_spanned! {
             span=>
+            #(#cfgs)*
             #unsafety_block extern "C++" {
                 #[cxx_name = #cpp_ident]
-                #(#cfgs)*
                 #(#doc_comments)*
                 #namespace
                 #unsafe_call fn #signal_ident_cpp(#parameter_signatures) #return_type;
@@ -155,6 +155,7 @@ pub fn generate_rust_signal(
         parse_quote_spanned! {
             span=>
             #[namespace = #namespace_str]
+            #(#cfgs)*
             extern "Rust" {
                 #[doc(hidden)]
                 fn #signal_handler_drop(handler: #signal_handler_alias);
--- a/crates/cxx-qt-lib/src/qml/qqmlapplicationengine.rs
+++ b/crates/cxx-qt-lib/src/qml/qqmlapplicationengine.rs
@@ -73,6 +73,11 @@ mod ffi {
             qobject: *mut QObject,
             url: &QUrl,
         );
+
+        #[qsignal]
+        #[rust_name = "object_creation_failed"]
+        #[cfg(any(cxxqt_qt_version_at_least_7, cxxqt_qt_version_at_least_6_4))]
+        fn objectCreationFailed(self: Pin<&mut QQmlApplicationEngine>, url: &QUrl);
     }
 
     unsafe extern "C++" {
--- a/crates/cxx-qt-gen/test_outputs/cfgs.rs
+++ b/crates/cxx-qt-gen/test_outputs/cfgs.rs
@@ -90,9 +90,9 @@ mod ffi {
         #[cfg(enabled)]
         fn inherit_enabled(self: &QObjectEnabled);
     }
+    #[cfg(not(enabled))]
     unsafe extern "C++" {
         #[cxx_name = "signal_disabled"]
-        #[cfg(not(enabled))]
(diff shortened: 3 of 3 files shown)


crates/cxx-qt-gen/src/naming/type_names.rs
```diff
@@ -119,6 +119,14 @@ impl TypeNames {
         type_names.populate_from_cxx_items(cxx_items, bridge_namespace, module_ident)?;
         type_names.populate_from_cxx_qt_data(cxx_qt_data, bridge_namespace, module_ident)?;
 
+        let qobject = Name {
+            rust: format_ident!("QObject"),
+            cxx: None,
+            module: None,
```
ahayzen-kdab, line 125: should this not be the `module_ident` ? So that if the type is used outside the bridge it is like `ffi::QObject` ?

crates/cxx-qt-lib/src/qml/qqmlapplicationengine.rs
```diff
@@ -65,6 +65,21 @@ mod ffi {
...
+        #[rust_name = "object_created"]
+        unsafe fn objectCreated(
+            self: Pin<&mut QQmlApplicationEngine>,
+            qobject: *mut QObject,
+            url: &QUrl,
+        );
+
+        #[qsignal]
+        #[rust_name = "object_creation_failed"]
+        #[cfg(cxxqt_qt_version_major = "6")]
```
ahayzen-kdab, line 79: this is since specifically 6.4 so we need the other check we have somewhere an at_least cfg flag instead

ahayzen-kdab: You should use

```
#[cfg(any(cxxqt_qt_version_at_least_7, cxxqt_qt_version_at_least_6_4))]
```

BenFordTytherington: should hopefully be fixed now

ahayzen-kdab: Due to the way our cfg currently works you need to specify at least 7+ and at least 6.4+

As at least 6.4+ will not be picked up by 7.0. Probably something we should improve

ahayzen-kdab:
Working on a CXX fix which should help with the CI issues.

ahayzen-kdab:
> Working on a CXX fix which should help with the CI issues.

So looks like #1222 has helped CI (only issue is around the MSRV now as versions have changed) using the patched CXX, so i'll try submitting that upstream.

crates/cxx-qt-lib/src/qml/qqmlapplicationengine.rs
```diff
@@ -73,6 +73,11 @@ mod ffi {
             qobject: *mut QObject,
             url: &QUrl,
         );
+
+        #[qsignal]
+        #[rust_name = "object_creation_failed"]
+        #[cfg(any(cxxqt_qt_version_at_least_7, cxxqt_qt_version_at_least_6_4))]
+        unsafe fn objectCreationFailed(self: Pin<&mut QQmlApplicationEngine>, url: &QUrl);
```
ahayzen-kdab, line 80: this doesn't need to be `unsafe` as it doesn't have a pointer in it's argument. Otherwise looks good.

crates/cxx-qt-gen/src/generator/rust/signals.rs
```diff
@@ -213,7 +221,9 @@ pub fn generate_rust_signal(
                 span =>
                 #(#cfgs)*
                 impl cxx_qt::signalhandler::CxxQtSignalHandlerClosure for #closure_struct {
+                    #(#cfgs)*
                     type Id = cxx::type_id!(#signal_handler_alias_namespaced_str);
+                    #(#cfgs)*
                     type FnType = dyn FnMut(#self_type_qualified, #(#parameters_qualified_type),*) + Send;
```
ahayzen-kdab, line 227: actually i think we don't need these cfgs they should be picked up by the impl cfg?

crates/cxx-qt-gen/src/generator/rust/signals.rs
```diff
@@ -174,6 +175,7 @@ pub fn generate_rust_signal(
                     #[doc = "Connect the given function pointer to the signal "]
                     #[doc = #signal_name_cpp]
                     #[doc = ", so that when the signal is emitted the function pointer is executed."]
+                    #(#cfgs)*
```
ahayzen-kdab, line 178: i think this one can still also go?

ahayzen-kdab (approved):

PR #1219, opened 2025-03-08
Serde implementations for mappable types

jnbooth:
This is a subset of #1151 that only adds serde support for the following:

- QByteArray
- QList
- QVector
- QStringList
- QUrl


LeonMatthesKDAB (approved):
Thank you @jnbooth for the contribution :)

Sorry it took me a while to approve this, I was at a trade show and sick the week after, but back to work now.

PR #1222, opened 2025-03-10
Test cfg cxx fix

ahayzen-kdab:
Test of #1217 with https://github.com/ahayzen-kdab/cxx/tree/cfg-on-type-alias


ahayzen-kdab:
Think this has all landed upstream now, closing.

PR #1223, opened 2025-03-12
replaced old generic qlist/qvector/qset headers where possible

QuentinLeeWeber:


LeonMatthesKDAB (approved):
LGTM, the caches will need a rebuild, but then this should be quicker to compile :crossed_fingers:

PR #1224, opened 2025-03-13
cxx-qt-build: return an Interface from CxxQtBuilder

ahayzen-kdab:
This is the first step towards #1125 towards a builder pattern with various stages.


crates/cxx-qt/build.rs
```diff
@@ -32,13 +32,16 @@ fn write_headers() {
 }
 
 fn main() {
+    // TODO: should we even have this ? Instead pass in an include_directory to our
+    // builder that handles this?
+    // As they are first being copied to the OUT_DIR/include/<crate> and then
+    // to the target/cxx-qt-build/../include
+    //
+    // This then causes problems because we try to symlink this dir now after
+    // the CXX files are written already
```
ahayzen-kdab, line 41: @LeonMatthesKDAB lets discuss this and the general way that we should handle how we have include dirs that use files from the source directory + have generated files (so is not as simple as a symlink). And whether there should be a default include dir already or not :thinking:

ahayzen-kdab: I note that in the CXX tutorials ( https://cxx.rs/tutorial.html ) they have the following

```cpp
#include "cxx-demo/include/blobstore.h"
#include "cxx-demo/src/main.rs.h"
```

For CXX the crate is already added as an include directory and the other folder is `target/cxxbridge/cxx-demo/src/main.rs.h`.

I wonder if we can do similar and use the fact that rust files are (always?) under `src/` ? As we could instead have

```
my-crate/
  include/
   myheader.h
  src/
    rustmod.rs
```

Then have in the export target folder
```
exporttarget/
  my-crate/
    include/   -- symlink to the my-crate/include/ (which avoids the current copy we always do)
    src/   -- where the generated files are written?
```

But probably needs more thought about how we can avoid collisions and if we can avoid a copy of the include like we do currently or not :thinking:

crates/cxx-qt-build/src/dir.rs
```diff
@@ -121,6 +121,10 @@ pub(crate) fn header_root() -> PathBuf {
     crate_target().join("include")
 }
 
+pub(crate) fn header_root_interface() -> PathBuf {
+    target().join("crates-include")
```
ahayzen-kdab, line 125: probably want a better name like `export-include` and sort out various naming across the codebase

crates/cxx-qt-build/src/lib.rs
```diff
@@ -1134,6 +1083,17 @@ extern "C" bool {init_fun}() {{
         // to the generated files without any namespacing.
         include_paths.push(header_root.join(&self.include_prefix));
 
+        // Automatically add any include/ directory from the cargo manifest dir
```
LeonMatthesKDAB, line 1086: This will mean any directory named `include/` will be available directly without prefix, right?

e.g. something like this:
```
include/test.h
Cargo.toml
```

Will mean it'll be possible to do:
```cpp
#include <test.h>
```

I think CXX behavior was to have this available as:
```cpp
#include <my_crate/include/test.h>
```

LeonMatthesKDAB: I think to specify certain include directories we wanted to add an `include` function that sets up the right include path with the given prefix and is passed to the Interface, right?

LeonMatthesKDAB (commented):
This seems to be going in the right direction, but some concepts around where which include dir is defined still require working out.

crates/cxx-qt-build/src/interface.rs
```diff
@@ -0,0 +1,208 @@
...
+        .chain(
+            dependencies
+                .iter()
+                .flat_map(|dep| &dep.manifest.exported_include_prefixes)
+                .cloned(),
+        )
+        .collect()
+}
+
+fn reexported_dependencies(interface: &Interface, dependencies: &[Dependency]) -> Vec<Dependency> {
```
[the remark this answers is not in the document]
ahayzen-kdab: Could be, this has just moved the existing code :-)

crates/cxx-qt-build/src/interface.rs
```diff
@@ -0,0 +1,208 @@
...
+        let dependencies = reexported_dependencies(&self, &self.dependencies);
+
+        self.manifest.exported_include_prefixes = all_include_prefixes(&self, &dependencies);
+
+        let manifest_path = dir::crate_target().join("manifest.json");
+        let manifest_json = serde_json::to_string_pretty(&self.manifest)
+            .expect("Failed to convert Manifest to JSON!");
+        std::fs::write(&manifest_path, manifest_json).expect("Failed to write manifest.json!");
+        println!(
+            "cargo::metadata=CXX_QT_MANIFEST_PATH={}",
```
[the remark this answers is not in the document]
ahayzen-kdab: A few lines further up (at the start of this export() function) we panic if the link name is empty?

crates/cxx-qt-build/src/lib.rs
```diff
@@ -417,15 +417,12 @@ impl CxxQtBuilder {
     /// included by later dependencies.
     ///
     /// The headers generated for this crate will be specified
-    pub fn library(interface_definition: Interface) -> Self {
-        let mut this = Self::new();
-        this.public_interface = Some(interface_definition);
-
+    pub fn library() -> Self {
```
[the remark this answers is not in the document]
ahayzen-kdab: Right i'll see if this can be removed now...

crates/cxx-qt-build/src/lib.rs
ahayzen-kdab: OK think i've sorted this out soooo

We now do the same as CXX with the best copy so any "header looking" files are copied from the source folder into the header root / include directory. So `include/test.h` ends up as `crate/include/test.h`.

If the `export_include_directory` is given a *relative* path then it is relative to the header root and is effectively copied across to the interface header directory.

This means that for `cxx-qt/include/header.h` these are automatically included and it only needs `.export_include_directory("cxx-qt", "cxx-qt")` to work, however note that now the path is `cxx-qt/include/header.h` rather than `cxx-qt/header.h` - but this then matches the same as CXX. This also works well in cxx-qt-lib-extras and in a similar way the include path has now changed.

For cxx-qt-lib we need to decide
- do we want to copy CXX include paths?
- should we just export all headers ignoring the features enabled?

If the answer is yes to both of these, then all the header writing/copying code in cxx-qt-lib can be removed and it becomes very simple.

If the answer is no to either/both of these, then we have to do the manual writing like currently in this code. And the way i've made this currently work is be passing an *absolute* path to `export_include_directory` which is pointing to the curated directory.

However i do see value in just simplifying this and copying CXX so that the include path patterns are very similar, this would mean a break with 0.8 if we went that route though or maybe we take that opportunity to have a single header you need to include instead like `cxx-qt-lib/include/qtcore.h` rather than a header for each type. @LeonMatthesKDAB what do we think?

crates/cxx-qt/build.rs
ahayzen-kdab: We have a second discussion further down that is newer lets close this one...

PR #1226, opened 2025-03-14
Add qobject extension trait

### LINKED ISSUE #562: Traits for CxxQtType/QtObject

BenFordTytherington:
- cxx-qt-lib trait for additional qobject wrappers

Closes #562 

The added methods are:
- blockSignals
- signalsBlocked
- setObjectName
- objectName
- parent
- setParent

### DIFF
--- a/crates/cxx-qt-lib/src/core/mod.rs
+++ b/crates/cxx-qt-lib/src/core/mod.rs
@@ -47,6 +47,9 @@ pub use cxx_qt::{QMetaObjectConnection, QMetaObjectConnectionGuard};
 mod qmodelindex;
 pub use qmodelindex::QModelIndex;
 
+mod qobject;
+pub use qobject::QObjectExt;
+
 mod qpersistentmodelindex;
 pub use qpersistentmodelindex::QPersistentModelIndex;
 
--- a/crates/cxx-qt-lib/src/core/qobject.rs
+++ b/crates/cxx-qt-lib/src/core/qobject.rs
@@ -1,0 +1,100 @@
+// SPDX-FileCopyrightText: 2025 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Ben Ford <ben.ford@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+pub use cxx_qt::QObject;
+use std::pin::Pin;
+
+#[cxx::bridge]
+pub mod ffi {
+    unsafe extern "C++" {
+        include!("cxx-qt-lib/qstring.h");
+        type QString = crate::QString;
+    }
+
+    unsafe extern "C++" {
+        include!("cxx-qt-lib/qobject.h");
+        #[rust_name = "QObjectExternal"]
+        type QObject;
+        #[rust_name = "block_signals"]
+        pub fn blockSignals(self: Pin<&mut Self>, block: bool) -> bool;
+
+        #[rust_name = "signals_blocked"]
+        pub fn signalsBlocked(&self) -> bool;
+
+        #[rust_name = "set_object_name"]
+        pub fn setObjectName(self: Pin<&mut Self>, name: &QString);
+
+        #[rust_name = "object_name"]
+        pub fn objectName(&self) -> QString;
+
+        pub fn parent(&self) -> *mut QObjectExternal;
+
+        #[rust_name = "set_parent"]
+        pub unsafe fn setParent(self: Pin<&mut Self>, parent: *mut QObjectExternal);
+    }
+}
+
+use ffi::{QObjectExternal, QString};
+
+/// Trait which exposes methods available on a `QObject`.
+/// Exposes some basic signals and methods for now, more to be added.
+pub trait QObjectExt {
+    fn block_signals(self: Pin<&mut Self>, block: bool) -> bool;
+
+    fn signals_blocked(&self) -> bool;
+
+    fn set_object_name(self: Pin<&mut Self>, name: &QString);
+
+    fn object_name(&self) -> QString;
+
+    fn parent(&self) -> *mut Self;
+
+    fn set_parent(self: Pin<&mut Self>, parent: &Self);
+}
+
+/// Used to convert the QObject type from the library type to the C++ type, as a pin
+fn cast_pin(obj: Pin<&mut QObject>) -> Pin<&mut QObjectExternal> {
+    unsafe {
+        let mut_ptr = obj.get_unchecked_mut() as *mut QObject as *mut QObjectExternal;
+        Pin::new_unchecked(&mut *mut_ptr)
+    }
+}
+
+/// Used to convert the QObject type from the library type to the C++ type
+fn cast(obj: &QObject) -> &QObjectExternal {
+    unsafe {
+        let ptr = obj as *const QObject as *const QObjectExternal;
+        &*ptr
+    }
+}
+
+impl QObjectExt for QObject {
+    fn block_signals(self: Pin<&mut Self>, block: bool) -> bool {
+        cast_pin(self).block_signals(block)
+    }
+
+    fn signals_blocked(&self) -> bool {
+        cast(self).signals_blocked()
+    }
+
+    fn set_object_name(self: Pin<&mut Self>, name: &QString) {
+        cast_pin(self).set_object_name(name)
+    }
+
+    fn object_name(&self) -> QString {
+        cast(self).object_name()
+    }
+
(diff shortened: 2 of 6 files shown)


crates/cxx-qt-lib/src/core/mod.rs
```diff
@@ -110,6 +113,10 @@ mod qvector;
 pub use qvector::{QVector, QVectorElement};
 
 mod qtlogging;
+mod qobject;
+
+pub use qobject::QObjectExt;
```
ahayzen-kdab, line 118: try to keep these grouped and ordered alphabetically, so should probably be higher up?

crates/cxx-qt-lib/src/core/mod.rs
```diff
@@ -39,6 +41,7 @@ mod qmarginsf;
 pub use qmarginsf::QMarginsF;
 
 // Reexport QMetaObjectConnection and guard from cxx-qt
+use cxx_qt::QObject;
```
ahayzen-kdab, line 44: probably not needed anymore?

crates/cxx-qt-lib/src/core/mod.rs
```diff
@@ -4,7 +4,9 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
 mod qbytearray;
+
 pub use qbytearray::QByteArray;
+use std::pin::Pin;
```
ahayzen-kdab, line 9: probably don't need the Pin or the extra line?

crates/cxx-qt-lib/src/core/mod.rs
```diff
@@ -142,4 +149,4 @@ mod ffi {
 /// ```
 ///
 /// See: <https://github.com/dtolnay/cxx/issues/1049>
-pub use ffi::c_void;
+pub use ffi::c_void;
```
ahayzen-kdab, line 152: can you configure your editor to end with a new line so it is like before? (same on the other file)

crates/cxx-qt-lib/src/core/qobject.rs
```diff
@@ -0,0 +1,37 @@
...
+use std::pin::Pin;
+
+#[cxx_qt::bridge]
+mod ffi {
+    unsafe extern "C++" {
+        #[rust_name = "QObjectExternal"]
+        type QObject;
+
+        #[cxx_name = "blockSignals"]
+        pub fn blockSignalsQObject(self: Pin<&mut QObjectExternal>, block: bool) -> bool;
```
ahayzen-kdab, line 16: Can this instead be

```rust
        #[rust_name = "block_signals"]
        pub fn blockSignals(self: Pin<&mut QObjectExternal>, block: bool) -> bool;
```

and then update below to `cast(self).block_signals(block)`

BenFordTytherington: yup

crates/cxx-qt-lib/src/core/qobject.rs
```diff
@@ -0,0 +1,95 @@
...
+mod ffi {
+    unsafe extern "C++" {
+        #[rust_name = "QObjectExternal"]
+        type QObject;
+
+        include!("cxx-qt-lib/qstring.h");
+        type QString = crate::QString;
+
+        #[rust_name = "block_signals"]
+        pub fn blockSignals(self: Pin<&mut QObjectExternal>, block: bool) -> bool;
```
LeonMatthesKDAB, line 19: Nit: You can use the shorthand syntax `self: Pin<&mut Self>`/`&self` here if you move the QString import into a separate `extern "C++"` block.

crates/cxx-qt-lib/src/core/qobject.rs
```diff
@@ -0,0 +1,95 @@
...
+
+        #[rust_name = "set_parent"]
+        pub unsafe fn setParent(self: Pin<&mut QObjectExternal>, parent: *mut QObjectExternal);
+    }
+}
+
+use crate::core::qobject::ffi::QString;
+use ffi::QObjectExternal;
+
+pub trait QObjectExt {
```
LeonMatthesKDAB, line 40: Please add documentation here, and on cxx_qt::QObject that you need to import this trait in order to access most QObject functions.
Otherwise it could be confusing that the functions aren't available automatically.

crates/cxx-qt-lib/src/core/qobject.rs
LeonMatthesKDAB: Also a good reminder that we don't support this in CXX-Qt yet...
I've opened an issue for it: #1245

LeonMatthesKDAB (commented):
Some nitpicks, and some documentation is missing.

Otherwise looks good :)

crates/cxx-qt-lib/src/core/qobject.rs
BenFordTytherington: Done, could maybe do with a touch more documentation, but I think it's fine for now

crates/cxx-qt-lib/src/core/qobject.rs
BenFordTytherington: Can fix once #1249 gets merged

LeonMatthesKDAB: Hm, that reminds me, this doesn't have to be a `#[cxx_qt::bridge]`, right?
You should be able to just use `#[cxx::bridge]` and fix this right away, no?

crates/cxx-qt-lib/src/core/qobject.rs
```diff
@@ -0,0 +1,95 @@
+// SPDX-FileCopyrightText: 2025 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Ben Ford <ben.ford@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use cxx_qt::QObject;
```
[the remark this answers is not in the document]
LeonMatthesKDAB: Why does this break the orphan rule? We just re-export a type from a dependency?

crates/cxx-qt-lib/src/core/qobject.rs
ahayzen-kdab: In it's current state it doesn't need a cxx_qt::bridge, but if eg the destroyed signal was added (which could be useful?) then it'd need a cxx_qt::bridge.

https://doc.qt.io/qt-6/qobject.html#destroyed

crates/cxx-qt-lib/src/core/qobject.rs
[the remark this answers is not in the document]
BenFordTytherington:
```
error[E0116]: cannot define inherent `impl` for a type outside of the crate where the type is defined
  --> crates/cxx-qt-lib/src/core/qobject.rs:32:16
   |
32 |         pub fn parent(&self) -> *mut QObject;
   |                ^^^^^^ impl for type defined outside of crate
   |
   = note: define and implement a trait or new type instead
   ```
   Since this is defined in cxx_qt_lib, and QObject is in cxx_qt

crates/cxx-qt-lib/src/core/qobject.rs
BenFordTytherington: Updated to `#[cxx::bridge]` and used self shorthand

LeonMatthesKDAB (approved):

PR #1227, opened 2025-03-14
cxx-qt-lib: Add two more QtLogging functions

redstrate:
This adds two new global functions available in <QtLogging>: [qFormatLogMessage](https://doc.qt.io/qt-6/qtlogging.html#qFormatLogMessage) and [qSetMessagePattern](https://doc.qt.io/qt-6/qtlogging.html#qSetMessagePattern).

qFormatLogMessage is similar to qt_message_output, but actually documented. Instead of going to the default message handler though, it outputs to a QString. This is probably not useful without message handler support, but it's easy to expose.

qSetMessagePattern is very useful, as it allows CXX-Qt applications to set custom Qt message patterns for logging.

### DIFF
--- a/crates/cxx-qt-lib/src/core/qtlogging.rs
+++ b/crates/cxx-qt-lib/src/core/qtlogging.rs
@@ -36,6 +36,21 @@ mod ffi {
         /// Outputs a message in the Qt message handler.
         fn qt_message_output(msgType: QtMsgType, context: &QMessageLogContext, string: &QString);
 
+        /// Generates a formatted string out of the type, context, str arguments.
+        #[cxx_name = "qFormatLogMessage"]
+        fn q_format_log_message(
+            msgType: QtMsgType,
+            context: &QMessageLogContext,
+            string: &QString,
+        ) -> QString;
+
+        /// Changes the output of the default message handler.
+        ///
+        /// # Safety
+        /// This function is marked as unsafe because it is not guaranteed to be thread-safe.
+        #[cxx_name = "qSetMessagePattern"]
+        unsafe fn q_set_message_pattern(pattern: &QString);
+
         #[cxx_name = "qmessagelogcontext_line"]
         #[doc(hidden)]
         fn line(context: &QMessageLogContext) -> i32;
@@ -131,4 +146,4 @@ unsafe impl ExternType for QMessageLogContext<'_> {
 }
 
 use crate::const_assert;
-pub use ffi::{qt_message_output, QtMsgType};
+pub use ffi::{q_format_log_message, q_set_message_pattern, qt_message_output, QtMsgType};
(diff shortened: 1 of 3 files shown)


crates/cxx-qt-lib/src/core/qtlogging.rs
```diff
@@ -36,6 +36,18 @@ mod ffi {
         /// Outputs a message in the Qt message handler.
         fn qt_message_output(msgType: QtMsgType, context: &QMessageLogContext, string: &QString);
 
+        /// Generates a formatted string out of the type, context, str arguments.
+        #[cxx_name = "qFormatLogMessage"]
+        fn format_log_message(
```
LeonMatthesKDAB, line 41: Both of these new functions are exported at the top-level of cxx-qt-lib as `cxx_qt_lib::format_log_message;` and `cxx_qt_lib::set_message_pattern` respectively.
Within that namespace, we usually stick to the names provided by Qt with only a case conversion.
So please name these methods `q_format_log_message` and `q_set_message_pattern` to indicate they are coming from Qt directly and aren't e.g. cxx-qt-lib specific functions.

crates/cxx-qt-lib/src/core/qtlogging.rs
```diff
@@ -36,6 +36,18 @@ mod ffi {
...
+            msgType: QtMsgType,
+            context: &QMessageLogContext,
+            string: &QString,
+        ) -> QString;
+
+        /// Generates a formatted string out of the type, context, str arguments.
```
LeonMatthesKDAB, line 47: This doc comment belongs to qFormatLogMessage, not qSetMessagePattern.

crates/cxx-qt-lib/src/core/qtlogging.rs
```diff
@@ -36,6 +36,18 @@ mod ffi {
...
+        #[cxx_name = "qSetMessagePattern"]
+        fn set_message_pattern(pattern: &QString);
```
LeonMatthesKDAB, line 49: The Qt documentation states that qFormatLogMessage is thread-safe.

However, it does **not** state that `set_message_pattern` is thread-safe!
Can you investigate whether this function is thread-safe?
Otherwise we may need to make it `unsafe` or enforce safety somehow...

LeonMatthesKDAB: Note that we're having similar issues with #1205

LeonMatthesKDAB:
Ping @redstrate, can you solve the remaining issues?

crates/cxx-qt-lib/src/core/qtlogging.rs
redstrate: Fixed

redstrate:
> Ping @redstrate, can you solve the remaining issues?

Done, sorry it's been a busy week - but I have time now :)

crates/cxx-qt-lib/src/core/qtlogging.rs
redstrate: I think it's fine to mark it as unsafe, I should probably add a comment explaining why it is right?

PR #1229, opened 2025-03-14
WIP add book entry for instantiating in Rust

BenFordTytherington:


LeonMatthesKDAB:
@BenFordTytherington Please also document that you can add constructor arguments if required :)

BenFordTytherington:
> @BenFordTytherington Please also document that you can add constructor arguments if required :)

I wasn't aware that was an option with these templates, what would that syntax look like?

LeonMatthesKDAB:
@BenFordTytherington , it follows the same pattern as our `construct` function, e.g. here:
https://github.com/KDAB/cxx-qt/blob/64d4ee6ba01af1902dcff56870af9b4ff898cabe/crates/cxx-qt-lib/src/core/qvector/qvector_i64.rs#L28

You just declare the appropriate parameters and if the declared parameters match an existing constructor it will just work :)
And you can even create multiple constructors that way by renaming them.

Most QObjects for example can be created with an optional `QObject* parent`, so that could be a good example to show, e.g.:
```rust
unsafe extern "C++" {
    include!("cxx-qt-lib/common.h");

    #[rust_name = "my_object_new"]
    fn make_unique() -> UniquePtr<MyObject>;
    
    #[rust_name = "my_object_new_with_parent"]
    fn make_unique(parent: *mut QObject) -> UniquePtr<MyObject>;
}
```

I think the docs should also mention that these functions are provided by CXX-Qt-lib as helper functions (which explains why they are in the rust::cxxqtlib1 namespace.
It's important to note that this is *not* `std::make_unique`.
These wrappers are specifically designed for better CXX compatibility.

LeonMatthesKDAB (dismissed):

PR #1231, opened 2025-03-17
Support QFlags

### LINKED ISSUE #1123: QFlags support
We need to figure out a plan for QFlags maybe something with https://crates.io/crates/bitflags and https://crates.io/crates/flagset etc

_Originally posted by @ahayzen-kdab in https://github.com/KDAB/cxx-qt/issues/555#issuecomment-1963612638_

jnbooth:
This PR defines a `QFlags<T: QFlag>` type to address the non-codegen parts of #1123. It mimics the API of Qt's own `QFlags`, including operators. The one exception is that because Rust doesn't have the `~` operator, the `~` operator in Qt is mapped to `!` in Rust, while the `!` operator in Qt is mapped to `QFlags::is_empty` in Rust.

The backing `QFlag` trait provides two things: a function for converting from the FFI enum type to its underlying representation (i.e. `value.repr`) and the `TypeId` of the corresponding `QFlags` type. For example:

```rs
#[cxx::bridge]
mod ffi {
    #[repr(i32)]
    enum MyFlag { A, B }; 
}
pub use ffi::MyFlag;

unsafe impl QFlag for MyFlag {
    type TypeId = type_id!("QFlags_MyFlag");
    type Repr = i32;
    fn to_repr(self) -> Self::Repr {
        self.repr
    }
}
pub type MyFlags = QFlags<MyFlag>;
```

It's unsafe because it has [the same safety contract as `cxx::ExternType::Id`](https://docs.rs/cxx/latest/cxx/trait.ExternType.html#integrating-with-bindgen-generated-types). There's also an internal utility macro to replace the above impl block with

```rs
impl_qflag!(MyFlag, "QFlags_MyFlag", i32);
```

Since `QFlags` can now be defined, this PR also adds [`QGuiApplication::keyboard_modifiers`](https://doc.qt.io/qt-6/qguiapplication.html#keyboardModifiers), [`QGuiApplication::mouse_buttons`](https://doc.qt.io/qt-6/qguiapplication.html#mouseButtons), and  [`QGuiApplication::query_keyboard_modifiers`](https://doc.qt.io/qt-6/qguiapplication.html#queryKeyboardModifiers),

Closes #1123

### DIFF
--- a/crates/cxx-qt-lib/src/core/qflags/mod.rs
+++ b/crates/cxx-qt-lib/src/core/qflags/mod.rs
@@ -8,11 +8,11 @@ use std::fmt::Debug;
 use std::hash::Hash;
 use std::ops::{BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Not};
 
-mod macros;
-
 mod qflag;
 pub use qflag::QFlag;
 use qflag::{QFlagExt, QFlagRepr};
+
+mod util;
 
 type QFlagInt<T> = <T as QFlagExt>::Int;
 
--- a/crates/cxx-qt-lib/src/core/qflags/mod.rs
+++ b/crates/cxx-qt-lib/src/core/qflags/mod.rs
@@ -106,14 +106,12 @@ impl<T: QFlag> QFlags<T> {
         }
     }
 
-    #[cfg(cxxqt_qt_version_at_least_6_2)]
     /// Returns `true` if any flag set in *flag* is also set in this flags object, otherwise
     /// `false`. If *flag* has no flags set, the return will always be `false`.
     pub fn test_any_flag(self, flag: T) -> bool {
         self.test_any_flags(Self::from(flag))
     }
 
-    #[cfg(cxxqt_qt_version_at_least_6_2)]
     /// Returns `true` if any flag set in *flags* is also set in this flags object, otherwise
     /// `false`. If *flags* has no flags set, the return will always be `false`.
     pub fn test_any_flags(self, flags: Self) -> bool {
@@ -303,7 +301,6 @@ mod test {
         assert!(flags.test_flags(other));
     }
 
-    #[cfg(cxxqt_qt_version_at_least_6_2)]
     #[test]
     fn qflags_test_any_flags() {
         let flags = KeyboardModifier::ControlModifier
--- a/crates/cxx-qt-lib/src/core/qflags/mod.rs
+++ b/crates/cxx-qt-lib/src/core/qflags/mod.rs
@@ -16,7 +16,7 @@ mod util;
 
 type QFlagInt<T> = <T as QFlagExt>::Int;
 
-/// The QFlags<T> class is a template class, where T is an enum type.
+/// The `QFlags<T>` class is a template class, where T is an enum type.
 /// QFlags is used throughout Qt for storing combinations of enum values.
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[repr(transparent)]
--- a/crates/cxx-qt-lib/src/core/qflags/mod.rs
+++ b/crates/cxx-qt-lib/src/core/qflags/mod.rs
@@ -10,18 +10,19 @@ use std::ops::{BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, N
 
 mod qflag;
 pub use qflag::QFlag;
-use qflag::{QFlagExt, QFlagRepr};
+use qflag::QFlagExt;
+
+mod repr;
+pub use repr::QFlagRepr;
 
 mod util;
 
-type QFlagInt<T> = <T as QFlagExt>::Int;
-
 /// The `QFlags<T>` class is a template class, where T is an enum type.
-/// QFlags is used throughout Qt for storing combinations of enum values.
+/// QFlags are used throughout Qt for storing combinations of enum values.
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[repr(transparent)]
 pub struct QFlags<T: QFlag> {
-    repr: QFlagInt<T>,
+    repr: <T::Repr as QFlagRepr>::Int,
 }
 
 impl<T: QFlag> Copy for QFlags<T> {}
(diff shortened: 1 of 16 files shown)


crates/cxx-qt-lib/src/gui/qguiapplication.rs
```diff
@@ -221,4 +236,39 @@ impl QGuiApplication {
...
     }
+
+    /// Returns the current state of the modifier keys on the keyboard. The current state is updated
+    /// synchronously as the event queue is emptied of events that will spontaneously change the
+    /// keyboard state (QEvent::KeyPress and QEvent::KeyRelease events).
+    ///
+    /// It should be noted this may not reflect the actual keys held on the input device at the time
+    /// of calling but rather the modifiers as last reported in an event.
+    /// If no keys are being held Qt::NoModifier is returned.
+    pub fn keyboard_modifiers() -> KeyboardModifiers {
```
LeonMatthesKDAB, line 247: Digging into the Qt sources reveals that these functions access a static variable directly to read the keyboard modifiers.

This is unfortunately not thread safe.
Please make this a member function, so we can at least ensure it's called in the same thread that created the QGuiApplication.

The same applies to mouse_buttons and query_keyboard_modifiers. (and [set_]desktop_file_name for that matter).

Related to #1205 and #1216

crates/cxx-qt-lib/src/core/qflags/qflag.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// ```ignore
+    /// fn to_repr(self) -> Self::Repr {
+    ///     self.repr
+    /// }
+    /// ```
+    fn to_repr(self) -> Self::Repr;
+}
+
+/// Internal utility trait for converting `T` in a `QFlag<T>` to the corresponding integer type.
+pub trait QFlagExt: QFlag {
```
LeonMatthesKDAB, line 93: Please mark this trait as `pub(super)` as it's an internal utility trait.

crates/cxx-qt-lib/src/core/qflags/qflag.rs
```diff
@@ -0,0 +1,106 @@
...
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use std::fmt::Debug;
+use std::hash::Hash;
+use std::ops::{BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Not};
+
+use cxx::ExternType;
+
+pub trait QFlagRepr: Sized {
```
LeonMatthesKDAB, line 12: Please mark this trait as private, as it should only be implemented here, and not by downstream consumers.

crates/cxx-qt-lib/src/core/qflags/qflag.rs
```diff
@@ -0,0 +1,106 @@
...
+    ///
+    /// ```ignore
+    /// fn to_repr(self) -> Self::Repr {
+    ///     self.repr
+    /// }
+    /// ```
+    fn to_repr(self) -> Self::Repr;
+}
+
+/// Internal utility trait for converting `T` in a `QFlag<T>` to the corresponding integer type.
```
LeonMatthesKDAB, line 92: Can you elaborate this comment a bit to explain the difference between the `Repr` and the `Int`.

It took me a while to understand why this trait is useful/necessary.

My current understanding is:

The enum (QFlag) that the QFlags type has a bit-field of may have a different representation (Repr) from the value the `QFlags` holds internally (Int).

The QFlagRepr trait represents this mapping, by assigning the correct `Int` type that QFlags holds to each possible repr type that the QFlag may contain.

From that perspective the trait makes a lot of sense to make this conversion more convenient, but it's documentation should explain this in more detail.

crates/cxx-qt-lib/src/core/qflags/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+    /// flags object only matches if it also has no flags set.
+    pub fn test_flags(self, flags: Self) -> bool {
+        if flags.is_empty() {
+            self.is_empty()
+        } else {
+            self.repr & flags.repr == flags.repr
+        }
+    }
+
+    #[cfg(cxxqt_qt_version_at_least_6_2)]
```
LeonMatthesKDAB, line 108: You can remove the `cfg` flag here and on test_any_flags.

While Qt only supports these functions after 6.2, we can support them for any Qt version, as the logic is still valid and we don't need to call any Qt functions that didn't exist at the time.

crates/cxx-qt-lib/src/core/qflags/util.rs
```diff
@@ -0,0 +1,34 @@
+// SPDX-FileCopyrightText: 2025 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Joshua Booth <joshua.n.booth@gmail.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+#[macro_export]
+macro_rules! impl_qflag {
```
LeonMatthesKDAB, line 7: Please rename this macro to `unsafe_impl_qflag`, as it otherwise hides the unsafety of the trait that it implements.

crates/cxx-qt-lib/src/core/qflags/qflag.rs
```diff
@@ -0,0 +1,106 @@
...
+/// Furthermore, the programmer asserts that `Repr` is the backing integer representation of the
+/// enum type that is the `Self` type of the impl.
+pub unsafe trait QFlag: Sized {
+    /// A type-level representation of the C++ namespace and type name of this type's `QFlags<T>`.
+    ///
+    /// This will always be defined using `type_id!` in the following form:
+    ///
+    /// ```ignore
+    /// type TypeId = cxx::type_id!("QFlags_MyType");
+    /// ```
```
LeonMatthesKDAB, line 74:
```suggestion
    /// This will always be defined using [cxx::type_id!].
```

I would remove the example here, as it suggests that the type_id! should always follow the form `QFlags_MyType`, which is often not the correct type_id! (e.g. MouseButtons uses `Qt::MouseButtons`).
So let's just refer to the type_id documentation on the CXX side.

LeonMatthesKDAB (commented):
Hi @jnbooth thank you for opening the PR on this.

I left some small comments, the approach itself is solid!

crates/cxx-qt-lib/src/core/qflags/qflag.rs
jnbooth: For some `T: QFlag`, `QFlag::Repr` is the backing integer representation of `T`, and `QFlagRepr::Int` is the backing integer representation of `QFlags<T>`. I've added documentation to clarify that.

crates/cxx-qt-lib/src/core/qflags/qflag.rs
jnbooth: I think it's useful for users to know which types implement `QFlagRepr`, so I've sealed the trait to prevent downstream implementations.

LeonMatthesKDAB: Cool, that works as well :+1:

LeonMatthesKDAB (approved):
Excellent work, thank you for the contribution @jnbooth

PR #1238, opened 2025-03-24
Fixed broken link to book in Rust generated code

whs31:
(Previous link leads to 404)


LeonMatthesKDAB (approved):
Thank you @whs31 for spotting the issue.
Our tests contained a few more instances of the invalid link, so I updated those as well.

PR #1241, opened 2025-03-28
fix: Allow generated methods to be reused by other properties

### LINKED ISSUE #1232: Unable to reference auto-generated qproperty flags in another flag
Currently, this does not work:
```rust
#[qobject]
#[qproperty(i32, number, READ, WRITE, NOTIFY)]
#[qproperty(QString, string, READ, WRITE, NOTIFY=number_changed)] // <-- CXX-Qt will error that number_changed doesn't exist.
```

However, the `number_changed` signal is being generated by the `number` property.
This sounds like an issue with name resolution, as the generated values are being added too late.

We got feedback from the training yesterday that this is pretty unexpected, especially as you can use the signal outside the bridge, and it is also an error to declare your own signal with that same name.

BenFordTytherington:
- Add list of methods pending generation
- Perform a lookup in the pending methods if they aren't found in the bridge
- Use name of pending method if found

Should fix #1232


crates/cxx-qt-gen/src/generator/naming/property.rs
```diff
@@ -43,7 +43,21 @@ impl NameState {
...
+                let name = if let Ok(name) = lookup {
+                    name
+                } else if let Some(name) = structured_qobject.pending_lookup(ident) {
+                    // If a bridge method isn't found, it may be pending generation
+                    name
+                } else {
+                    // return the not found error if the method wasn't pending either
+                    lookup?
+                };
+                Self::Custom(name)
```
LeonMatthesKDAB, line 59: Why is the lookup of auto-generated functions/signals split into the separate pending_lookup function?
These are normal functions, just auto-generated. So why treat them any different in lookup?

IMO method_lookup and signal_lookup should just return them.

crates/cxx-qt-gen/src/parser/property.rs
```diff
@@ -224,6 +227,25 @@ impl ParsedQProperty {
...
+            pending.push(getter_name_from_property(&self.name));
+        }
+
+        if let Some(FlagState::Auto) = &self.flags.write {
+            pending.push(setter_name_from_property(&self.name));
+        }
+
+        if let Some(FlagState::Auto) = &self.flags.notify {
+            pending.push(notify_name_from_property(&self.name));
+        }
```
LeonMatthesKDAB, line 245: This is a signal, not a method, so needs to be treated differently during lookup.
We wouldn't want someone to e.g. use:

```
#[qproperty(i32, number, READ, NOTIFY=number)]
```
Which cxx-qt-gen would now accept.

crates/cxx-qt-gen/src/generator/structuring/qobject.rs
```diff
@@ -40,17 +41,29 @@ impl<'a> StructuredQObject<'a> {
...
             constructors: vec![],
+            pending_generation,
             threading: false,
         }
     }
 
+    /// Searches for a name in the list of functions pending generation
+    pub fn pending_lookup(&self, id: &Ident) -> Option<Name> {
+        lookup(&self.pending_generation, id, |pending| pending)
+    }
```
LeonMatthesKDAB, line 65: I don't think this should be a separate function, the lookup should just work via method_lookup and signal_lookup respectively.
If you want to look up whether a certain function can be named, then I don't think you need to care whether it's a generated function or a manually written one.

LeonMatthesKDAB (changes requested):
The overall approach is good, but it needs to be split cleanly between signals and methods, and I don't see the need to separate the lookup.

crates/cxx-qt-gen/src/generator/naming/property.rs
BenFordTytherington: Because at this stage, the actual function signatures don't exist, the pending_lookup is simply searching through a set of names of the functions which are pending generation, whereas the signal and method lookups are finding actual methods which have been associated during structuring, so they have a declaration, and all the associated data already, whereas the pending ones just have a name, and will be generated later.

BenFordTytherington: Fixed

crates/cxx-qt-gen/src/generator/structuring/qobject.rs
BenFordTytherington: Resolved

crates/cxx-qt-gen/src/parser/property.rs
LeonMatthesKDAB: Looks like this is resolved now.

LeonMatthesKDAB (approved):

PR #1242, opened 2025-03-28
Add shorthand for cxx_qt::Initialize

### LINKED ISSUE #1236: Add `impl cxx_qt::Initialize for X {}` as shorthand for `impl cxx_qt::Constructor<()> for X {}`
It can be confusing that you can use Initialize outside the bridge, but not inside.
We could just have that be an alias for the expected Constructor declaration to improve ease-of-use.

BenFordTytherington:
- `impl cxx_qt::Initialize for x {}` can now be written in the bridge as shorthand
- It is shorthand for `impl cxx_qt::Constructor<()> for x {}`
- Closes #1236

### DIFF
--- a/crates/cxx-qt-gen/src/parser/trait_impl.rs
+++ b/crates/cxx-qt-gen/src/parser/trait_impl.rs
@@ -3,7 +3,7 @@ use indoc::formatdoc;
 // SPDX-FileContributor: Leon Matthes <leon.matthes@kdab.com>
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
-use syn::{Error, Ident, ItemImpl, Path, Result, Token, Type, TypePath};
+use syn::{parse_quote, Error, Ident, ItemImpl, Path, Result, Token, Type, TypePath};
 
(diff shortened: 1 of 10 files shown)


crates/cxx-qt-gen/src/parser/trait_impl.rs
```diff
@@ -55,6 +61,7 @@ impl TraitKind {
                     CXX-Qt currently only supports:
                       - cxx_qt::Threading
                       - cxx_qt::Constructor
+                      - cxx_qt::Initialize (as shorthand for Constructor<()>)
```
LeonMatthesKDAB, line 64: This also needs updating in the book and in the docs for cxx_qt::Initialize.

LeonMatthesKDAB (commented):
Works well, but is missing a bit of documentation.

Please also update e.g. qml_features to use the shorthand where appropriate.

crates/cxx-qt-gen/src/parser/trait_impl.rs
BenFordTytherington: Done

ahayzen-kdab (approved):
LGTM :-)

PR #1243, opened 2025-03-29
Replace static method wrappers with aliases

jnbooth:
This is how we currently handle static methods:

```cpp
::std::int32_t
qpointDotProduct(const QPoint& p1, const QPoint& p2)
{
  return QPoint::dotProduct(p1, p2);
}
```

We can cut down on code generation by doing this instead: 

```cpp
inline ::std::int32_t (*qpointDotProduct)(const QPoint&, const QPoint&) = QPoint::dotProduct;
```


LeonMatthesKDAB (approved):
Very nice, ~270 lines of code removed without a change in functionality :+1:

Issue #1237, opened 2025-03-20
Long path names break Windows build (originally: MSVC Generator can break CMake builds)


LeonMatthesKDAB:
During yesterdays training @narnaud  could only build with the Ninja generator, not the MSVC one.

We need to investigate whether we can reproduce this and how to fix it.

@narnaud can you reproduce the error and provide us with the error message please?

narnaud:
The error log is quite long, here is the beginning:
```
[...]

PR #1246, opened 2025-03-31
TEMP: Try to use MSVC generator in CI.

LeonMatthesKDAB:
Trying to replicate #1237.

DO NOT MERGE, as this will likely break sccache!


LeonMatthesKDAB:
It turned out this was about long path names after all, so this isn't needed.

PR #1247, opened 2025-04-01
Automatically implement Upcast<Self> for all types

jnbooth:
Suppose a user creates a function like this:

```rust
fn do_something<T: Upcast<MyObject>>(object: &T);
```

It should be possible to call `do_something(object)` where `object` *is* `&MyObject`.

The `Upcast<Self>` implementations are all no-ops.

### DIFF
--- a/crates/cxx-qt/src/lib.rs
+++ b/crates/cxx-qt/src/lib.rs
@@ -267,6 +267,28 @@ pub trait Upcast<T> {
     }
 }
 
+impl<T> Upcast<T> for T {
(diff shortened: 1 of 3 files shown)


LeonMatthesKDAB (approved):
Thank you @jnbooth.
Together with the changes that @BenFordTytherington is working on in #1252 this means we'll have to use somewhat ugly/verbose disambiguation syntax to specify the type to upcast to, but that's just how it has to be I guess.

PR #1249, opened 2025-04-04
Add shorthand for &self in RustQt blocks where the type can be inferred

### LINKED ISSUE #1245: Support for shortand `&self` in bridge
CXX supports using `&self` and `self: Pin<&mut Self>` instead of the longer: `self: &T` as long as there is only a single type in a given block.

This is very handy and we should support the same in CXX-Qt:

```rust
extern "C++Qt" {
    #[qobject]

    // No need to specify the type explicitly here, MyObject is the only type defined in this `extern` block, so `Self` refers to `MyObject` implicitly.
    fn my_function(&self) -> i32;
    fn my_other_function(self: Pin<&mut Self>) -> i32;
}
```

The open question we currently have is how this would deal with name resolution in the structuring phase...
So we may need to refactor our parser to return which objects/methods were defined in which block, instead of adding them all into a single structure like we currently do.

BenFordTytherington:
Allows `Pin<&mut Self>` and `&self` inside the bridge, as shorthand.
- Closes #1245 
- Refactors parser to separate RustQt and C++Qt blocks better, in a more consistent way
- Adds shorthand module, for other shorthands such as #1278 and #1242

### DIFF
--- a/crates/cxx-qt-gen/src/lib.rs
+++ b/crates/cxx-qt-gen/src/lib.rs
@@ -11,7 +11,7 @@
 mod generator;
 mod naming;
 mod parser;
-mod preprocessor;
+mod shorthand;
 mod syntax;
 mod writer;
 
@@ -21,7 +21,7 @@ pub use generator::{
     GeneratedOpt,
 };
 pub use parser::Parser;
-pub use preprocessor::self_inlining::qualify_self_types;
+pub use shorthand::self_inlining;
 pub use syntax::{parse_qt_file, CxxQtFile, CxxQtItem};
 pub use writer::{cpp::write_cpp, rust::write_rust};
 
@@ -88,6 +88,7 @@ mod tests {
             $(assert!($parse_fn(syn::parse_quote! $input).is_err());)*
         }
     }
+    use crate::self_inlining::qualify_self_types;
     pub(crate) use assert_parse_errors;
 
     /// Helper for formating C++ code
--- a/crates/cxx-qt-gen/src/lib.rs
+++ b/crates/cxx-qt-gen/src/lib.rs
@@ -21,6 +21,7 @@ pub use generator::{
     GeneratedOpt,
 };
 pub use parser::Parser;
+pub use preprocessor::self_inlining::qualify_self_types;
 pub use syntax::{parse_qt_file, CxxQtFile, CxxQtItem};
 pub use writer::{cpp::write_cpp, rust::write_rust};
 
@@ -175,7 +176,8 @@ mod tests {
         expected_cpp_header: &str,
         expected_cpp_source: &str,
     ) {
-        let parser = Parser::from(syn::parse_str(input).unwrap()).unwrap();
+        let mut parser = Parser::from(syn::parse_str(input).unwrap()).unwrap();
+        qualify_self_types(&mut parser).unwrap();
 
         let mut cfg_evaluator = CfgEvaluatorTest::default();
         cfg_evaluator.cfgs.insert("crate", Some("cxx-qt-gen"));
--- a/crates/cxx-qt-gen/src/lib.rs
+++ b/crates/cxx-qt-gen/src/lib.rs
@@ -11,6 +11,7 @@
 mod generator;
 mod naming;
 mod parser;
+mod preprocessor;
 mod syntax;
 mod writer;
 
--- a/crates/cxx-qt-gen/src/parser/method.rs
+++ b/crates/cxx-qt-gen/src/parser/method.rs
@@ -9,7 +9,9 @@ use crate::{
     syntax::{foreignmod, types},
 };
 use core::ops::Deref;
+use quote::format_ident;
 use std::collections::{BTreeMap, HashSet};
+use std::ops::DerefMut;
 use syn::{Attribute, ForeignItemFn, Ident, Result};
 
 /// Describes a C++ specifier for the Q_INVOKABLE
@@ -146,6 +148,12 @@ impl Deref for ParsedMethod {
     }
 }
 
+impl DerefMut for ParsedMethod {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.method_fields
+    }
+}
+
 /// Struct with common fields between Invokable types.
 /// These types are ParsedSignal, ParsedMethod and ParsedInheritedMethod
 #[derive(Clone)]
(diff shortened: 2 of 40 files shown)


BenFordTytherington:
Currently only supported in RustQt blocks, C++Qt blocks to come

BenFordTytherington:
Tried to implement in the qml_features example and there is a cxx error about there being multiple types in the block, even though there is only one qobject in the example, which may be because it adds the QObject import?

crates/cxx-qt-gen/src/preprocessor/self_inlining.rs
```diff
@@ -0,0 +1,172 @@
...
+}
+
+/// For a given parser, attempt to inline the `Self` type used in any of the blocks with that blocks unique QObject
+pub fn qualify_self_types(parser: &mut Parser) -> Result<()> {
+    let data = &mut parser.cxx_qt_data;
+    for block in 0..data.qobjects.len() {
+        let methods = &mut data.methods[block];
+        let signals = &mut data.signals[block];
+        let inherited = &mut data.inherited_methods[block];
+        let qobjects = &data.qobjects[block];
```
ahayzen-kdab, line 53: almost seems like you could use `zip` here but seems it'd need to be a nested as it only supports combining two iterators? :-/

it'd probably end up as like
```rust
for (((qobject, method), signal), inherit_method) in zip(zip(zip(data.qobjects, data.methods), data.signals), data.inherited_methods)
```

But then does that end up being less readable :rofl:

crates/cxx-qt-gen/src/preprocessor/self_inlining.rs
```diff
@@ -0,0 +1,172 @@
...
+
+        let inline_self = qobjects.len() == 1;
```
ahayzen-kdab, line 55: do we only ever need to care about QObjects? eg do we support non-qobject attributed types in extern C++Qt / RustQt blocks ?

and do we need to support normal `type` items in C++/Rust blocks if we start avoiding passthrough, this would be a separate change though?

ahayzen-kdab: Right it's fine for now, but might need to change later depending what we do

crates/cxx-qt-gen/src/preprocessor/self_inlining.rs
BenFordTytherington: To be fair, the zip solution isn't too bad

crates/cxx-qt-gen/src/preprocessor/self_inlining.rs
```diff
@@ -0,0 +1,191 @@
...
+        .collect()
+}
+
+/// For a given parser, attempt to inline the `Self` type used in any of the blocks with that blocks unique QObject
+pub fn qualify_self_types(parser: &mut Parser) -> Result<()> {
+    for (qobjects, methods, inherited, signals) in separate_blocks(&mut parser.cxx_qt_data) {
+        let inline_self = qobjects.len() == 1;
+        let inline_ident = qobjects
+            .last()
+            .map(|obj| format_ident!("{}", obj.declaration.ident_left));
```
LeonMatthesKDAB, line 77: The `inline_self` feels like it's carrying redundant information.
The `Option` should either be `Some`, in which case we can inline, or `None` if we cannot inline.

How about getting rid of `inline_self` and instead doing:
```rust
let mut qobjects = qobjects.iter();
let mut inline_ident = qobjects.next().map(|obj| obj.declaration.ident_left.clone());
if qobjects.next().is_some() {
    // More than one ident found
    inline_ident = None;
}
```

crates/cxx-qt-gen/src/preprocessor/self_inlining.rs
```diff
@@ -0,0 +1,191 @@
...
+) -> syn::Result<()> {
+    for method in invokables.iter_mut() {
+        if method.self_unresolved {
+            if inline {
+                if let Some(inline_type) = type_to_inline.clone() {
+                    method.qobject_ident = inline_type;
+                } else {
+                    return Err(Error::new(
+                        method.method.span(),
+                        "Expected a type to inline, no `qobject` typename was passed!",
```
LeonMatthesKDAB, line 36: This error appears if there is no QObject type within a Rust block that uses `Self`.

However, in this context it is confusing to talk about "types to inline" and "`qobject` typename".
Please use an error message that is understandable by the end user in their context.

crates/cxx-qt-gen/src/lib.rs
```diff
@@ -20,6 +21,7 @@ pub use generator::{
     GeneratedOpt,
 };
 pub use parser::Parser;
+pub use preprocessor::self_inlining::qualify_self_types;
```
LeonMatthesKDAB, line 24: I'm not too happy with the name `preprocessor` here, as it's reminiscent of the C preprocessor, which does something very different and is invoked before the parser, not after.

Why not just move the `qualify_self_types` function into the `naming` module and then just call `naming::qualify_self_types`?

Side-note: In Rust it's best-practice to only `use` the module that a function is in and then call the function with `module_name::function_name`, so please stick to that convention. (I know we haven't done this ourselves consistently before :sweat_smile: :see_no_evil: )

LeonMatthesKDAB (changes requested):
Hi @BenFordTytherington, I'm very happy to see this work.

There are however some changes I want to see before we merge this. Especially I think it would be better to use an Array of Structures instead of a Structure of Arrays.

Other than that it's just some minor nitpicks left.

crates/cxx-qt-gen/src/preprocessor/self_inlining.rs
BenFordTytherington: Sorted

crates/cxx-qt-gen/src/parser/method.rs
```diff
@@ -156,6 +164,7 @@ pub struct MethodFields {
     pub parameters: Vec<ParsedFunctionParameter>,
     pub safe: bool,
     pub name: Name,
+    pub self_unresolved: bool,
```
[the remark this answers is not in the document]
BenFordTytherington: Done

examples/qml_basics/src/main.rs
```diff
@@ -31,7 +31,7 @@ mod qobject {
         type Greeter = super::GreeterRust;
 
         #[qinvokable]
-        fn greet(self: &Greeter) -> QString;
+        fn greet(&self) -> QString;
```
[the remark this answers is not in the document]
BenFordTytherington: Not sure how we would go about permissions best there, but I've changed it back

LeonMatthesKDAB (approved):
Beautiful :fire:

PR #1252, opened 2025-04-08
Add Transitive casting and testing code

BenFordTytherington:
Adds macros for implementing transitive casting, for 3 types, or a longer chain.
For 3 types A, B and C, `impl_transitive_cast!` will implement casting between A -> C, provided the relations between exist. For a longer chain, use `chain_cast!`.

These are scoped to cxx_qt due to `#[macro_export]`, but Upcast has now also become unsafe and moved into a casting module, use it like so: `use cxx_qt::casting::Upcast`.


BenFordTytherington:
Macro now written in example for testing purposes, I wasn't sure where to put it / export it publically. This will implement casting to all the types written after the first one, provided there is casting between each type, e.g chain_cast(A, B, C, D, E) will, provided A -> B, B -> C, C -> D and D -> can cast, generate A -> C, A -> D, A -> E.

In a future commit, could maybe add more recursion so that this also generates B -> D, B -> E, etc... Also probably needs some testing.

LeonMatthesKDAB (approved):

PR #1253, opened 2025-04-10
Use shorter paths in the OUT_DIR

### LINKED ISSUE #1237: Long path names break Windows build (originally: MSVC Generator can break CMake builds)

LeonMatthesKDAB:
As described in #1237, on Windows we're sometimes hitting the 260 character limit
for filepaths.

This is mostly due to a lot of nesting by Qt Creator, Corrosion and
Cargo who all include the target triple and other configuration information in the
build path.

However, we can at least reduce the character count a little bit on our
part.

For example, for qqmlengine.cxxqt.h, the path length is now reduced by
~40 characters from:

out/cxx-qt-build/target/crates/cxx-qt-lib/include/cxx-qt-lib-internals/src/qml/qqmlengine.cxxqt.h

to now:

out/cxxqtbuild/include/private/src/qml/qqmlengine.cxxqt.h

Closes #1237

### DIFF
--- a/crates/cxx-qt-build/src/dir.rs
+++ b/crates/cxx-qt-build/src/dir.rs
@@ -21,6 +21,11 @@ pub(crate) const INCLUDE_VERB: &str = "create symlink";
 #[cfg(not(unix))]
 pub(crate) const INCLUDE_VERB: &str = "deep copy files";
 
+pub(crate) fn gen() -> PathBuf {
+    // Use a short name due to the Windows file path limit!
+    out().join("cxxqtgen")
+}
+
 // Clean a directory by removing it and recreating it.
 pub(crate) fn clean(path: impl AsRef<Path>) -> Result<()> {
     let result = std::fs::remove_dir_all(&path);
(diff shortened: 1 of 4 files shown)


crates/cxx-qt-build/src/dir.rs
```diff
@@ -37,14 +42,26 @@ pub(crate) fn clean(path: impl AsRef<Path>) -> Result<()> {
...
 pub(crate) fn crate_target() -> PathBuf {
-    target().join("crates").join(crate_name())
+    let path = target();
+    if is_exporting_crate() {
+        path.join("crates").join(crate_name())
+    } else {
+        // If we're not exporting, use a shortened path
+        // The paths for files in the OUT_DIR can get pretty long, especially if combined with
+        // Corrosion/CMake.
+        // This is an issue, as Windows has a maximum path length of 260 characters.
```
ahayzen-kdab, line 52: can we somewhere detect if we think the path you are building in is too long and warn on Windows?

ahayzen-kdab (approved):
LGTM :-)

crates/cxx-qt-build/src/dir.rs
ahayzen-kdab: Lets do this later :-)

PR #1254, opened 2025-04-10
Don't set MultiThreadedDLL for MSVC Debug builds

### LINKED ISSUE #1234: Windows: Qt Debug dlls conflict with CXX-Qt
On Windows, due to [Rust always linking to the non-Debug runtime](https://corrosion-rs.github.io/corrosion/common_issues.html#linking-debug-cc-libraries-into-rust-fails-on-windows-msvc-targets), Corrosion users need to do the same with all linked in C++ dependencies.
This includes Qt!

On Windows, Qt ships debug dlls that are built in debug mode and link to the debug runtime.
This is not compatible with CXX-Qt, and we must ensure that our users always link to the right Qt binaries (e.g. the non-Debug ones, even in a Debug build).

We may be able to fix this in cxx-qt-cmake for our users, otherwise we should emit a warning/error, like we do if you have set the incorrect  MSVC runtime.

Otherwise, we may run into UB, as the debug and non-debug runtimes on Windows are not ABI-compatible!

Encountered by @narnaud in yesterdays training.

LeonMatthesKDAB:
Once https://github.com/KDAB/cxx-qt-cmake/pull/12 is merged, this should
no longer be necessary.

Closes #1234

### DIFF
--- a/.github/workflows/github-cxx-qt-tests.yml
+++ b/.github/workflows/github-cxx-qt-tests.yml
@@ -285,6 +285,7 @@ jobs:
             clang_format_path: /home/runner/.local/bin/clang-format
             cargo_dir: ~/.cargo
             rustc_wrapper: sccache
+            build_type: Release
           - name: Ubuntu 24.04 (gcc) Qt6
             os: ubuntu-24.04
             qt_version: 6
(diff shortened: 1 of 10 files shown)


LeonMatthesKDAB:
Abandoned in favor of #1256

PR #1255, opened 2025-04-12
Use QAnyStringView for QString::arg

jnbooth:
As of [Qt 6.9](https://doc.qt.io/qt-6/whatsnew69.html), [`QString::arg`](https://doc.qt.io/qt-6/qstring.html#arg) now accepts any parameter that can convert to `QAnyStringView`. That means we can skip unnecessary allocations:

```rs
// old (still works, but allocates a QString)
mystring.arg(&QString("myarg"));

// new (no allocation, Qt processes UTF8)
mystring.arg("myarg");

```


jnbooth:
Actually, it would probably be better to wait until support for Qt 5 is dropped, so the cfg pragmas aren't necessary.

PR #1257, opened 2025-04-15
QUuid compatibility with Qt6.9

LeonMatthesKDAB:


ahayzen-kdab:
Bumped to 6.9 in this change #1258

PR #1260, opened 2025-04-29
Remove cxx-qt dependency of cxx-qt-macro

LeonMatthesKDAB:
cxx-qt-macro needed cxx-qt as a dev-dependency for the documentation.
This didn't allow me to release the crate, as cxx-qt-macro depended on
cxx-qt 0.7.2, which wasn't released yet, but cxx-qt itself also depended
on cxx-qt-macro 0.7.2 which also wasn't released yet.

We can solve this conundrum by moving the documentation of cxx-qt-macro
into cxx-qt itself.
The documentation of cxx-qt-macro should not be used anyway, as it is an
implementation detail.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -442,8 +442,6 @@ dependencies = [
 name = "cxx-qt-macro"
 version = "0.7.1"
 dependencies = [
- "cxx",
- "cxx-qt",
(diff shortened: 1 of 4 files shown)


ahayzen-kdab (approved):
LGTM :-)

PR #1263, opened 2025-05-04
Add a lifetime parameter to QMetaObjectConnectionGuard

jnbooth:
Instead of requiring `'static` closures, connection signal handlers now accept lifetime parameters:

```rs
pub fn on_mysignal<'a, F>(self: Pin<&mut Self>, closure: F) -> QMetaObjectConnectionGuard<'a>
where
    F: FnMut(Pin<&mut Self>) + 'a + Send,
```

This makes it possible to attach callbacks to signals that use scoped variables. Only `QMetaObjectConnectionGuard<'static>` can use `.release()`, so the Drop implementation will ensure all other closures are disconnected once their lifetimes have expired.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,10 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Casting is automatically implemented for qobjects or types which have `#[base = T]` in `"RustQt"` or `"C++Qt"` blocks
 - Support for `QMessageLogContext` and sending log messages to the Qt message handler.
 - Serde support for further types: `QByteArray`, `QSet`, `QStringList`, `QVector`, `QUrl`
-
-### Changed
-
-- Add lifetime parameter to `QMetaObjectConnectionGuard`.
+- Add `QScopedMetaObjectConnectionGuard`, which is `QMetaObjectConnectionGuard` with a scoped lifetime. `QMetaObjectConnectionGuard` is now a type alias for `QScopedMetaObjectConnectionGuard<'static>`.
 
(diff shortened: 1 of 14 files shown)


LeonMatthesKDAB:
Hi @jnbooth, that sounds like a nice restriction to lift.

IMO, the `'static` lifetime will still be used in most cases, so ideally we should keep this case simple and ideally backwards compatible.

Can you try turning `QMetaObjetConnectionGuard` into a type alias, e.g.:
```rust
type QMetaObjetConnectionGuard = QScopedMetaObjectConnectionGuard<'static>;
```

And then rename the old QMetaObjectConnectionGuard to QScopedMetaObjectConnectionGuard (or another name if you have a better idea).

That should allow us to keep backwards compatibility.

LeonMatthesKDAB:
Hi, this mostly looks good, but @ahayzen-kdab and I just discussed whether there's any safety concerns we need to think about here.

We found that this can probably introduce UB if the QScopedMetaObjectConnectionGuard is `Send`.
In theory, you could send the guard, then drop it, which could cause a race condition between the thread running the disconnection and any slot invocations inside the thread that created the connection.

One way to solve this would be to make `QMetaObjectConnection` not Send.
The easiest way to do this is to change it to:

```
#[repr(C)]
pub struct QMetaObjectConnection {
    _space: MaybeUninit<*const std::ffi::c_void>,
}
```

As any pointer member type will make it no longer Send and Sync.

jnbooth:
I don't follow how that would result in UB, but I will take your word for it!

LeonMatthesKDAB (approved):

PR #1264, opened 2025-05-04
Add QEventLoop to cxx-qt-lib-extras

jnbooth:
I've been finding [`QEventLoop`](https://doc.qt.io/qt-6/qeventloop.html) useful in my own projects, especially for unit testing, so I figured I'd make a PR to add it to cxx-qt-lib-extras if other people would find it useful as well. The neat thing about `QEventLoop` is that it can be used for nested local event processing rather than needing to rely on `QApplication::exec` and all the baggage it brings. For example, you can run the event loop in Rust without needing to delegate to C++ at all, which is very useful for unit tests:

```rs
fn wait_for_encrypted(socket: &QSslSocket, timeout: Duration) -> bool {
    let now = Instant::now();
    let mut event_loop = QEventLoop::new();
    loop {
        event_loop.pin_mut().process_all_events();
        if socket.is_encrypted() {
            return true;
        }
        if now.elapsed() > timeout {
            return false;
        }
    }
}
```

Furthermore, `QEventLoop::exec_with` can be used to run a closure inside Qt's event processing so that signals will fire. Without those helper functions, there is no way to make signals fire from Rust's side, which means it's impossible to use them in tests.

In general, I think the ability to process events from Rust rather than C++ could have safety and performance benefits down the line. Maybe even async compatibility? At the very least, it should make it easier to use blocking sync like mutexes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Casting is automatically implemented for qobjects or types which have `#[base = T]` in `"RustQt"` or `"C++Qt"` blocks
 - Support for `QMessageLogContext` and sending log messages to the Qt message handler.
 - Serde support for further types: `QByteArray`, `QSet`, `QStringList`, `QVector`, `QUrl`
+- Added `QEventLoop` to cxx-qt-lib-extras.
 
 ### Removed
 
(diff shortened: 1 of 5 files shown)


LeonMatthesKDAB (approved):
@jnbooth thank you for the contribution.

As is policy for cxx-qt-lib-extras, every added binding is helpful and welcome :rocket:

PR #1265, opened 2025-05-07
Remove unnecessary constraints from QFlags derived implementations

jnbooth:
A `QFlags` is just a wrapper around an integer, so it should always implement `Eq`, `Ord`, etc. However, derived implementations apply their constraints to all type arguments, which means that currently, `QFlags<T>` only implements `Ord` if `T` implements `Ord`, and so on. There's no reason to have those bounds, so I removed them.


LeonMatthesKDAB (approved):
Thank you for the contribution @jnbooth :rocket:

PR #1267, opened 2025-05-11
Use i64 instead of qint64 for public APIs

jnbooth:
Public functions that previously accepted a `qint64` now accept an `i64`.

### DIFF
--- a/crates/cxx-qt-lib/src/core/qdate.rs
+++ b/crates/cxx-qt-lib/src/core/qdate.rs
@@ -26,10 +26,8 @@ mod ffi {
         include!("cxx-qt-lib/qdate.h");
         type QDate = super::QDate;
 
-        /// Returns a QDate object containing a date ndays later than the date of this object (or earlier if ndays is negative).
-        ///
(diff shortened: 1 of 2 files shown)


LeonMatthesKDAB:
@ahayzen-kdab I remember you dealt with qint before, did we want to have those as public API or not?
I think I remember you adding those types at some point :thinking:

ahayzen-kdab:
> @ahayzen-kdab I remember you dealt with qint before, did we want to have those as public API or not? I think I remember you adding those types at some point 🤔

This likely came from the problem we have of `std::int64_t` / `std::uint64_t` being subtly different to Qt's `qint64` and `quint64` from a CXX perspective so we couldn't use `u64` in the bridge as CXX complained the type was different, but they are actually the same size.

Now that we have the `qint64` type we can keep CXX happy but the Rust side isn't as nice. Guess the question is there any overhead from the u64 <-> quint64 conversion, and looks like no from the code. So likely better to go this route now that we have all the pieces in place.

ahayzen-kdab (approved):
LGTM, lets go with this as there is minimal overhead and it makes the Rust side feel more native :-)

PR #1278, opened 2025-05-20
Add  attribute which generates wrappers for a function

### LINKED ISSUE #1235: Shorthand to directly wrap functions on the inner Rust struct
Currently, if you want to expose a function on the inner Rust struct to C++/QML, you always have to write a wrapper function outside the bridge.

E.g.:
```rust
#[cxx_qt::bridge]
mod qobject {
extern "RustQt" {

#[qobject]

fn my_fun(self: &MyObject) -> i32;

}
}

impl qobject::MyObject {
    fn my_fun(&self) -> i32 {
        self.rust().my_fun();
    }
}
```

We could consider automating this if you provide a given attribute to the declaration (e.g. `#[inner]`).

Example:
```rust
#[cxx_qt::bridge]
mod qobject {
extern "RustQt" {

#[qobject]

#[inner] // auto-generates the wrapper outside the bridge
fn my_fun(self: &MyObject) -> i32;

}
}
```

Note that this may cause issues with `Result<>`, as the Error type wouldn't necessarily be known.

BenFordTytherington:
Closes #1235.

Adds the `auto_wrap` attribute which will generate the wrapper function for methods defined in the bridge.

### DIFF
--- a/crates/cxx-qt-gen/src/generator/rust/method.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/method.rs
@@ -3,7 +3,7 @@
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
-use crate::generator::rust::get_params_tokens;
+use crate::generator::rust::{get_call_params_tokens, get_params_tokens};
 use crate::{
     generator::{naming::qobject::QObjectNames, rust::fragment::GeneratedRustFragment},
     parser::method::ParsedMethod,
@@ -31,6 +31,8 @@ pub fn generate_rust_methods(
                 cpp_class_name_rust,
             );
 
+            let call_parameters = get_call_params_tokens(&invokable.parameters);
+
             let return_type = &invokable.method.sig.output;
 
             let cfgs = &invokable.cfgs;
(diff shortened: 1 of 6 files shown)


crates/cxx-qt-gen/test_outputs/passthrough_and_naming.rs
```diff
@@ -780,6 +780,9 @@ cxx_qt::static_assertions::assert_eq_size!(
     cxx_qt::signalhandler::CxxQtSignalHandler<SecondObjectCxxQtSignalClosurepropertyNameChanged>,
     [usize; 2]
 );
+fn my_function(self: &SecondObject, param: i32) {
+    self.rust().my_function(param)
+}
```
ahayzen-kdab, line 785: this generated code still needs to end up being wrapped in a impl.

So it should be

```rust
impl ffi::SecondObject {
    fn my_function(self: &SecondObject, param: i32) {
        self.rust().my_function(param)
    }
}
```

As this is effectively a shorthand so the developer doesn't need to write that, as it's just a pass through

book/src/bridge/attributes.md
```diff
@@ -43,3 +43,8 @@ For [`#[qproperty]`](./extern_rustqt.md#properties), a CXX or Rust name can be p
 The `#[auto_cxx_name]` and `#[auto_rust_name]` attributes can be used to automatically rename cxx and rust names.
 These are placed at a block level on `extern "RustQt"` or `extern "C++Qt"` blocks, and will automatically case convert the items inside, unless they specify either a `rust_name` or `cxx_name`.
 By default `#[auto_cxx_name]` will generate a camelCase conversion for`cxx_name` and `#[auto_rust_name]` will generate a snake_case conversion for `rust_name`.
+
+### Automatic wrapping
```
ahayzen-kdab, line 47: an example that is compiled would be good too. I think the load_color here could be changed to use it https://github.com/KDAB/cxx-qt/blob/main/examples/qml_features/rust/src/invokables.rs#L99

So that you have load_color auto_wrapped to read the inner as_qcolor on the Rust type (which could be renamed)

crates/cxx-qt-gen/test_outputs/passthrough_and_naming.rs
```diff
@@ -780,6 +780,11 @@ cxx_qt::static_assertions::assert_eq_size!(
     cxx_qt::signalhandler::CxxQtSignalHandler<SecondObjectCxxQtSignalClosurepropertyNameChanged>,
     [usize; 2]
 );
+impl ffi::SecondObject {
+    fn my_function(self: &SecondObject, param: i32) {
```
ahayzen-kdab, line 784: aha this type should be `ffi::SecondObject` that's why it's failing :-)

BenFordTytherington: Fixed

crates/cxx-qt-gen/src/generator/rust/method.rs
```diff
@@ -54,26 +58,57 @@ pub fn generate_rust_methods(
...
+
+            let wrapper_fn = if invokable.wrap {
+                vec![parse_quote_spanned! {
+                    invokable.method.span() =>
                     #(#cfgs)*
-                    #[doc(hidden)]
-                    #unsafe_call fn #invokable_ident_rust(#parameter_signatures) #return_type;
-                }
+                    impl #qualified_impl {
+                        #method_safety fn #invokable_ident_rust(#parameter_signatures) #return_type {
```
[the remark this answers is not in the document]
BenFordTytherington: Do we need to account for pub super? I also think if we don't put pub it might not be callable, but I can experiment

crates/cxx-qt-gen/src/generator/rust/mod.rs
```diff
@@ -146,6 +146,48 @@ pub fn get_params_tokens(
...
+        quote! { &#class_name_path }
+    };
+    if parameters.is_empty() {
+        quote! { self: #struct_sig }
+    } else {
+        let parameters = parameters
+            .iter()
+            .map(|parameter| {
+                let ident = &parameter.ident;
+                let ty = &parameter.ty;
```
ahayzen-kdab, line 167: i think you need to run `syn_type_cxx_bridge_to_qualified` over this? Otherwise if the type is `T` from the bridge it won't be converted to `ffi:T` or whatever etc

LeonMatthesKDAB (approved):

PR #1280, opened 2025-05-21
Add _default constructors and rename constructors returning UniquePtr

### LINKED ISSUE #1203: Standardize constructor method names
There are currently no standardized naming conventions for constructor methods. For example, consider these methods:

```rs
QDate::from_string_enum(string: &QString, format: DateFormat) -> Option<Self>
QDateTime::from_string(string: &QString, format: DateFormat) -> Option<Self>
QTime::from_string_enum(string: &QString, format: DateFormat) -> Self
```

`QDate::from_string_enum` and `QDateTime::from_string` have the same signature, but different names. `QDate::from_string_enum` and `QTime::from_string_enum` have the same name, but different signatures.

Proposal 1: constructor methods should be the names of each parameter type, prefixed by `from_`. If they return `Option<Self>` rather than `Self`, the name should be suffixed by `_opt`, which is a common pattern in other crates. So the above would be replaced by:

```rs
QDate::from_string_dateformat_opt(string: &Qstring, format: DateFormat) -> Option<Self>

QDateTime::from_string_dateformat_opt(string: &QString, format: DateFormat) -> Option<Self>

QTime::from_string_dateformat(string: &QString, format: DateFormat) -> Self
```

Proposal 2: Deprecate all the constructor methods and just use `From` and `TryFrom` with tuples. So the above would be replaced by

```rs
impl TryFrom<(&QString, DateFormat)> for QDate

impl TryFrom<(&QString, DateFormat)> for QDateTime

impl From<(&QString, DateFormat)> for QTime
```

Proposal 3: Do proposal 1 *and* the implementations from proposal 2.

BenFordTytherington:
Not entirely sure if I understood the discussion, but should potentially close #1203

### DIFF
--- a/crates/cxx-qt-lib/src/core/qdate.rs
+++ b/crates/cxx-qt-lib/src/core/qdate.rs
@@ -186,7 +186,12 @@ impl QDate {
         }
     }
 
-    /// Returns the time represented in the string as a QTime using the format given, or None if this is not possible.
+    /// Returns the QTime represented by the string, using the format given, or the Qt invalid date if not possible.
+    pub fn from_string_or_default(string: &ffi::QString, format: &ffi::QString) -> Self {
+        ffi::qdate_from_string(string, format)
+    }
+
+    /// Returns the QTime represented by the string, using the format given in the enum, or None if this is not possible.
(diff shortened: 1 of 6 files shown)


jnbooth:
There is also https://github.com/KDAB/cxx-qt/pull/1220

PR #1282, opened 2025-05-21
Place all objects and passthrough items in the same block, to pass th…

BenFordTytherington:
Ultimately, it would be good to preserve the blocks things appear in, so if you place a cfg or doc attribute on the top of a block, it is preserved in the final code, but this requires significant rewriting, so work is beginning by slowly grouping things into the same extern block, and passing through docs.

### DIFF
--- a/crates/cxx-qt-gen/bat
+++ b/crates/cxx-qt-gen/bat
@@ -1,0 +1,1 @@
+/home/ben/cxx-qt/crates/cxx-qt-gen/src/generator/naming/property.rs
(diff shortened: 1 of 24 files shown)


LeonMatthesKDAB:
@BenFordTytherington what is the use-case for this, especially the `#[doc()]` attribute on extern blocks, what does this translate to in the final `cargo doc`?

An extern block is not a module, so does the documentation even end up anywhere readable?

BenFordTytherington:
This seemed like too much of a task, to reorganise code placement in generated output, just for the sake of it, as I had hoped some minor changes could make things easier to place in the same output. I had hoped this would allow for putting cfg attributes on the block, and passing them through to generation more easily, but I think this is not the way, so will close.

PR #1284, opened 2025-05-22
qt-build-utils can invoke qmake when it is a script

SanderVocke:
Ensure qmake is called either through `cmd` (Windows) or `sh` (Linux/MacOS).

The main change is this invocation. The rest of the code change is:

- There were two places qmake were previously invoked and they didn't share code. To prevent duplicating this shell wrapper, refactored a bit to ensure they both call the same method to query qmake.
- Due to said refactoring, also had to change the error handling a bit because the error generation moved. Added error types.

### DIFF
--- a/crates/qt-build-utils/src/utils.rs
+++ b/crates/qt-build-utils/src/utils.rs
@@ -54,3 +54,19 @@ pub(crate) fn is_windows_target() -> bool {
 pub(crate) fn is_emscripten_target() -> bool {
     std::env::var("CARGO_CFG_TARGET_OS") == Ok("emscripten".to_owned())
 }
+
+/// Create a command to run qmake with a query.
+///
+/// On Windows, this wraps the invocation in `cmd /C` so that qmake can be a script.
+pub(crate) fn qmake_query_command(qmake_path: &Path, var_name: &str) -> Command {
+    let mut cmd;
+    if cfg!(target_os = "windows") {
+        cmd = Command::new("cmd");
+        let qmake_command = format!("\"{}\" -query {}", qmake_path.display(), var_name);
+        cmd.args(["/C", &qmake_command]);
+    } else {
+        cmd = Command::new(qmake_path);
+        cmd.args(["-query", var_name]);
+    }
+    cmd
+}
--- a/crates/qt-build-utils/src/installation/qmake.rs
+++ b/crates/qt-build-utils/src/installation/qmake.rs
@@ -116,8 +116,8 @@ impl TryFrom<PathBuf> for QtInstallationQMake {
 
     fn try_from(qmake_path: PathBuf) -> anyhow::Result<Self> {
         // Attempt to read the QT_VERSION from qmake
-        let qmake_version = match Command::new(&qmake_path)
-            .args(["-query", "QT_VERSION"])
+        let mut cmd = crate::utils::qmake_query_command(&qmake_path, "QT_VERSION");
+        let qmake_version = match cmd
             // Binaries should work without environment and this prevents
             // LD_LIBRARY_PATH from causing different Qt version clashes
             .env_clear()
@@ -211,9 +211,9 @@ impl QtInstallation for QtInstallationQMake {
 
 impl QtInstallationQMake {
     fn qmake_query(&self, var_name: &str) -> String {
+        let mut cmd = crate::utils::qmake_query_command(&self.qmake_path, var_name);
         String::from_utf8_lossy(
-            &Command::new(&self.qmake_path)
-                .args(["-query", var_name])
+            &cmd
                 // Binaries should work without environment and this prevents
                 // LD_LIBRARY_PATH from causing different Qt version clashes
                 .env_clear()


SanderVocke:
I only tested this so far on the "happy case" on Windows (that is to say: when using a qmake .bat wrapper). Need to test on other platforms too.

SanderVocke:
I marked as ready for review.
Tested all combinations of `QMAKE=qmake6` and `QMAKE=(a qmake wrapper).[sh|bat]` on Windows, Ubuntu and MacOS GitHub runners and cxx-qt builds fine with these combinations.

Please advise how to solve the rust formatting error in CI: a simple `cargo fmt` in the project did not succeed:

```
C:\sdev\cxx-qt> cargo fmt
error: expected item, found `..`
 --> \\?\C:\sdev\cxx-qt\examples\cargo_without_cmake\src\cxxqt_object.rs:1:1
  |
1 | ../../qml_minimal/rust/src/cxxqt_object.rs
  | ^^ expected item
  |
  = note: for a full list of items that can appear in modules, see <https://doc.rust-lang.org/reference/items.html>

Error writing files: failed to resolve mod `cxxqt_object`: cannot parse \\?\C:\sdev\cxx-qt\examples\cargo_without_cmake\src\cxxqt_object.rs
```

crates/qt-build-utils/src/lib.rs
```diff
@@ -390,18 +402,47 @@ impl QtBuild {
...
+                Ok(output_string)
+            }
+            Err(e) if e.kind() == std::io::ErrorKind::NotFound => Err(QtBuildError::QtMissing),
+            Err(e) => Err(QtBuildError::RunQmakeFailed(e)),
+        }
+    }
+
+    /// Query the qmake executable which was previously
+    /// verified. Panic on error.
     pub fn qmake_query(&self, var_name: &str) -> String {
```
ahayzen-kdab, line 444: note as part of refactorings going on to the build system, this code has moved to https://github.com/KDAB/cxx-qt/blob/96d323343fc8b59617e8902aa6f61360634d5a62/crates/qt-build-utils/src/installation/qmake.rs#L105

The problem itself appears to be that the path to qmake might not be a binary executable but a script? Wonder if we should have a general Command wrapper to always call things via sh or not :thinking: 

Or whether instead the user should have to provide a binary, doesn't seem to tricky to just write a bash/batch script and execute on that?

SanderVocke: Where this led to failure was with a .bat script on Windows. I don't think there is a way to make a .bat "more executable".  Probably, executable scripts on Linux already work, although I didn't test that.

Thanks for the heads-up, I will adapt to the refactor that was done. My PR so far solved it locally. Would you prefer such a generic Command wrapper, or do you think the current approach in the PR is OK?

I ran into some build failures on your CI I still need to investigate. Locally the PR works on Linux and Windows, with .sh and .bat qmake wrapper scripts.

SanderVocke:
@ahayzen-kdab is this still on your radar?

crates/qt-build-utils/src/utils.rs
```diff
@@ -14,3 +16,16 @@ pub(crate) fn is_apple_target() -> bool {
 pub(crate) fn is_emscripten_target() -> bool {
     std::env::var("CARGO_CFG_TARGET_OS") == Ok("emscripten".to_owned())
 }
+
+/// Wrap a command in a native subshell
+pub(crate) fn native_shell_command(command: &str) -> Command {
+    let mut result: Command;
+    if cfg!(target_os = "windows") {
```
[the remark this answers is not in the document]
LeonMatthesKDAB: @ahayzen-kdab I actually prefer sticking with `cfg!` here.
AFAIK, the `cfg!` macro evaluates to a bool literal.
This should allow the compiler to optimize out the branches that are not used by the current cfg, but will still check that the alternative branches compile.

With `#[cfg()]` if you compile on Linux you will not notice if the windows part of the branch has an error that could be caught at compile-time.

crates/qt-build-utils/src/utils.rs
```diff
@@ -14,3 +16,16 @@ pub(crate) fn is_apple_target() -> bool {
 pub(crate) fn is_emscripten_target() -> bool {
     std::env::var("CARGO_CFG_TARGET_OS") == Ok("emscripten".to_owned())
 }
+
+/// Wrap a command in a native subshell
+pub(crate) fn native_shell_command(command: &str) -> Command {
```
LeonMatthesKDAB, line 21: I believe passing the command as a plain string here will fail if the path to qmake contains a space.

As this path is provided by the user, we must ensure this works as well.
Please make sure that we can handle paths with spaces.

I suggest adding a testcase for this by adding a `test scripts/` folder that contains a `qmake.sh` and `qmake.bat` script that just echo a string or forward to `qmake`.

crates/qt-build-utils/src/utils.rs
```diff
@@ -14,3 +16,16 @@ pub(crate) fn is_apple_target() -> bool {
...
+    let mut result: Command;
+    if cfg!(target_os = "windows") {
+        result = Command::new("cmd");
+        result.args(["/C", command]);
+    } else {
```
LeonMatthesKDAB, line 26: Please use an `else if` that actively lists the operating systems that we expect to have an `sh` binary and fall back to a plain `Command` in the other cases, as otherwise this will break on any OS that does not have an `sh` binary.

LeonMatthesKDAB (changes requested):
Hi, @SanderVocke, overall I'm in favor of adding support for this, especially for the Windows case, where you can't just make a script an executable.

I'm somewhat concerned about possible regressions to invoking plain qmake binaries (especially on Unix where it is already possible to just make any script an executable).

We definitely need to be careful around handling of spaces, as I think that is now broken.

Maybe a good way to handle this could be:
Only wrap the executable in a native shell command on Windows, and only if it matches the `.bat` extension.
Checking the file extension would also allow us to add support for e.g. powershell instead of `cmd` via checking for the `.ps1` extension.

Do you think this would fit the use case on Windows? As I think on Windows we can rely on these extensions to be there, which is not necessarily the case on Linux.

SanderVocke:
Hi, thanks for the review! I agree, on second thought this change is really only needed on windows as executable scripts make this a non-issue on other platforms. I will update the pr.

PR #1287, opened 2025-05-23
cxx-qt-build: return an Interface from CxxQtBuilder

LeonMatthesKDAB:
Another take on #1285.

The main difference is in how include paths are handled for the CXX-style automatic include of the header directory.

In this PR, the root of the automatic inclue can be configured.
This leads to less changes for downstream crates.
It also allows us to move to an include system of "cxx-qt-lib/core/qstring.h" while keeping backwards-compatibility with forwarding-headers

- **cxx-qt-build: split interface into a separate file**
- **cxx-qt-build: move write_manifest to export of Interface**
- **cxx-qt-build: move write of exported include directories to interface**
- **cxx-qt-build: store manifest and dependencies in Interface for now**
- **cxx-qt-build: return an Interface rather than taking as input**
- **cxx-qt-build: check the links key when exporting**
- **cxx-qt-build: remove CxxQtBuilder::library instead use new and export**
- **cxx-qt-build: Add CXX-style automatic inclusion of header files**
- **cxx-qt-build: Remove Interface::export_include_dir**


ahayzen-kdab (approved):
As discussed think this is a good step in the right direction, didn't know if you want to use this in it's current state or keep working on it. Going to approve so you can land this if required and iterate separately :-)

PR #1295, opened 2025-06-18
Fix: Constructors no longer fail silently

### LINKED ISSUE #1233: Missing `impl Constructor<()> for ... {}` in the bridge silently fails
If you miss adding the `impl Constructor for ...` shim trait impl in the bridge, but implement the trait outside the bridge, you don't get a warning, CXX-Qt will just skip using your constructor.

This is especially bad in combination with `cxx_qt::Initialize`, as that requires you to declare a default constructor, which you also get if you don't add the declaration, but the one generated by CXX-Qt won't actually call your `initialize` function.

## Idea to remedy this

We could introduce a new trait: `ConstructorDeclaration<()>` (name TBD), where `Constructor` requires `ConstructorDeclaration`.
And ConstructorDeclaration would only be implemented by the bridge if you add `impl Constructor<()> for ... {}`.

That way you would get a compiler error if you implement the Constructor or Initialize trait without declaring it in the bridge.

BenFordTytherington:
- Closes #1233 
- A shim trait is used to check whether the user declared a constructor
- This is implemented by the bridge if you add the constructor trait in bridge
- This will then allow you to implemented constructor for your type
- This prevents constructors not being used silently

### DIFF
--- a/book/src/bridge/extern_rustqt.md
+++ b/book/src/bridge/extern_rustqt.md
@@ -111,7 +111,7 @@ is equivalent to writing
 impl cxx_qt::Constructor<()> for x {}
 ```
 
-inside the bridge.
+inside the bridge. You can then implement your constructors outside the bridge, using either of these traits.
 
 For further documentation see the [traits page](./traits.md).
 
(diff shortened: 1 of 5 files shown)


LeonMatthesKDAB (dismissed):
This works well, but needs some better user-facing documentation.

Please also update the Changelog to mention this new feature :)

ahayzen-kdab (dismissed):
Looks good, lets land this any further improvements can be done separately

ahayzen-kdab (approved):
LGTM :-)

PR #1297, opened 2025-06-24
Replace handwritten documentation with Qt documentation

jnbooth:
Currently, cxx-qt-lib has many errors in its documentation. Most of them are minor. Rather than writing documentation by hand, I think it is probably simpler and lower-maintenance to simply copy in the documentation provided by Qt.


LeonMatthesKDAB (approved):
Hi @jnbooth , thank you for this PR, this is a great improvement to the documentation.

Keeping up-to-date with the official Qt documentation could end up being quite a time-consuming task.
Did you manage to automate this process or did you just manually migrate the documentation?

jnbooth:
@LeonMatthesKDAB No, I just copied them. Since cxx-qt-lib supports a wide range of Qt versions, I think we probably don't need to worry about keeping its documentation "up to date".

PR #1299, opened 2025-06-25
README: Add sccache instructions

the-ssd:


LeonMatthesKDAB:
Hi @the-ssd , thank you for the contribution.
sccache is indeed highly recommended, we use it extensively in CI and the CXX-Qt repository itself, so I have no issue in adding it to the documentation.

However, I think there are better ways of using sccache than to modify the global cc binary.
For cargo builds, you can simply add this section to your `~/.cargo/config.toml`
```toml
[build]
rustc-wrapper="sccache"
```

And for CMake, you can set the variables `CMAKE_C_COMPILER_LAUNCHER` and `CMAKE_CXX_COMPILER_LAUNCHER` to `sccache`.
The CXX-Qt repository actually contains a cmake script under `cmake/CompilerCaching.cmake` that will automatically detect if ccache or sccache is available and set it up accordingly.

Can you please try these steps and confirm that it still works for you as expected?

the-ssd:
> Hi @the-ssd , thank you for the contribution. sccache is indeed highly recommended, we use it extensively in CI and the CXX-Qt repository itself, so I have no issue in adding it to the documentation.
> 
> However, I think there are better ways of using sccache than to modify the global cc binary. For cargo builds, you can simply add this section to your `~/.cargo/config.toml`
> 
> ```toml
> [build]
> rustc-wrapper="sccache"
> ```
> 
> And for CMake, you can set the variables `CMAKE_C_COMPILER_LAUNCHER` and `CMAKE_CXX_COMPILER_LAUNCHER` to `sccache`. The CXX-Qt repository actually contains a cmake script under `cmake/CompilerCaching.cmake` that will automatically detect if ccache or sccache is available and set it up accordingly.
> 
> Can you please try these steps and confirm that it still works for you as expected?

It probably works, but it doesn't cover cargo_without_cmake example. Maybe CxxQtBuilder can be modified to use sccache for c++ also? But that seems like more of a problem with cc, since I didn't see a way to use a wrapper for c. I myself will be switching to using cmake anyway, so I can use KDE's translation infrastructure.

LeonMatthesKDAB:
Setting `rustc-wrapper="sccache"` in the `~/.cargo/config.toml` should enable sccache automatically for all C++ compilation done by CxxQtBuilder, so that should cover `cargo_without_cmake`.

We internally use the `cc` crate, which will automatically pick this up, see: https://github.com/rust-lang/cc-rs/blob/6f0dd80fa36c9533b27bfa0f513dccd65a650836/src/lib.rs#L3036

PR #1300, opened 2025-06-25
Use Qt's public logging APIs rather than undocumented internals

jnbooth:
Instead of exposing undocumented internal features of Qt, qtlogging.rs now exports `q_critical`, `q_debug`, `q_fatal`, `q_info`, and `q_warning`, equivalent to Qt's [qCritical, qDebug, qFatal, qInfo, and qWarning](https://doc.qt.io/qt-6/qtlogging.html#qCritical). So instead of doing this:

```rs
qt_message_output(
  QtMsgType::QtInfoMsg,
  &QMessageLogContext::new(c"myfile.rs", 1, c"", c"default"),
  &QString::from(&format!("Test message: {x} / {y}"))
);
```

You can simply do this:
```rs
q_info!("Test message: {x} / {y}");
```

The file and line number are automatically filled in by the [`file`](https://doc.rust-lang.org/std/macro.file.html) and [`line`](https://doc.rust-lang.org/std/macro.line.html) built-in macros.


LeonMatthesKDAB:
Hi @jnbooth,
overall I'm in favor of adding the appropriate logging functions.
However, I think the internal API that was already exposed should not be removed.

The internal API is suited to a different use-case and allows more fine-grained access.
One good use-case for it would be to implement a custom `log` backend with the Qt logging infrastructure.
This would be quite difficult with just `q_info!`, etc.

So please add the already exposed API back.

PR #1302, opened 2025-07-03
Require the argument passed to QObjectExt::set_parent to be mutable

### LINKED ISSUE #1294: Is it really safe for QObjectExt::set_parent to accept an immutable reference?
Currently, QObjectExt::set_parent accepts a `parent: &QObject` argument, which it then converts to a mutable pointer: https://github.com/KDAB/cxx-qt/blob/511ec27922c53b19cddd89c2dc035925d7bca7e1/crates/cxx-qt-lib/src/core/qobject.rs#L94-L99

Is this really safe? QObject::setParent() requires a mutable pointer on Qt's side, and its implementation mutates the parent by appending the object to its list of children. It also triggers the ChildAdded event on the parent, which may perform arbitrary code in its event handler.

I would have expected something more like this:

```rs
fn set_parent<T>(self: Pin<&mut Self>, parent: Option<Pin<&mut T>>)
where
    T: Upcast<QObject>,
{
    unsafe {
        let parent = match parent {
            Some(parent) => ptr::from_mut(parent.upcast_pin().get_unchecked_mut()).cast(),
            None => ptr::null_mut(),
        };
        cast_pin(self).set_parent(parent);
    }
}
```

jnbooth:
Closes #1294.

### DIFF
--- a/crates/cxx-qt-lib/src/core/qobject.rs
+++ b/crates/cxx-qt-lib/src/core/qobject.rs
@@ -3,8 +3,10 @@
 //
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
+use cxx_qt::casting::Upcast;
 pub use cxx_qt::QObject;
 use std::pin::Pin;
+use std::ptr;
 
 #[cxx::bridge]
 pub mod ffi {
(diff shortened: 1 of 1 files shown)


LeonMatthesKDAB (approved):

LeonMatthesKDAB:
Thank you for all the contributions @jnbooth .

jnbooth:
@LeonMatthesKDAB Is there anything I need to do for this to be merged?

PR #1315, opened 2025-07-21
Improve CI log readability by removing sccache debug statements

jnbooth:
This gets rid of all those messages like

```
warning: qml_features@0.1.0: [2025-07-19T16:17:37Z DEBUG sccache::config] Attempting to read config file at "/home/runner/.config/sccache/config"
```

that make it incredibly difficult to spot why tests fail.


LeonMatthesKDAB (approved):
Approved, We can always add it back when needed.

Issue #578, opened 2023-06-07
Should signals only be mutable ?


ahayzen-kdab:
Determine if signals should always require a `Pin<&mut Self>`.

Eg is `Q_SIGNAL void mySignal() const;` a usecase in Qt ?

LeonMatthesKDAB:
For now, we'll only allow mutable signals, if someone needs const signals, we can reopen this.

ahayzen-kdab:
Appears there are such things as const signals used in Qt you can see this here https://doc.qt.io/qt-6/qvideosink.html#signals where `videoFrameChanged` is const. Therefore reopening this issue.

LeonMatthesKDAB:
Note that the signal handler of a const signal should then also only get const access to the emitter.
And I propose to only allow const signals in C++Qt blocks for now, not RustQt.

kingIzi:
Hah, no way! I'm still very new to Rust. For a workaround, I created a C++ wrapper that the QML_ELEMENT inherits from, which connects to `videoFrameChanged` a QVideoSink property, and emits a custom non-const `videoFrameChanged` that can be caught inside RustQt

PS: If there is another way to connect to const signals inside C++Qt, I will have to learn as I go

PR #1317, opened 2025-07-25
Allow immutable signals

BenFordTytherington:
Fix for #578

### DIFF
--- a/crates/cxx-qt-gen/src/generator/rust/signals.rs
+++ b/crates/cxx-qt-gen/src/generator/rust/signals.rs
@@ -85,9 +85,7 @@ pub fn generate_rust_signal(
     let self_type_cxx = if signal.mutable {
         parse_quote_spanned! {span => Pin<&mut #qobject_name_rust> }
     } else {
-        // CODECOV_EXCLUDE_START
-        unreachable!("Signals cannot be immutable right now so this cannot be reached")
-        // CODECOV_EXCLUDE_STOP
+        parse_quote_spanned! {span => &#qobject_name_rust }
     };
     let self_type_qualified = syn_type_cxx_bridge_to_qualified(&self_type_cxx, type_names)?;
     let qualified_impl = qobject_name.rust_qualified();
--- a/crates/cxx-qt-gen/src/parser/signals.rs
+++ b/crates/cxx-qt-gen/src/parser/signals.rs
@@ -9,7 +9,8 @@ use crate::{
 };
 use core::ops::Deref;
 use std::ops::DerefMut;
-use syn::{spanned::Spanned, Attribute, Error, ForeignItemFn, Result, Visibility};
+use syn::spanned::Spanned;
+use syn::{Attribute, Error, ForeignItemFn, Result, Visibility};
 
 #[derive(Clone)]
 /// Describes an individual Signal
@@ -26,6 +27,11 @@ pub struct ParsedSignal {
     pub cfgs: Vec<Attribute>,
 }
 
+pub enum ImmutabilityConstraint {
+    Allowed,
+    Disallowed,
+}
+
 impl ParsedSignal {
     const ALLOWED_ATTRS: [&'static str; 6] =
         ["cfg", "cxx_name", "rust_name", "inherit", "doc", "qsignal"];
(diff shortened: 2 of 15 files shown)


crates/cxx-qt-gen/src/generator/rust/signals.rs
```diff
@@ -85,9 +85,10 @@ pub fn generate_rust_signal(
     let self_type_cxx = if signal.mutable {
         parse_quote_spanned! {span => Pin<&mut #qobject_name_rust> }
     } else {
-        // CODECOV_EXCLUDE_START
-        unreachable!("Signals cannot be immutable right now so this cannot be reached")
-        // CODECOV_EXCLUDE_STOP
+        // // CODECOV_EXCLUDE_START
+        // unreachable!("Signals cannot be immutable right now so this cannot be reached")
+        // // CODECOV_EXCLUDE_STOP
```
ahayzen-kdab, line 90: can remove the commented code :-)

ahayzen-kdab: guess we need to ensure there is a unit test that reaches this for coverage though

crates/cxx-qt-gen/src/parser/signals.rs
```diff
@@ -42,12 +42,13 @@ impl ParsedSignal {
         let fields = MethodFields::parse(method, auto_case)?;
         let attrs = require_attributes(&fields.method.attrs, &Self::ALLOWED_ATTRS)?;
 
-        if !fields.mutable {
-            return Err(Error::new(
-                fields.method.span(),
-                "signals must be mutable, use Pin<&mut T> instead of T for the self type",
-            ));
-        }
+        // TODO: add proper checks
```
ahayzen-kdab, line 45: right, as @LeonMatthesKDAB suggested, we probably only want const signals support inside `unsafe extern "C++Qt"` blocks not `unsafe extern "RustQt"`, so we need to determine that info somehow here :thinking:

crates/cxx-qt-gen/test_inputs/signals.rs
```diff
@@ -26,6 +26,9 @@ mod ffi {
         #[qsignal]
         fn ready(self: Pin<&mut Self>);
 
+        #[qsignal]
+        fn const_ready(&self);
```
ahayzen-kdab, line 30: this would need to move up into the `C++Qt` block to be on the `QTimer`

examples/qml_features/rust/src/externcxxqt.rs
```diff
@@ -22,6 +22,11 @@ pub mod ffi {
         #[qsignal]
         fn triggered(self: Pin<&mut Self>);
 
+        /// const signal that is emitted when trigger is fired
+        #[qsignal]
+        #[rust_name = "triggered_const_signal"]
+        fn triggeredConstSignal(&self);
```
ahayzen-kdab, line 28: you should add a call to this here

https://github.com/KDAB/cxx-qt/blob/main/examples/qml_features/cpp/external_qobject.cpp#L18

And copy the connection here 

https://github.com/KDAB/cxx-qt/blob/main/examples/qml_features/rust/src/externcxxqt.rs#L69

And then you can add a signal spy for it and check that they are triggered here so it's tested :-) 

https://github.com/KDAB/cxx-qt/blob/main/examples/qml_features/tests/tst_externcxxqt.qml#L41

crates/cxx-qt-gen/src/generator/rust/signals.rs
BenFordTytherington: Resolved

crates/cxx-qt-gen/src/parser/signals.rs
```diff
@@ -33,19 +34,24 @@ impl ParsedSignal {
...
     pub fn mock(method: &ForeignItemFn) -> Self {
-        Self::parse(method.clone(), CaseConversion::none()).unwrap()
+        Self::parse_rust_qt_signal(method.clone(), CaseConversion::none()).unwrap()
     }
 
-    pub fn parse(method: ForeignItemFn, auto_case: CaseConversion) -> Result<Self> {
+    pub fn parse_with_mutability(
+        method: ForeignItemFn,
+        auto_case: CaseConversion,
+        immutable_allowed: bool,
```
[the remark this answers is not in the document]
BenFordTytherington: Added, let me know what you think of the naming though.

crates/cxx-qt-gen/src/parser/signals.rs
```diff
@@ -26,27 +27,42 @@ pub struct ParsedSignal {
...
         let cfgs = extract_cfgs(&method.attrs);
         let fields = MethodFields::parse(method, auto_case)?;
         let attrs = require_attributes(&fields.method.attrs, &Self::ALLOWED_ATTRS)?;
 
-        if !fields.mutable {
-            return Err(Error::new(
-                fields.method.span(),
-                "signals must be mutable, use Pin<&mut T> instead of T for the self type",
-            ));
+        match mutability_allowed {
```
ahayzen-kdab, line 55: could potentially have one if statement?

```rust
if matches!(mutability_allowed, MutabilityContraint::Disallowed) && !fields.mutable {
...
}
```

Although this then reads curiously should it be `ImmutabilityConstraint` ? As when the constraint is disallowed we are not allowed to have immutability but we are allowed to have mutability. Whereas when you read this is sounds a little like the opposite?

Issue #1138, opened 2024-12-03
CXX-Qt-lib: Implement `QHash<K,V>` instead of `QHash<QHashPair_K_V>`


LeonMatthesKDAB:
# Idea

It turns out there is an exception to the orphan rule that we can exploit to make `QHash<K,V>` work that we didn't think of before.

Basically, whilst you generally cannot implement a foreign trait for a foreign type, you can actually do so, if the trait is templated and one of the template arguments is a local type!

So if we make the QHashPair trait templated over the Key and the Value, we can actually make this work:

```rust
pub struct QHash<Key, Value>
where
    QHash<Key, Value>: QHashPair<Key, Value>,
{
    _space: MaybeUninit<usize>,
    _key: PhantomData<Key>,
    _value: PhantomData<Value>,
}
```

It would then be possible for users to:
```rust
impl cxx_qt_lib::QHashPair<cxx_qt_lib::QString, MyType>
    for cxx_qt_lib::QHash<cxx_qt_lib::QString, MyType> {
// ....
}
```
because `MyType` is a local type, this is possible, even if the target type and trait are coming from cxx_qt_lib.

# Backwards compatibility

The most problematic part of this refactoring is backwards-compatibility.
Our current tutorial videos already show `QHash<QHashPair_K_V>` and we want to keep those tutorials working as much as possible.

We may be able to solve this issue by using default type paramters, e.g. by setting `V=()`, we would allow users to use a QHashPair_K_V as the `Key` whilst leaving the value blank.

Because all behavior in the end is delegated to the trait anyway, we can implement `QHashPair<QHashPair_K_V, ()>`, while also adding a deprecation notice to the `QHashPair_K_V`.

Note that we likely want to remove this entirely for  the final 1.0 release, as this solution is likely to require some contortion on the implementation of QHash.

jnbooth:
>Basically, whilst you generally cannot implement a foreign trait for a foreign type, you can actually do so, if the trait is templated and one of the template arguments is a local type!

However, it's useful that the `QHashPair` approach allows you to implement the trait for pairs of foreign types, e.g. creating `QHashPair_QString_QString`.

LeonMatthesKDAB:
@jnbooth Hi, while that is true you could still do something very close with the newtype pattern.

e.g. just add your own QString type:
```rust
#[repr(transparent)]
type MyQString(QString);

// To/From impls

impl QHashable for QHash<MyQString, QString> { ... }
```

[...]

PR #1318, opened 2025-07-29
cxx-qt-lib: add loadData and setInitialProperties to QQmlApplicationEngine

ahayzen-kdab:


crates/cxx-qt-lib/src/qml/qqmlapplicationengine.rs
```diff
@@ -6,6 +6,10 @@
 #[cxx_qt::bridge]
 mod ffi {
     unsafe extern "C++" {
+        include!("cxx-qt-lib/qbytearray.h");
+        type QByteArray = crate::QByteArray;
+        include!("cxx-qt-lib/qmap.h");
+        type QMap_QString_QVariant = crate::QMap<crate::QMapPair_QString_QVariant>;
```
ahayzen-kdab, line 12: wonder if we can create a `QVariantMap` alias that works to make this a bit nicer, but it works :-)

LeonMatthesKDAB: I think this is good enough for now, let's not introduce too many types around QMap, as we were thinking about refactoring that anyhow similar to https://github.com/KDAB/cxx-qt/issues/1138

LeonMatthesKDAB (approved):

PR #1321, opened 2025-08-01
Remove unnecessary invocations of .to_string()

jnbooth:
`x.to_string()` is generally equivalent  to `write!(s, "{x}")`, which often isn't what you actually want to do:

- Instead of `"".to_string()`, use `String::new()`.
- For `cow: Cow<str>`, instead of `cow.to_string()`, use `cow.into_owned()`.
- For `s: &str`, instead of `s.to_string()`, use `s.to_owned()`.
- For `qstring: QString`, instead of `qstring.to_string()`, use `String::from(&qstring)`.
- For `string: String`, instead of `string.to_string()`, use `string.clone()`.
- For `path: &Path`, instead of `format!("{}", path.to_string_lossy())`, use `format!("{}", path.display())`.

### DIFF
--- a/crates/cxx-qt-lib/src/core/qbytearray.rs
+++ b/crates/cxx-qt-lib/src/core/qbytearray.rs
@@ -262,7 +262,7 @@ impl From<&QByteArray> for Vec<u8> {
 
 impl<const N: usize> From<&[u8; N]> for QByteArray {
     #[inline]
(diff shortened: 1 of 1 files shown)


LeonMatthesKDAB (approved):
@jnbooth Great, thank you for the contribution.

PR #1324, opened 2025-08-07
Use early return/continue and pattern matching instead of nested if-else blocks

jnbooth:
Adds the following lints:

- [clippy::if_not_else](https://rust-lang.github.io/rust-clippy/master/index.html#if_not_else)
- [clippy::manual_let_else](https://rust-lang.github.io/rust-clippy/master/index.html#manual_let_else)
- [clippy::redundant_else](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_else)
- [clippy::single_match_else](https://rust-lang.github.io/rust-clippy/master/index.html#single_match_else)

By using early return/continue and pattern matching, nested if-else blocks can be flattened out.

### DIFF
--- a/crates/cxx-qt-gen/src/naming/cpp.rs
+++ b/crates/cxx-qt-gen/src/naming/cpp.rs
@@ -65,19 +65,12 @@ pub(crate) fn syn_type_to_cpp_return_type(
             }
 
             let arg = args.pop().unwrap();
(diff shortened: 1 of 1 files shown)


LeonMatthesKDAB (approved):
Thank you for the cleanup @jnbooth

PR #1327, opened 2025-08-19
Add comprehensive linting

jnbooth:
Enables most Clippy lints.

Most of these changes were automatically applied using `cargo clippy --fix`. Nearly all of them are simple syntactic adjustments like adding a semicolon at the end of a line. The remaining changes are things like removing unused `&self` parameters and accepting `&str` instead of `String` if ownership isn't necessary.

I can break those up into multiple PRs if that's preferred, but I figured I'd start by putting them all in one place so people can take a look and decide what they think.


Cargo.toml
```diff
@@ -65,8 +65,26 @@ serde_json = "1.0"
 thiserror = "1.0"
 
 [workspace.lints.clippy]
```
ahayzen-kdab, line 67: i guess my question would be why are these lints not enabled by default? Do we want to enable so many? I usually follow the defaults for lints as there is sometimes a reason why they are not enabled yet. As otherwise this can become an explosion of debates.

.github/workflows/github-cxx-qt-tests.yml
```diff
@@ -426,12 +426,12 @@ jobs:
...
       uses: actions/checkout@v5
     # Ensure clippy and rustfmt is installed, they should come from github runner
-    # clippy version needs to be 1.78.0 for the MSRV lint
+    # clippy version needs to be 1.87.0 for all lints
     #
     # Note we still need rustfmt for the cxx-qt-gen tests
     - name: "Install Rust toolchain"
       run: |
-        rustup toolchain add 1.78.0 --component clippy
+        rustup toolchain add 1.87.0 --component clippy
```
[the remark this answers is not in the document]
LeonMatthesKDAB: IMO, clippy could also just be separate runner, as I think it doesn't share many build artifacts with the main build anyhow.

Cargo.toml
jnbooth: No clippy lints are enabled by default as far as I'm aware.

PR #1337, opened 2025-10-02
Revert "Add a lifetime parameter to QMetaObjectConnectionGuard (#1263)"

### LINKED ISSUE #1336: QScopedMetaObjectConnectionGuard is unsound
#1263 introduced the QScopedMetaObjectConnectionGuard, which allowed connecting lambdas to signals that contain lifetime-bound items.

The idea is that the guard is dropped, which will disconnect the signal, ensuring the lifetime bounds are upheld.

However, this *does not work*, as Rust doesn't give any guarantees that an object must be dropped at all!
This then allows leaking the guard, thereby allowing unsound memory access!

For example this modified qml_minimal code compiles:

```rust
    pub fn increment_number(mut self: Pin<&mut Self>) {
        let previous = *self.number();
        self.as_mut().set_number(previous + 1);

        let guard = self.on_number_changed(|_| println!("{previous}"));
        std::mem::forget(guard);
    }
```

But leads to `previous` becoming a dangling pointer, which ends up printing garbage on my machine.

LeonMatthesKDAB:
This reverts commit 20ee58f4208882395f1f6a7254f2eae1b51daf2f.

Closes #1336

### DIFF
--- a/crates/cxx-qt-gen/test_outputs/signals.rs
+++ b/crates/cxx-qt-gen/test_outputs/signals.rs
@@ -232,10 +232,8 @@ mod ffi {
     unsafe extern "C++" {
         #[doc(hidden)]
         #[namespace = "cxx_qt::my_object::rust::cxxqtgen1"]
-        type QTimerCxxQtSignalHandlerconst_ready<'a> = cxx_qt::signalhandler::CxxQtSignalHandler<
-            'a,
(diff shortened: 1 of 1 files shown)


ahayzen-kdab (approved):
Looks good assuming CI likes it

PR #1338, opened 2025-10-02
qt-build-utils: add QmlDir and QmlUri builders

ahayzen-kdab:


crates/qt-build-utils/src/lib.rs
```diff
@@ -15,6 +15,11 @@
 
 #![allow(clippy::too_many_arguments)]
 
+mod builder;
```
ahayzen-kdab, line 18: maybe these are more generators than builders as they are creating strings rather than building anything :thinking:

crates/qt-build-utils/src/builder/qmluri.rs
```diff
@@ -0,0 +1,34 @@
...
+pub struct QmlUriBuilder {
+    uri: Vec<String>,
+}
+
+impl QmlUriBuilder {
+    /// Construct a [QmlUriBuilder] from a given string
+    pub fn new(uri: impl IntoIterator<Item = impl Into<String>>) -> Self {
+        Self {
+            uri: uri.into_iter().map(Into::into).collect(),
+        }
```
LeonMatthesKDAB, line 17: This needs checks on the strings to ensure they only contain valid characters.

Unsure what exactly we want to limit it to, but to me it sounds like the whole point of such a builder is that it checks things like this for us...

Also, for now I'm happy to be more restrictive about this than potentially needed, just so we're sure we don't allow anything invalid.

crates/qt-build-utils/src/builder/qmluri.rs
```diff
@@ -0,0 +1,34 @@
+// SPDX-FileCopyrightText: 2025 Klarälvdalens Datakonsult AB, a KDAB Group company <info@kdab.com>
+// SPDX-FileContributor: Andrew Hayzen <andrew.hayzen@kdab.com>
+//
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+/// A builder for representing a QML uri
+#[derive(Clone)]
+pub struct QmlUriBuilder {
```
LeonMatthesKDAB, line 8: This is not a builder, but just a `QmlUri`.

crates/qt-build-utils/src/builder/qmluri.rs
```diff
@@ -0,0 +1,34 @@
...
+    uri: Vec<String>,
+}
+
+impl QmlUriBuilder {
+    /// Construct a [QmlUriBuilder] from a given string
+    pub fn new(uri: impl IntoIterator<Item = impl Into<String>>) -> Self {
```
LeonMatthesKDAB, line 14: We should probably have a `From<&str>` impl that splits on dots, so it's possible to do: `"com.kdab.com.cxx_qt".into()` when a QmlUri is needed.

crates/qt-build-utils/src/builder/qmldir.rs
```diff
@@ -0,0 +1,145 @@
...
+    /// name is the plugin library name. This is usually not the same as the file
+    /// name of the plugin binary, which is platform dependent. For example, the
+    /// library MyAppTypes would produce libMyAppTypes.so on Linux and MyAppTypes.dll
+    /// on Windows.
+    ///
+    /// Zero or more C++ plugin declarations may exist in the qmldir file.
+    /// However, since plugin loading is a relatively expensive operation, clients
+    /// are advised to specify at most a single plugin.
+    //
+    // TODO: Zero or more but only one is recommended so is enforcing one ok for now?
```
LeonMatthesKDAB, line 98: If we only want to support a single plugin for now, then document it and add a check that it's not already been set. (panicking in that case is fine).

crates/qt-build-utils/src/builder/qmldir.rs
```diff
@@ -0,0 +1,145 @@
...
+            .unwrap_or_default();
+        let class_name = self
+            .class_name
+            .map(|name| format!("classname {name}"))
+            .unwrap_or_default();
+        let type_info = self
+            .type_info
+            .map(|file| format!("typeinfo {file}"))
+            .unwrap_or_default();
+        let prefer = format!("prefer :/qt/qml/{}/", self.uri.as_dirs());
```
LeonMatthesKDAB, line 56: As these are all just optionally appended to the writer, this should be refactored into a series of if statements:

```rust
if let Some(class_name) = self.class_name {
    write!(writer, "classname {class_name}");
}
```

That is IMO more readable and avoids a bunch of string allocations.

crates/qt-build-utils/src/builder/qmlplugincpp.rs
```diff
@@ -0,0 +1,87 @@
...
+    plugin_class_name: String,
+    qml_cache: bool,
+    uri: QmlUriBuilder,
+}
+
+impl QmlPluginCppBuilder {
+    /// Construct a [QmlPluginCppBuilder] from a uri and plugin class name
+    pub fn new(uri: QmlUriBuilder, plugin_class_name: impl Into<String>) -> Self {
+        Self {
+            plugin_class_name: plugin_class_name.into(),
```
LeonMatthesKDAB, line 21: Anything we should check here for validity of the plugin_class_name?

crates/qt-build-utils/src/builder/qresource.rs
```diff
@@ -0,0 +1,213 @@
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<T: Into<QResourceFile>> From<T> for QResource {
+    fn from(value: T) -> Self {
+        Self::new().file(value)
+    }
+}
```
LeonMatthesKDAB, line 67: I would like to see a similar implementation for `QResources` itself.

Most likely we want a `From` impl for `T: Into<QResourceFile>`, as well as `T: Iterator<Item=I>, I: Into QResourceFile`,
That way we could have an API on CxxQtBuilder that is simply:

```rust
.resources([my_image.png, my_other_image.png])
```
Which would boil down to a QResources, with a single QResources and two QResourceFile instances.

crates/qt-build-utils/src/lib.rs
LeonMatthesKDAB: Hm, most of these are just for generating QML-specific things, so why not split it into a `qml` and a `qrc` module?

crates/qt-build-utils/src/builder/qmldir.rs
ahayzen-kdab: done

crates/qt-build-utils/src/builder/qresource.rs
ahayzen-kdab: done

LeonMatthesKDAB (dismissed):

LeonMatthesKDAB (approved):

PR #1345, opened 2025-10-08
qt-build-utils: add Qml Language Server ini builder

ahayzen-kdab:


LeonMatthesKDAB (commented):
Looks good, I just have some questions about what Qmlls expects.

PR #1364, opened 2025-11-26
Export qmltypes and qmldir for each plugin

ahayzen-kdab:


crates/cxx-qt-build/src/lib.rs
```diff
@@ -878,22 +878,39 @@ impl CxxQtBuilder {
...
             cc_builder.define("QT_STATICPLUGIN", None);
 
             // If any of the files inside the qml module change, then trigger a rerun
             for file in qml_module.qml_files {
                 println!("cargo::rerun-if-changed={}", file.path().display());
             }
 
+            // Export the .qmltypes and qmldir files into a stable path, so that tools like
+            // qmllint/qmlls can find them.
+            let plugin_dir = dir::module_export(&qml_module.uri);
```
ahayzen-kdab: should this export be happening here inside the build_qml_module ? Or should this happen somehow somewhere else as part of the export phase?

ahayzen-kdab: Right lets add that in another PR to keep things moving...

PR #1369, opened 2025-12-12
Add an example of a dynamically loaded QML plugin

### LINKED ISSUE #1266: What happened to the `QmlExtensionPlugin` example
I remember an example about it, but it disappeared after 0.5.
This is confusing to me what is the right way to write a "plugin" nowadays?

It seems cxx-qt is almost dealing with everything, all the symbols are generated, the `QmlExtensionPlugin` class, the `qmltypes`, etc... It just I don't understand how to make it a loadable plugin.

I'm happy with a cargo only build or a cmake based one.

## Some use case

I want to introduce a new module to test cxx-qt for a cpp based project, but I want to make the changes as minimal as possible. So I thought the plugin approach sound nice.

## What I've done so far

Everything I've tried is based of the `qml_minimal` example

- cargo only: build a dynlib, move the dynlib file in the same folder as where `qmldir` is located. I get `rust/target/debug/build/qml_minimal-b6d7b9673e0f279f/out/qt-build-utils/qml_modules/com/kdab/cxx_qt/demo/libcom_kdab_cxx_qt_demo.dylib' is not a Qt plugin`.
- Many variant from CMake
- Write my own plugin:

```
qt_add_plugin(example_qml_minimal
    cpp/plugin.cpp
)
target_link_libraries(example_qml_minimal PRIVATE qml_minimal_qml_module)
```

And in `plugin.cpp`:

```cpp
#include <QtQml/QQmlExtensionPlugin>
#include <QtQml/QQmlEngine>

#include <qml_minimal/src/cxxqt_object.cxxqt.h>

class CoreQmlpluginPlugin : public QQmlExtensionPlugin
{
    Q_OBJECT
    Q_PLUGIN_METADATA(IID QQmlExtensionInterface_iid)

public:
    void registerTypes(const char* uri) override
    {
        qmlRegisterType<my_object::MyObject>(uri, 1, 0, "MyObject");
    }
};

#include "plugin.moc"
```

The latter works, but then I'm loosing the benefit of what cxx-qt is doing: automation (qml_element) and tooling files (qmltypes)

## Conclusion

I feel I'm missing an obvious step somewhere. Using the keyword `plugin` doesn't yield any interesting issue here, or relevant doc in the book.

LeonMatthesKDAB:
Closes #1266 

This mostly needs some adjustments to how we build the QML plugin to be built as a dynamic plugin, instead of static, and then needs CMake to copy everything into the right place.


examples/qml_minimal_plugin/rust/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+  "Leon Matthes <leon.matthes@kdab.com>"
+]
+edition = "2021"
+license = "MIT OR Apache-2.0"
+
+# This will instruct Cargo to create a static
+# library which CMake can link against
+[lib]
+crate-type = ["cdylib"]
+# ANCHOR_END: book_static_lib
```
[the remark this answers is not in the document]
LeonMatthesKDAB: Right, I will need to update the book in general, so the anchors here and in the CMakeLists.txt will likely change. (was just copied from qml_minimal)

crates/cxx-qt-build/src/lib.rs
```diff
@@ -688,6 +693,15 @@ impl CxxQtBuilder {
         for include_path in include_paths {
             builder.include(include_path);
         }
+
+        if !Self::is_debug() {
```
ahayzen-kdab: should we only do this on Windows ? Or do we also need to on other platforms :thinking:

ahayzen-kdab: Seems when building a plugin you use `if cc_builder.get_compiler().is_like_msvc() {` to set this.

Does this *also* need to be enabled in all cc builders too then ?

examples/CMakeLists.txt
```diff
@@ -9,9 +9,10 @@
 # When using `cargo test`
 add_subdirectory(qml_features)
 add_subdirectory(qml_minimal)
-add_subdirectory(qml_multi_crates)
 
 # TODO: get demo_threading working for wasm builds
 if(NOT BUILD_WASM)
+    add_subdirectory(qml_minimal_plugin)
+    add_subdirectory(qml_multi_crates)
```
ahayzen-kdab: does multi crates now not build somehow when it did before for WASM?

examples/qml_minimal_plugin/CMakeLists.txt
```diff
@@ -0,0 +1,131 @@
...
+
+# Link to the qml module, which in turn links to the Rust qml_minimal library
+target_link_libraries(example_qml_minimal_plugin PRIVATE Qt::Core Qt::Gui Qt::Qml Qt::QuickControls2)
+
+# If we are using a statically linked Qt then we need to import any qml plugins
+# qt_import_qml_plugins(example_qml_minimal_plugin)
+
+if(BUILD_TESTING)
+    #
+    # Unit test
```
ahayzen-kdab: can you copy the `qmllint` test from the `qml_minimal` example so we check that too here

LeonMatthesKDAB: done

PR #1383, opened 2025-12-23
Add missing QVariantValue implementations

jnbooth:
Implements QVariantValue for the following types:

- QFont
- QImage
- QLine
- QLineF
- QPen
- QPolygon
- QPolygonF
- QRegion
- QVariantHash
- QVariantList
- QVariantMap


ahayzen-kdab (approved):
I think initially we only added implementations for types that had a QVariant constructor, but makes sense to do all as this cannot be implemented externally for any cxx-qt-lib types, and these can be done via QVariant::fromValue in C++.

PR #1405, opened 2026-02-09
Add bindings for QChar

jnbooth:
A `QString` is a list of `QChar`s, but cxx-qt-lib doesn't currently have bindings for `QChar`. That means any time you want to interact with the `QString` as a list (split, add, remove, etc.), you need to allocate a new `QString`. This PR adds bindings for `QChar` so it can be used directly. That also makes it much easier to find out what's inside a `QString` without needing to convert it to a `String`.

A `QChar` is just a standard UTF-16 double byte, so there's no need for FFI calls. In fact, they can be constructed at compile time. That means instead of doing this:

```rust
fn split_items(s: &QString) -> QStringList {
    s.split(&QString::from("|"), /* other args */)
}
```

You can now do this:

```rust
const DELIM: QChar = QChar::from_char('|').unwrap();

fn split_items(s: &QString) -> QStringList {
    s.split(DELIM, /* other args */)
}
```

This is both more efficient (no allocations) and more readable (no "magic" strings).


crates/cxx-qt-lib/src/core/qchar.rs
```diff
@@ -0,0 +1,150 @@
...
+    extern "C++" {
+        include!("cxx-qt-lib/qchar.h");
+        #[allow(unused)]
+        type QChar = super::QChar;
+    }
+}
+
+#[repr(C)]
+#[derive(Copy, Clone, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct QChar {
```
ahayzen-kdab, line 21: would be good to add QChar as a test in `qt_types_standalone` as this does a round trip of construct, copy, destruct to check for memory leaks

PR #1430, opened 2026-03-10
feat: add QQmlImageProviderBase

dphaldes:
supercedes #1142


dphaldes:
I have "working" implementation here https://github.com/dphaldes/compose/blob/master/src/bridge.rs.  
image

Not particularly fond of the solution since it uses an unchecked cast since I wasn't able to figure out how to cast `Pin<&mut QQmlImageProviderBase>` into `*mut QQmlImageProviderBase`. The application also segfaults when closing probably because it is trying to release the ImageProvider. 

Any ideas appreciated!

dphaldes:
I didn't see that method. I will look into using that. It would work but it still wouldnt work as a safe cast will it? as in using the Upcast trait instead of `as` keyword

crates/cxx-qt-lib/src/qml/qqmlimageproviderbase.rs
```diff
@@ -0,0 +1,34 @@
...
+    #[repr(i32)]
+    #[namespace = "rust::cxxqtlib1"]
+    #[derive(Debug)]
+    enum QQmlImageProviderBaseImageType {
+        Invalid = 0,
+        Image,
+        Pixmap,
+        Texture,
+        ImageResponse,
+    }
```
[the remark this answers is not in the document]
dphaldes: I am going to enable it only on qt6 for now.

crates/cxx-qt-lib/src/qml/qqmlengine.rs
```diff
@@ -98,6 +103,23 @@ mod ffi {
...
+        ///
+        /// Sets the provider to use for images requested via the image: url scheme, with host providerId.
+        /// The QQmlEngine takes ownership of provider.
+        #[cfg(cxxqt_qt_version_major = "6")]
+        #[rust_name = "add_image_provider"]
+        unsafe fn addImageProvider(
+            self: Pin<&mut QQmlEngine>,
+            provider_id: &QString,
+            provider: *mut QQmlImageProviderBase,
+        );
```
ahayzen-kdab, line 119: clippy seems to be confused and thinking there is no `# Safety` section ? Does this happen locally too ? Wonder if the order needs to be different or something else is confusing it

dphaldes: I cant seem to fix this in anyway. clippy crashes before it reaches this locally

ahayzen-kdab: Looks like clippy gets confused with the layers of macros, elsewhere `#[allow(clippy::missing_safety_doc)]` was added to suppress this, so probably best to just copy that

https://github.com/KDAB/cxx-qt/blob/2cb6df179b37e760878413d480edd623c61cfdd7/crates/cxx-qt-lib/src/core/qtlogging.rs#L53-L60

dphaldes: Done

ahayzen-kdab (approved):
Looks good, thanks for working that through to the end! :-)

PR #1441, opened 2026-04-19
Add Q(Gui)Application::applicationDisplayName

nicolasfella:


ahayzen-kdab (approved):
LGTM, these may be able to use the newer casting but the others are implemented in the same way

PR #1444, opened 2026-05-04
Use `#[Self = "..."]` instead of manual code bindings

jnbooth:
This is a cxx feature unlocked by #1416.


ahayzen-kdab (approved):
LGTM :-)

PR #1446, opened 2026-05-06
Implement `PartialEq<[u8]>` for QByteArray and `PartialEq<str>` for QString

jnbooth:
In Qt, you can test equality for QByteArrays and QStrings against raw data without needing to allocate memory, e.g. `QStringLiteral("test") == "test"`. This PR makes it possible to do the same in Rust by implementing `PartialEq` for the corresponding types.

### DIFF
--- a/crates/cxx-qt-lib/src/core/qstring.rs
+++ b/crates/cxx-qt-lib/src/core/qstring.rs
@@ -247,6 +247,39 @@ impl PartialEq for QString {
 
 impl Eq for QString {}
 
+impl PartialEq<str> for QString {
+    fn eq(&self, other: &str) -> bool {
+        self.as_slice().iter().copied().eq(other.encode_utf16())
+    }
+}
+impl PartialEq<QString> for str {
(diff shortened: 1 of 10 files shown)


crates/cxx-qt-lib/src/core/qstring.rs
```diff
@@ -247,6 +247,39 @@ impl PartialEq for QString {
 
 impl Eq for QString {}
 
+impl PartialEq<str> for QString {
+    fn eq(&self, other: &str) -> bool {
+        self.as_slice().iter().copied().eq(other.encode_utf16())
+    }
+}
```
ahayzen-kdab, line 254: hmm wonder if we want this one, as most of the time the API ends up being explicit about when there is a copy or conversion to/from utf16 or runtime cost ? 🤔

nicolasfella: https://github.com/KDAB/cxx-qt/pull/1473 does this without the conversion

PR #1448, opened 2026-05-07
Fix memory leaks caused by `std::move` on types with no move constructor

jnbooth:
`cxx` uses the `IsRelocatable` struct to detect whether a C++ type is safe to move by `memcpy`. It permits only types that have trivial move constructors and no destructor. `cxx-qt-lib` and `cxx-qt-lib-extras` sometimes override that check, like so:

```cpp
namespace rust {
template<>
struct IsRelocatable<MyType> : ::std::true_type
{};
}
```

This is a violation of `cxx`'s assumptions, but it's valid if two conditions are met:

1. `QTypeInfo<MyType>::isRelocatable` must be true. This ensures `memcpy` is safe to use.
2. `MyType` must have a move constructor (i.e. `MyType(MyType&& other)`). This ensures that if the type is passed from Rust into C++ by value, C++ will take ownership of it, and vice versa.

The first condition can be checked at compile time with `static_assert(QTypeInfo<MyType>::isRelocatable)`, which is exactly what `cxx-qt-lib` and `cxx-qt-lib-extras` do. However, as far as I know, there is no way to check the second condition at compile time. Due to this, `cxx-qt-lib` and `cxx-qt-lib-extras` declare `IsRelocatable` for several types that do not have move constructors: `QFont`, `QMessageLogContext`, `QCommandLineOption`, and `QCommandLineParser`. If any of these types are passed by value into C++, their reference counter will not decrement, which means they will never deallocate.

However, when those invalid `IsRelocatable` declarations are removed, another problem arises: it is no longer possible for Rust to bind to C++ functions that return those types, since those are *also* passing by value. That includes the FFI bindings we use for `MyType::default()`, `MyType::clone()`, and so on. The solution is to construct the types in-place using `new (p) MyType(...)`, which is the C++ syntax for applying a constructor to the existing place in memory at pointer address `p`, rather than constructing a new value on the stack and returning it. (`cxx` makes heavy use of that syntax in generated code, which is how I learned about it.) So now `common.h` has a new constructor function:

```cpp
// (CURRENT) Constructs a value on the stack and returns it.
template<typename T, typename... Args>
T
construct(Args... args)
{
  return T(std::forward<Args>(args)...);
}

// (ADDED) Constructs a value in-place at the specified pointer address.
template<typename T, typename... Args>
void
constructInPlace(T* uninit, Args... args)
{
  new (uninit) T(std::forward<Args>(args)...);
}
```

On the Rust side, those types now use the `crate::util::new_in_place` helper function to construct their values, which uses `MaybeUninit` under the hood.

```rs
pub(crate) unsafe fn new_in_place<T, F: FnOnce(*mut T)>(constructor: F) -> T
{
    let mut uninit = MaybeUninit::uninit();
    constructor(uninit.as_mut_ptr());
    unsafe { uninit.assume_init() }
}
```

So instead of this:

```rust
impl Clone for MyType {
    fn clone(&self) -> Self {
        ffi::mytype_clone(self)
    }
}
```

Corresponding implementations now do this:

```rust
impl Clone for MyType {
    fn clone(&self) -> Self {
        unsafe { new_in_place(|uninit| ffi::mytype_clone(uninit, self)) }
    }
}
```

It's a bit unwieldy, but I think it's the cleanest approach. The only other way to prevent memory leaks that I can think of is to wrap those types in `UniquePtr`s, but that would be a significant breaking change for those types and wouldn't really be an improvement anyway (in my opinion). This way, it's all kept internal behind the API.

EDIT: Also to note, this is how `cxx` handles it. When an initializer is defined in a bridge, this is the Rust side:

```rs
extern "C" {
void rust$cxxqtlib1$cxxbridge1$192$qfont_init_default(::QFont *return$) noexcept {
  ::QFont (*qfont_init_default$)() = ::rust::cxxqtlib1::construct;
  new (return$) ::QFont(qfont_init_default$());
}
```

And this is the C++ side:

```cpp
    pub fn qfont_init_default() -> QFont {
        unsafe extern "C" {
            #[link_name = "rust$cxxqtlib1$cxxbridge1$192$qfont_init_default"]
            fn __qfont_init_default(__return: *mut QFont);

        }
        unsafe {
            let mut __return = ::cxx::core::mem::MaybeUninit::<QFont>::uninit();
            __qfont_init_default(__return.as_mut_ptr());
            __return.assume_init()
        }
    }
```

### DIFF
--- a/crates/cxx-qt-lib-extras/include/core/qcommandlineoption.h
+++ b/crates/cxx-qt-lib-extras/include/core/qcommandlineoption.h
@@ -7,15 +7,3 @@
 #pragma once
 
 #include <QtCore/QCommandLineOption>
-
-#include "rust/cxx.h"
-
-// Define namespace otherwise we hit a GCC bug
-// https://gcc.gnu.org/bugzilla/show_bug.cgi?id=56480
-namespace rust {
-
-template<>
-struct IsRelocatable<QCommandLineOption> : ::std::true_type
-{};
-
-} // namespace rust
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -378,9 +378,9 @@ dependencies = [
 
 [[package]]
 name = "codespan-reporting"
-version = "0.12.0"
+version = "0.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fe6d2e5af09e8c8ad56c969f2157a3d4238cebc7c55f0a517728c38f7b200f81"
+checksum = "af491d569909a7e4dee0ad7db7f5341fef5c614d5b8ec8cf765732aba3cff681"
 dependencies = [
  "serde",
  "termcolor",
@@ -500,11 +500,12 @@ dependencies = [
 
 [[package]]
 name = "cxx"
-version = "1.0.176"
+version = "1.0.194"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d5e9fd2958ba90283c0398c7cef1972aa23db8ab6e4323759a4fa8affb0299c4"
+checksum = "747d8437319e3a2f43d93b341c137927ca70c0f5dabeea7a005a73665e247c7e"
 dependencies = [
  "cc",
+ "cxx-build",
  "cxxbridge-cmd",
  "cxxbridge-flags",
  "cxxbridge-macro",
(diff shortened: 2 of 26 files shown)


crates/cxx-qt-lib-extras/include/core/qcommandlineoption.h
```diff
@@ -7,15 +7,3 @@
 #pragma once
```
ahayzen-kdab, line 7: So all (?) of these types were missing the usually `qt_types_standalone` tests that check for the memory leaks and would have spotted the incorrect usage of being marked relocatable/trivial type.

Are you able to add `qt_types_standalone` tests for these so we ensure that they are not leaking with the in-place route?

Note for QTimeZone this was spotted and we used an opaque type instead.

PR #1456, opened 2026-05-29
qt-build-utils: do not add whole QML module folder to qrc

ahayzen-kdab:


crates/qt-build-utils/src/lib.rs
```diff
@@ -347,9 +347,6 @@ impl QtBuild {
             let qml_uri_dirs_prefix = format!("/qt/qml/{qml_uri_dirs}");
             let mut qrc = File::create(&qrc_path).expect("Could not create qrc file");
             QResources::new()
-                .resource(QResource::new().prefix("/".to_string()).file(
-                    QResourceFile::new(qml_module_dir_str).alias(qml_uri_dirs_prefix.clone()),
```
ahayzen-kdab, line 351: seems like the folder is not required to be added to qrc, but why was this originally here?

ahayzen-kdab: Looks like this has always been there since we have had QML module support, unclear what the usecase was maybe before we had API to register files to the QRC with an API as the tests still work ...

https://github.com/ahayzen-kdab/cxx-qt/blob/ded676dca59cc1f7132f3c7eeab0cda4d65e726c/crates/qt-build-utils/src/lib.rs#L663

BenFordTytherington (approved):
Doesn't seem like it's used for anything, so good optimisation to remove it, save space

PR #1459, opened 2026-06-15
qt-build-utils: qt minimal fallback if other versions are available

ahayzen-kdab:
Do not try just the first matching version try others too


BenFordTytherington (approved):
Looks good to me, informative comments

PR #1465, opened 2026-07-03
qt_minimal: enable feature in github runner not always in cargo

ahayzen-kdab:


BenFordTytherington (approved):
looks good, helpful to enable in CI, but keep out of regular cargo builds while it is still a new feature

PR #1470, opened 2026-07-27
cxx-qt-lib: implement QVariantValue for QList of primitive types

yuknig:
Add implementation of `QVariantValue` trait for `QList<T>` where `T` is a primitive type (`bool`, `f32`, `f64`, `i8`, `i16`, `i32`, `i64`, `u8`, `u16`, `u32`, `u64`). This enables conversion of these list types to and from QVariant.

This functionality is needed as part of the migration from the in-house qtbridge type library to the type library provided by cxx-qt.


ahayzen-kdab (approved):
LGTM thanks!

PR #1471, opened 2026-07-30
Implement Qlist of QObject pointers

yuknig:
Add a `QList` specialization for `QObjectMutPtr`.
`QObjectMutPtr` is a newtype wrapper around `*mut QObject` that works around Rust's orphan rule limitations.


crates/cxx-qt-lib/include/core/qlist/qlist.h
```diff
@@ -23,6 +23,7 @@
 #include "qlist_QLineF.h"
 #include "qlist_QMargins.h"
 #include "qlist_QMarginsF.h"
+#include "qlist_QObjectPtr.h"
```
ahayzen-kdab: wonder if `QObjectMutPtr` is a better name as this is a `*mut T` not a `*const T` ? And elsewhere in std / cxx things like `as_mut_ptr` are used and there is event `SimdMutPtr` in std

yuknig: Fixed.
That makes sense. In qtbridge-rust, this specialization is actually named QList_ptr_mut_QObject.

ahayzen-kdab (approved):
LGTM, fun how the new type avoids the unsafe methods in the CXX bridge that you usually need for ptrs, but likely fine as you need to convert to/from the inner type.

@yuknig were you going to look at adding support in `QVariant` for `QList<QObject*>` too or were they not needed there ? and i guess `QVariant<QObject*>` might be generally more useful ?

yuknig:
@ahayzen-kdab Yes, I'm looking into it. I'm going to add it in follow-up PR.

PR #1473, opened 2026-08-03
cxx-qt-lib: Implement equality comparison for QString to String/str

nicolasfella:
Using QString's comparability to QUtf8StringView

### DIFF
--- a/crates/cxx-qt-lib/src/core/qstring.cpp
+++ b/crates/cxx-qt-lib/src/core/qstring.cpp
@@ -186,5 +186,11 @@ qstringTrimmed(const QString& string)
   return string.trimmed();
 }
 
+bool
+qstringEqStr(const QString& lhs, rust::Str rhs)
+{
+  return lhs == QUtf8StringView(rhs.data(), rhs.size());
+}
+
 }
 }
--- a/crates/cxx-qt-lib/src/core/qstring.rs
+++ b/crates/cxx-qt-lib/src/core/qstring.rs
@@ -139,6 +139,9 @@ mod ffi {
         #[doc(hidden)]
         #[rust_name = "qstring_cmp"]
         fn operatorCmp(a: &QString, b: &QString) -> i8;
+        #[doc(hidden)]
+        #[rust_name = "compare_qstring_str"]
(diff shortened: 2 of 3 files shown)


crates/cxx-qt-lib/src/core/qstring.cpp
```diff
@@ -186,5 +186,11 @@ qstringTrimmed(const QString& string)
   return string.trimmed();
 }
 
+bool
+qstringEqStr(const QString& lhs, rust::Str rhs)
+{
+  return lhs == QUtf8StringView(rhs.data(), rhs.size());
```
ahayzen-kdab, line 192: this only exists since Qt 6 ? So needs to be put behind a QT_VERSION_CHECK on the C++ side and a eg `#[cfg(cxxqt_qt_version_major = "6")]` or similar for the Rust code

crates/cxx-qt-lib/src/core/qstring.rs
```diff
@@ -245,6 +248,18 @@ impl PartialEq for QString {
...
+    fn eq(&self, other: &String) -> bool {
+        ffi::compare_qstring_str(self, other)
+    }
+}
+
+impl PartialEq<str> for QString {
+    fn eq(&self, other: &str) -> bool {
+        ffi::compare_qstring_str(self, other)
+    }
+}
```
ahayzen-kdab, line 261: appears implementing this causes a load of clippy warnings elsewhere as owned strings aren't needed for comparison now.

```
error: this creates an owned instance just for comparison
  --> tests/qt_types_standalone/rust/src/qcoreapplication.rs:30:5
   |
30 |     String::from(&app.application_name()) == "kdab"
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try: `&app.application_name()`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#cmp_owned
   = note: `-D clippy::cmp-owned` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(clippy::cmp_owned)]`
```

Are you able to fix those too ?

PR #1474, opened 2026-08-04
Introduce QJsonArray, QJsonObject and QJsonValue

yuknig:
Adds initial Rust bindings for JSON types: QJsonArray, QJsonObject, and QJsonValue.


crates/cxx-qt-lib/src/core/qjsonvalue.rs
```diff
@@ -0,0 +1,246 @@
...
+        #[rust_name = "to_string"]
+        fn qjsonvalueToQString(value: &QJsonValue) -> QString;
+
+        #[doc(hidden)]
+        #[rust_name = "to_array"]
+        fn qjsonvalueToArray(value: &QJsonValue) -> QJsonArray;
+
+        #[doc(hidden)]
+        #[rust_name = "to_object"]
+        fn qjsonvalueToObject(value: &QJsonValue) -> QJsonObject;
```
ahayzen-kdab, line 125: can these all not be direct bindings rather than having a wrapper ? 

Eg why not
```rust
/// Converts the value to a `bool`. Returns `false` if the value is not a boolean.
#[rust_name = "to_bool"]
fn toBool(self: &QJsonValue) -> bool;
```

crates/cxx-qt-lib/src/core/qjsonvalue.rs
```diff
@@ -0,0 +1,246 @@
...
+    }
+}
+
+impl From<&QJsonObject> for QJsonValue {
+    fn from(value: &QJsonObject) -> Self {
+        ffi::qjsonvalue_init_from_qjsonobject(value)
+    }
+}
+
+unsafe impl cxx::ExternType for QJsonValue {
```
ahayzen-kdab, line 211: elsewhere on all of these unsafe externtype blocks we have been putting something similar to

```
// Safety:
//
// Static checks on the C++ side to ensure the size is the same.
```

crates/cxx-qt-lib/src/core/mod.rs
```diff
@@ -25,6 +25,15 @@ pub use qflags::{QFlag, QFlagRepr, QFlags};
...
 pub use qhash::{QHash, QHashPair, QHashPair_QString_QVariant, QHashPair_i32_QByteArray};
 
+mod qjsonarray;
+pub use qjsonarray::QJsonArray;
+
+mod qjsonobject;
+pub use qjsonobject::QJsonObject;
+
+mod qjsonvalue;
+pub use qjsonvalue::QJsonValue;
```
ahayzen-kdab, line 35: would be good to have these added as types to `qt_types_standalone` (they just do a simple construct/clone etc across C++/Rust) to ensure there is no incorrect memory management, copy one of the other types examples

ahayzen-kdab: as then those tests are run through valgrind to look for memory leaks, so it's good to have all the cxx-qt-lib types that are marked as trivial checked for this, -extras are less supported so more optional for them

crates/cxx-qt-lib/src/core/qjsonvalue.rs
yuknig: I tried creating a direct binding as shown in your snippet, but unfortunately it does not compile because the Rust and C++ function signatures do not match (The C++ method has a parameter with a default value):
```
F:\dev\repos\fork\cxx-qt\target\debug\build\cxx-qt-lib-0d38c4151faaaa4a\out\cxxqtgen/src/core/qjsonvalue.cxx.cpp(107): error C2440: 'initializing': cannot convert from 'bool (__cdecl QJsonValue::* )(bool) const' to 'bool (__cdecl QJsonValue::* )(void) const'
F:\dev\repos\fork\cxx-qt\target\debug\build\cxx-qt-lib-0d38c4151faaaa4a\out\cxxqtgen/src/core/qjsonvalue.cxx.cpp(107): note: Types pointed to are unrelated; conversion requires reinterpret_cast, C-style cast or parenthesized function-style cast
```

crates/cxx-qt-lib/src/core/qjsonvalue.rs
yuknig: Fixed in f0a46d1.

crates/cxx-qt-lib/src/core/mod.rs
yuknig: Fixed in 2f9c41d.

crates/cxx-qt-lib/src/core/qjsonvalue.rs
ahayzen-kdab: So this is due to the default parameter, which is then tricky for CXX/Rust. 

Another option could be to have two methods
- Use `rust_name` to change this to `to_bool_or` (unsure of the best name there) so there can be a direct CXX binding as the types will then be correct
- Then create a pure Rust wrapper for `to_bool` that calls `to_bool_or(false)`

ahayzen-kdab: But there are also probably loads of places that could have this, so it's probably fine to keep it as-is for n ow i guess to stay consistent with other APIs that won't have the or variant

ahayzen-kdab (approved):
LGTM, thanks for the tests too!

crates/cxx-qt-lib/src/core/qjsonvalue.rs
yuknig: Implemented in https://github.com/KDAB/cxx-qt/pull/1479. Please consider that change if it is Ok from the API perspective.

PR #1475, opened 2026-08-04
cxx-qt-lib: implement QVariantValue for QList of QObject pointers

yuknig:
Add implementation of `QVariantValue` trait for `QList<QObjectMutPtr>`.


ahayzen-kdab (approved):
LGTM :-) Note if there are no conflicts you don't have to rebase, i can either do it via the Github UI, or when adding to the merge queue github can rebase it too.

PR #1478, opened 2026-08-10
cxx-qt-lib: add binding for QVariant::typeName()

yuknig:
Add binding for `QVariant::typeName()`.

### DIFF
--- a/crates/cxx-qt-lib/src/core/qvariant/mod.rs
+++ b/crates/cxx-qt-lib/src/core/qvariant/mod.rs
@@ -294,3 +294,37 @@ unsafe impl ExternType for QVariant {
     type Id = type_id!("QVariant");
     type Kind = cxx::kind::Trivial;
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{QList, QString};
+
+    #[test]
+    fn test_type_name() {
+        let variant = QVariant::from(&true);
+        assert_eq!(variant.type_name(), "bool");
+
+        let variant = QVariant::from(&123_i32);
+        assert_eq!(variant.type_name(), "int");
+
+        let variant = QVariant::from(&0.25f32);
+        assert_eq!(variant.type_name(), "float");
+
+        let variant = QVariant::from(&0.75f64);
+        assert_eq!(variant.type_name(), "double");
+
+        let variant = QVariant::from(&QList::<i32>::from_iter(&[1, 2, 3]));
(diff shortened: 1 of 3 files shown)


crates/cxx-qt-lib/src/core/qvariant/qvariant.cpp
```diff
@@ -56,6 +58,18 @@ namespace rust {
...
+qvariantTypeName(const QVariant& variant)
+{
+  if (const char* name = variant.typeName()) {
+    const ::std::size_t len = ::std::strlen(name);
+    return ::rust::Slice{ reinterpret_cast<const ::std::uint8_t*>(name),
+                          len + 1 };
+  }
+
+  return {};
+}
```
ahayzen-kdab, line 70: Would it be possible to use say `::rust::Str(const char *);` here and end up returning a `&str` in Rust ? As this is likely UTF8 compliant (as it's only ascii names?) there is no UTF16 vs UTF8 conversion required like QString, so then this could end up being `type_name(&self) -> &str` in Rust with no overhead still?

yuknig: I think that construction of `rust::Str` on the C++ side does not come for free. There are some checks and panic if string in not UTF-8:
https://docs.rs/crate/cxx/1.0.176/source/src/cxx.cc#331

What if we still return slice of bytes from C++ and create a `&str` on the Rust side (done in fad2915)? IMO this way is a bit better because handling of potentially invalid UTF-8 strings happens on the Rust side (even though such cases should never occur). So it feels a bit more Rust native.

crates/cxx-qt-lib/src/core/qvariant/mod.rs
```diff
@@ -119,6 +126,16 @@ impl QVariant {
         self.user_type().into()
     }
 
+    /// Returns the name of the type stored in the variant.
+    ///
+    /// The returned string describes the C++ datatype used to store the data.
+    /// An invalid variant returns an empty string.
+    pub fn type_name(&self) -> &str {
```
ahayzen-kdab, line 133: looks good could we have a unit test either here in this rust file as a cfg(test) or in the qt_types_standalone qvariant test to check it works ?

yuknig: Added a few tests in 3929152.

ahayzen-kdab (approved):
LGTM thanks!

PR #1479, opened 2026-08-11
cxx-qt-lib: add bindings for QJsonValue methods `toTypeX` 

yuknig:
… accepting a default value

Use these overloads instead of auxiliary helper functions that provide default-value handling.


ahayzen-kdab (approved):
Lets go with this, it removes C++ code and gives a Rust like API (doing an Option would be too complicated)

PR #1481, opened 2026-08-16
cxx-qt-lib: Add bindings for model/view-related enums in Qt namespace

### LINKED ISSUE #1480: cxx-qt-lib: Missing bindings to model/view-related Qt namespace enums
Subclasses of QAbstractItemModel implement a few methods (`data()`, `headerData()`, `roleNames()`) that take/return `role: i32` values corresponding to some Qt::ItemDataRole enum value. Additionally, for a model to indicate to a view which operations should be supported, it must implement the `flags()` method, which returns a value of type Qt::ItemFlags, an alias for QFlags<Qt::ItemFlag>.

It would be useful to be able to write the names of these enumeration values when working with model/view item data roles (i.e. writing 'ItemDisplayRole::DisplayRole' instead of '0'). It would also be nice to be able to use the bindings to QFlags/QFlag with ItemFlag values.

bumbledbees:
- Add binding for Qt::ItemDataRole
- Add binding for Qt::ItemFlag
- Add type alias for Qt::ItemFlags (QFlags<Qt::ItemFlag>)

fixes: https://github.com/KDAB/cxx-qt/issues/1480

### DIFF
--- a/crates/cxx-qt-lib/src/core/mod.rs
+++ b/crates/cxx-qt-lib/src/core/mod.rs
@@ -99,9 +99,9 @@ pub use qstringlist::QStringList;
 mod qt;
 pub use qt::{
     AspectRatioMode, BGMode, CaseSensitivity, ClipOperation, ConnectionType, DateFormat, FillRule,
-    GlobalColor, KeyboardModifier, KeyboardModifiers, LayoutDirection, MouseButton, MouseButtons,
-    Orientation, Orientations, PenCapStyle, PenJoinStyle, PenStyle, SizeMode, SplitBehaviorFlags,
-    TimeSpec, TransformationMode,
+    GlobalColor, ItemDataRole, ItemFlag, ItemFlags, KeyboardModifier, KeyboardModifiers,
+    LayoutDirection, MouseButton, MouseButtons, Orientation, Orientations, PenCapStyle,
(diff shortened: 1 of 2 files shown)


crates/cxx-qt-lib/src/core/mod.rs
```diff
@@ -99,9 +99,9 @@ pub use qstringlist::QStringList;
 mod qt;
 pub use qt::{
     AspectRatioMode, BGMode, CaseSensitivity, ClipOperation, ConnectionType, DateFormat, FillRule,
-    GlobalColor, KeyboardModifier, KeyboardModifiers, LayoutDirection, MouseButton, MouseButtons,
-    Orientation, Orientations, PenCapStyle, PenJoinStyle, PenStyle, SizeMode, SplitBehaviorFlags,
-    TimeSpec, TransformationMode,
+    GlobalColor, ItemDataRole, ItemFlag, ItemFlags, KeyboardModifier, KeyboardModifiers,
```
ahayzen-kdab, line 102: approach looks good are you able to rust format over the files as CI is failing on that

bumbledbees: yep, done!

ahayzen-kdab (approved):
LGTM :-)

PR #1482, opened 2026-08-16
Implement QVariantValue for QList<QString>

yuknig:
Add a `QVariantValue` implementation for `QList<QString>`.

`QStringList` already has an implementation, but in Qt 5 `QStringList` is a subclass of `QList<QString>` but not an alias for it. So, these two are distinct types and each needs its own implementation.


ahayzen-kdab (approved):
LGTM and tests seem to pass for both Qt 5 and Qt 6. Assume there isn't any confusion between QList<QString> and QStringList when you retrieve on the Rust side as they are an alias in Qt 6 so the "same" anyway, it's only Qt 5 where this can be fun ?

yuknig:
Thanks! 
Yes, in Qt6 they are the same type. In Qt 5 however, `QStringList` inherits `QList<QString>` and has a separate meta type.

PR #1484, opened 2026-08-18
cxx-qt-lib: mirror PartialEq impls added in the previous commit

yuknig:
so that QList operand may appear on either side of an equality comparison.

### DIFF
--- a/crates/cxx-qt-lib/src/core/qlist/mod.rs
+++ b/crates/cxx-qt-lib/src/core/qlist/mod.rs
@@ -112,6 +112,44 @@ where
     }
 }
 
+// The `PartialEq` implementations below mirror the ones above with the operands swapped,
+// allowing supported types to appear on either side.
+impl<T, U> PartialEq<QList<U>> for [T]
+where
+    U: QListElement + PartialEq<T>,
+{
+    fn eq(&self, other: &QList<U>) -> bool {
+        other == self
+    }
+}
+
+impl<T, U> PartialEq<QList<U>> for &[T]
+where
+    U: QListElement + PartialEq<T>,
+{
+    fn eq(&self, other: &QList<U>) -> bool {
+        other == self
+    }
+}
+
+impl<T, U, const N: usize> PartialEq<QList<U>> for [T; N]
+where
+    U: QListElement + PartialEq<T>,
+{
+    fn eq(&self, other: &QList<U>) -> bool {
+        other == self
+    }
+}
+
+impl<T, U, const N: usize> PartialEq<QList<U>> for &[T; N]
+where
+    U: QListElement + PartialEq<T>,
+{
+    fn eq(&self, other: &QList<U>) -> bool {
+        other == self
+    }
+}
+
 impl<T> Eq for QList<T> where T: QListElement + Eq {}
 
 impl<T> fmt::Debug for QList<T>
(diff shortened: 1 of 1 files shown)


crates/cxx-qt-lib/src/core/qlist/mod.rs
```diff
@@ -112,6 +112,48 @@ where
...
 
+// The `PartialEq` implementations below mirror the ones above with the operands swapped,
+// allowing supported types to appear on either side.
+impl<T, U> PartialEq<QList<U>> for [T]
+where
+    T: PartialEq<U>,
+    U: QListElement,
+{
+    fn eq(&self, other: &QList<U>) -> bool {
+        self.len() == other.len() as usize && self.iter().eq(other.iter())
```
ahayzen-kdab: as these are the opposite operands, could these all just do like `other == self` as the implementation to flip the order back and reuse the code from above rather than repeating the same implementation code again? or does that not work ? 🤔  Or even just this final inner eq() that they all eval to could reuse the existing code like this ?

yuknig: Changed to use `other == self` comparison in 64e89d1.

ahayzen-kdab (commented):
Looks good, a bit unfortunate that because the types are different we need this reversal.

I guess another simpler way could be instead to make a `as_slice(&self) -> &[T]` method like `Vec` has then you could do like `&[1,2] == list.as_slice()`.

I note that when looking at `Vec<T>` against `&[T]` it only allows you to compare a vec to slice and not slice to vec ? Instead you can do `as_slice()` to cause a slice to slice comparison which then works

@yuknig So i wonder if we should do the same here, keep only the PartialEq you have in the previous change and then add as_slice(), this would then be consistent with how a Vec works in std i think ?

yuknig:
@ahayzen-kdab I'm afraid we can't implement `as_slice()` for `QList` with Qt 5. As I understand the [documentation](https://doc.qt.io/archives/qt-5.15/qlist.html#details), `QList` does not guarantee a contiguous memory layout like one in `Vec`. It might have something like linked list layout:
> Note: QVector and QVarLengthArray both guarantee C-compatible array layout. QList does not. This might be important if your application must interface with a C API.

UPD: The lack of `data()` and `constData()` functions in QList in Qt 5 also hints that we should not rely on contiguity of the memory storage.

ahayzen-kdab:
> @ahayzen-kdab I'm afraid we can't implement `as_slice()` for `QList` with Qt 5. As I understand the [documentation](https://doc.qt.io/archives/qt-5.15/qlist.html#details), `QList` does not guarantee a contiguous memory layout like one in `Vec`. It might have something like linked list layout:
> 
> > Note: QVector and QVarLengthArray both guarantee C-compatible array layout. QList does not. This might be important if your application must interface with a C API.
> 
> UPD: The lack of `data()` and `constData()` functions in QList in Qt 5 also hints that we should not rely on contiguity of the memory storage.

@yuknig right in Qt 5 a Qt 6 QList is actually a QVector so likely put `as_slice()` if possible (?) only for Qt 6 to QList, and in theory would be added to QVector too. But tbh Qt 5 is EOL anyway so it's a matter of just keeping it building not blocking new features because of it so using the cfg flags like we do elsewhere is the route. Or whether we just ignore this feature entirely as Rust std doesn't tend to implement reverse PartialEq like this from what i can see ?

yuknig:
Yes, I think I can adapt codebase on the client side codebase so that it compiles without this change.

PR #1485, opened 2026-08-20
cxx-qt-lib: add binding for processEvents() method of QGuiApplication, QCoreApplication, QApplication

yuknig:


ahayzen-kdab (approved):
LGTM, we can probably add the QEventLoop flags later if required and those classes probably need rechecking at some point if the new inheritance stuff means that the derived classes don't need the duplicate methods, but problems for later

PR #1486, opened 2026-08-21
cxx-qt-lib: harden conversion of QObjectMutPtr from / to raw pointer.

yuknig:
Replace `From` trait implementations with dedicated conversion methods to make raw pointer conversions a bit more explicit.

### DIFF
--- a/crates/cxx-qt-lib/src/core/qobjectmutptr.rs
+++ b/crates/cxx-qt-lib/src/core/qobjectmutptr.rs
@@ -7,20 +7,37 @@ use cxx_qt::QObject;
 
 /// A thin wrapper around `*mut QObject`.
 /// Using the 'newtype' idiom lets us implement `ExternType`
-/// for it, which wouldn't possible for pure `*mut QObject` itself.
+/// for it, which wouldn't be possible for pure `*mut QObject` itself.
 #[repr(transparent)]
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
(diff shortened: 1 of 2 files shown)


crates/cxx-qt-lib/src/core/qobjectmutptr.rs
```diff
@@ -7,20 +7,27 @@ use cxx_qt::QObject;
...
     }
-}
 
-impl From<QObjectMutPtr> for *mut QObject {
-    fn from(value: QObjectMutPtr) -> Self {
-        value.0
+    /// Return wrapped raw pointer.
+    pub fn as_raw(self) -> *mut QObject {
+        self.0
     }
```
ahayzen-kdab, line 41: nice to make the wrapper more robust, wonder if it should follow a similar API style to CXX's UniquePtr and SharedPtr ?
https://docs.rs/cxx/latest/cxx/struct.UniquePtr.html
https://docs.rs/cxx/latest/cxx/struct.SharedPtr.html

Eg they have methods like this
- `pub unsafe fn from_raw(raw: *mut T) -> Self`
- `pub fn as_ptr(&self) -> *const T`
- `pub fn as_mut_ptr(&self) -> *mut T`

So `from_raw` and `as_mut_ptr` probably make most sense here as names?

yuknig: Good point. Fixed in 5ba284c.

ahayzen-kdab (approved):
LGTM :-)

PR #1488, opened 2026-09-11
book: remove deprecated 'multilingual' key in book.toml

bumbledbees:
In version 0.5 of mdBook, support for the `book.multilingual` key in book.toml was removed. Trying to build the CXX-Qt book with a current version of mdBook will fail because of this, so the field should be removed.

(see: https://github.com/rust-lang/mdBook/pull/2775)


ahayzen-kdab (approved):
LGTM :-)

PR #1489, opened 2026-09-11
book: fix example code in Inheritance section

bumbledbees:
When trying to subclass `QAbstractItemModel` based on the example in the [Inheritance section of the
book](https://kdab.github.io/cxx-qt/book/concepts/inheritance.html), my code wouldn't compile. The book's code doesn't have an `extern "C++"` or `extern "C++Qt"` block to provide the definition for `QAbstractItemModel`, which is what I had to add to get my code to compile.

There were already mdBook anchors at the relevant parts of the code that the book is referencing, so this commit includes them when necessary.


ahayzen-kdab (approved):
LGTM :-) Oops good spot probably missed in one of the many migrations 😅

